PR #6, opened 2017-04-11
more description in documentation

maximd33:


README.md
```diff
@@ -7,7 +7,7 @@ Intel® Media SDK provides an API to access hardware-accelerated video decode, e
 
 # Important note
 The current version of Intel Media SDK is only compatible with the Linux* Graphics Driver from [Intel® Media Server Studio for Linux*](https://software.intel.com/en-us/intel-media-server-studio). You need to download and install Intel Media Server Studio 2017 R2 _before_ you build Intel Media SDK.
-Intel Media SDK depends on a special version of [LibVA](https://www.freedesktop.org/wiki/Software/vaapi/) which comes with Intel Media Server Studio installation and is not in upstream, so this version is not compatible with the LibVA/driver available at [01org](https://01.org/linuxgraphics/downloads ).
+Intel Media SDK depends on a special version of [LibVA](https://www.freedesktop.org/wiki/Software/vaapi/) which comes with Intel Media Server Studio installation and is not in upstream, so this version is not compatible with the LibVA/driver available at [01org](https://01.org/linuxgraphics/downloads). Installed libav-devel,libdrm-devel/libX11-devel should be sufficient for build environment to compile.
```
amaslenn: Two important moments:
 - Cmake handles dependencies already.
 - Using mentioned packages different from what comes with Intel MSS installation may lead to broken functionality of Media SDK, so I can't agree with this addition.

README.md
```diff
@@ -67,6 +67,8 @@ perl tools/builder/build_mfx.pl --cmake=intel64.make.release
 ```
 This will build MSDK binaries and MSDK samples.
 
+Please note build_mfx.pl optional flags, like one example of: --no-warn-as-error to disable default: treat Warning As Error.
```
amaslenn: Please mention, that using this option may be required when using newer version (`>4.8`) of compiler (probably better place it after paragraph about using CLang compiler). And this option is `--no_warn_as_error`, with underscores, not dashes.

maximd33: right,
see my commit 813f9ef

README.md
maximd33: - it has not much to do with cmake but packages already installed
- it is only about scenario like "build PC" where you dont run but just compile. it is not much covered by documentation but proven to work though.

amaslenn: It is still misleading: you had to install this list of dependencies, but it is really up to current environment and distribution you use. And functionality matters! Such comment in main document may mislead people to create incorrect environment for build and run.

README.md
```diff
@@ -71,6 +71,8 @@ If you want to configure build with CLang compiler use the following command:
 ```sh
 perl tools/builder/build_mfx.pl --cmake=intel64.make.release.clang
 ```
+Please note build_mfx.pl optional flags, like one example of: --no_warn_as_$error to disable default: treat Warning As Error,
```
amaslenn: Please add formatting options for `build_mfx.pl` and option name.

PR #18, opened 2017-12-05
cmake: allow MFX_HOME to be optional

### LINKED ISSUE #16: Add a default if MFX_HOME is not set
MFX_HOME should not be required since it usually just points to the same directory as is being configured by cmake (i.e. the top of source tree).  Consider setting the default to CMAKE_HOME_DIRECTORY when MFX_HOME is not defined in the environment.

uartie:
Often, MFX_HOME will just point back to the source tree that
is being configured by cmake.

Thus, set MFX_HOME to CMAKE_HOME_DIRECTORY if user does not
set it.

Fixes #16

Signed-off-by: U. Artie Eoff <ullysses.a.eoff@intel.com>


uartie:
any comments?

PR #19, opened 2017-12-05
cmake: using GNUInstallDirs for mfx libraries and headers

### LINKED ISSUE #17: Allow INSTALL_LIB_PATH to be customized by developer
The mfx, dispatch_shared, dispatch_trace, and pkgconfig files and libraries are installed into non-GNU standard directories (lib/lin_x86 or lib/lin_x64).  Please allow developer to optionally configure this path instead of forcing non-standards upon them.  Generally, GNUInstallDirs (https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) would be a better approach for maximum compatibility with most Linux distributions.

uartie:
Use GNUInstallDirs to determine standard location to
install libraries and include headers.  This also
enables developers to override them, too.  For example,

-DCMAKE_INSTALL_LIBDIR=lib/lin_x64 will install into
$CMAKE_INSTALL_PREFIX/lib/lin_x64 and so on.

Fixes #17

Signed-off-by: U. Artie Eoff <ullysses.a.eoff@intel.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-cmake_minimum_required( VERSION 2.6.2 )
+cmake_minimum_required( VERSION 2.8.5 )
 project( mediasdk )
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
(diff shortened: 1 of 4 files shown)


uartie:
any comments?

onabiull:
Sorry for late reply. I've started some internal discussion about this change. Actually tradeoff is between product-like binaries location like MSS currently use and library-like location as proposed in this change. And having MediaSDK dispatcher, samples and plugins complicate things a bit. I'd like to have same approach among all components of media stack and this communication takes time.

PR #21, opened 2017-12-07
Fix build failures on ubuntu 16.04

ruiling:
This request is used to address the issues met during compiling ffmpeg-qsv against msdk. This would improve the compatibility of msdk and also make it easy to build ffmpeg-qsv.


onabiull:
Thank you for contribution! It would take some time to verify these fixes. There are some pieces of code that didn't made opensource due to various reasons (Windows, some plugins), but based on MSDK code and should be source-level compatible. Once I get confirmation that changes won't break anything I'll merge these changes.

ruiling:
@onabiull That makes sense. wait for your verification result. If verification failed, it's better you could give me some hint, so I can further refine the patch. Thanks!

dbermond:
ffmpeg expects mfx headers to be located at `<prefix>/include/mfx` instead of `<prefix>/include`.

It also expects the pkgconfig file to be named `libmfx.pc` instead of `mfx.pc`.

At least, this is the behavior on newer ffmpeg versions. I'm not sure about ffmpeg expectations in the old version provided by Ubuntu 16.04 (ffmpeg 2.8?). I needed to do these modifications in order to get current ffmpeg (3.4.1 and git master) to compile with `--enable-libmfx`.

ruiling:
yeah, it's true. I have written some documentation that hasn't been upstreamed to FFmpeg, currently it is located at: https://github.com/Intel-FFmpeg-Plugin/Intel_FFmpeg_plugins/wiki/Building-against-open-source-MediaSDK-and-iHD-driver-for-QSV-codecs
Personally I agree with you to install the header files into `<prefix>/include/mfx/` and name pkg_config file as `libmfx.pc` to stay compatible with FFmpeg. Current solution is not quite user friendly. But I am not sure what MediaSDK developer thinks about this? @onabiull

ruiling:
@onabiull I have rebased to the latest master and fix one more link error caused by recent commit. I am not sure can you give a review? It is really appreciated if we can go through the internal validation of this PR once you think the patch is ok.

dbermond:
The patch 'Add runtime library that we depend on. ' currently fails to be applied on the latest git master.

ruiling:
@dbermond rebased.

onabiull:
Could you please rebase this change?

ruiling:
Sorry for late response. Just returned from vacation. the PR has been rebased.

Issue #23, opened 2017-12-20
sample_decode failed when decoding hevc clips.


lizhong1008:
H264 decoder can work well, but hevc failed seems it can't find the plugin.

To reproduce: sample_decode h265 -i ~/fate-suite/hevc-conformance/MVEDGE_A_qualcomm_3.bit -o test.yuv

plugin_loader.h :168 [ERROR] Failed to load plugin from GUID, sts=-9: { 0x33, 0xa6, 0x1c, 0x0b, 0x4c, 0x27, 0x45, 0x4c, 0xa8, 0xd8, 0x5d, 0xde, 0x75, 0x7c, 0x6f, 0x8e } (Intel (R) Media SDK HW plugin for HEVC DECODE)
Default plugin cannot be loaded (possibly you have to define plugin explicitly)

[ERROR], sts=MFX_ERR_UNSUPPORTED(-3), Init, Plugin load failed at /root/lizhong/opensource/MediaSDK/samples/sample_decode/src/pipeline_decode.cpp:379

[ERROR], sts=MFX_ERR_UNSUPPORTED(-3), main, Pipeline.Init failed at /root/lizhong/opensource/MediaSDK/samples/sample_decode/src/sample_decode.cpp:661

lizhong1008:
After check the source code: "api/opensource/mfx_dispatch/src/mfx_plugin_hive_linux.cpp",
plugins.cfg file path is set to hard code: "const char rootPluginPath[] = "/opt/intel/mediasdk/plugins/plugins.cfg". Could it to be configurable as CMAKE_INSTALL_PREFIX?

eero-t:
This seems a fixed issue, could it be closed?

PR #25, opened 2017-12-21
mfx_plugin: make plugins path configurable on Linux

lizhong1008:
Allow plugins path can be configured by environment variable
"MFX_PLUGINS_PATH".
Default path is "/opt/intel/mediasdk/plugins/plugins.cfg"

Signed-off-by: Zhong Li <zhong.li@intel.com>

### DIFF
--- a/api/opensource/mfx_dispatch/src/mfx_plugin_hive_linux.cpp
+++ b/api/opensource/mfx_dispatch/src/mfx_plugin_hive_linux.cpp
@@ -38,9 +38,15 @@
 #define TRACE_HIVE_INFO(str, ...) DISPATCHER_LOG_INFO((("[HIVE]: "str), __VA_ARGS__))
 #define TRACE_HIVE_WRN(str, ...) DISPATCHER_LOG_WRN((("[HIVE]: "str), __VA_ARGS__))
 
+#ifndef MFX_PLUGINS_DIR
+#define MFX_PLUGINS_DIR "/opt/intel/mediasdk/plugins"
+#endif
+
 namespace 
 {
-    const char rootPluginPath[] = "/opt/intel/mediasdk/plugins/plugins.cfg";
+    const char *rootPluginPath[2] = { getenv("MFX_PLUGINS_PATH" ),
+                                      MFX_PLUGINS_DIR"/plugins.cfg"
+                                    };
     //const wchar_t rootDispatchPath[] = L"Software\\Intel\\MediaSDK\\Dispatch";
     const char pluginSubkey[] = "Plugin";
     const char TypeKeyName[] = "Type";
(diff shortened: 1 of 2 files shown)


api/opensource/mfx_dispatch/src/mfx_plugin_hive_linux.cpp
```diff
@@ -40,7 +40,9 @@
 
 namespace 
 {
-    const char rootPluginPath[] = "/opt/intel/mediasdk/plugins/plugins.cfg";
+    const char *rootPluginPath[2] = { getenv("MFX_PLUGINS_PATH" ),
+                                      "/opt/intel/mediasdk/plugins/plugins.cfg"
+                                    };
```
uartie: Can you make this configurable at compile time, too?  For example, add -DMFX_PLUGINS_DIR=... to compiler and use `MFX_PLUGINS_DIR "/plugins.cfg"` here instead of hardcoded /opt/intel/mediasdk/plugins/plugins.cfg.  In cmake files, MFX_PLUGINS_DIR is set to `$CMAKE_INSTALL_PREFIX/plugins` and `CMAKE_INSTALL_PREFIX=/opt/intel/mediasdk` by default.  LibVA uses a similar technique for the VA_DRIVERS_PATH.

lizhong1008: Good idea, will update this patch. : )

dbermond:
Shouldn't it also modify the actual plugins installation path?

Plugins are still being installed at `/usr/plugins` (which is not a standard directory).

lizhong1008:
@dbermond , plugins is installed to $CMAKE_INSTALL_PREFIX/plugins and the default path is /opt/intel/mediasdk/plugins. If you want to change the MSDK installation path, you can run "perl tools/builder/build_mfx.pl --cmake=intel64.make.debug --prefix=$LIBMSDK_PREFIX" and change $LIBMSDK_PREFIX as your requirement.

dbermond:
@lizhong1008 Thank you for the reply.

I'm already using `--prefix='/usr'`. But the environment variable setting is not working for me.

After applying this patch and running `export MFX_PLUGINS_PATH='/usr/lib'` I'm still getting the plugins installed at `/usr/plugins`. It also have the same result with `export MFX_PLUGINS_PATH='lib'`.

lizhong1008:
@dbermond , I think you should  "export MFX_PLUGINS_PATH=/usr/lib/plugins/plugins.cfg".  And the environment variable "MFX_PLUGINS_PATH" should not be necessary since I've updated the patch with "add_definitions( -DMFX_PLUGINS_DIR=\"${MFX_PLUGINS_DIR}\" )"

lizhong1008:
@onabiull  @dvrogozh Any comment to this patch? 
It can fix the bug that application can't find plugins if MSDK has not been installed to default path, just like https://github.com/Intel-Media-SDK/MediaSDK/issues/23.

dvrogozh:
@onabiull should review and approve this. He is out of office now due to public holidays. Please, expect answer after January 9.

uartie:
@lizhong1008 it looks like @onabiull added some of these features in d1a12d8018134

lizhong1008:
@uartie Got it, thanks for remind. It works much better than previous hard code, but still can't be configurable on run time.

onabiull:
Looks like this code already merged as part of previous PR.

PR #37, opened 2018-01-12
Add HEVC FEI support

onabiull:


_studio/hevc_fei/h265_fei/src/mfx_h265_fei_encode_hw.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (!surface) return MFX_ERR_NONE; // In case of frames draining in display order
+
+    MFX_CHECK(ctrl, MFX_ERR_UNDEFINED_BEHAVIOR);
+
+    mfxPlatform p = {};
+
+    mfxStatus sts = m_core.QueryPlatform(&p);
+    MFX_CHECK_STS(sts);
+
+    bool isSKL = p.CodeName == MFX_PLATFORM_SKYLAKE, isICLplus = false;
```
artem-shaporenko: ICL should not be published here

PR #41, opened 2018-01-19
[vpp] Add support of p010 to rgb4 conversion

rkazants:


_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -389,6 +389,17 @@ mfxStatus mfxDefaultAllocatorVAAPI::SetFrameData(const VAImage &va_image, mfxU32
         }
         else mfx_res = MFX_ERR_LOCK_MEMORY;
         break;
+    case VA_FOURCC_P010:
+        if (mfx_fourcc == MFX_FOURCC_P010)
+        {
+            ptr->PitchHigh = (mfxU16)(va_image.pitches[0] / (1 << 16));
```
lakulako: '>>16' and '& 0xffff' could be more clear.
Better use static_cast<mfxU16>.
At 399 better ptr->U + sizeof(mfxU16)

rkazants: The formula for pitch high is written the same as for other formats. So propose to stay the current one.
I can correct line 339 as you proposed.

lakulako: Don't multiply drawbacks.

rkazants: Done.

PR #46, opened 2018-01-24
[build, cmake] Enabled out-of-source build for Eclipse CDT with makef…

alariono:
…iles and project generation

### DIFF
--- a/tools/builder/build_mfx.pl
+++ b/tools/builder/build_mfx.pl
@@ -141,12 +141,16 @@ sub get_cmake_gen_cmd {
     push @cmake_cmd_gen, '-DWARNING_FLAGS="-Wall -Werror"' if not $no_warn_as_error;
 
     # This is the main different between "make" and "eclipse"" generators.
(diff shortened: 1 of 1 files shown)


tools/builder/build_mfx.pl
```diff
@@ -257,32 +263,30 @@ sub _opt_cmake_handler {
...
 } else {
     rmtree($target_path) if -d $target_path  and $clean;
     mkpath($target_path) if !-d $target_path and -d $build_root;
     chdir($target_path) or die "can't chdir -> $target_path: $!";
+    $cmake_gen_cmd = get_cmake_gen_cmd($build_root, %build);
 }
 
 my $exit_code = 0;
-my $cmake_gen_cmd = get_cmake_gen_cmd($build_root, %build);
+#my $cmake_gen_cmd = get_cmake_gen_cmd($build_root, %build);
```
otsirule: Never understand why need comment code and leave it without explanation. Wouldn't it better to remove it?

alariono: Fine. Let's remove

PR #52, opened 2018-01-27
sample_fei changes merge from internal

artem-shaporenko:
Multisession NN support
MFE support
build check for MFX_VESION


otsirule:
Could you please do not use a colon at the end of commit message? It's much easier to read in history without it.

PR #56, opened 2018-02-01
[sample_fei] Clean up buffers to avoid possible memory growth after p…

mgonchar:
…ipeline recreation

If CEncodingPipeline::Run() returns MFX_ERR_DEVICE_LOST or MFX_ERR_DEVICE_FAILED app will try to reset the pipeline and repeat encoding. But FEI buffers won't be cleaned. This patch prevents memory growth in such situations.


otsirule:
Please squash commits: it's expected that all commits in pull requests can be built, so it will be easier to use git bisect.

PR #61, opened 2018-02-03
[vc1d] Fix for b_picture_fraction usage

dmitryermilov:
Fix for b_picture_fraction usage


dmitryermilov:
This should be merge together with https://github.com/intel/media-driver/pull/22

PR #68, opened 2018-02-08
[h264e, h264e fei] Removed usage of VAEncPackedHeaderH264_SEI

dmitryermilov:
VAEncPackedHeaderH264_SEI has been depricated so
we replace it with VAEncPackedHeaderRawData


dmitryermilov:
This should be merged together with https://github.com/intel/media-driver/pull/91

PR #71, opened 2018-02-09
Timeout option in sample_hevc_fei

otsirule:
Added timeout option for HEVC FEI testing and made changes required for enabling this option.

### DIFF
--- a/samples/sample_hevc_fei/include/hevc_fei_preenc.h
+++ b/samples/sample_hevc_fei/include/hevc_fei_preenc.h
@@ -40,6 +40,8 @@ public:
 
     virtual mfxStatus PreEncFrame(HevcTask * task) = 0;
 
+    virtual mfxStatus ResetIOState() = 0;
+
 protected:
     mfxStatus ResetExtBuffers(const MfxVideoParamsWrapper & videoParams);
 
@@ -64,6 +66,8 @@ public:
     virtual mfxStatus QueryIOSurf(mfxFrameAllocRequest* request);
 
     virtual mfxStatus PreEncFrame(HevcTask * task);
+
+    virtual mfxStatus ResetIOState();
 
 private:
     mfxU8  m_numOfStructuresPerFrame; // number of mfxExtFeiPreEncMVMB structures per frame
@@ -92,6 +96,8 @@ public:
 
     mfxStatus PreEncFrame(HevcTask * task);
 
+    mfxStatus ResetIOState();
+
 private:
     mfxStatus PreEncMultiFrames(HevcTask * task);
     mfxStatus PreEncOneFrame(HevcTask & currTask, const RefIdxPair& dpbRefIdxPair, const RefIdxPair& activeRefIdxPair, const bool bDownsampleInput);
@@ -112,6 +118,7 @@ private:
     std::auto_ptr<FileHandler> m_pFile_MBstat_out;
 
     bool m_isMVoutFormatted;
+    mfxU32 m_processedFrames;
 
     DISALLOW_COPY_AND_ASSIGN(FEI_Preenc);
 
@@ -131,6 +138,7 @@ public:
     mfxStatus PreEncFrame(HevcTask * task);
 
     MfxVideoParamsWrapper GetVideoParam();
+    mfxStatus ResetIOState();
 
 private:
     mfxStatus DownSampleFrame(HevcTask & task);
--- a/samples/sample_hevc_fei/src/fei_utils.cpp
+++ b/samples/sample_hevc_fei/src/fei_utils.cpp
@@ -192,6 +192,13 @@ mfxStatus YUVReader::GetFrame(mfxFrameSurface1* & pSurf)
     return sts;
 }
 
+mfxStatus YUVReader::ResetIOState()
+{
+    m_FileReader.Reset();
+
+    return MFX_ERR_NONE;
+}
+
 /**********************************************************************************/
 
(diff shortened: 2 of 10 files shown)


samples/sample_hevc_fei/include/ref_list_manager.h
```diff
@@ -117,6 +107,21 @@ class EncodeOrderControl
         m_encoding.resize(0);
 
         m_lastFieldInfo.Reset();
+        m_lastIDR = 0;
+        m_frameOrder = 0;
+
+        m_lastTask.Reset();
+        // make m_lastTask invalid:
+        Fill(static_cast<HevcDpbFrame & >(m_lastTask), IDX_INVALID);
```
mgonchar: shouldn't all the code below be part of task.Reset()?

samples/sample_hevc_fei/src/fei_utils.cpp
```diff
@@ -533,6 +547,13 @@ void FieldSplitter::Close()
     return m_pTarget->Close();
 }
 
+mfxStatus FieldSplitter::ResetIOState()
+{
+    m_pTarget->ResetIOState();
```
mgonchar: you should return status of ResetIOState() here

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -320,25 +311,52 @@ mfxStatus CEncodingPipeline::Execute()
...
     {
-        if (m_inParams.nNumFrames <= numSubmitted) // frame encoding limit
+        if (m_inParams.nNumFrames <= numSubmitted                               // frame encoding limit
+            || (m_inParams.nTimeout && time(0) - start >= m_inParams.nTimeout)) // encoding time limit
             break;
 
         sts = m_pYUVSource->GetFrame(pSurf);
+        if (MFX_ERR_MORE_DATA == sts && m_inParams.nTimeout && time(0) - start < m_inParams.nTimeout)
+        {
+            // reached end of input file, drain encoder and start from the beginning
```
mgonchar: that won't work. You shouldn't drain frames in case of loop encoding. Check presets for B frames, for AVC codec you would see that encoded sequence after draining follows some other preset. The right thing to do here is ONLY reset file pointers to emulate long file, other components should think that they process the next frame from stream

mgonchar: Forcing encoder to insert IDR is also required, of course

samples/sample_hevc_fei/src/fei_utils.cpp
otsirule: Agree. Thanks.

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
otsirule: Ok. I'll check that reorder class will work correctly without draining and fix.

samples/sample_hevc_fei/include/ref_list_manager.h
otsirule: Looks like it should. I'll check the code where Reset is called and submit the change if there are no conflicts with this change.

samples/sample_hevc_fei/include/hevc_fei_preenc.h
```diff
@@ -92,6 +96,8 @@ class FEI_Preenc : public IPreENC
 
     mfxStatus PreEncFrame(HevcTask * task);
 
+    mfxStatus ResetIOState();
```
lakulako: Add virtual because it is virtual. Same for 141.

samples/sample_hevc_fei/include/ref_list_manager.h
```diff
@@ -129,10 +130,20 @@ class EncodeOrderControl
 
     HevcTask* GetTask(mfxFrameSurface1 * surface)
     {
-        HevcTask* task = ReorderFrame(surface);
-        if (task)
+        HevcTask* task = NULL;
+        if (m_reorder)
```
lakulako: I don't like unnecessary renaming. Also now class called 'Reorder' can be not a reorder.

otsirule: The same for EncodeOrderControl - it can be not encode order.

samples/sample_hevc_fei/include/hevc_fei_preenc.h
otsirule: See no reason to make it virtual because other functions in this class are non-virtual and there's no need to derive from it now.

otsirule:
Changed behavior after offline discussion. Unnecessary changes were removed.

samples/sample_hevc_fei/include/hevc_fei_preenc.h
lakulako: This method overrides parent virtual IPreENC::ResetIOState. It is already virtual.

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
```diff
@@ -155,8 +155,12 @@ mfxStatus FEI_Encode::Query()
 
 mfxStatus FEI_Encode::Init()
 {
-    mfxStatus sts = m_FileWriter.Init(m_dstFileName.c_str());
-    MSDK_CHECK_STATUS(sts, "FileWriter Init failed");
+    mfxStatus sts = MFX_ERR_NONE;
+    if (m_dstFileName.length())
```
lakulako: Better use !empty()

PR #79, opened 2018-02-19
[vpp] Enable low power mode scaling

rkazants:
Added run-time check for APL to use low
power scaling mode by default due to power
consumption consideration.
Document scaling mode in sample vpp.

### DIFF
--- a/_studio/shared/src/mfx_vpp_vaapi.cpp
+++ b/_studio/shared/src/mfx_vpp_vaapi.cpp
@@ -458,6 +458,7 @@ mfxStatus VAAPIVideoProcessing::QueryCapabilities(mfxVppCaps& caps)
     }
 
     caps.uMirroring = 1;
+    caps.uScaling = 1;
 
     return MFX_ERR_NONE;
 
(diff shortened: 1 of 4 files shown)


_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1158,13 +1162,13 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
...
-        /* Use SFC by default on BXT platforms due to power consumption considerations */
-        m_pipelineParam[0].filter_flags |= VA_FILTER_SCALING_DEFAULT;
-#else
-        /* Force AVS by default for all platforms except BXT */
-        m_pipelineParam[0].filter_flags |= VA_FILTER_SCALING_HQ;
-#endif
+        if(MFX_HW_APL == hwType) {
+            /* Use HW fixed function scaling engine by default on APL due to power consumption considerations */
+            m_pipelineParam[0].filter_flags |= VA_FILTER_SCALING_DEFAULT;
+        } else {
```
dmitryermilov: please keep the usual MSDK style of {} blocks like:
```
}
else
{
```

PR #88, opened 2018-02-27
Add support setting -n > input's real num frames

otsirule:


samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -331,20 +331,25 @@ mfxStatus CEncodingPipeline::Execute()
             sts = m_pYUVSource->ResetIOState();
             MSDK_BREAK_ON_ERROR(sts);
 
-            // execution timeout can be too big for available disk space, so reset output file pointers
-            // despite the fact that stream can be undecodable
-            if (m_pFEI_PreENC.get())
+            if (time(0) - start < m_inParams.nTimeout)
```
mgonchar: 1) What is the reason to check the same thing as on line 321 once more? It doesn't bring much difference as I see.
2) Would it work as expected during nNumFrames != 0 ? The default for timeout is 0, so you just prohibit output file reset. If this is desired behavior, what if nNumFrames is used together with nTimeout ? It should work in the way that the first happened event: timeout or frame number limit reaching should interrupt processing. How the output file should look like in such situation

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -69,7 +69,7 @@ void PrintHelp(const msdk_char *strAppName, const msdk_char *strErrorMessage)
     msdk_printf(MSDK_STRING("                          if ds_strength parameter is missed or equal 1, PREENC is used on the full resolution\n"));
     msdk_printf(MSDK_STRING("                          otherwise PREENC is used on downscaled (by VPP resize in ds_strength times) surfaces\n"));
     msdk_printf(MSDK_STRING("   [-EncodedOrder] - use app-level reordering to encoded order (default is display; ENCODE only)\n"));
-    msdk_printf(MSDK_STRING("   [-n number] - number of frames to process\n"));
+    msdk_printf(MSDK_STRING("   [-n number] - number of input frames to process\n"));
```
mgonchar: Please, add some comment about behavior when n > number of frames in input

otsirule:
@mgonchar thank you for comments. Fixed in the last commit.

PR #89, opened 2018-02-27
Enable build for Android platform

dmitrybrazhkin:


_studio/mfx_lib/CMakeLists.txt
```diff
@@ -137,6 +137,11 @@ foreach( prefix ${MSDK_STUDIO_ROOT}/shared/src )
     ${prefix}/mfx_mfe_adapter.cpp
   )
 endforeach()
+foreach( prefix ${MSDK_LIB_ROOT}/genx/field_copy/src )
```
dvrogozh: Android does not use cmake. Thus, this is unrelated change. Could you, please, extract it to the separate commit?

_studio/mfx_lib/vpp/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ mfx_include_dirs()
 include_directories( ${MSDK_LIB_ROOT}/vpp/include )
 include_directories( ${MSDK_LIB_ROOT}/cmrt_cross_platform/include )
 include_directories( ${MSDK_LIB_ROOT}/genx/copy_kernels/include )
-include_directories( ${MSDK_LIB_ROOT}/genx/field_copy/src )
```
dvrogozh: Again, separate commit.

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -41,11 +41,6 @@
 #include "va/va.h"
 #include <va/va_backend.h>
 
-#if (MFX_VERSION >= MFX_VERSION_NEXT)
```
dvrogozh: @onabiull : is that OK to delete this? recently I am afraid of any changes to versions stuff.

_studio/shared/umc/io/umc_va/src/umc_va_linux.cpp
```diff
@@ -310,10 +310,10 @@ Status LinuxVideoAccelerator::Init(VideoAcceleratorParams* pInfo)
                                   && ((m_Profile & VA_CODEC) != UMC::VA_H265)
                                   && ((m_Profile & VA_CODEC) != UMC::VA_VP8)
                                   && ((m_Profile & VA_CODEC) != UMC::VA_VP9)
-#ifndef ANDROID
```
dvrogozh: I understand that Android may not support these codecs, but why disable them here? maybe move somewhere to the upper level (MFX)?

android/build/mfx_defs_internal.mk
```diff
@@ -0,0 +1,34 @@
...
+MFX_CFLAGS_INTERNAL := \
+    $(MFX_CFLAGS) \
+    $(MFX_CFLAGS_STL)
+
+MFX_CFLAGS_INTERNAL_32 := -DLINUX32
+MFX_CFLAGS_INTERNAL_64 := -DLINUX32 -DLINUX64
+
+MFX_CFLAGS_INTERNAL_HW := \
+    $(MFX_CFLAGS_INTERNAL) \
+    -Wno-error=non-virtual-dtor
```
dvrogozh: can you check whether we really still need this?

_studio/hevce_hw/Android.mk
```diff
@@ -0,0 +1,86 @@
...
+MFX_LOCAL_STATIC_LIBRARIES := \
+    libmfx_trace_hw \
+    libmfx_lib_merged_hw \
+    libumc_core_merged_hw \
+    libsafec
+
+MFX_LOCAL_LDFLAGS := \
+    $(MFX_LDFLAGS) \
+    -Wl,--version-script=$(LOCAL_PATH)/libmfx_h265e_plugin.map \
+    -Wl,--no-warn-shared-textrel
```
dvrogozh: Why you need this?! Looks like something is very wrong in the build for this target.

_studio/hevce_hw/libmfx_h265e_plugin.map
```diff
@@ -0,0 +1,6 @@
+LIBMFXSW_PLUGIN {
+  global:
+   CreatePlugin;
+   local:
+   *;
+};
```
dvrogozh: add newline, please

_studio/shared/umc/io/umc_va/src/umc_va_linux.cpp
dmitrybrazhkin: Android supports all these codecs and they are not disabled here. needAllocatedSurfaces shows possibility of Adaptive Playback feature

_studio/shared/src/libmfx_core_vaapi.cpp
dmitrybrazhkin: Lines 44-47 is a dummy code. I just removed it

android/build/mfx_defs_internal.mk
dmitrybrazhkin: It was added by @dmitryermilov with the following message: "Fix for "shared library text segment is not shareable" warrning treated as error."
I checked the build without this line, and the build was completed without such error.

_studio/hevce_hw/Android.mk
dmitrybrazhkin: It was added by @dmitryermilov with the following message: "Fix for "shared library text segment is not shareable" warrning treated as error."
I checked the build without this line, and the build was completed without such error.

dvrogozh: If you don't see build errors, I suggest to remove.

_studio/mfx_lib/Android.mk
```diff
@@ -0,0 +1,208 @@
...
+    libumc_codecs_merged \
+    libumc_io_merged_hw \
+    libumc_core_merged_hw \
+    libmfx_trace_hw \
+    libsafec
+
+MFX_LOCAL_LDFLAGS_HW += \
+    $(MFX_LDFLAGS) \
+    -Wl,--version-script=$(LOCAL_PATH)/mfx_lib/libmfx.map \
+    -Wl,--no-warn-shared-textrel
```
dvrogozh: Here is one more. I think we need a grep results  to eliminate all such warning flags in Android.mk files

_studio/shared/src/libmfx_core_vaapi.cpp
dvrogozh: got it, my mistake

_studio/shared/umc/io/umc_va/src/umc_va_linux.cpp
dvrogozh: ok.

_studio/shared/umc/io/umc_va/src/umc_va_linux.cpp
```diff
@@ -310,10 +310,10 @@ Status LinuxVideoAccelerator::Init(VideoAcceleratorParams* pInfo)
...
+                                  && ((m_Profile & VA_CODEC) != UMC::VA_VC1)
                                   && ((m_Profile & VA_CODEC) != UMC::VA_MPEG2)
+#ifndef ANDROID
                                   && ((m_Profile & VA_CODEC) != UMC::VA_JPEG)
```
dvrogozh: What about JPEG? we have adaptive playback support on Android and does not have it on Linux? This looks weird. Can you check whether driver supports this on Linux as well? I would prefer to have no difference between Linux/Android here.

dmitrybrazhkin: needAllocatedSurfaces = true means that Adaptive Playback is unsupported (pre-allocated surfaces are required in this case). So, Adaptive Playback is supported for JPEG on Linux and unsupported on Android. We do not deliver JPEG decoder on Android, so it can be excluded here.

_studio/mfx_lib/CMakeLists.txt
dmitrybrazhkin: done

_studio/mfx_lib/vpp/CMakeLists.txt
dmitrybrazhkin: done

android/build/mfx_defs_internal.mk
dmitrybrazhkin: done

_studio/hevce_hw/libmfx_h265e_plugin.map
dmitrybrazhkin: done

dmitrybrazhkin:
@dvrogozh , any concerns?

_studio/hevce_hw/libmfx_h265e_plugin.map
```diff
@@ -0,0 +1,6 @@
+LIBMFXSW_PLUGIN {
```
dvrogozh: new file? why

dmitrybrazhkin: We did not have existing .map file for HEVCE plug-in

dvrogozh: that's a bug in linux implementation then. Could you, please, fix within this series, please? otherwise we will forget.

onabiull: We don't need one. All plugins share same map file. Please use following code snippet:

`set(MFX_LDFLAGS "${MFX_ORIG_LDFLAGS} -Wl,--version-script=${MFX_ROOTDIR}/mfx_lib/plugin/libmfxsw_plugin.map" )`

dmitrybrazhkin: Done

android/build/mfx_defs_internal.mk
dvrogozh: -Wno-error=non-virtual-dtor still presents. If build is ok without it, need to remove.

android/build/mfx_stl.mk
```diff
@@ -0,0 +1,7 @@
+# Purpose:
+#   Defines STL support flags, include paths, etc.
+#
+# Defined variables:
+#   MFX_CFLAGS_STL - cflags for STL support
+
+MFX_CFLAGS_STL := -fexceptions -frtti
```
dvrogozh: Is this still needed?

dmitrybrazhkin: Yes. Build fails without these options

android/build/mfx_defs_internal.mk
dmitrybrazhkin: Build fails without this option

dvrogozh: Error look valid. Instead of adding this warning, need to add virtual destructors in the appropriate classes!

dmitrybrazhkin: The error occurs in CM headers, but we cannot modify these classes as they used in API and we can get wrong input. To minimize potential risk we can keep the warning for common code, but disable it for particular files

dvrogozh: Ok, let's try to disable warning for particular files.

dmitrybrazhkin: done

_studio/mfx_lib/cmrt_cross_platform/include/cmrt_cross_platform.h
```diff
@@ -49,6 +49,11 @@
 #endif
 #endif
 
+#if defined(ANDROID)
```
dvrogozh: I think this should be done both on Linux and Android, regardless. Moreover, on Linux we need to add -Wvirtual-dtor to force the check if it is not the default (I believe it is not, need to add here: builder/FindGlobals.cmake). In that way we will have aligned build between Linux/Android and avoid build regressions on Android. I believe such a change should be a separate preceding patch.

dmitrybrazhkin: done

builder/FindGlobals.cmake
```diff
@@ -92,8 +92,8 @@ if( Linux OR Darwin )
     set(no_warnings "-Wno-deprecated-declarations -Wno-unknown-pragmas -Wno-unused")
   endif()
 
-  set(CMAKE_C_FLAGS "-pipe -fPIC")
-  set(CMAKE_CXX_FLAGS "-pipe -fPIC")
+  set(CMAKE_C_FLAGS "-pipe -fPIC -Wnon-virtual-dtor")
```
dvrogozh: I agree that this is much more appropriate place for these flags than DEBUG/RELEASE flags below. However, could you, please, make 1 more enhancement? Add:
`set(warnings "-Wformat -Wformat-security -Wnon-virtual-dtor")`
right before, similar way we have 'no_warnings'. And eventually eliminate these flags below in DEBUG/RELEASE.

_studio/mfx_lib/cmrt_cross_platform/include/cmrt_cross_platform.h
dvrogozh: This apparently should be:
```
#if defined(__clang__)
  #pragma clang diagnostic push
  #pragma clang diagnostic ignored "-Wnon-virtual-dtor"
#elif defined(__GNUC__)
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wnon-virtual-dtor"
#endif
```
This will eliminate the need of ANDROID define. And will not introduce issues on Windows (in case we add this).

dmitrybrazhkin: done

builder/FindGlobals.cmake
dmitrybrazhkin: done

PR #93, opened 2018-03-05
[vpp] Enable RGB565 to YUY2 color conversion

rkazants:
Successfully run sample_vpp for validation on SKL.

### DIFF
--- a/_studio/shared/src/libmfx_allocator_vaapi.cpp
+++ b/_studio/shared/src/libmfx_allocator_vaapi.cpp
@@ -42,6 +42,8 @@
 #define VASurfaceAttribUsageHint 8
 #define VAEncMacroblockMapBufferType 29
 
+// TODO: remove this internal definition once it appears in VAAPI
+#define VA_FOURCC_R5G6B5 MFX_MAKEFOURCC('R','G','1','6')
 
 unsigned int ConvertMfxFourccToVAFormat(mfxU32 fourcc)
 {
@@ -53,6 +55,10 @@ unsigned int ConvertMfxFourccToVAFormat(mfxU32 fourcc)
         return VA_FOURCC_YUY2;
     case MFX_FOURCC_YV12:
         return VA_FOURCC_YV12;
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+    case MFX_FOURCC_RGB565:
+        return VA_FOURCC_R5G6B5;
+#endif
     case MFX_FOURCC_RGB4:
         return VA_FOURCC_ARGB;
     case MFX_FOURCC_BGR4:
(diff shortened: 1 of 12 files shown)


_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -42,6 +42,11 @@
 #define VASurfaceAttribUsageHint 8
 #define VAEncMacroblockMapBufferType 29
 
+// TODO: remove this internal definition once it appears in VAAPI
+#define VA_FOURCC(ch0, ch1, ch2, ch3) \
```
lakulako: MFX_MAKEFOURCC should be used. Isn't it the same?
What is the reason to introduce HERE local VA_FOURCC_R5G6B5? Isn't it defined elsewhere?

_studio/shared/src/libmfx_core.cpp
```diff
@@ -1354,6 +1354,19 @@ mfxStatus CoreDoSWFastCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc, int c
 
         break;
 
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+    case MFX_FOURCC_RGB565:
+        {
+            mfxU8* ptrSrc = MFX_MIN(MFX_MIN(pSrc->Data.R, pSrc->Data.G), pSrc->Data.B);
```
lakulako: That is why  R G B should be equal: to get rid of this unclear operations.

samples/sample_common/src/sysmem_allocator.cpp
```diff
@@ -130,6 +130,13 @@ mfxStatus SysMemFrameAllocator::LockFrame(mfxMemId mid, mfxFrameData *ptr)
         ptr->V = ptr->Y + 3;
         ptr->Pitch = 2 * Width2;
         break;
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+    case MFX_FOURCC_RGB565:
+        ptr->G = ptr->B;
+        ptr->R = ptr->B + 1;
```
lakulako: The same comments as above for the similar code below.

_studio/shared/src/libmfx_allocator_vaapi.cpp
rkazants: VA_FOURCC_R5G6B5 is not defined in VA API.

rkazants:
Fixed Leonid's comments

_studio/shared/src/libmfx_core.cpp
lakulako: const mfxU8* ptrSrc = pSrc->Data.B;
mfxU8* ptrDst = pDst->Data.B;

rkazants:
Fixed two Leonid's comments yet.

PR #104, opened 2018-03-21
[sample_hevc_fei] Fix MVP block size setting for PREENC

vshampor:


otsirule:
Confusing commit title. sample_hevc_fei doesn't support MV predictors for PreENC

PR #107, opened 2018-03-22
Update API to 1.26

onabiull:
Add MCTF, major refactoring of scene change detector, updated samples.


onabiull:
Hope this would be last BLOB-like update.

_studio/mfx_lib/plugin/CMakeLists.txt
```diff
@@ -51,7 +53,7 @@ foreach( enc ${vencs} )
   list( APPEND vdirs_hw encode_hw/${enc} )
 endforeach()
 
-list( APPEND vdirs_impl vpp cmrt_cross_platform genx/h264_encode genx/copy_kernels)
```
artem-shaporenko: so copy kernels are excluded now?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -1267,6 +1306,7 @@ mfxU8 MfxHwH264Encode::DetermineQueryMode(mfxVideoParam * in)
 /*
 Setting default value for LowPower option.
 By default LowPower is OFF (using DualPipe)
+For LKF: use LowPower by default i.e. if LowPower is Unknown then LowPower is ON
```
artem-shaporenko: ups

PR #110, opened 2018-03-23
[mpeg2e] Change macroblock_address calculation formula

onabiull:
This change should be merged along with respective driver change [pr87](https://github.com/intel/media-driver/pull/87)

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2018 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -910,8 +910,8 @@ mfxStatus VAAPIEncoder::FillSlices(ExecuteBuffers* pExecuteBuffers)
         ENCODE_SET_SLICE_HEADER_MPEG2&  ddiSlice = pExecuteBuffers->m_pSlice[i];
         assert(ddiSlice.NumMbsForSlice == width_in_mbs);
         sliceParam = &m_sliceParam[i];
-        //sliceParam->macroblock_address = i * width_in_mbs;
-        sliceParam->macroblock_address = i;
+        sliceParam->macroblock_address = i * width_in_mbs;
+        //sliceParam->macroblock_address = i;
```
dmitryermilov: it doesn't make sense to keep this line

PR #122, opened 2018-03-28
Fix default cmake configure build (cmake ..)

dvrogozh:
The goal is to be able to configure/build mediasdk with:
```
mkdir build && cd build
cmake ..
make
make install
```
This PR provides minimal required changes.


builder/FindGlobals.cmake
```diff
@@ -116,20 +115,20 @@ if( Linux OR Darwin )
...
+    append("-m32 -g" CMAKE_C_FLAGS)
+    append("-m32 -g" CMAKE_CXX_FLAGS)
+    append("-m32 -g" LINK_FLAGS)
   endif()
 
-  if(__ARCH MATCHES ia32)
-    link_directories(/usr/lib)
-  else()
+  if(CMAKE_SYSTEM_PROCESSOR MATCHES x86_64)
     link_directories(/usr/lib64)
```
alariono: Ubuntu use another one-library dir. Can we make differentiation between CentOS and Ubuntu here?

DavidMarec: Is this statement not what '$(LIBDIR)' stands for, running CMAKE ?

_studio/hevce_hw/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ else()
   set( hevc_version $ENV{MFX_HEVC_VERSION} )
 endif()
 
-if( __ARCH MATCHES intel64 )
+if(CMAKE_SYSTEM_PROCESSOR MATCHES x86_64)
```
DavidMarec: better use:
````CMAKE
if("${CMAKE_SIZEOF_VOID_P}" EQUAL "8")
````

To make sure **CMake** hits a 64bits architecture.

builder/FindGlobals.cmake
onabiull: This may conflict with @uartie change introduced by [pr19](https://github.com/Intel-Media-SDK/MediaSDK/pull/19)

DavidMarec: is there any need to precise the system libraries ?

_studio/hevce_hw/CMakeLists.txt
dvrogozh: How this is better and why? CMAKE_SYSTEM_PROCESSOR is a variable which can be overwritten in cross-compiling to make sure we build for the right architecture. CMAKE_SIZEOF_VOID_P corresponds to target architecture or to host architecture or to what?

builder/FindGlobals.cmake
dvrogozh: The current change focuses on removing __ARCH. So, we are having here a side discussion. Which is good, but will be addressed in a separate change. I also don't understand this place, it needs a review and rewrite. I will create and issue to continue discussion.

_studio/hevce_hw/CMakeLists.txt
DavidMarec: How to make sure that any of the "64 bits" archs  that may exist will match the expression "x86_64"  ? This is not so obvious.
CMAKE_SIZEOF_VOID_P is computed from a test-run; It's what the compiler told to cmake.

builder/FindGlobals.cmake
dvrogozh: Let's track this in #126.

PR #125, opened 2018-03-29
Disallow base class deletion for derivates

DavidMarec:
To avoid 'non-virtual-dtor' warning and possible mistakes, do not allow the deletion of  "CmDevice" derivate instances from its base class.
If this feature is required, remove the 'protected' keyword and set the destructor virtual.

### DIFF
--- a/_studio/mfx_lib/cmrt_cross_platform/include/cmrt_cross_platform.h
+++ b/_studio/mfx_lib/cmrt_cross_platform/include/cmrt_cross_platform.h
@@ -1534,6 +1534,8 @@ namespace CmLinux
 
         CM_RT_API virtual int32_t GetVISAVersion(uint32_t& majorVersion, uint32_t& minorVersion) = 0;
         //adding new functions in the bottom is a must
+protected: // do not allow  base deletion for derivates
+        ~CmDevice() {};
     };
 }
 #endif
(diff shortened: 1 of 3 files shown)


DavidMarec:
heck, there is too many of those virtual classes with unprotected destructor

dvrogozh:
Guys, here is the problem. These definitions should match definitions of cmrt library. If you will change table of virtual functions (add virtual destructor) we will have seg.faults. I totally agree that this is a proper way to address the issue, but if we want to have virtual destructors we need to change cmrt API first which is here: https://github.com/intel/media-driver/blob/master/cmrtlib/agnostic/share/cm_rt.h. @DavidMarec : could you try to suggest such a change to media-driver?

DavidMarec:
@dvrogozh 
By the fact, if you just enclose the destructor as "protected", there will be no signature changes and this will make the C++ compiler happy.
I can also address the issue to the media-driver team  too.

dvrogozh:
@DavidMarec : yep, I understand. I just commented from the "what I would like to see" perspective.

DavidMarec:
@dvrogozh  Looking deeper, I found that the only one that **should** be virtual is `SurfaceIndex`.
There are classes that provide a cast for it. 

Fortunately it does not trigger any error, while is has never been derived actually.

The classes from `cmvm.h` may also require virtual destructor; but these _rewrites_ of `basic_stream` classes and others std templates sound a bit weird .

PR #136, opened 2018-04-05
[hevce] Enabled LowPower mode on CNL

DenWolf:


_studio/hevce_hw/h265/include/mfx_h265_encode_vaapi.h
```diff
@@ -247,7 +247,7 @@ mfxStatus SetSkipFrame(
         virtual
         VAEntrypoint GetVAEntryPoint()
         {
-            return VAEntrypointEncSlice;
+            return IsOn(m_videoParam.mfx.LowPower) ? VAEntrypointEncSliceLP : VAEntrypointEncSlice;
```
onabiull: Did your check that this correctly handled on pre-CNL platforms?

DenWolf:
@onabiull , I will update this patch today

onabiull:
My original question was related to runtime behavior with API 1.27 or higher. Does HEVC Encoder return correct MFX_ERR_UNSUPPORTED on pre-CNL platforms with this change?

DenWolf:
This patch doesn't change behavior of this function on pre-CNL platforms - in this case GetVAEntryPoint() returns  VAEntrypointEncSlice as previously. Or I miss something?
So, if we cannot get MFX_ERR_UNSUPPORTED on pre-CNL platforms for HEVC VDEnc, I think this is an incorrect behavior and it need to be fixed by another patches.

PR #143, opened 2018-04-06
Fix API version handling during build process

dvrogozh:
This PR assumes that HEVC FEI should have been enabled at API=1.26. If that's not so, few changes in this commit should be extracted and landed, but change itself should appear later with appropriate API version for HEVC FEI.


dvrogozh:
Ok, I cleaned up API requirements for HEVC FEI in the code to MFX_VERSION_NEXT.

PR #146, opened 2018-04-10
Merge BDW/CFL/BXT configs into default one

dvrogozh:
After this change only 2 configs in mediasdk tree will be left: mfx_config.h and mfx_common_linux_bxtmin.h. The change should not modify behavior for any platform, well, at least I very much tried that to be so.


dvrogozh:
@onabiull : could you, please, review?

alariono:
ok for me

PR #156, opened 2018-04-17
sample_fei KW fixes

mgonchar:
Improved memory management for cmd parser; fixed compilation warnings

### DIFF
--- a/samples/sample_fei/src/sample_fei.cpp
+++ b/samples/sample_fei/src/sample_fei.cpp
@@ -1186,7 +1186,7 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, AppConfig* pCon
     return MFX_ERR_NONE;
 }
 
-mfxStatus ParseParFile(FILE *parFile, std::vector<AppConfig*> &config, msdk_char *parBuf, mfxU32 fileSize)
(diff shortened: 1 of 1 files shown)


samples/sample_fei/src/sample_fei.cpp
```diff
@@ -1244,13 +1243,18 @@ mfxStatus ParseParFile(FILE *parFile, std::vector<AppConfig*> &config, msdk_char
...
+
+        for(mfxU32 i = 0; i < argc; i++)
         {
             msdk_printf(MSDK_STRING("%s "), argv[i]);
         }
-        config.push_back(new AppConfig);
-        sts = ParseInputString(argv, (mfxU8)argc, config[configCount]);
+
+        std::unique_ptr<AppConfig> tmp(new AppConfig);
+        config.emplace_back(std::move(tmp));
```
knikolae: You should just pass new AppConfig to emplace_back to avoid local.

mgonchar: No, it's not exception safe way, see here http://kayari.org/cxx/antipatterns.html#emplace-new
If it so happens that container is full and reallocation is required, dynamic memory could be lost. I haven't found strict requirement for compiler to call new element constructor first before container resizing, so consider this is the safest way. Also it is possible to get rid of tmp and use rvalue directly at emplace_back but that would overcomplicate the code line. Maybe make_unique is the best option here, but I'm not sure that we support c++14 features

PR #169, opened 2018-04-22
[vp8] macroblock_offset alignment

dmitryermilov:
The change depends on a corresponding driver fix.


dmitryermilov:
@onabiull, @XinfengZhang,
This should be merged together with https://github.com/intel/media-driver/pull/86

PR #171, opened 2018-04-24
[codeowners] Add reviewers for /api/include/

adydychk:
Reviewers added by request of Oleg Nabiullin.


adydychk:
I added all existing reviewers, will invite Solodkov and Leonenko.

PR #178, opened 2018-04-25
[vpp] fix regression caused by new ADI_SCD interface

snowpiel:
Regression from 20/44 pass rate to 8/44 pass rate.

Old interface provides new and last information with one field delay.
New interface provides new information on current field.
New interface has been used with external BRC.

There is no delay in new SCD algorithm.
SCD decision tree needs to be retrained to detect scene change with one field duration.
ADI_SCD state machine needs to be updated.

New passrate: 20/44

Issue: MDP-40579
Test: vpp_sc_di, beh_avce_extbrc, beh_avce_mbext_brc


_studio/shared/asc/src/asc.cpp
```diff
@@ -1366,8 +1367,8 @@ mfxStatus ASC::RunFrame(SurfaceIndex *idxFrom, mfxU32 parity) {
     res = m_queue->DestroyEvent(e);
     SCD_CHECK_CM_ERR(res, MFX_ERR_DEVICE_FAILED);
 
-    AscFrameAnalysis();
 
+    AscFrameAnalysis();
```
dvrogozh: You don't modify anything in this file actually. Please, remove the file from the review.

snowpiel:
I have pushed the following changes as suggested by Dima:
- Change decision tree file license provided by Sebastian with the one from original file in master
- remove ^M in tree.cpp
- revert _studio/shared/asc/src/asc.cpp to what was on master

PR #181, opened 2018-04-26
[vpp] MCTF behavior test fail on Linux

snowpiel:
Open source code assumes VPP is only implemented on HW.
MCTF is a SW/kernel VPP feature.
This results in an error in QueryIOSurf().

Resolution: copied fix from close source code to open source.
Fix improves pass rate of beh_vpp_mctf_douse from 0/4 to 3/4.

Issue: MDP-40851
Test: beh_vpp_mctf_douse


_studio/shared/include/mfx_config.h
```diff
@@ -166,6 +166,10 @@
         #undef MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE
         #undef MFX_ENABLE_VP8_VIDEO_DECODE_HW
     #endif // #if defined(AS_HEVCD_PLUGIN)
+    #if defined(AS_CAMERA_PLUGIN)
```
dvrogozh: I don't think we have camera plugin in open source. If so, these change in config is useless. We don't build it and thus we can't support it. Please, drop the change.

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
```diff
@@ -426,8 +426,30 @@ mfxStatus VideoVPPBase::QueryIOSurf(VideoCORE* core, mfxVideoParam *par, mfxFram
 
         mfxSts = CheckIOPattern_AndSetIOMemTypes(par->IOPattern, &(request[VPP_IN].Type), &(request[VPP_OUT].Type), bSWLib);
         MFX_CHECK_STS(mfxSts);
+#if defined (MFX_ENABLE_HW_ONLY_VPP)
         return (bSWLib)? MFX_ERR_UNSUPPORTED : MFX_ERR_NONE;
+#else
+        return (bSWLib)? MFX_WRN_PARTIAL_ACCELERATION : MFX_ERR_NONE;
     }
+    else
+    {
```
dvrogozh: do you really need this 'else' part? and if yes - why?

snowpiel: The else if for the case where core->GetPlatformType() does not return  MFX_PLATFORM_HARDWARE.  
I suspect that the wrong piece of code (corresponding to camera plugin) had been copied to the open source code.  It assumes that only HW VPP is supported.  which is not true for MCTF.
I I'll remove the code corresponding to camera plugin as it is not in open source.

dvrogozh: Yes, this looks to be so. From the scripts stripping closed source version:
```
# fgrep -rsn AS_CAMERA_PLUGIN  builder/
builder/open_source_symbols_undefine.list:16:AS_CAMERA_PLUGIN
# fgrep -rsn MFX_ENABLE_HW_ONLY_VPP builder/
builder/open_source_symbols_define.list:6:MFX_ENABLE_HW_ONLY_VPP
```
@onabiull: will a strip script left both macro paths for MFX_ENABLE_HW_ONLY_VPP or not under the above conditions?

I don't think that you use this 'else' path for MCTF... Actually I am not sure whether we have any use cases for that... Oleg, how do you think - do we need this 'else' at all in open source?

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
```diff
@@ -426,8 +426,27 @@ mfxStatus VideoVPPBase::QueryIOSurf(VideoCORE* core, mfxVideoParam *par, mfxFram
 
         mfxSts = CheckIOPattern_AndSetIOMemTypes(par->IOPattern, &(request[VPP_IN].Type), &(request[VPP_OUT].Type), bSWLib);
         MFX_CHECK_STS(mfxSts);
-        return (bSWLib)? MFX_ERR_UNSUPPORTED : MFX_ERR_NONE;
+        return (bSWLib)? MFX_WRN_PARTIAL_ACCELERATION : MFX_ERR_NONE;
     }
```
onabiull: We have neither SW lib nor SW fallbacks in open source, so unsupported should be returned here.

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
```diff
@@ -426,8 +426,27 @@ mfxStatus VideoVPPBase::QueryIOSurf(VideoCORE* core, mfxVideoParam *par, mfxFram
...
+    else
+    {
+        // since MSDK_3.0 asyncDepth is mandatory
```
onabiull: Looks like we should replace this whole else block with `return MFX_ERR_UNSUPORTED;`

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
snowpiel: MCTF is not HW implementation so bSWLib is set to true.
In close source, when camera plugin is not requested we have:
return (bSWLib)? MFX_WRN_PARTIAL_ACCELERATION : MFX_ERR_NONE; 

Are you suggesting that we do not follow the logic of close source to support MCTF ?
Feel free to provide your own suggestion.

PR #184, opened 2018-04-27
[HEVC FEI ENCODE] Add repackctrl & repackstat supports

jsunintel:
Signed-off-by: Jing SUN <jing.a.sun@intel.com>

### DIFF
--- a/_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_hw.h
+++ b/_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_hw.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2018 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -118,12 +118,24 @@ protected:
     {
         mfxExtFeiHevcEncFrameCtrl* EncFrameCtrl = reinterpret_cast<mfxExtFeiHevcEncFrameCtrl*>(GetBufById(task.m_ctrl, MFX_EXTBUFF_HEVCFEI_ENC_CTRL));
 
-        if (EncFrameCtrl && (!!EncFrameCtrl->PerCuQp != !!m_vpar.m_pps.cu_qp_delta_enabled_flag))
-        {
-            SoftReset(task);
-        }
+        mfxExtFeiHevcRepackCtrl* repackctrl = reinterpret_cast<mfxExtFeiHevcRepackCtrl*>(GetBufById(task.m_ctrl, MFX_EXTBUFF_HEVCFEI_REPACK_CTRL));
 
-        return;
+        if (m_vpar.m_pps.cu_qp_delta_enabled_flag)
+        {
+            if (!repackctrl && !EncFrameCtrl->PerCuQp)
+            {
+                // repackctrl or PerCuQp is disabled, so insert PPS and turn the flag off.
+                SoftReset(task);
+            }
+        }
+        else
+        {
+            if (repackctrl || EncFrameCtrl->PerCuQp)
+            {
+                // repackctrl or PerCuQp is enabled, so insert PPS and turn the flag on.
+                SoftReset(task);
+            }
+        }
     }
 
     void SoftReset(MfxHwH265Encode::Task& task)
--- a/samples/sample_hevc_fei/include/hevc_fei_encode.h
+++ b/samples/sample_hevc_fei/include/hevc_fei_encode.h
@@ -1,5 +1,5 @@
 /******************************************************************************\
-Copyright (c) 2017, Intel Corporation
+Copyright (c) 2017-2018, Intel Corporation
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
@@ -29,9 +29,11 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 class FEI_Encode
 {
 public:
-    FEI_Encode(MFXVideoSession* session, mfxHDL hdl, MfxVideoParamsWrapper& encode_pars,
-        const mfxExtFeiHevcEncFrameCtrl& def_ctrl, const msdk_char* dst_output,
-        const msdk_char* mvpInFile, PredictorsRepaking* rpck);
+    FEI_Encode(MFXVideoSession* session, mfxHDL hdl,
+        MfxVideoParamsWrapper& encode_pars, const mfxExtFeiHevcEncFrameCtrl& encodeCtrl,
+        const msdk_char* strDstFile, const msdk_char* mvpInFile,
+        const msdk_char* repackctrlFile, const msdk_char* repackstatFile,
+        PredictorsRepaking* repacker);
 
     ~FEI_Encode();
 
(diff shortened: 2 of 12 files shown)


jsunintel:
Please help me review this and thanks a lot!

samples/sample_hevc_fei/include/hevc_fei_encode.h
```diff
@@ -29,9 +29,11 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 class FEI_Encode
 {
 public:
-    FEI_Encode(MFXVideoSession* session, mfxHDL hdl, MfxVideoParamsWrapper& encode_pars,
-        const mfxExtFeiHevcEncFrameCtrl& def_ctrl, const msdk_char* dst_output,
-        const msdk_char* mvpInFile, PredictorsRepaking* rpck);
+    FEI_Encode(MFXVideoSession* session, mfxHDL hdl,
+        MfxVideoParamsWrapper& encode_pars, const mfxExtFeiHevcEncFrameCtrl& encodeCtrl,
+        const msdk_char* strDstFile, const msdk_char* mvpInFile,
```
otsirule: I think we need refactoring here for input/output files, but it's out of scope for this commit.

samples/sample_hevc_fei/include/sample_hevc_fei_defs.h
```diff
@@ -30,6 +30,8 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 
 #define CHECK_STS_AND_RETURN(X, MSG, RET) {if ((X) < MFX_ERR_NONE) {MSDK_PRINT_RET_MSG(X, MSG); return RET;}}
 
+#define MSDK_ZERO_ARRAY(VAR, NUM) {memset(VAR, 0, sizeof(VAR[0])*NUM);}
```
otsirule: Please use template Zero

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
```diff
@@ -197,6 +229,45 @@ mfxStatus FEI_Encode::Reset(mfxVideoParam& par)
...
+    MSDK_CHECK_STATUS(sts, "FEI Encode Read repack ctrl MaxFrameSize failed");
+
+    sts = m_pFile_repack_ctrl->Read(&(pRepackCtrl->NumPasses), sizeof(mfxU32), 1);
+    MSDK_CHECK_STATUS(sts, "FEI Encode Read repack ctrl NumPasses failed");
+
+    sts = m_pFile_repack_ctrl->Read(pRepackCtrl->DeltaQP, sizeof(mfxU8), 8);
+    MSDK_CHECK_STATUS(sts, "FEI Encode Read repack ctrl DeltaQP failed");
+
+    if (pRepackCtrl->NumPasses > 8)
+        msdk_printf(MSDK_STRING("WARNING: HEVC NumPasses should be less than or equal to 8\n"));
```
otsirule: Why is it warning, but not error?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
```diff
@@ -238,6 +315,12 @@ mfxStatus FEI_Encode::EncodeFrame(mfxFrameSurface1* pSurf)
         if (MFX_ERR_NONE <= sts && m_syncPoint) // ignore warnings if output is available
         {
             sts = SyncOperation();
+
+            if ((MFX_ERR_NONE == sts) && m_pFile_repack_stat.get())
```
otsirule: Maybe it's better to move this block in SyncOperation wrapper?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
```diff
@@ -197,6 +229,45 @@ mfxStatus FEI_Encode::Reset(mfxVideoParam& par)
...
+    mfxI8 repackStatChar[64];
+
+    mfxExtFeiHevcRepackStat *pRepackStat = m_encodeCtrl.GetExtBuffer<mfxExtFeiHevcRepackStat>();
+    MSDK_CHECK_POINTER(pRepackStat, MFX_ERR_NOT_INITIALIZED);
+
+    snprintf(repackStatChar, sizeof(repackStatChar),
+             "%d NumPasses\n", pRepackStat->NumPasses);
+    sts = m_pFile_repack_stat->Write(repackStatChar, strlen(repackStatChar), 1);
+
+    pRepackStat->NumPasses = 0;
```
otsirule: Why should it be zeroed after write operation?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
```diff
@@ -223,6 +294,12 @@ mfxStatus FEI_Encode::EncodeFrame(mfxFrameSurface1* pSurf)
     for (;;)
     {
 
+        if (m_pFile_repack_ctrl.get() && pSurf)
+        {
```
otsirule: All ctrl parameters are set in SetCtrlParams function. Why is this block here? I see that it's implemented to work only in Encoded Order.

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
jsunintel: I am modifying it.

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
jsunintel: I prefer to keep this block in EncodeFrame, because in this way it is in the same function as SetRepackCtrl is, to let code viewer get the SetRepackCtrl/GetRepackStat relationship easier. How do you think about it?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
jsunintel: It is reset for the next frame.

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
jsunintel: For the flushed frame encode request, don't need to do SetRepackCtrl again. Yes, the repack ctrl tests only work in encoded order, for the repack ctrl data is always in encoded order. AVC works in this way as well.

samples/sample_hevc_fei/include/sample_hevc_fei_defs.h
jsunintel: Did you mean that I should use MSDK_ZERO_MEMORY to replace MSDK_ZERO_ARRAY? There are many MSDK_ZERO_MEMORY called in this file.

_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_hw.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2018 Intel Corporation
```
lakulako: should become 2017-2018

samples/sample_hevc_fei/include/sample_hevc_fei_defs.h
mgonchar: no, I think Olga meant template function Zero, but I don't see any sense in using it. I also don't like such macro (I know that we used them in previous samples, but that isn't really good way to do zeroing data). I suggest to use std::fill for initialization and std::copy for copying and not to use macro

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
otsirule: Can we add comment here?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
otsirule: It makes sense. But I'm afraid it will be a too big function if we always write here all blocks which dump output files.  Let's leave as is for now, but it's better to refactor and move sync operation here and move dumping all output files in separate logical block (i.e. function)

PR #186, opened 2018-04-27
[H264e HW] CFL Support, fix ASCVidSample init, MFX_VERSION >= 1026

spossos:
Added support for CFL in ASC::InitGPUsurf(), it was causing
mfx_err_device_failed when ran on a CFL platform.
Correct initialization of ASCVidSample in ASC::Init, which in case
of failure during initialization of GPU surfaces in InitGPUsurf()
will make Init to exit without initializing m_videoData array,
which then cause the system to crash at ASC::Close() because it
will identify m_videoData array to hold info (not NULL) and then
try to deallocate unallocated memory.
BRC LTR code now Wrapped in MFX_VERSION >= 1026, replaces
(MFX_VERSION >= MFX_VERSION_NEXT)

Issue: MDP-41086, MDP-40692
Test: h264e_extbrc_adaptive_ltr

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
+++ b/_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
@@ -1268,7 +1268,7 @@ namespace MfxHwH264Encode
         par.EncodedOrder = task->m_encOrder;
         par.PyramidLayer = (mfxU16)task->m_loc.level;
         par.NumRecode = (mfxU16)task->m_repack;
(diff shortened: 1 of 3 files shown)


tests/test_results.json
```diff
@@ -0,0 +1,568 @@
+{
```
dvrogozh: What is this file and where it is being used? Isn't that an accidental commit?

spossos: I may have misunderstood the "How to contribute" instructions, there in the last step (6) it says "Attach generated test_results.json to your pull request". Is this only intended for when you have not bit exact results?

PR #188, opened 2018-05-03
Fixes for compilation warnings on gcc 7.3

dmitryermilov:
The change addresses the following types of warnings treated as errors:
* this ‘if’ clause does not guard this statement
* ‘*’ in boolean context
* dereferencing type-punned pointer will break strict-aliasing rules


_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -1034,33 +1034,31 @@ mfxStatus VAAPIEncoder::FillUserDataBuffer(mfxU8 *pUserData, mfxU32 userDataLen)
 
 mfxStatus VAAPIEncoder::FillVideoSignalInfoBuffer(ExecuteBuffers* pExecuteBuffers)
 {
-    VAEncMiscParameterExtensionDataSeqDisplayMPEG2 &
-        miscSeqInfo  = (VAEncMiscParameterExtensionDataSeqDisplayMPEG2 &)m_pMiscParamsSeqInfo->data;
-
-    VAStatus                      vaSts;
+    VAEncMiscParameterExtensionDataSeqDisplayMPEG2* miscSeqInfo  =
```
onabiull: Indirection level lost here

PR #191, opened 2018-05-07
Samples bugfixes and features - 2

fzhar:


samples/sample_common/include/current_date.h
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************\
-Copyright (c) 2005-2017, Intel Corporation
+Copyright (c) 2005-2018, Intel Corporation
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
```
dvrogozh: Do we still use '#define PRODUCT_COPYRIGHT' in this file below? If yes - please, update it from 2014 to 2018. If we don't use it - delete.

samples/sample_common/include/sample_defs.h
```diff
@@ -45,6 +45,27 @@ enum {MCTF_BITRATE_MULTIPLIER = 100000};
...
+
+enum {
+    MFX_HANDLE_DEVICEWINDOW  = 0x101 /* A handle to the render window */
+}; //mfxHandleType
+
+#ifndef D3D_SURFACES_SUPPORT
+#define D3D_SURFACES_SUPPORT 1
+#endif
+
+#if defined(_WIN32) && !defined(MFX_D3D11_SUPPORT)
```
dvrogozh: 1. Do we use WIN32 or _WIN32? This is inconsistent with line 49.
2. Probably MFX_D3D11_SUPPORT should be enough by itself here? We are already covered by WIN32/64 above and we define D3D11_SUPPORT only for windows...

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1571,6 +1571,13 @@ mfxStatus CDecodingPipeline::SyncOutputSurface(mfxU32 wait)
 
     mfxStatus sts = m_mfxSession.SyncOperation(m_pCurrentOutputSurface->syncp, wait);
 
+    if (MFX_ERR_GPU_HANG == sts) {
+        msdk_printf(MSDK_STRING("GPU hang happened\n"));
+        // Output surface can be corrupted
+        // But should be delivered to output anyway
+        sts = MFX_ERR_NONE;
```
dvrogozh: Will/how sample_decode stop working after this error? Printing message is cool, but error handling looks strange to me.

samples/sample_multi_transcode/include/pipeline_transcode.h
```diff
@@ -131,6 +139,10 @@ namespace TranscodingSample
     struct sMctfRunTimeParam
     {
 #ifdef ENABLE_MCTF_EXT
+#if 0
```
dvrogozh: Why do we have #if 0 here? We already have define right above.

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -19,6 +19,9 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 
 #include "mfx_samples_config.h"
 
+#if defined(_WIN32) || defined(_WIN64)
+#include <windows.h>
```
dvrogozh: Strange placement. Could we move it somewhere to sample_common?

samples/sample_plugins/rotate_opencl/include/opencl_filter.h
```diff
@@ -26,7 +26,12 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "mfxvideo++.h"
 #include "logger.h"
 
+#if !defined(_WIN32) && !defined(_WIN64)
```
dvrogozh: We can use such construction if we have some code for Linux only, but coupled with else this looks terrible. Can you, please, use "#if defined(_WIN32) || defined(_WIN64)" here?

Also, can you clarify why we need to differentiate here between Windows/Linux?

samples/sample_common/include/current_date.h
fzhar: This is standard header for all samples code. I don't see any PRODUCT_COPYRIGHT in this change

samples/sample_common/include/sample_defs.h
fzhar: This is synchronization patch and it will include windows parts as well. After full migration of samples to open source windows-related parts of the code will be uploaded into GitHub. So, no need to remove them now

samples/sample_decode/src/pipeline_decode.cpp
fzhar: This was the requirements from library team. By default library and samples will use "soft recovery" - so nothing should be done in case of GPU hang (library will recover by itself), just print error message notifying that GPU hang occurred and several frames might be missing.

samples/sample_multi_transcode/include/pipeline_transcode.h
fzhar: Will be removed.

samples/sample_multi_transcode/src/pipeline_transcode.cpp
fzhar: Nope. It is used for particular purpose in this paticular file, no need to move it to common files

samples/sample_common/include/current_date.h
dvrogozh: @fzhar : It is not changed in your change and that's a problem. Look here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/0eeeccf96cd0f7f867a17152c8a44cc6a0893a08/samples/sample_common/include/current_date.h#L25

samples/sample_multi_transcode/include/pipeline_transcode.h
```diff
@@ -138,12 +138,6 @@ namespace TranscodingSample
     // that can be changed in run-time;
     struct sMctfRunTimeParam
     {
-#ifdef ENABLE_MCTF_EXT
```
dvrogozh: This should not be a separate change, but squashed.

samples/sample_plugins/rotate_opencl/include/opencl_filter.h
dvrogozh: Use _normal_ layout here, please. What you have provides useless difficulty.

fzhar: seems *normal* for me

PR #198, opened 2018-05-12
[H264e HW] KW hits in Opensource

spossos:
Fix for issues reported by KlockWork for PV5 release.

Issue: MDP-42116


_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -759,9 +759,6 @@ mfxStatus ImplementationAvc::Init(mfxVideoParam * par)
         return MFX_WRN_PARTIAL_ACCELERATION;
 
     GUID guid;
-#if defined(MFX_ENABLE_MFE)
```
onabiull: @artem-shaporenko can you confirm that  this is safe change?

artem-shaporenko: It's ok for now.

PR #201, opened 2018-05-14
[HEVCe FEI] Add external BRC control

vmaksimo:


samples/sample_hevc_fei/include/sample_hevc_fei_defs.h
```diff
@@ -225,8 +230,8 @@ template<>struct mfx_ext_buffer_id<mfxExtHEVCRefLists>{
 template<>struct mfx_ext_buffer_id<mfxExtFeiHevcRepackCtrl>{
     enum {id = MFX_EXTBUFF_HEVCFEI_REPACK_CTRL};
 };
-template<>struct mfx_ext_buffer_id<mfxExtFeiHevcRepackStat>{
```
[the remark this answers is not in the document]
vmaksimo: Oops, merge problems. Will fix

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -701,6 +712,11 @@ void AdjustOptions(sInputParams& params)
         msdk_printf(MSDK_STRING("WARNING: Encoded order is enabled by force in repackctrl.\n"));
         params.bEncodedOrder = true;
     }
+
+    if (!params.bExtBRC && params.TargetKbps) {
+        msdk_printf(MSDK_STRING("WARNING: Target bitrate is ignored as external BRC is disabled\n"));
+        params.TargetKbps = 0;
+    }
```
otsirule: What if I set ExtBRC and QP?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
```diff
@@ -63,6 +67,7 @@ FEI_Encode::~FEI_Encode()
         {
             m_buf_allocator.Free(pQP);
         }
+        HEVCExtBRC::Destroy(m_ExtBRC);
```
otsirule: Why create called in pipeline but destroy in encode?

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
vmaksimo: Not sure. Is it OK to return with error, if they are set together?

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
vmaksimo: Create is called in pipeline because it depends on input params, when Destroy doesn't. 
Also, destroyable pointer is stored in encode, not in pipeline.

Do you think it's better to move destroy step to pipeline destructor?

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
otsirule: Yep, because you can't read users mind

samples/sample_hevc_fei/src/hevc_fei_encode.cpp
otsirule: I think that Create should be called in Encode because it's Encode's attribute, not pipeline

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -457,7 +457,6 @@ MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFr
     // default settings
     pars.mfx.RateControlMethod = MFX_RATECONTROL_CQP;
```
otsirule: It looks strange that RateControlMethod is CQP when TargetKbps != 0

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -472,6 +471,7 @@ MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFr
     pars.mfx.NumRefFrame  = user_pars.nNumRef;
     pars.mfx.NumSlice     = user_pars.nNumSlices;
     pars.mfx.EncodedOrder = user_pars.bEncodedOrder;
+    pars.mfx.TargetKbps   = user_pars.TargetKbps;
```
otsirule: It looks really strange that RateControlMethod is still CQP when TargetKbps can be non-zero

vmaksimo: TargetKbps is set to 0 if ExtBRC is not specified.

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -701,6 +717,16 @@ void AdjustOptions(sInputParams& params)
         msdk_printf(MSDK_STRING("WARNING: Encoded order is enabled by force in repackctrl.\n"));
         params.bEncodedOrder = true;
     }
+
+    if (!params.bExtBRC && params.TargetKbps) {
+        msdk_printf(MSDK_STRING("WARNING: Target bitrate is ignored as external BRC is disabled\n"));
+        params.TargetKbps = 0;
```
vmaksimo: TargetKbps is set to 0, so it's not possible to have RateControl =CQP and non-zero TargetKbps at the same time.

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
otsirule: It's no obvious in this code

vmaksimo: agree. not sure how to do it more clear and to minimize the number of different checking

PR #204, opened 2018-05-15
[HEVCe FEI] Use mfxExtHEVCParam to pass width/height to encoder

vmaksimo:
When height and/or width divided with 8 but not divided with 16,
it's recommended to add extended buffer to increase performance.

### DIFF
--- a/samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
+++ b/samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
@@ -498,6 +498,17 @@ MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFr
     // qp offset per pyramid layer, default is library behavior
     pCO3->EnableQPOffset = user_pars.bDisableQPOffset ? MFX_CODINGOPTION_OFF : MFX_CODINGOPTION_UNKNOWN;
 
+    // When height and/or width divided with 8 but not divided with 16
+    // add extended parameter to increase performance
+    if ((pars.mfx.FrameInfo.CropW & 15) == 8 || (pars.mfx.FrameInfo.CropH & 15) == 8) {
+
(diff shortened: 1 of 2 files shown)


samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -498,6 +498,17 @@ MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFr
...
     pCO3->EnableQPOffset = user_pars.bDisableQPOffset ? MFX_CODINGOPTION_OFF : MFX_CODINGOPTION_UNKNOWN;
 
+    // When height and/or width divided with 8 but not divided with 16
+    // add extended parameter to increase performance
+    if ((pars.mfx.FrameInfo.CropW & 15) == 8 || (pars.mfx.FrameInfo.CropH & 15) == 8) {
+
+        mfxExtHEVCParam* pHP = pars.AddExtBuffer<mfxExtHEVCParam>();
+        if (!pHP) throw mfxError(MFX_ERR_NOT_INITIALIZED, "Failed to attach mfxExtHEVCParam");
+
+        pHP->PicWidthInLumaSamples  = pars.mfx.FrameInfo.CropW;
```
otsirule: From library code I see that these options will be aligned by 16, so I don't understand why we set them and how it should have the impact on performance

mgonchar: Yes, this is good note.
1) This buffer is not related to performance optimization, this buffer just sends real size of "coding window", but crops in surface set us "display window".
2) You fairly mention that alignment now makes no difference between these values and crops in surfaces, but for future platforms picWidthInLumaSamples (and height) will have a 8 pixels alignment

So, the if statement is useless and the comment should be changed to express the real meaning of passing this values

adydychk:
@vmaksimo Please make a rebase on top of master. Android CI will be fixed.

PR #206, opened 2018-05-16
[HEVCe FEI] Update checking for minimum search window size

vmaksimo:
Minimum supported custom Ref Window (RefWidth x RefHeight) is 20x20, according to driver limitations


samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -112,9 +112,9 @@ void PrintHelp(const msdk_char *strAppName, const msdk_char *strErrorMessage)
     msdk_printf(MSDK_STRING("   [-qrep] - quality predictor MV repacking before encode\n"));
     msdk_printf(MSDK_STRING("   [-SearchWindow value] - specifies one of the predefined search path and window size. In range [1,5] (5 is default).\n"));
     msdk_printf(MSDK_STRING("                           If zero value specified: -RefWidth / RefHeight, -LenSP are required\n"));
-    msdk_printf(MSDK_STRING("   [-RefWidth width] - width of search region (should be multiple of 4), maximum allowed search window is 64x32 for\n"));
-    msdk_printf(MSDK_STRING("                       one direction and 32x32 for bidirectional search\n"));
-    msdk_printf(MSDK_STRING("   [-RefHeight height] - height of search region (should be multiple of 4), maximum allowed is 32\n"));
+    msdk_printf(MSDK_STRING("   [-RefWidth width] - width of search region (should be multiple of 4),\n"));
+    msdk_printf(MSDK_STRING("                       valid range is [20, 64] for one direction and [20,32] for bidirectional search\n"));
+    msdk_printf(MSDK_STRING("   [-RefHeight height] - height of search region (should be multiple of 4), valid range is [20, 32]\n"));
```
mgonchar: that is not exactly true. the limitation for one/bi directional search is on overall search region area, so correct restrictions are: both should be multiply of 4, valid range for each is [20, 64], overall area size limit 2048 for one directional search and 1024 for bi-directional.

vmaksimo: According to the library code, we have a certain restriction on RefWidth/Height for 32 as maximum value for B-frames.
So, it's not possible to set, for example, 44x20 window for bidirectional search, although it would be less than 1024.

mgonchar: The problem is for P frames. Limits for w and h are [20, 64], but 64*64 > 2048

PR #225, opened 2018-05-28
[AVC FEI][sample] adding option "-recon" to dump recon surfaces

djie1:


samples/sample_fei/src/fei_encpak.cpp
```diff
@@ -223,6 +224,8 @@ FEI_EncPakInterface::~FEI_EncPakInterface()
 
     m_pAppConfig->PipelineCfg.pEncVideoParam = NULL;
     m_pAppConfig->PipelineCfg.pPakVideoParam = NULL;
+
+    m_pMFXAllocator = NULL;
```
mgonchar: No need of that

samples/sample_fei/src/fei_encpak.cpp
```diff
@@ -1254,6 +1282,12 @@ mfxStatus FEI_EncPakInterface::ResetState()
         // reset FileWriter
         sts = m_FileWriter.Init(m_pAppConfig->dstFileBuff[0]);
         MSDK_CHECK_STATUS(sts, "FEI ENCPAK: FileWriter.Init failed");
+
+        if (m_pAppConfig->reconFile)
+        {
+            sts = m_ReconWriter.Init(m_pAppConfig->reconFile, 0);
```
mgonchar: At line 654 you used 1 as number of views. As I see from the code, since it is not multiview mode, second parameter doesn't affect anything. But better to be consistent and use same value across similar calls

samples/sample_fei/src/fei_encpak.cpp
djie1: Removed.

samples/sample_fei/src/fei_encpak.cpp
djie1: Done.

PR #226, opened 2018-05-28
Android make files clean-up + refactoring

dmitrybrazhkin:


adydychk:
@dmitrybrazhkin Commit was re-built current status is success as you see. The previous fail and long "pseudo in progress" status were due to some network issues.

PR #228, opened 2018-05-29
[mpeg2d] Add check for headers corruption

lakulako:
When picture data is extracted from bs, verifies for proper picture
type, structure and temporal order using current and next picture headers.
If corrupted picture is detected it is dropped and the next picture
will be extracted.


lakulako:
Not ready yet.

lakulako:
Now ready.

PR #229, opened 2018-05-29
Variable default initialization to avoid gcc warnings

dmitryermilov:


adydychk:
Android failed in this commit due to infrastructure reasons. Build in master is in status success.

PR #247, opened 2018-06-01
[mfe, fei] fix resource leak

artem-shaporenko:
fixed resource leak in sample_fei due to wrong order of joined sessions destruction

### DIFF
--- a/samples/sample_fei/src/sample_fei.cpp
+++ b/samples/sample_fei/src/sample_fei.cpp
@@ -1447,16 +1447,16 @@ int main(int argc, char *argv[])
         RunPipeline(pipelines[0].get(), 0);
     }
 
-    for(std::vector<std::thread>::iterator t = pipe_threads.begin(); t != pipe_threads.end(); t++)
+    for(auto t = pipe_threads.begin(); t != pipe_threads.end(); t++)
     {
         t->join();
     }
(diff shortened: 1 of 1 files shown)


samples/sample_fei/src/sample_fei.cpp
```diff
@@ -1454,7 +1454,7 @@ int main(int argc, char *argv[])
     msdk_printf(MSDK_STRING("\nProcessing finished after %.2f sec \n"), MSDK_GET_TIME(msdk_time_get_tick(), startTime, frequency));
     pipe_threads.clear();
 
-    for (auto & pipeline : pipelines)
+    for(std::vector<std::unique_ptr<CEncodingPipeline>>::reverse_iterator pPipeline = pipelines.rbegin(); pPipeline != pipelines.rend(); pPipeline++)
```
mgonchar: you can use "auto pPipeline" to simplify things. But maybe iPipeline is better, since it is not a pointer

samples/sample_fei/src/sample_fei.cpp
```diff
@@ -1454,7 +1454,7 @@ int main(int argc, char *argv[])
...
     {
         pipeline->Close();
```
mgonchar: this line should be changed as well

PR #252, opened 2018-06-01
[h264d] Prevent crop info to be printed in sample_decode

lakulako:
During init h264 decoder wiped crop info in provided params structure.
Added code to keep and verify the info.
Now sample_decode prints proper Crops.
(output color format to be corrected in sample)

Issue: MDP-42346
Test: sample_decode with cropped h264 streams


adydychk:
@lakulako Make a rebase with the latest master branch please. The fail of Android is because of broken commit form master.

Issue #220, opened 2018-05-24
Remove perl wrapper for compiling


chivakker:
Can MediaSDK remove the perl wrapper and stick to cmake as the compiler?  This can help to standardize building with dependencies like media-driver and gmmlib.

onabiull:
Hello,

We gradually improve cmake defaults to make such scenario viable. Actually on my configuration default cmake build scenario 
```cmake 
   mkdir build && cd build
   cmake ../
   make
   sudo make install
```
already work as expected. There are some work in progress to make X11 headers optional for samples and probably relax other restrictions. After all perl wrapper just set a number of cmake options.

dvrogozh:
@onabiull : can we switch github build scenario which we trigger during each commit check from perl to what you have described?

PR #258, opened 2018-06-04
Document direct cmake configure build in README

dvrogozh:
Fixes: #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,13 @@ You can find answers for the most frequently asked questions [here](https://soft
 
   * [License](#license)
   * [How to contribute](#how-to-contribute)
-  * [Documentation](#Documentation)
+  * [Documentation](#documentation)
   * [Products which use Media SDK](#products-which-use-media-sdk)
   * [System requirements](#system-requirements)
   * [How to build](#how-to-build)
-    * [Requirements](#requirements)
     * [Build steps](#build-steps)
     * [Enabling Instrumentation and Tracing Technology](#enabling-instrumentation-and-tracing-technology)
+  * [Recommendations](#recommendations)
   * [See also](#see-also)
 
 # License
@@ -32,6 +32,8 @@ Intel Media SDK is licensed under MIT license. See [LICENSE](./LICENSE) for deta
 See [CONTRIBUTING](./CONTRIBUTING.md) for details. Thank you!
 
 # Documentation
+
+To get copy of Media SDK documentation use Git* with [LFS](https://git-lfs.github.com/) support.
 
 Please find full documentation under the `doc/` folder. Key documents:
 * [Media SDK Developer Reference](./doc/mediasdk-man.pdf)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,34 +62,30 @@ Media SDK test and sample applications may require additional software packages
 # How to build
 ## Requirements
  - Git* (with [LFS](https://git-lfs.github.com/) support)
- - Perl* v5.16+
  - Cmake* v2.8+
  - GCC* v4.8+
```
onabiull: We switched to 6.3.1

README.md
```diff
@@ -60,42 +60,37 @@ You may also wish to visit Intel Media Server Studio [support page](https://soft
 Media SDK test and sample applications may require additional software packages (for example, X Server, Wayland, LibDRM, etc.) to be functional.
 
 # How to build
-## Requirements
- - Git* (with [LFS](https://git-lfs.github.com/) support)
```
adydychk: Are you sure to delete requirements at all?
Especially in case of git **lfs**. I see that some people don`t install lfs and after that they wonder that the documentation is "not work".  :)

README.md
```diff
@@ -60,42 +60,37 @@ You may also wish to visit Intel Media Server Studio [support page](https://soft
...
 
 ## Enabling Instrumentation and Tracing Technology
 To enable the Instrumentation and Tracing Technology API you need either Intel® VTune™ Amplifier installed or to manually build an open source version. You can get ITT source files from [GitHub](https://github.com/01org/IntelSEAPI/tree/master/ittnotify) and build it on your own.
 
 **Please note** that auto detection of the Intel VTune Amplifier configuration is not supported. The next step is mandatory if you want to use this feature: set `$ITT_PATH` so `$ITT_PATH/include/ittnotify.h` and `$ITT_PATH/libittnotify64.a` will be valid paths. MSDK build system will automatically detect it.
 
+# Recommendations
+
+* In case of GCC compiler it is strongly recommended to use GCC version 6 or later since that's the first GCC version which has non-experimental support of C++11 being used in Media SDK.
+
```
adydychk: In addition we can add link to the infrastructure documentation https://github.com/Intel-Media-SDK/infrastructure#deploy-build-box-worker-buildbot
There is described needed packages and its versions which are 100% stable. I can modify infrastructure documentation with the needs of product documentation.

README.md
dvrogozh: I am absolutely sure to remove "requirements" section since it in any case don't list all the requirements. The only up to date list of the requirements is configure scripts. However, let's preserve note about Git with LFS. I will move it to the section where we describe how to get sources.

README.md
dvrogozh: You have a section in README now where you can add it. Let's do that in some of the next commit.

adydychk: Great, will do that.

PR #260, opened 2018-06-05
[AVCe] Add missing checks of libva return statuses to AVC encoder

mgonchar:
Introduced some more VA sts checks for libVA calls. Only AVC and FEI Encode for now


adydychk:
@mgonchar Please make a rebase on top of master. Android CI will be fixed.

PR #279, opened 2018-06-07
[HEVC FEI] Get different videoParams for PreENC and Encode

vmaksimo:
PreENC doesn't support all the ext buffers, as Encode does.
This commit fixes the problem of getting more buffers for PreENC, than it supports.

### DIFF
--- a/samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
+++ b/samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
@@ -81,7 +81,7 @@ mfxStatus CEncodingPipeline::Init()
 
         if (m_inParams.bEncodedOrder)
         {
-            MfxVideoParamsWrapper param = GetEncodeParams(m_inParams, frameInfo);
+            MfxVideoParamsWrapper param = GetEncodeParams(m_inParams, frameInfo, PIPELINE_COMPONENT::ENCODE);
 
             sts = m_pParamChecker->Query(param);
             MSDK_CHECK_STATUS(sts, "m_pParamChecker->Query failed");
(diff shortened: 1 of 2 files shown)


samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -445,7 +445,7 @@ mfxStatus CEncodingPipeline::DrainBufferedFrames()
 
 // Function generates common mfxVideoParam ENCODE
 // from user cmd line parameters and frame info from upstream component in pipeline.
-MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFrameInfo& in_fi)
+MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFrameInfo& in_fi, const PIPELINE_COMPONENT component)
```
mgonchar: no need for "component" to be const

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -481,7 +481,7 @@ MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFr
     // configure B-pyramid settings
     pCO2->BRefType = user_pars.BRefType;
 
-    if (user_pars.bExtBRC) {
+    if (user_pars.bExtBRC && PIPELINE_COMPONENT::ENCODE) {
```
mgonchar: missing "component ==" here

samples/sample_hevc_fei/src/pipeline_hevc_fei.cpp
```diff
@@ -514,13 +514,13 @@ MfxVideoParamsWrapper GetEncodeParams(const sInputParams& user_pars, const mfxFr
     // This buffer is a correct way to pass coding window size
```
mgonchar: // This buffer is a correct way to pass coding window size to HEVC encoder

PR #280, opened 2018-06-07
[AVC FEI, ENC, PAK, PreENC] Sync output data via vaMapBuffer call

mgonchar:
Call of vaSyncSurface looks to be excessive, according to change in AVC encoder, vaMapBuffer perform blocking synchronization and should be enough to get the output data.

Manual runs showed no difference in approaches


dvrogozh:
Do we have 1N workloads for the FEI to try? If we don't, we actually will not see the effect of this change...

artem-shaporenko:
No unfortunately, we don't have 1:N FEI app.

dvrogozh:
In that case this will be a blind fix I afraid. To land it better to have some feedback from the customers which use FEI in any 1N workload, even paired with legacy encoder will do.

artem-shaporenko:
Well, for FEI it probably won't take affect, as in most of cases there is pre-enc before encode, so input surface will be different for encode

PR #285, opened 2018-06-08
Fix videowall tm

vvkalinin24:


_studio/hevce_hw/Android.mk
```diff
@@ -1,87 +0,0 @@
-LOCAL_PATH:= $(call my-dir)
```
dvrogozh: How changes to h.265 relates to videowall tm?

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1647,7 +1647,14 @@ mfxStatus VAAPIVideoProcessing::Execute_Composition_TiledVideoWall(mfxExecutePar
                                *outputSurface);
         MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
-        outputparam         = m_pipelineParam[0];
```
dvrogozh: This looks to be initial setup of outputparam. You have wiped it out. Are you sure that we will not get uninitialized values?  Why you did not simply add code afterwards?

_studio/hevce_hw/Android.mk
vvkalinin24: Sorry, nothing, need to delete this file from this commit

PR #286, opened 2018-06-08
[VPP] Fix videowall using different transformation matrix

vvkalinin24:


vvkalinin24:
need to redesign commit

Issue #266, opened 2018-06-06
Compilation fails with GCC 8.x


jwang11:
MediaSDK: e1f5bb4a72a060fd850e06825710390ede027076

Fedora 28 and ClearLinux 22xxx recently introduce gcc 8.x, which cause compilation issue for MediaSDK

```
/home/jwang/devel/MediaSDK/_studio/shared/umc/core/umc/src/umc_video_decoder.cpp: In constructor ‘UMC::VideoDecoderParams::VideoDecoderParams()’:
/home/jwang/devel/MediaSDK/_studio/shared/umc/core/umc/src/umc_video_decoder.cpp:30:46: error: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘UMC::VideoStreamInfo’ {aka ‘struct UMC::sVideoStreamInfo’}; use assignment or value-initialization instead [-Werror=class-memaccess]
     memset(&info, 0, sizeof(sVideoStreamInfo));
                                              ^
In file included from /home/jwang/devel/MediaSDK/_studio/shared/umc/core/umc/include/umc_video_decoder.h:24,
                 from /home/jwang/devel/MediaSDK/_studio/shared/umc/core/umc/src/umc_video_decoder.cpp:22:
/home/jwang/devel/MediaSDK/_studio/shared/umc/core/umc/include/umc_structures.h:568:20: note: ‘UMC::VideoStreamInfo’ {aka ‘struct UMC::sVideoStreamInfo’} declared here
     typedef struct sVideoStreamInfo : public StreamInfo
                    ^~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [_studio/shared/umc/CMakeFiles/umc.dir/build.make:128: _studio/shared/umc/CMakeFiles/umc.dir/core/umc/src/umc_video_decoder.cpp.o] Error 1
make[2]: Leaving directory '/home/jwang/devel/MediaSDK/__cmake/intel64.make.release'
make[1]: *** [CMakeFiles/Makefile2:1874: _studio/shared/umc/CMakeFiles/umc.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 28%] Building CXX object api/opensource/mfx_dispatch/CMakeFiles/mfx.dir/src/mfx_load_plugin.cpp.o
[...]

PR #289, opened 2018-06-11
Fix a wrong VAAPI check in FillCUQPDataDDI

jwang11:
According to MFX_HW_VAAPI definition below, it is 4. It make
((coreParams.Impl & 0xF00) == MFX_HW_VAAPI) always get False.Obviouly
it should be MFX_IMPL_VIA_VAAPI
enum eMFXVAType
{
    MFX_HW_NO       = 0,
    MFX_HW_D3D9     = 1,
    MFX_HW_D3D11    = 2,
    MFX_HW_VAAPI    = 4,// Linux VA-API
};

Signed-off-by: jwang <jing.j.wang@intel.com>


jwang11:
This PR is also related to #266.

jwang11:
the PR has kept pending for 10 days, if someone can help review it, thanks？

artem-shaporenko:
There are more such places, I'm working on redesigning encoder to use internal Core interface and will fix all those places

PR #295, opened 2018-06-13
Fix HEVC FEI build checks

onabiull:


_studio/hevc_fei/CMakeLists.txt
```diff
@@ -18,7 +18,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-if( API_USE_LATEST )
+if("1.27" VERSION_LESS API_VERSION)
```
dvrogozh: Guys, can we right this in an "normal" way? I.e. `if (API_VERSION VERSION_GREATER "1.27")`?

onabiull: It should be `if (API_VERSION VERSION_GREATER_EQUAL "1.27")`  then. Can't say which version scare me more :)

onabiull:
Looks like `VERSION_GREATER_EQUAL` supported since cmake 3.7 only.

PR #303, opened 2018-06-14
[cmake] Change Release to release

adydychk:


adydychk:
Build fails only on `list artifacts` action (as should be). The builds and tests are successful.

PR #314, opened 2018-06-16
Small fixes hevce

akokoshn:


_studio/hevce_hw/h265/include/mfx_h265_encode_plugin_hw.h
```diff
@@ -163,6 +163,12 @@ class Plugin : public MFXEncoderPlugin
     virtual ~Plugin()
     {}
 
+    static Plugin* GetInstance()
+    {
+        static Plugin instance(false);
+        return &instance;
```
mgonchar: In previous design each Create call returned new object, currently it return same(!) object for everyone who wants hevc encoder, i.e. they will share all their resources. Is it ok?

_studio/hevce_hw/h265/include/mfx_h265_encode_plugin_hw.h
```diff
@@ -34,7 +34,7 @@ class Plugin : public MFXEncoderPlugin
 public:
     static MFXEncoderPlugin* Create()
     {
-        return new Plugin(false);
+        return GetInstance();
```
mgonchar: I guess FEI also suffers from such issue?

mgonchar: + Old design also had such behavior of function create. Why now there is a memory leak?

akokoshn: FEI shouldn't be affected by this issue.
Early (actual for FEI) this object delete by  mfx_sessing during closing, but  
after moving HW HEVCe to library  USER_Register() for HW HEVCe guid do nothing, so mfx_session don't know about plugin.
So nobody deletes object Plugin(), issue reproduced each time when HE HEVCe loaded as plugin (used guid)

_studio/hevce_hw/h265/include/mfx_h265_encode_plugin_hw.h
akokoshn: Yes, this object uses just for check exist of required methods

PR #321, opened 2018-06-19
[h265] Fix GPU hang

dmitryermilov:
Accord. to ITU-T H.265 7.4.2.4 "Any PPS NAL unit containing the value
of pps_pic_parameter_set_id for the active PPS RBSP for a coded
picture (and consequently for the layer containing the coded picture)
shall have the same content as that of the active PPS RBSP for the coded
picture, unless it follows the last VCL NAL unit of the coded picture
and precedes the first VCL NAL unit of another coded picture". The same
is applicable for SPS. So, we just complete current AU if we met SPS\PPS
which differ from active ones.

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_task_supplier.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_task_supplier.cpp
@@ -1,15 +1,15 @@
 // Copyright (c) 2018 Intel Corporation
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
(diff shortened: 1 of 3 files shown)


_studio/shared/umc/codec/h265_dec/src/umc_h265_task_supplier.cpp
```diff
@@ -1799,15 +1799,18 @@ UMC::Status TaskSupplier_H265::AddOneFrame(UMC::MediaData * pSource)
...
+                        if (umsRes == UMC::UMC_WRN_REPOSITION_INPROGRESS)
+                        {
+                            umsRes = UMC::UMC_OK;
+                        }
+                        else if (umsRes == UMC::UMC_NTF_NEW_RESOLUTION ||
                             (nut == NAL_UT_SPS && umsRes == UMC::UMC_ERR_INVALID_STREAM))
                         {
-
-                            int32_t nalIndex = pMediaDataEx->index;
-                            int32_t size = pMediaDataEx->offsets[nalIndex + 1] - pMediaDataEx->offsets[nalIndex];
```
dmitry-gurulev: We don't need this change. Otherwise, we'll loose [MFX_WRN_VIDEO_PARAM_CHANGED] notification.

dmitryermilov: Thanks. It's addressed in https://github.com/Intel-Media-SDK/MediaSDK/pull/324

PR #328, opened 2018-06-20
[vpp] Added RGBP format support

alariono:
- This is 8 bit planar format without alpha channel
- RGBP supported for VPP output only
- Main usage case is NV12->RGBP conversion
- Decoders and encoders do not work with RGBP format!

Signed-off-by: Andrey Larionov <andrey.larionov@intel.com>

### DIFF
--- a/_studio/shared/include/mfxstructures-int.h
+++ b/_studio/shared/include/mfxstructures-int.h
@@ -95,7 +95,10 @@ enum
     MFX_FOURCC_YUV422H      = MFX_MAKEFOURCC('4','2','2','H'),
     MFX_FOURCC_YUV422V      = MFX_MAKEFOURCC('4','2','2','V'),
     MFX_FOURCC_YUV444       = MFX_MAKEFOURCC('4','4','4','P'),
-    MFX_FOURCC_RGBP         = MFX_MAKEFOURCC('R','G','B','P')
+#if (MFX_VERSION <= 1027)
+    MFX_FOURCC_RGBP         = MFX_MAKEFOURCC('R','G','B','P'),
+#else
+#endif
 };
 
 #ifdef __cplusplus
--- a/_studio/shared/include/mfxstructures-int.h
+++ b/_studio/shared/include/mfxstructures-int.h
@@ -86,7 +86,13 @@ enum {
     // AVC Profiles & Levels
     MFX_PROFILE_AVC_HIGH10      =110,
 };
-
+/*
+Some components (samples, JPEG decoder) has used MFX_FOURCC_RGBP already.
+So, for API 1.27 and below "MFX_FOURCC_RGBP" defined inside of msdk library
+and samples.
+Since next version of API (1.28) "MFX_FOURCC_RGBP" should officially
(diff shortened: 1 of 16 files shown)


_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -452,6 +452,7 @@ mfxStatus VAAPIVideoProcessing::QueryCapabilities(mfxVppCaps& caps)
         if (MFX_FOURCC_NV12    == g_TABLE_SUPPORTED_FOURCC[indx] ||
```
onabiull: At least Query and execute need runtime check if this color format supported by HW.

_studio/shared/include/mfxstructures-int.h
```diff
@@ -95,7 +95,10 @@ enum
     MFX_FOURCC_YUV422H      = MFX_MAKEFOURCC('4','2','2','H'),
     MFX_FOURCC_YUV422V      = MFX_MAKEFOURCC('4','2','2','V'),
     MFX_FOURCC_YUV444       = MFX_MAKEFOURCC('4','4','4','P'),
-    MFX_FOURCC_RGBP         = MFX_MAKEFOURCC('R','G','B','P')
+#if (MFX_VERSION <= 1027)
```
onabiull: This is unusual for MSDK code, please add comment, to justify  <= usage.

_studio/shared/src/mfx_vpp_vaapi.cpp
alariono: Media driver do not report to application back list of supported to color format.
According to driver's implementation RGBP works from GEN8 (media-driver repo, search for RGBP)

_studio/shared/include/mfxstructures-int.h
alariono: This is "unusual" feature too. Some components (samples, JPEG decoder) has used MFX_FOURCC_RGBP already. So, for API 1.27 and below "MFX_FOURCC_RGBP" defined inside of msdk library and samples. Since next version of API (1.28) "MFX_FOURCC_RGBP" should officially defined in API "mfxstructures.h".

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1037,6 +1040,10 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
     switch (refFourcc)
     {
     case MFX_FOURCC_RGB4:
+    case MFX_FOURCC_RGBP:
```
onabiull: Look like this one missed.

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1051,6 +1058,10 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
     switch (targetFourcc)
     {
     case MFX_FOURCC_RGB4:
```
onabiull: And this one

PR #329, opened 2018-06-20
[Avce][sample] fix potential mem leak issue.

dlin2:


samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -1767,6 +1765,7 @@ mfxStatus CEncodingPipeline::AllocExtBuffers(sInputParams *pInParams)
             }
         }
 
+        tmpForInit = new bufSet(numOfFields);
```
mgonchar: this memory could still leak, since push_back is not exception free (I guess, AddSet also). Suggest to wrap tmpForInit to unique_ptr and release owned pointer after manipulations

PR #334, opened 2018-06-21
[samples] Changed Locked in atomic manner

vvkalinin24:


samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -596,12 +596,12 @@ mfxStatus drmRenderer::render(mfxFrameSurface1 * pSurface)
 
     /* Unlock previous Render Target Surface (if exists) */
     if (NULL != m_pCurrentRenderTargetSurface)
-        m_pCurrentRenderTargetSurface->Data.Locked--;
+        msdk_atomic_dec16(&m_pCurrentRenderTargetSurface->Data.Locked);
```
mgonchar: In other places there also was volatile specificator
msdk_atomic_dec16((volatile mfxU16*)(&(*it)->Data.Locked));

dmitryermilov: Is vaapi_utils_drm.cpp the only place which requires this change?

PR #347, opened 2018-06-27
[AVC FEI] Timeout option was added for multi-session mode

romka0017:


samples/sample_fei/src/sample_fei.cpp
```diff
@@ -1383,6 +1383,20 @@ int main(int argc, char *argv[])
             msdk_printf(MSDK_STRING("ERROR: ParFile \"%s\" reading failed\n"), parFileName);
             return MFX_ERR_UNSUPPORTED;
         }
+
+        mfxU32 i = 0;
```
mgonchar: Please make i signed (mfxI32) and remove casts to uint from argc

samples/sample_fei/src/sample_fei.cpp
```diff
@@ -1383,6 +1383,20 @@ int main(int argc, char *argv[])
...
+        mfxI32 i = 0;
+        mfxU32 timeout = 0;
+        while (i < argc && !strstr(argv[i], "-timeout")) ++i;
+        if (i < argc - 1)
+        {
+            timeout = atoi(argv[++i]);
+        }
+        // sample should use the same timeout for all sessions
+        // because sessions should work in parallel
+        for (auto & config : Configs)
```
mgonchar: this "for" should be inside above "if"

PR #357, opened 2018-06-28
[h265d] Fixed potential null dereference

dmitryermilov:
MDP-43868


_studio/shared/umc/codec/h265_dec/src/umc_h265_task_supplier.cpp
```diff
@@ -2179,9 +2179,16 @@ UMC::Status TaskSupplier_H265::AddSlice(H265Slice * pSlice, bool )
         // and precedes the first VCL NAL unit of another coded picture
         // if the slices belong to different AUs or SPS/PPS was changed,
         // close the current AU and start new one.
+
+        H265SeqParamSet * sps = m_Headers.m_SeqParams.GetHeader(pSlice->GetSeqParam()->GetID());
```
dmitry-gurulev: const pointer to SPS\PPS is enough here, we don't change headers' content

PR #387, opened 2018-07-04
[core] Add GPU copy support for RExt

mvafin:


_studio/shared/include/cm_mem_copy.h
```diff
@@ -67,6 +67,11 @@ typedef mfxI32 cmStatus;
 #define min(a,b)            (((a) < (b)) ? (a) : (b))
 #endif
 
+static bool operator < (const mfxHDLPair & l, const mfxHDLPair & r)
```
dmitryermilov: what is the idea of < operator ? How is it used?

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -2125,7 +2578,7 @@ mfxStatus CmCopyWrapper::Initialize(eMFXHWType hwtype)
     if (m_HWType == MFX_HW_UNKNOWN)
         return MFX_ERR_UNDEFINED_BEHAVIOR;
 
-    m_timeout = CM_MAX_TIMEOUT_MS;
+    m_timeout = m_HWType >= MFX_HW_ICL ? CM_MAX_TIMEOUT_SIM : CM_MAX_TIMEOUT_MS;
```
dmitryermilov: why? please add a comment in code.

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -2162,6 +2615,14 @@ mfxStatus CmCopyWrapper::InitializeSwapKernels(eMFXHWType hwtype)
     case MFX_HW_CFL:
         cmSts = m_pCmDevice->LoadProgram((void*)skl_copy_kernel_genx,sizeof(skl_copy_kernel_genx),m_pCmProgram,"nojitter");
         break;
+#if !defined(ANDROID)
```
dmitryermilov: why is it under #if !defined(ANDROID)?

mvafin: Removed.

_studio/shared/src/cm_mem_copy.cpp
mvafin: Removed. It was needed for simulation.

_studio/shared/include/cm_mem_copy.h
mvafin: It is used to add pair to std::map in CmCopyWrapper::CreateCmSurface2D

dmitryermilov: Still not clear to me. Oleg, do you see the point?

onabiull: std::map keys should be comparable using operator <. Still I'd request to add explicit comment in the code to reduce confusion.

dmitryermilov: I see now.  Thanks. Yes, Maxim, please add a comment.

mvafin: Done.

onabiull:
I believe PR #490 already contains this change.

PR #388, opened 2018-07-04
Prevent crash when DecodeFrameCheck is called when m_isInit is false

russelltg:
Before this patch, it would segfault if it was called before initialization. Now, it returns cleanly.


russelltg:
The problem is that `session->m_pDECODE->m_pH264VideoDecoder` is null, not `m_pDECODE`

dvrogozh:
@onabiull : do we have similar problems for other decoders? can we have test coverage for this?

onabiull:
We do test uninitialized sessions. And I'm still curious on how can we get h264 decoder half-configured. Still I see no harm in doing additional check.

PR #393, opened 2018-07-05
Sample hevc fei abr refactor

mgonchar:
Introduced new class for Look Ahead queue (not storing all history anymore, only required frames)
Clean up variable/class names, removed dead code
Added comments

Test results for LA BRC relative to HW BRC are following:
+4.54% of BD-Rate Y for real MSE (requires original YUV)
+3.19% for algorithm which approximate distortion with number of nonzero transform coefficients

### DIFF
--- a/samples/sample_hevc_fei_abr/include/hevc_sw_dso.h
+++ b/samples/sample_hevc_fei_abr/include/hevc_sw_dso.h
@@ -29,10 +29,12 @@ class HevcSwDso : public IYUVSource
 public:
     HevcSwDso(const SourceFrameInfo& inPars,
               SurfacesPool* sp,
-              std::shared_ptr<MVPPool> mvpPool,
-              std::shared_ptr<CTUCtrlPool> ctuCtrlPool,
+              std::shared_ptr<MVPPool> & mvpPool,
+              std::shared_ptr<CTUCtrlPool> & ctuCtrlPool,
               bool calc_BRC_stat = false,
-              bool dump_mvp = false)
+              bool dump_mvp      = false,
+              bool est_dist      = false,
+              DIST_EST_ALGO alg  = NNZ)
         : IYUVSource(inPars, sp)
         , m_inPars(inPars)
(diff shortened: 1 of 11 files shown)


samples/sample_hevc_fei_abr/include/hevc_sw_dso.h
```diff
@@ -77,8 +83,11 @@ class HevcSwDso : public IYUVSource
     std::shared_ptr<CTUCtrlPool>        m_ctuCtrlPool;
 
 private:
-    bool m_bCalcBRCStat = false;
-    bool m_bDumpFinalMVPs = false;
+    bool m_bCalcBRCStat        = false;
+    bool m_bDumpFinalMVPs      = false;
+    bool m_bEstimateDistortion = false;
+
+    DIST_EST_ALGO m_AlgorythmType = NNZ;
```
onabiull: Algorithm?

samples/sample_hevc_fei_abr/include/task.h
```diff
@@ -21,38 +21,66 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+        {
+            NumPixelsPredictedFromRef[fo] = amount;
+        }
+    }
+
+    void FillProp()
+    {
+        for (auto & item : NumPixelsPredictedFromRef)
+        {
+            ShareOfPredictedPixelsFromRef[item.first] = mfxF64(item.second) / ((1.0 - ShareIntra) * NPixelsInFrame);
```
onabiull: Static analysis would complain about potential division by zero here.

samples/sample_hevc_fei_abr/include/hevc_sw_dso.h
mgonchar: yes! sure)

samples/sample_hevc_fei_abr/include/task.h
mgonchar: it's the only place where static analysis complains? will add a check

PR #412, opened 2018-07-13
[HEVC E]: fix for interlace (ref list and SWBRC)

ogolikov:
Issue: MDP-46419
Test: manually

reflist was changed if B pyramid with one forward reference. L1 is correspondence field. Needed for better quality.
SW BRC:
-changes for inner bitrate: MaxFrameSize calculation was changed for B pyr field case (reduced), GopParams changing
-changes for all SW BRC: Layer+1 for second field.

### DIFF
--- a/_studio/hevce_hw/h265/src/mfx_h265_encode_hw_utils.cpp
+++ b/_studio/hevce_hw/h265/src/mfx_h265_encode_hw_utils.cpp
@@ -2941,8 +2941,9 @@ void ConstructRPL(
 #elif (HEVCE_FIELD_MODE == 2)
                         MFX_SORT_COMMON(RPL[1], numRefActive[1], (Abs(DPB[RPL[1][_i]].m_poc/2 - poc/2)  + ((DPB[RPL[1][_i]].m_secondField == bSecondField) ? 0 : 16)) > (Abs(DPB[RPL[1][_j]].m_poc/2 - poc/2)  + ((DPB[RPL[1][_j]].m_secondField == bSecondField) ? 0 : 16)));
 #elif (HEVCE_FIELD_MODE == 3 || HEVCE_FIELD_MODE == 4)
-                    if (par.isBPyramid())
+                    if (par.isBPyramid() && l0 > 1)
                     {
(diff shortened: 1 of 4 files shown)


_studio/hevce_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -2941,8 +2941,9 @@ void ConstructRPL(
 #elif (HEVCE_FIELD_MODE == 2)
                         MFX_SORT_COMMON(RPL[1], numRefActive[1], (Abs(DPB[RPL[1][_i]].m_poc/2 - poc/2)  + ((DPB[RPL[1][_i]].m_secondField == bSecondField) ? 0 : 16)) > (Abs(DPB[RPL[1][_j]].m_poc/2 - poc/2)  + ((DPB[RPL[1][_j]].m_secondField == bSecondField) ? 0 : 16)));
 #elif (HEVCE_FIELD_MODE == 3 || HEVCE_FIELD_MODE == 4)
-                    if (par.isBPyramid())
+                    if (par.isBPyramid() && l0 > 1)
```
lakulako: Is it OK that a field can have 1 L0 and 1 L1 reference fields of the same polarity?

_studio/mfx_lib/shared/include/mfx_brc_common.h
```diff
@@ -196,6 +196,7 @@ class cBRCParams
     mfxU16 gopPicSize;
     mfxU16 gopRefDist;
     bool   bPyr;
+    bool   bFieldMode;
```
lakulako: In fact it is specific for HEVC only joining of fields to frames. May be call it bHevcFields or bJoinFields?

ogolikov: I think, it is OK. for H264 SW BRC works in progressive mode for both interlace and progressive

_studio/hevce_hw/h265/src/mfx_h265_encode_hw_utils.cpp
ogolikov: if the same polarity - must be ok

PR #415, opened 2018-07-16
[sample_hevc_fei_abr] Handle errors from encoders

dmitryermilov:
MDP-46410

### DIFF
--- a/samples/sample_hevc_fei_abr/include/hevc_fei_encode.h
+++ b/samples/sample_hevc_fei_abr/include/hevc_fei_encode.h
@@ -28,6 +28,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 class IEncoder
 {
 public:
+
     virtual ~IEncoder() {};
     virtual mfxStatus Init() = 0;
     virtual mfxStatus Query() = 0;
@@ -56,6 +57,9 @@ public:
     }
     // Submit new data to BRC
     virtual void UpdateBRCStat(FrameStatData& stat_data) = 0;
+
+    typedef std::function< void() > NotifyErrorCallback;
+    virtual void SetNotifyErrorCallback(NotifyErrorCallback callback) = 0;
 };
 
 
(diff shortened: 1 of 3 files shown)


samples/sample_hevc_fei_abr/include/pipeline_hevc_fei.h
```diff
@@ -72,6 +72,10 @@ class EncoderContext
     {
         return m_encoder->SubmitFrame(task);
     }
+    void SetNotifyErrorCallback(IEncoder::NotifyErrorCallback callback)
+    {
+        return m_encoder->SetNotifyErrorCallback(callback);
```
mgonchar: it's void type return, so just don't return anything

samples/sample_hevc_fei_abr/include/pipeline_hevc_fei.h
```diff
@@ -154,6 +157,8 @@ class CFeiTranscodingPipeline
     mfxStatus FillInputFrameInfo(mfxFrameInfo& fi);
 
     DISALLOW_COPY_AND_ASSIGN(CFeiTranscodingPipeline);
+
+    bool m_error; // used by asynchronous encoders to signal an error
```
mgonchar: shouldn't it be some sort of atomic?

samples/sample_hevc_fei_abr/include/hevc_fei_encode.h
```diff
@@ -83,8 +87,17 @@ class FEI_Encode : public IEncoder
...
-
+            try
+            {
+                mfxStatus sts = DoWork(task);
+                if (MFX_ERR_NONE != sts)
+                    m_error_callback();
+            }
+            catch(...)
+            {
+                MSDK_TRACE_ERROR("Caught an exception from DoWork");
```
mgonchar: maybe a little bit more details: FEI_Encode::DoWork

samples/sample_hevc_fei_abr/include/pipeline_hevc_fei.h
dmitryermilov: Okay

samples/sample_hevc_fei_abr/include/pipeline_hevc_fei.h
dmitryermilov: I think it would be excessive. Let's just add 'volatile'

samples/sample_hevc_fei_abr/include/hevc_fei_encode.h
dmitryermilov: Okay

samples/sample_hevc_fei_abr/include/pipeline_hevc_fei.h
mgonchar: sure

PR #416, opened 2018-07-16
Install include files to $prefix/include/mfx

dvrogozh:


CMakeLists.txt
```diff
@@ -104,7 +104,9 @@ message("Debug flags:")
 message("  CMAKE_C_FLAGS_DEBUG                     : ${CMAKE_C_FLAGS_DEBUG}")
 message("  CMAKE_CXX_FLAGS_DEBUG                   : ${CMAKE_CXX_FLAGS_DEBUG}")
 message("Install:")
-message("  CMAKE_INSTALL_RREFIX                    : ${CMAKE_INSTALL_RREFIX}")
+message("  CMAKE_INSTALL_RREFIX                    : ${CMAKE_INSTALL_PREFIX}")
```
onabiull: I believe it should be PREFIX :)

dvrogozh: arggh. I did not check the spelling on the left side of the expression... Thank you for the catch.

PR #424, opened 2018-07-17
Change install location for plugins and samples

dvrogozh:


dvrogozh:
@adydychk, @azhogov : CI test fails, however, this change runs successfully on my side. Question: do you run apps/plugins from the _installed_ locations or you move them elsewhere?

azhogov:
It uses not installed location. I agree that the test scripts should be updated.

dvrogozh:
@azhogov : this PR is supposed to address potential concerns for mediasdk system installation and packaging. Thus, this is a priority. Can you, please, give an ETA when test scripts will be updated?

azhogov:
@dvrogozh We are going to start fix it this week. Next week the fix will be deployed.

dvrogozh:
Ok. This PR is finally unlocked. Folks, pay attention that it will change layout of mediasdk library on the system for those who run 'make install'!!

dvrogozh:
@onabiull : do we merge this?

PR #455, opened 2018-07-26
Fix android warnings 1 3

akokoshn:


_studio/shared/umc/codec/h265_dec/src/umc_h265_heap.cpp
```diff
@@ -343,9 +343,6 @@ uint8_t * CumulativeArraysAllocation(int n, int align, ...)
     int cumulativeSize = 0;
     for (int i = 0; i < n; i++)
     {
-        void * ptr = va_arg(args, void *);
```
dvrogozh: You can't remove this va_arg call since va_arg reads an argument and advances to the next one. So, you will read not a currSize, but pointer instead.

_studio/shared/umc/codec/h265_dec/src/umc_h265_heap.cpp
```diff
@@ -343,9 +343,6 @@ uint8_t * CumulativeArraysAllocation(int n, int align, ...)
...
-        ptr; // just skip it
-
         int currSize = va_arg(args, int);
```
dvrogozh: But actually, you don't need to fix the above since this whole function is simply not used:
```
fgrep -rsn CumulativeArraysAllocation _studio/
_studio/shared/umc/codec/h265_dec/include/umc_h265_heap.h:36:uint8_t * CumulativeArraysAllocation(int n, int align, ...);
_studio/shared/umc/codec/h265_dec/include/umc_h265_heap.h:37:// Free memory allocated by CumulativeArraysAllocation
_studio/shared/umc/codec/h265_dec/src/umc_h265_heap.cpp:338:uint8_t * CumulativeArraysAllocation(int n, int align, ...)
_studio/shared/umc/codec/h265_dec/src/umc_h265_heap.cpp:374:// Free memory allocated by CumulativeArraysAllocation
```
Just remove it as a whole.

_studio/mfx_lib/cmrt_cross_platform/include/cmrt_cross_platform.h
```diff
@@ -69,9 +69,6 @@ class SurfaceIndex
 
 private:
     unsigned int index;
```
dvrogozh: Alexey, please, break this change into 3: 1) CMRT, 2) MCTF, 3) VPP. These are too far things to be combined together.

_studio/shared/umc/codec/h265_dec/src/umc_h265_heap.cpp
```diff
@@ -343,8 +343,7 @@ uint8_t * CumulativeArraysAllocation(int n, int align, ...)
     int cumulativeSize = 0;
     for (int i = 0; i < n; i++)
     {
-        void * ptr = va_arg(args, void *);
-        ptr; // just skip it
+        va_arg(args, void *);
```
dvrogozh: Maybe we will remove the whole function if we don't use it? Why you have left it?

akokoshn: I plane to remove all unused functions/methods in the next PR (for keep structure: PR - type of issue, commit - component)

_studio/mfx_lib/cmrt_cross_platform/include/cmvm.h
```diff
@@ -967,7 +967,6 @@ matrix<T, R, C>::matrix(const matrix<T2, R2, C2>& src, const uint sat)
     : array_2d_(data)
 #endif
 {
-    static const bool conformable = check_true<R*C == R2*C2>::value;
```
dvrogozh: We talked with Oleg today and he noted this, but had no time to post yet. This wonderful code is actually a static_error:)! And should be used as such: CM_STATIC_ERROR(R*C == R2*C2, "matrices have different dimensions");

_studio/mfx_lib/cmrt_cross_platform/include/cmvm.h
```diff
@@ -984,7 +983,6 @@ matrix<T, R, C>::matrix(const matrix_ref<T2, R2, C2>& src, const uint sat)
...
     assert(R*C == R2*C2);
```
dvrogozh: And of the above was intended to be a STATIC_ERROR, then question is why we have same assert here? Answer to this is simple, CM_STATIC_ERROR is defined to nothing:)! I suggest to define it properly and remove assert() calls altogether.

dvrogozh:
@onabiull : sorry, I thought I reviewed it yesterday.

akokoshn:
@onabiull , re base done

PR #459, opened 2018-07-27
Pass pie LD flag for executables

dmitryermilov:
CMAKE_EXEC_LINKER_FLAGS is a misprint.


dvrogozh:
Why you make this merge from the pie branch rather than a mss2018_r2 merge to master with all the changes done in this release branch? I think that's the way we follow now...

PR #460, opened 2018-07-30
Libmfx refactor: component inits

dvrogozh:
That's a refactor towards a bigger goal which I will submit if we will agree on this series: https://github.com/dvrogozh/MediaSDK/tree/scheduler_threads. And even further I am trying to get single place where mediasdk components are being initialized/closed to effectively add a hook to the scheduler.

### DIFF
--- a/_studio/mfx_lib/shared/src/libmfxsw_enc.cpp
+++ b/_studio/mfx_lib/shared/src/libmfxsw_enc.cpp
@@ -43,36 +43,33 @@
 #include "mfx_h264_enc.h"
 #endif
 
-
-
-VideoENC *CreateENCSpecificClass(mfxVideoParam *par, VideoCORE *pCore)
+template<>
+VideoENC* _mfxSession::Create<VideoENC>(mfxVideoParam& par)
 {
-    VideoENC *pENC = (VideoENC *) 0;
+    VideoENC* pENC = nullptr;
+    VideoCORE* pCore = m_pCORE.get();
     mfxStatus mfxRes = MFX_ERR_MEMORY_ALLOC;
-    mfxU32 codecId = par->mfx.CodecId;
+    mfxU32 codecId = par.mfx.CodecId;
 
     switch (codecId)
     {
 #if ( defined(MFX_ENABLE_LA_H264_VIDEO_HW) || defined(MFX_ENABLE_H264_VIDEO_FEI_PREENC) || defined(MFX_ENABLE_H264_VIDEO_FEI_ENC))
     case MFX_CODEC_AVC:
 #if defined(MFX_ENABLE_LA_H264_VIDEO_HW)
-        if (bEnc_LA(par))
+        if (bEnc_LA(&par))
             pENC = (VideoENC*) new VideoENC_LA(pCore, &mfxRes);
 #endif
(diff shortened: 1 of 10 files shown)


_studio/mfx_lib/shared/src/libmfxsw_pak.cpp
```diff
@@ -141,13 +141,6 @@ mfxStatus MFXVideoPAK_Init(mfxSession session, mfxVideoParam *par)
...
     try
     {
-        // close the existing PAK unit,
-        // if it is initialized.
-        if (session->m_pPAK.get())
-        {
-            MFXVideoPAK_Close(session);
-        }
-
         // create a new instance
```
dvrogozh: @mgonchar : we need your input here: is that save?

_studio/mfx_lib/shared/src/libmfxsw_enc.cpp
```diff
@@ -202,30 +202,11 @@ mfxStatus MFXVideoENC_Init(mfxSession session, mfxVideoParam *par)
...
-            session->m_bIsHWENCSupport = true;
-
-        // SW fallback if EncodeGUID is absence
-        if (MFX_PLATFORM_HARDWARE == session->m_currentPlatform &&
-            !session->m_bIsHWENCSupport &&
-            MFX_ERR_NONE <= mfxRes)
-        {
-            mfxRes = MFX_WRN_PARTIAL_ACCELERATION;
-        }
     }
```
dvrogozh: @mgonchar : please, pay attention.

_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
```diff
@@ -201,6 +201,12 @@ mfxStatus MFXVideoUSER_Register(mfxSession session, mfxU32 type,
             return MFX_ERR_UNDEFINED_BEHAVIOR;
         }
 
+#ifndef MFX_ENABLE_USER_VPP
```
dvrogozh: I rearranged checks under MFX_ENABLE_USER_VPP and MFX_ENABLE_VPP to make more sense...

However, I completely don't understand why someone may wish to generate the library completely without VPP or user VPP plugin support?! Well, I understand that someone may wish to avoid some features, especially CM kernels, but we need to have separate ENABLE for that. Thus, I suggest to actually remove both of these ENABLES completely from the whole mediasdk.

@onabiull : how do you think? who we may need to ask about that? maybe @jonrecker?

_studio/mfx_lib/shared/src/libmfxsw_pak.cpp
mgonchar: Not sure, the difference is only in call of
// wait until all tasks are processed
        session->m_pScheduler->WaitForTaskCompletion(session->m_pPAK.get());

Without this piece of code that call won't be performed. Cleanup of allocated resources by PAK interface should be correctly performed by destructor

I believe it looks reasonable to have this call, but I'm not really familiar with scheduler code

_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
onabiull: I believe this is artifact of attempts to create bare runtime library. In current state there is no need in these defines.

PR #461, opened 2018-07-31
 Brc sliding window frame tolerance mode on APL and CFL turns on via GetHWType() call

IlyaSergeyev:


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1697,12 +1698,16 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
...
+    eMFXHWType hwType = hwCore->GetHWType();
+    bool isFrameToleranceModeRequired = (MFX_HW_APL == hwType);
 
     FillSps(par, m_sps);
     FillBrcStructures(par, m_vaBrcPar, m_vaFrameRate);
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetHRD(par, m_vaDisplay, m_vaContextEncode, m_hrdBufferId),                              MFX_ERR_DEVICE_FAILED);
-    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetRateControl(par, m_mbbrc, 0, 0, m_vaDisplay, m_vaContextEncode, m_rateParamBufferId), MFX_ERR_DEVICE_FAILED);
+    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetRateControl(par, m_mbbrc, 0, 0, m_vaDisplay, m_vaContextEncode, m_rateParamBufferId, false, isFrameToleranceModeRequired), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, m_frameRateId),                        MFX_ERR_DEVICE_FAILED);
```
dmitryermilov: what is FrameToleranceMode? Tolerance to what? It sounds weird to me.

IlyaSergeyev: For me to. I just trying to avoid ifdefs. SetRateControl function is using ifdef LINUX_TARGET_PLATFORM to set rc_flags.bits.frame_tolerance_mode = 1 now. I don't know what it means for codec, i just try to detect BXT platform via  GetHWType() call.

dmitryermilov: Ahh, it's BRC stuff.  And it seems isFrameToleranceModeRequired==true in current implementation always means frame_tolerance_mode = 1 (sliding window). According to va.h there are actually 3 modes: 0, 1, 2.  Let's call it then something like IsBrcSlidingWindowToleranceMode or IsBrcSlidingWindowFrameToleranceMode.

IlyaSergeyev: Ok, renamed

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -204,12 +205,12 @@ mfxStatus SetRateControl(
     if(par.calcParam.maxKbps)
         rate_param->target_percentage = (unsigned int)(100.0 * (mfxF64)par.calcParam.targetKbps / (mfxF64)par.calcParam.maxKbps);
 
-#if defined(LINUX_TARGET_PLATFORM_BXTMIN) || defined(LINUX_TARGET_PLATFORM_BXT) || defined(LINUX_TARGET_PLATFORM_CFL)
-    // Activate frame tolerance sliding window mode
-    if (extOpt3->WinBRCSize) {
-        rate_param->rc_flags.bits.frame_tolerance_mode = 1;
+    if (isFrameToleranceModeRequired) {
```
dmitryermilov: please rename

PR #462, opened 2018-08-01
Fix hevce plugin build: exclude unused stuff

dvrogozh:


dvrogozh:
@onabiull : I can break this PR into 2 if you don't want to deal with moving files back to mfx_lib for now. How do you think?

onabiull:
Sound like a good idea. I prefer to merge gen 10/11 features first and cleanup file structure after.

PR #463, opened 2018-08-01
Fastboot

IlyaSergeyev:


IlyaSergeyev:
1)      how this works?

It builds full libmfx, than it builds all configurations described in _studio/fastboot

2)      how many libs I will build?

In this version – 8, I will remove all unneeded versions except one, to be more clear

3)      how default build is changed?

Used codecs flags moved from mfx_config*.h into builder/profiles/*.cmake

4)      which config options I have?

builder/profiles/*.cmake
4) how I can use them?
                      You can change lists of codecs in this files, if you want to build custom library, or you can make new config

PR #466, opened 2018-08-01
switch to std::thread

dvrogozh:


_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -82,20 +81,24 @@ mfxSchedulerCore::~mfxSchedulerCore(void)
...
-        vm_thread_linux_schedparams params;
+        if (handle.joinable()) {
+            struct sched_param param{};
 
-        memset(&params, 0, sizeof(params));
-        params.schedtype = m_param.params.SchedulingType;
-        params.priority = m_param.params.Priority;
-        return vm_thread_set_scheduling(&handle, &params);
+            param.sched_priority = m_param.params.Priority;
+            return !pthread_setschedparam(handle.native_handle(), m_param.params.SchedulingType, &param);
```
alariono: This is not part of std::thread...
So, MSDK still keep dependency from pthread and code is not cross platform

mgonchar: @alariono I thought the intent of this change not to make code cross-platform, but to get rid of own thread wrapper

dvrogozh: @mgonchar : exactly. std::thread as all std library don't support everything in the world. Priorities operations are still missing there and should be handle thru native handles. They even have the corresponding example: https://en.cppreference.com/w/cpp/thread/thread/native_handle. The major idea is to remove our wrapper or rather remove VM wrapper and if we really need any wrapper move it to the mediasdk core. In this specific case I don't believe we will need the above operation anywhere else outside the scheduler, thus, this is implemented right in the scheduler, not as a wrapper inside mediasdk core.

mgonchar: @dvrogozh I think next candidate for clean up is our mutexes/lockers

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -82,20 +81,24 @@ mfxSchedulerCore::~mfxSchedulerCore(void)
 
 } // mfxSchedulerCore::~mfxSchedulerCore(void)
 
-bool mfxSchedulerCore::SetScheduling(vm_thread& handle)
+bool mfxSchedulerCore::SetScheduling(std::thread& handle)
 {
-    handle;
+    (void)handle;
```
onabiull: Why do we need this line?

PR #468, opened 2018-08-02
genx: kernels unification in a single folder and ENABLE_KERNELS option

dvrogozh:


dvrogozh:
@onabiull : rebase is done

PR #469, opened 2018-08-02
Detection of rate control type for bxt platform changed to runtime

IlyaSergeyev:


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -4398,18 +4398,21 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
...
+            par.mfx.RateControlMethod == MFX_RATECONTROL_LA  ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_LA_HRD ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_LA_EXT ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_VBR ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_QVBR ||
+            (mfxRateControlCbrOnBxt && !IsOn(extOpt2->ExtBRC));
+
      if (extOpt3->WinBRCMaxAvgKbps || extOpt3->WinBRCSize)
      {
-         if ((par.mfx.RateControlMethod != MFX_RATECONTROL_LA  &&
```
dvrogozh: @stasefremov : Stas, do you understand what is written here?! or do you know someone who does? I very much want to see list of modes for which we try to apply settings rather than list of modes for which we don't! I simply can't say what we should do without knowing the intent...

dmitryermilov: @takulako , perhaps you can help. Can you please clarify ?

eistarov: The old check looks clear enough for me:
Sliding window supported by limited number of RC modes (LA, LA_HRD, LA_EXT, VBR, QVBR, HW CBR).
If the current mode is not one of them - zero all winbrc parameters and return warning.

What's unclear here is the proposed changes - mfxRateControl variable name looks confusing and doesn't reflect its actual sense.

IlyaSergeyev: mfxRateControl  renamed to slidingWindowSupported

dvrogozh: @eistarov : for me the code is not clear. It becomes clear after your comment which is entirely missing in the code. I suggest to add exactly what you have wrote: "Sliding window is supported by limited number of RC modes (LA, LA_HRD, LA_EXT, VBR, QVBR, HW CBR)." After that code starts to make some sense. @IlyaSergeyev : please, add.

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -4398,18 +4398,21 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
...
+        (platform == MFX_HW_APL || platform == MFX_HW_CFL) && 
+        (par.mfx.RateControlMethod == MFX_RATECONTROL_CBR);
+
+    bool mfxRateControl =
+            par.mfx.RateControlMethod == MFX_RATECONTROL_LA  ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_LA_HRD ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_LA_EXT ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_VBR ||
+            par.mfx.RateControlMethod == MFX_RATECONTROL_QVBR ||
+            (mfxRateControlCbrOnBxt && !IsOn(extOpt2->ExtBRC));
```
dvrogozh: @eistarov : what is this BXT specific check about? What exactly they tried to change for BXT? Is my understanding correct that they enabled sliding window for CBR mode on BXT? But what the check for ExtBrc is about?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -4402,7 +4402,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         (platform == MFX_HW_APL || platform == MFX_HW_CFL) && 
         (par.mfx.RateControlMethod == MFX_RATECONTROL_CBR);
 
-    bool mfxRateControl =
+    bool slidingWindowSupported  =
```
dvrogozh: Squash this!

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
eistarov: It looks like they're trying to say: "Sliding window + CBR is supported on BXT with HW BRC, but still not with SW one (aka ExtBRC)".

onabiull:
@dvrogozh are you ok with latest version?

PR #470, opened 2018-08-02
VPP max resolution detection via caps query

IlyaSergeyev:


_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -423,14 +424,24 @@ mfxStatus VAAPIVideoProcessing::QueryCapabilities(mfxVppCaps& caps)
...
-    caps.uMaxHeight = 4096;
-#endif
+    if (m_pipelineCaps.max_output_width && m_pipelineCaps.max_output_height)
+    {
+        caps.uMaxWidth = m_pipelineCaps.max_output_width;
+        caps.uMaxHeight = m_pipelineCaps.max_output_height;
+    }
+    else {
+        /* NB! The code below should to be replaced with querying caps from driver*/
+        VAAPIVideoCORE *hwCore = dynamic_cast<VAAPIVideoCORE *>(m_core);
```
dvrogozh: See comment from the driver team: https://github.com/intel/media-driver/commit/1479294aa643af67693bad0e5eb6d8eea9ea0b25#commitcomment-29930527. They actually say that Gen8+ supports 16K-32 (16352) and Gen9+ supports 16K (16384). 

I suggest to be as simple as possible here and rely on driver caps. Our open source mediasdk supports Gen8+ platforms. Thus, let's check platforms as >=BDW and hardcode 16K-32 limit here without differentiation between 16K-32.

@onabiull, @dmitryermilov : how do you think?

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -573,12 +584,14 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
                 else /* For BFF, second field is Top */
                     deint.flags = VA_DEINTERLACING_BOTTOM_FIELD_FIRST;
 
-                #if defined(LINUX_TARGET_PLATFORM_BXT) || defined(LINUX_TARGET_PLATFORM_BXTMIN)
-                if (MFX_PICSTRUCT_FIELD_TFF & pCurSurf_frameInfo->frameInfo.PicStruct)
-                    deint.flags = VA_DEINTERLACING_ONE_FIELD;
-                else /* For BFF case required to set all bits  */
-                    deint.flags = VA_DEINTERLACING_BOTTOM_FIELD_FIRST | VA_DEINTERLACING_BOTTOM_FIELD | VA_DEINTERLACING_ONE_FIELD;
-                #endif // defined(LINUX_TARGET_PLATFORM_BXTMIN) || defined(LINUX_TARGET_PLATFORM_BXT)
+                if(MFX_HW_APL == hwType)
```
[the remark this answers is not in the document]
dmitryermilov: Ilya, I fully agree with Dmitry R. Since the PR is called "VPP max resolution detection via caps query" it should be only about caps.  Please move the rest to another PR.

_studio/shared/src/mfx_vpp_vaapi.cpp
dmitryermilov: I agree. Let's make it simple and use 16K-32 for >=BDW . 
@IlyaSergeyev , @alariono , since we change behavior here, can you please manually check workability on  16K-32 resolution on BDW ?

IlyaSergeyev:
splited into #476

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -423,14 +423,10 @@ mfxStatus VAAPIVideoProcessing::QueryCapabilities(mfxVppCaps& caps)
...
 
-    /* NB! The code below should to be replaced with querying caps from driver*/
-#if defined(LINUX_TARGET_PLATFORM_BXTMIN) || defined(LINUX_TARGET_PLATFORM_BXT)
-    caps.uMaxWidth  = 8192;
-    caps.uMaxHeight = 8192;
-#else
-    caps.uMaxWidth  = 4096;
-    caps.uMaxHeight = 4096;
-#endif
+    MFX_CHECK_WITH_ASSERT(m_pipelineCaps.max_output_width && m_pipelineCaps.max_output_height, MFX_ERR_DEVICE_FAILED);
```
dvrogozh: In that way we will not work with the Q2 driver release. That's bad. Let's have at least some default if these values are absent. I suggest to use 4K (forget about those things I wrote about BDW and 16K-32 - that's too complex). Just use 4K default since that's actually what most users had before building our code.

alariono:
@IlyaSergeyev, test this change with driver's fix:
https://github.com/intel/media-driver/pull/263

IlyaSergeyev:
@alariono , @dvrogozh 
With driver with patch https://github.com/intel/media-driver/pull/263 works fine.

PR #473, opened 2018-08-02
Sliding window frame tolerance mode for BXT

IlyaSergeyev:
Hardware-type detected in CreateAuxilliaryDevice(),
than it used to switch set the frame tolerance mode in SetRateControl() function
(for APL or CFL it is a "sliding window")

### DIFF
--- a/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
+++ b/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
@@ -297,7 +297,15 @@ typedef struct tagENCODE_CAPS
             UINT    HMEOffsetSupport        : 1;
             UINT    DirtyRectSupport        : 1;
             UINT    MoveRectSupport         : 1;
-            UINT                            : 21;
+            UINT    FrameSizeToleranceSupport     : 1; // eFrameSizeTolerance_Low (Sliding window) supported
+            UINT    HWCounterAutoIncrementSupport : 2;
+            UINT    MBControlSupport              : 1;
+            UINT    ForceReparationCheckSupport   : 1;
+            UINT    CustomRoundingControl         : 1;
+            UINT    LLCStreamingBufferSupport     : 1;
+            UINT    DDRStreamingBufferSupport     : 1;
+            UINT    LowDelayBRCSupport            : 1; // eFrameSizeTolerance_ExtremelyLow (Low delay) supported
+            UINT                            : 12;
         };
         UINT      CodingLimits2;
     };
@@ -501,6 +509,31 @@ typedef enum tagENCODE_ARGB_COLOR
     eColorSpace_P2020 = 2
 }ENCODE_ARGB_COLOR;
 
+typedef enum tagENCODE_FRAME_SIZE_TOLERANCE
+{
+    eFrameSizeTolerance_Normal = 0,//default
+    eFrameSizeTolerance_Low = 1,//Sliding Window
+    eFrameSizeTolerance_ExtremelyLow = 2//low delay
+}ENCODE_FRAME_SIZE_TOLERANCE;
+
+typedef enum tagENCODE_SCENARIO
+{
+    eScenario_Unknown = 0,
+    eScenario_DisplayRemoting = 1,
+    eScenario_VideoConference = 2,
+    eScenario_Archive = 3,
+    eScenario_LiveStreaming = 4,
+    eScenario_VideoCapture = 5,
+    eScenario_VideoSurveillance = 6
+} ENCODE_SCENARIO;
+
+typedef enum tagENCODE_CONTENT
+{
+    eContent_Unknown = 0,
+    eContent_FullScreenVideo = 1,
+    eContent_NonVideoScreen = 2
+} ENCODE_CONTENT;
+
 typedef struct tagENCODE_SET_SEQUENCE_PARAMETERS_H264
 {
     USHORT  FrameWidth;
@@ -566,8 +599,12 @@ typedef struct tagENCODE_SET_SEQUENCE_PARAMETERS_H264
     };
 
     UINT    UserMaxFrameSize;
-    USHORT  AVBRAccuracy;
-    USHORT  AVBRConvergence;
+    USHORT  ICQQualityFactor;
+
+    ENCODE_ARGB_COLOR ARGB_input_color;
+    ENCODE_SCENARIO scenario_info;
+    ENCODE_CONTENT  content_info;
+    ENCODE_FRAME_SIZE_TOLERANCE frame_size_tolerance;
 
 } ENCODE_SET_SEQUENCE_PARAMETERS_H264;
 
@@ -638,22 +675,6 @@ typedef struct tagENCODE_ROI
     CHAR   PriorityLevelOrDQp; // [-3..3] or [-51..51]
 } ENCODE_ROI;
 
-typedef enum tagENCODE_SCENARIO
-{
-	eScenario_Unknown = 0,
-	eScenario_DisplayRemoting = 1,
-	eScenario_VideoConference = 2,
-	eScenario_Archive = 3,
-	eScenario_LiveStreaming = 4
-} ENCODE_SCENARIO;
-
-typedef enum tagENCODE_CONTENT
-{
-	eContent_Unknown = 0,
-	eContent_FullScreenVideo = 1,
-	eContent_NonVideoScreen = 2
-} ENCODE_CONTENT;
-
 typedef struct tagMOVE_RECT
 {
     USHORT SourcePointX;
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -2730,9 +2736,11 @@ mfxStatus VAAPIEncoder::Execute(
         }
     }
 
+    bool isBrcSlidingWindowFrameToleranceMode = (MFX_HW_APL == m_hwType || MFX_HW_CFL == m_hwType);
```
dvrogozh: That's much better than before! However, this should not be a variable you hold on stack and calculate each time. I wrote this before, here is explicitly once more time:
1. Add 'UCHAR m_BrcSlidingWindowFrameToleranceMode' to https://github.com/Intel-Media-SDK/MediaSDK/blob/5d9da414de71a2c38374ed932295b6786a3c5863/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h#L305 (ENCODE_CAPS structure)
2. Calculate it once in the https://github.com/Intel-Media-SDK/MediaSDK/blob/5d9da414de71a2c38374ed932295b6786a3c5863/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp#L1332 (VAAPIEncoder::CreateAuxilliaryDevice)

_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
```diff
@@ -295,6 +295,7 @@ namespace MfxHwH264Encode
         mfxU32 m_userMaxFrameSize;  // current MaxFrameSize from user.
         mfxU32 m_mbbrc;
         ENCODE_CAPS m_caps;
+        eMFXHWType m_hwType;
```
dvrogozh: Please, look into the comment below. If you will address it, you may not need to store hwType here in the class. Though if you will use it in some other place (another patch?), you may...

IlyaSergeyev:
m_hwType removed from VAAPIEncoder class, BrcSlidingWindowFrameToleranceMode added to m_caps and calculated once in CreateAuxilliaryDevice

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -242,6 +242,7 @@ typedef struct tagENCODE_CAPS
             UINT    NoESS                       : 1;
             UINT    Color420Only                : 1;
             UINT    ICQBRCSupport               : 1;
+            UCHAR   BrcSlidingWindowFrameToleranceMode: 1;
```
dvrogozh: The rest of the patch looks good. Now we need to correctly define this variable. @stasefremov : can you, please, advise hot to define it?

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1344,6 +1345,9 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     {
         mfxStatus mfxSts = hwcore->GetVAService(&m_vaDisplay);
         MFX_CHECK_STS(mfxSts);
+        eMFXHWType m_hwType = hwcore->GetHWType();
+        m_caps.BrcSlidingWindowFrameToleranceMode =
```
GlaUgrub: "memset(&m_caps, 0, sizeof(m_caps));" few lines below!

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -168,7 +168,8 @@ mfxStatus SetRateControl(
     VADisplay    vaDisplay,
     VAContextID  vaContextEncode,
     VABufferID & rateParamBuf_id,
-    bool         isBrcResetRequired = false)
+    bool         isBrcResetRequired = false,
+    bool         isBrcSlidingWindowFrameToleranceMode = false)
```
GlaUgrub: I think it's better to pass m_caps to the function instead of this parameter with huge name.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -204,12 +205,12 @@ mfxStatus SetRateControl(
...
         rate_param->target_percentage = (unsigned int)(100.0 * (mfxF64)par.calcParam.targetKbps / (mfxF64)par.calcParam.maxKbps);
 
-#if defined(LINUX_TARGET_PLATFORM_BXTMIN) || defined(LINUX_TARGET_PLATFORM_BXT) || defined(LINUX_TARGET_PLATFORM_CFL)
-    // Activate frame tolerance sliding window mode
-    if (extOpt3->WinBRCSize) {
-        rate_param->rc_flags.bits.frame_tolerance_mode = 1;
+    if (isBrcSlidingWindowFrameToleranceMode) {
+        // Activate frame tolerance sliding window mode
+        if (extOpt3->WinBRCSize) {
+            rate_param->rc_flags.bits.frame_tolerance_mode = 1;
```
GlaUgrub: AFAIK there are at least 3 modes of frame size tolerance. Isn't it better to introduce enum for these modes instead of using magic number "1"?

dvrogozh: Yep, I agree. Here is extract from VAAPI:
```
            /** Frame tolerance mode.
             *
             *  Indicates the tolerance the application has to variations in the frame size.
             *  For example, wireless display scenarios may require very steady bit rate to
             *  reduce buffering time. It affects the rate control algorithm used,
             *  but may or may not have an effect based on the combination of other BRC
             *  parameters.  Only valid when the driver reports support for
             *  #VAConfigAttribFrameSizeToleranceSupport.
             *
             *  equals 0    -- normal mode;
             *  equals 1    -- maps to sliding window;
             *  equals 2    -- maps to low delay mode;
             *  other       -- invalid.
             */
            uint32_t frame_tolerance_mode   : 2;

```

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
dvrogozh: Should be either :2 or some enum. @GlaUgrub : still need to place this correctly.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1339,19 +1338,22 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 {
     m_core = core;
 
+    m_width  = width;
+    m_height = height;
+
+    memset(&m_caps, 0, sizeof(m_caps));
```
dvrogozh: where these 3 lines came from and why? they look suspicious. Maybe they are required, maybe not... Looks to be a separate bug fix. @stasefremov : can you, please, comment?

IlyaSergeyev: These rows are came from the 1350 row. I set m_caps.FrameToleranceMode, so initializing of m_caps should be before it, not after.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -243,6 +253,7 @@ typedef struct tagENCODE_CAPS
             UINT    NoESS                       : 1;
             UINT    Color420Only                : 1;
             UINT    ICQBRCSupport               : 1;
+            UCHAR   FrameToleranceMode          : 2;
```
stasefremov: 1) Please move to CodingLimits2. Let's keep proper alignment within unions.
2) Please rename to "FrameSizeTolerance".
3) Do we really need 2 bits? Is it possible that driver supports one tolerance mode and doesn't support another?

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
dvrogozh: I managed to miss this, thank you.

dvrogozh:
@stasefremov : were your concerns addressed?

stasefremov:
My concerns for implementation details were addressed. But there is still concern regarding number of bits allocated for "FrameSizeTolerance" field. In Windows CAPS structure (created by driver arch) we have 1 bit for this field (and there is some reason behind this). I would like to keep Windows and Linux structures in sync. So my suggestion is to talk with driver uArch team and figure which is correct number of bits and why.

PR #474, opened 2018-08-02
[samples] Add standalone build scripts

onabiull:


samples/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+cmake_minimum_required( VERSION 2.6.2 FATAL_ERROR )
```
dvrogozh: I don't think that's a good idea to go below cmake required by mediasdk itself. And we check 2.8.5 on our side, right?

samples/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+project( mediasdk-samples )
+
+option( CMAKE_MFX_HOME "Where is Media SDK?" OFF )
+option( CMAKE_VTUNE_HOME "Where is VTune?" OFF )
+
+option( ENABLE_ITT "Build targets with ITT instrumentation support (requires VTune)?" OFF )
+
+option( ENABLE_SW "Build SW versions of the targets?" OFF )
```
dvrogozh: do we need this?

samples/build.pl
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/perl
```
dvrogozh: can we avoid this wrapper altogether, please? Just default configuration of cmake should be enough.

onabiull:
It is just for MSS, to provide consistent experience with previous releases. This change won't appear in master.

PR #479, opened 2018-08-05
[core] fix MFXClose()

akokoshn:
-Add waitong for task complation for all registred components

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_session.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_session.cpp
@@ -601,6 +601,29 @@ void _mfxSession::Clear(void)
 
 void _mfxSession::Cleanup(void)
 {
+    // wait until all task are processed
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_session.cpp
```diff
@@ -601,6 +601,29 @@ void _mfxSession::Clear(void)
 
 void _mfxSession::Cleanup(void)
 {
+    // wait untill all task are processed
```
onabiull: Small typo in a comment. 
I wonder if we should rearrange wait sequence. In a typical transcode with vpp scenario, decode tasks would be stopped first, then vpp, then encode. @dvrogozh, @artem-shaporenko, @eistarov do you see any point in sorting wait sequence?

onabiull:
@akokoshn could you please cherry-pick this one to mss2018_r2 ?

PR #480, opened 2018-08-06
[TOOLS] Building process for val tools was removed from default build

romka0017:
If you want to build tools, you should add --tools=on for build_mfx.pl


onabiull:
@adydychk could you please add -DENABLE_TOOLS=ON to CI?

adydychk:
Ok, but it needs to make some manual checks with this parameter but without this PR.

adydychk:
@onabiull CI updated. Updated all builders except android and clang builders.

CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ if( CMAKE_SYSTEM_NAME MATCHES Linux )
 endif()
 
 option( ENABLE_ITT "Build targets with ITT instrumentation support (requires VTune)?" OFF )
+option( ENABLE_TOOLS "Build targets with additional tools" OFF )
```
dvrogozh: How about the code:
```
if (ENABLE_TOOLS)
  add_subdirectory(${CMAKE_HOME_DIRECTORY}/tools)
endif
```
right after line 66 in this file below. In that way we will avoid any changes and checks in the tools/* cmake files...

tools/asg-hevc/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if (NOT ENABLE_TOOLS)
```
dvrogozh: I believe this change is not needed - see above comment

tools/bs_parser_hevc/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if (NOT ENABLE_TOOLS)
```
dvrogozh: I believe this change is not needed - see above comment

tools/bs_parser_hevc/tools/hevc_fei_extractor/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if (NOT ENABLE_TOOLS)
```
dvrogozh: I believe this change is not needed - see above comment

tools/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_subdirectory(asg-hevc)
```
dvrogozh: Copyright header is missed. Please, add.

tools/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
...
+add_subdirectory(bs_parser_hevc)
+add_subdirectory(bs_parser_hevc/tools/hevc_fei_extractor)
```
dvrogozh: add new line at and of file

tools/bs_parser_hevc/CMakeLists.txt
```diff
@@ -7,4 +7,4 @@ file( GLOB_RECURSE sources "${CMAKE_CURRENT_SOURCE_DIR}/src/*.c" "${CMAKE_CURREN
 set( defs " -DMFX_VERSION_USE_LATEST " )
 
 make_library( bs_parser_hevc_static none static )
-set( defs "" )
+set( defs "" )
```
dvrogozh: add new line at end of file

Issue #484, opened 2018-08-06
MediaSDK fails to build with glibc 2.28


dbermond:
I'm getting the following error when trying to build MediaSDK with the newly released glibc 2.28:
```
[ 79%] Building CXX object samples/metrics_monitor/CMakeFiles/test_monitor.dir/test/igt_load.cpp.o
/storage/linux/abs/intel-media-sdk-git/src/MediaSDK/samples/metrics_monitor/test/igt_load.cpp:30:10: fatal error: libio.h: No such file or directory
 #include <libio.h>
          ^~~~~~~~~
compilation terminated.
make[2]: *** [samples/metrics_monitor/CMakeFiles/test_monitor.dir/build.make:76: samples/metrics_monitor/CMakeFiles/test_monitor.dir/test/igt_load.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2072: samples/metrics_monitor/CMakeFiles/test_monitor.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

- It was building fine with glibc 2.27.

- glibc 2.28 no longer provides the header `libio.h`.

**System Information:**
**OS:** Arch Linux
**Compiler:** gcc 8.2.0
**glibc:** 2.28
**libva:** git master
**MediaSDK:** git master

dvrogozh:
libio.h looks very strange. I guess this should be stdio.h.

dvrogozh:
@dbermond : could you, please, try the above fix? I don't have your environment, so can't be 100% sure...

dbermond:
@dvrogozh Thank you for the reply and for working on this issue.

I can confirm that you patch fixes it.

Tested in MediaSDK git master and in stable release 2018Q2.1. Both versions are building fine with glibc 2.28.

dvrogozh:
Cool. @onabiull : can you, please, merge and cherry-pick to 2018Q2.1?

onabiull:
Done: https://github.com/Intel-Media-SDK/MediaSDK/releases/tag/MediaSDK-2018-Q2.2

PR #485, opened 2018-08-06
metrics_monitor: use stdio.h instead of libio.h in the tests

dvrogozh:
Fixes: #484


dbermond:
This works fine for me in fixing the build issue with glibc 2.28.

PR #500, opened 2018-08-09
[HEVCe][LINUX]Added profiles for rext

iveselov:


iveselov:
incorrect deffinitions

PR #502, opened 2018-08-09
2 more enable options: BUILD_RUNTIME and BUILD_SAMPLES

dvrogozh:
To be honest I don't like who we named these options. They should be BUILD_TOOLS, BUILD_SAMPLES, etc. I remained with ENABLE_* just for consistency with ENABLE_TOOLS. @onabiull : do we have consensus to rename these to BUILD_*? I will update PR then...


dvrogozh:
This PR was merged as part of https://github.com/Intel-Media-SDK/MediaSDK/pull/503.

PR #503, opened 2018-08-09
Do not build mfx_reflect if we do not need to and other trace fixes

dvrogozh:
That's one more step forward towards cleaner build and fastboot support. We actually did not use mfx_reflect on Linux (requires textlog to be enabled), but we compiled and linked against it. This resulted in an increase of build time for ~20 seconds and increase of stripped/unstrapped binaries. Change in library sizes is:
* Regular build (stripped): 10729880 -> 10168600 (5%); and non-stripped: 46026560 -> 39843840 (15%)
* Fastboot (striped): 2963864 -> 2406680 (23%)


dvrogozh:
Makes sense. Indeed, reflector mechanism should be under CI or we will lose it with uAPI updates. I will add ENABLE_TEXTLOG today. Let's finalize the PR early next week.

_studio/mfx_lib/CMakeLists.txt
```diff
@@ -164,7 +164,6 @@ endif()
```
dvrogozh: That's the first patch of this PR since all prev. 5 patches actually belong to another PR: https://github.com/Intel-Media-SDK/MediaSDK/pull/502. I based current PR on that one since they significantly overlap in top level cmake and readme.

PR #489, opened 2018-08-07
Replace UMC_ENABLE_* defines with MFX_ENABLE_*

IlyaSergeyev:
1) removed AUDIO defs
2) UMC defs changed to MFX
3) implemented minimal mfxconfig.h.in
4) several steps in moving definitions from mfx_config.h to mfxconfig.h.in/cmake options

PR #506, opened 2018-08-10
Replace UMC_ENABLE_* defines with MFX_ENABLE_* 

IlyaSergeyev:
I can't understand why this one #489 fails tests. Commit 06cc9ea is a diff between #506 and #489. I will squash commits after review.


IlyaSergeyev:
Name "do not remove changed" @dvrogozh

_studio/shared/umc/codec/h264_enc/include/umc_h264_common.h
```diff
@@ -19,7 +19,6 @@
 // SOFTWARE.
 
 #include "umc_defs.h"
-#if defined(UMC_ENABLE_H264_VIDEO_ENCODER)
```
dvrogozh: Why this was removed rather than changed to MFX_ENABLE_...?

_studio/shared/include/mfx_config.h
```diff
@@ -115,14 +115,11 @@
...
-
-        // aac
-        #define MFX_ENABLE_AAC_AUDIO_DECODE
-
-        //mp3
-        #define MFX_ENABLE_MP3_AUDIO_DECODE
-
-        // av1
+        // user plugin for decoder, encoder, and vpp
+        #define MFX_ENABLE_USER_DECODE
```
dvrogozh: This should not be here. Looks to be remnant of some rework. Remove.

_studio/shared/umc/codec/h264_enc/include/umc_h264_common.h
IlyaSergeyev: It causes compilation error in mfx_h264_enc_common_hw.cpp (UMC_H264_ENCODER’ has not been declared) without enable option for h264 encoder patch. MFX_ENABLE_H264_VIDEO_ENCODE has to be defined somewhere.

dvrogozh: Ok, let's wait for the #516 then.

IlyaSergeyev:
Rebased

_studio/shared/umc/codec/brc/src/umc_h264_brc.cpp
```diff
@@ -23,6 +23,8 @@
 #include "umc_video_data.h"
 #include "umc_h264_common.h"
 
+#if defined(MFX_ENABLE_H264_VIDEO_ENCODE)
```
IlyaSergeyev: In case of fastboot we have not defined namespace UMC_H264_ENCODER.

PR #517, opened 2018-08-14
[bs_parser_hevc] Support for I-slices with enabled IBC

ndomnina:
Standard allows I-slices with enabled IBC, which is effectively ignored.
Skip ref-list modification for I-slices to avoid exception.


mgonchar:
@eistarov could you please have a look?

PR #521, opened 2018-08-15
[sample_hevc_fei] Fix missed or incorrect setting cli options

otsirule:
- Restored missing AdaptiveSearch in sample_hevc_fei_abr because it should be enabled when compared to legacy

- Set SearchWindow for PreENC separately for more accurate testing of this control's influence on HEVC FEI performance and quality


adydychk:
Test failed because of infrastructure reasons. We are working on fix.

PR #523, opened 2018-08-16
Delete hardcode value for VAConfigAttribRTFormat

dmenshov:
- Delete hardcode value for VAConfigAttribRTFormat because
function vaGetConfigAttributes get valid params for
attributes.


dmenshov:
I think you should do the same for AVC and MPEG2.

PR #530, opened 2018-08-17
[sample_hevc_fei_abr] Added wait in GetSurface

mgonchar:
This fixes possible synchronization issues in situations when decoder is too fast and actually consumes all surfaces from pool


mgonchar:
@onabiull looks like it worked

PR #533, opened 2018-08-20
Fixes android build

IlyaSergeyev:
mfx_android_defs.h hasn't MFX_ENABLE_VP9_VIDEO_DECODE,
,that was in mfx_config.h and is needed for NUM_REF_FRAMES

Other variant is a changing ifdef MFX_ENABLE_VP9_VIDEO_DECODE in umc_vp9_dec_defs.h
to MFX_ENABLE_VP9_VIDEO_DECODE_HW


dvrogozh:
The problem is that there is still no android build in the list of checks. @adydychk : can you, please, trigger it?

adydychk:
This case is under investigation

onabiull:
Looks like this fix already merged

IlyaSergeyev:
1ec56c1 already fixed this issue

PR #551, opened 2018-08-23
WA [HEVCe] Added internal sync

iveselov:


iveselov:
We got sporadic core dumps with any scenario for HEVCe. Looks like race condition. So I changed sync mechanism.

PR #552, opened 2018-08-23
Fixes 'invalid use of non-static data member' error on clang 6.0

IlyaSergeyev:


dvrogozh:
Why no CI build checks were ran? This change breaks build against gcc-8.1 with:
`MediaSDK/tools/bs_parser_hevc/include/bs_mem+.h:56:19: error: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘struct BS_HEVC2::TU’; use assignment or value-initialization instead [-Werror=class-memaccess]
`

mgonchar:
@dvrogozh  this patch changed code which has no relation to bs_parser.

Bs_parser just missing some changes which were made in same code for parser in sample_hevc_fei_abr (that sample has it's own copy of bs_parser and code between those two somehow misaligned)

dvrogozh:
Yeah... Oleg noted to me that this change could not introduce this issue and instead we had misconfigured CI which tested something else rather than gcc-8.1... Well, bad things happen. But now we have a blocker issue.

GoAlexander:
We used gcc 7 instead of gcc 8... But now it is fixed.

PR #558, opened 2018-08-23
Add Media SDK Tracer code

onabiull:


tools/tracer/config/config.cpp
```diff
@@ -0,0 +1,193 @@
...
+        }
+        return path;
+    }
+#endif
+
+Config* Config::conf = NULL;
+
+Config::Config()
+{
+    const char* home =
```
mgonchar: who will delete dynamic memory returned from get_path_reg() (on windows)?

onabiull: Windows support out of scope for this branch. But code in question contains memory leak. Not that critical, considering debugging nature of the tracer, but still should be fixed.

PR #560, opened 2018-08-24
Certain changes in linux mediasdk dispatcher

dvrogozh:
This change refactors mediasdk dispatcher. Dispatcher for Linux is
separated out. Linux operating system implies certain things which
can and can not be done. Previous  version of the dispatcher did not
quite satisfy Linux requirements for integration into the OS.

Key changes to pay attention on:
 * Audio support is dropped, user will get MFX_ERR_UNSUPPORTED or
 MFX_ERR_NOT_FOUND depending on the uAPI
 * Video library is loaded by either libmfxhw64.so.1 or libmfxsw64.so.1
 depending on requested implementation
 * Search for the libraries is done in this order: first libraries are
 searched in system paths, the MFX_MODULES_DIR passed during compilation
 time is favored
 * All plugins, including developer plugins, should be centrally
 registered in MFX_PLUGINS_DIR/plugins.cfg. MFX_PLUGINS_DIR is passed
 during compilation time.
 * Having custom plugins side-by-side to applicaiton is not allowed.


api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+  } else {
+    *session = nullptr;
+  }
+
+  return mfx_res;
+}
+
+mfxStatus MFXClose(mfxSession session)
+{
+  std::unique_ptr<MFX::LoaderCtx> loader((MFX::LoaderCtx*)session);
```
mgonchar: why do you wrapped it to unique_ptr at all? somebody allocated memory and you need to clean up it? if so, why don't do it if Close succeed? Looks tricky

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+  }
+  return mfx_res;
+}
+
+mfxStatus MFXVideoUSER_Load(mfxSession session, const mfxPluginUID *uid, mfxU32 version)
+{
+  MFX::LoaderCtx* loader = (MFX::LoaderCtx*)session;
+  std::string path;
+
+  if (!loader) return MFX_ERR_INVALID_HANDLE;
```
mgonchar: I think it is better to do parameters check before any other manipulation

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+      return std::find_if(
+        MFX::g_GlobalCtx.m_plugins.begin(),
+        MFX::g_GlobalCtx.m_plugins.end(),
+        [&puid](MFX::PluginInfo& item){ return item.getUID() == puid; }
+        );
+    };
+
+    // checking whether we have already cached this plugin for this application
+    auto it = find_uid(*uid);
+    if (it == MFX::g_GlobalCtx.m_plugins.end() && !MFX::g_GlobalCtx.m_plugins.size()) {
```
mgonchar: use g_GlobalCtx.m_plugins.empty() to check that container is (isn't) empty

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+  mfxStatus mfx_res = MFXInit(loader->getImpl(), &version, clone);
+  if (MFX_ERR_NONE != mfx_res) {
+    return mfx_res;
+  }
+
+  // join the sessions
+  mfx_res = MFXJoinSession(session, *clone);
+  if (MFX_ERR_NONE != mfx_res) {
+    MFXClose(*clone);
+    *clone = NULL;
```
mgonchar: nullptr

api/opensource/mfx_dispatch/linux/mfxparser.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <list>
+
+#include "mfxloader.h"
+
+namespace MFX {
+
+static bool parseGUID(const char* src, mfxPluginUID* uid)
+{
+    mfxPluginUID plugin_uid{};
+    mfxU8* p = uid->Data;
```
mgonchar: check uid pointer first

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+{
+  std::mutex m_mutex;
+  std::list<PluginInfo> m_plugins;
+};
+
+static GlobalCtx g_GlobalCtx;
+
+std::shared_ptr<void> make_dlopen(const char* filename, int flags)
+{
+  return std::shared_ptr<void>(
```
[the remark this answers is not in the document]
dvrogozh: std::shared_ptr is copy assignable.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+  inline mfxIMPL getImpl() {
+    return m_implementation;
+  }
+
+  inline mfxVersion getVersion() {
+    return m_version;
+  }
+
+private:
+  std::shared_ptr<void> m_dlh;
```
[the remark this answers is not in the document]
dvrogozh: You know, as of now it is not really shared. However, it is done via std::shared_ptr on purpose (I considered and tried std::unique_ptr) because it is more difficult to wrap it in std::unique_ptr because of the difference in their implementations. Essentially, you will be forced to declare:
`typedef std::unique_ptr<void,  int (*)(void *)> unique_dlopen_ptr;`
and make other changes which are simply hidden by std::shared_ptr. In other words, std::shared_ptr seems more easy to use for this particular purpose to smart-wrap dlopen.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+mfxStatus LoaderCtx::Close()
+{
+  auto proc = (decltype(MFXClose)*)m_table[eMFXClose];
+  mfxStatus mfx_res = (proc)? (*proc)(m_session): MFX_ERR_NONE;
+
+  m_implementation = {};
+  m_version = {};
+  m_session = nullptr;
+  memset(&m_table, 0, sizeof(m_table));
```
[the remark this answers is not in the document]
dvrogozh: ok

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+    }
+  } catch(...) {
+    return MFX_ERR_UNKNOWN;
+  }
+
+  mfxStatus mfx_res = MFX_ERR_NONE;
+  std::list<PluginCtx> ctx;
+
+  try {
+    ctx.emplace_front(PluginCtx(*this));
```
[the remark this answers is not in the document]
dvrogozh: ok

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+  return MFX_ERR_NONE;
+}
+
+mfxStatus LoaderCtx::UnloadPlugin(const mfxPluginUID& uid)
+{
+  mfxStatus mfx_res = MFX_ERR_NONE;
+
+  try {
+    std::list<PluginCtx> ctx;
```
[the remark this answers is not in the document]
dvrogozh: It is complicated on purpose, I think you did not catch the idea which is to move bottom half of plugin ctx destroy (which may contain potentially long dlclose) out of the mutex protection. I will add the comment and will use std::find_if instead of the loop, but splice() will remain.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+mfxStatus MFXInit(mfxIMPL impl, mfxVersion *ver, mfxSession *session)
+{
+    mfxInitParam par;
+
+    memset(&par, 0, sizeof(par));
```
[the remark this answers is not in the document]
dvrogozh: yep.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,582 @@
...
+    par.Implementation = impl;
+    if (ver) {
+        par.Version = *ver;
+    } else {
+        par.Version.Major = MFX_VERSION_MAJOR;
```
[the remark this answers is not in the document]
dvrogozh: vice versa, but ok.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
dvrogozh: yep. thank you for the catch.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
dvrogozh: ok

api/opensource/mfx_dispatch/linux/mfxparser.cpp
dvrogozh: this function will/should never be called with null src or uid, check omitted on purpose

api/opensource/mfx_dispatch/linux/mfxloader.cpp
dvrogozh: I don't want to use 'delete' directly:).

api/opensource/mfx_dispatch/linux/mfxloader.cpp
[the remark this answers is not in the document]
mgonchar: yes, I know. It's generally same question about whether this resource is really shared or not. You gave pretty detailed explanation in previous comment

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,583 @@
...
+  } else if (par.Implementation & MFX_IMPL_SOFTWARE) {
+    libs.emplace_back(LIBMFXSW);
+    libs.emplace_back(MFX_MODULES_DIR "/" LIBMFXSW);
+  } else {
+    return MFX_ERR_UNSUPPORTED;
+  }
+
+  mfxStatus mfx_res = MFX_ERR_UNSUPPORTED;
+
+  for (auto lib: libs) {
```
mgonchar: these strings may be quite large, so I think better to use references here:

for (auto& libs: libs)

api/opensource/mfx_dispatch/linux/mfxloader.cpp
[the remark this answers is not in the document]
mgonchar: ok, got it

api/opensource/mfx_dispatch/linux/mfxloader.cpp
mgonchar: Ok, now I see how it works. Memory was allocated at line 414

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,583 @@
...
+  }
+
+  return mfx_res;
+}
+
+mfxStatus MFXClose(mfxSession session)
+{
+  std::unique_ptr<MFX::LoaderCtx> loader((MFX::LoaderCtx*)session);
+
+  if (!loader) return MFX_ERR_INVALID_HANDLE;
```
mgonchar: if you check the raw pointer before wrapping it to unique_ptr you will save one constructor call

_studio/shared/src/libmfx_core.cpp
```diff
@@ -1348,6 +1348,8 @@ mfxStatus CoreDoSWFastCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc, int c
         MFX_CHECK_STS(sts);
 
         roi.height >>= 1;
+        printf(">>> dvrogozh: pDst->Data.UV=%p, dstPitch=%d, pSrc->Data.UV=%p, srcPitct=%d\n",
```
mgonchar: looks like debug print, which you forgot to remove

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,583 @@
...
+
+mfxStatus LoaderCtx::Close()
+{
+  auto proc = (decltype(MFXClose)*)m_table[eMFXClose];
+  mfxStatus mfx_res = (proc)? (*proc)(m_session): MFX_ERR_NONE;
+
+  m_implementation = {};
+  m_version = {};
+  m_session = nullptr;
+  std::fill_n(m_table, sizeof(m_table)/sizeof(m_table[0]), nullptr);
```
mgonchar: I suggest to use c++11 power instead :)

std::fill(std::begin(m_table), std::end(m_table), nullptr)

api/opensource/mfx_dispatch/linux/mfxloader.h
```diff
@@ -0,0 +1,125 @@
...
+{
+  uint8_t* data = reinterpret_cast<uint8_t*>(&id);
+  std::stringstream ss;
+  for (size_t i=0; i < sizeof(id); ++i) ss << data[i];
+  return ss.str();
+}
+
+static void print(std::list<PluginInfo>& plugins)
+{
+  for (auto plg: plugins) {
```
mgonchar: PluginInfo may be a big object, so better to use reference

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,577 @@
...
+
+  for (auto& lib: libs) {
+    std::shared_ptr<void> hdl = make_dlopen(lib.c_str(), RTLD_LOCAL|RTLD_NOW);
+    if (hdl) {
+      do {
+        /* Loading functions table */
+        bool wrong_version = false;
+        for (int i = 0; i < eFunctionsNum; ++i) {
+          assert(i == g_mfxFuncTable[i].id);
+          m_table[i] = dlsym(hdl.get(), g_mfxFuncTable[i].name);
```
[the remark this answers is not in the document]
dvrogozh: no, it is not. The goal of the dispatcher is to properly load a library of the _lower_ version than the one used to build the dispatcher itself. In that way some functions will be missed, but application will be able to work if it requests the API for which all functions can be found.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,577 @@
...
+  }
+  if (par.Implementation & MFX_IMPL_AUDIO) {
+    return MFX_ERR_UNSUPPORTED;
+  }
+
+  std::vector<std::string> libs;
+
+  if (MFX_IMPL_BASETYPE(par.Implementation) == MFX_IMPL_AUTO ||
+      MFX_IMPL_BASETYPE(par.Implementation) == MFX_IMPL_AUTO_ANY) {
+    libs.emplace_back(LIBMFXHW);
```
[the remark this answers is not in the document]
dvrogozh: This is a tiny optimization which I believe hardly cost it.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,577 @@
...
+  return MFXVideoUSER_Unregister((mfxSession)&m_loader, m_plugin_param.Type);
+}
+
+mfxStatus LoaderCtx::LoadPlugin(const mfxPluginUID &uid, mfxU32 version, const char *path)
+{
+  if (!path) return MFX_ERR_NULL_PTR;
+
+  std::lock_guard<std::mutex> lock(m_guard);
+
+  for (auto it: m_plugins) {
```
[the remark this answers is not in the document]
dvrogozh: ok

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,577 @@
...
+{
+  std::list<PluginCtx> ctx;
+  {
+    // We will move plugin ctx which we are going to delete to the
+    // array allocated on stack. In this way we will move bottom half
+    // of plugin ctx destroy, including potentially long dlclose, out
+    // of the mutex.
+    std::lock_guard<std::mutex> lock(m_guard);
+    auto it = std::find_if(std::begin(m_plugins), std::end(m_plugins),
+      [uid](const PluginCtx& cur_ctx){ return cur_ctx.getUID() == uid; });
```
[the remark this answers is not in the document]
dvrogozh: ok. But actually there is a sense which is to avoid pointer, and potential need to check it for null.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,577 @@
...
+        return std::find_if(
+          MFX::g_GlobalCtx.m_plugins.begin(),
+          MFX::g_GlobalCtx.m_plugins.end(),
+          [&puid](MFX::PluginInfo& item){ return item.getUID() == puid; }
+          );
+      };
+
+      // checking whether we have already cached this plugin for this application
+      auto it = find_uid(*uid);
+      if (it == MFX::g_GlobalCtx.m_plugins.end() && MFX::g_GlobalCtx.m_plugins.empty()) {
```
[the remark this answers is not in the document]
dvrogozh: No, condition is correct. The idea was that that if we don't have plugins registry information available, then we should read it.

However, this place was written when I thought that we will support few locations where plugins could be registered. But now only one is left. As a result this place is over complicated. I will simplify.

api/opensource/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -0,0 +1,577 @@
...
+      path = it->getPath();
+    }
+
+    return loader->LoadPlugin(*uid, version, path.c_str());
+  } catch(...) {
+    return MFX_ERR_MEMORY_ALLOC;
+  }
+}
+
+mfxStatus MFXVideoUSER_LoadByPath(mfxSession session, const mfxPluginUID *uid, mfxU32 version, const mfxChar *path, mfxU32 len)
```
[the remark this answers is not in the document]
dvrogozh: yep

api/opensource/mfx_dispatch/linux/mfxloader.h
dvrogozh: yep

chivakker:
lgtm, please include this patch in the release.

dvrogozh:
@mgonchar : source code did not change at all. I have moved a code a little bit:
- Dispatcher moved to api/mfx_dispatch (opensource folder dropped)
- Linux specific files including CMakeLists.txt, *.pc files moved to api/mfx_dispatch/linux folder

PR #574, opened 2018-08-27
fastboot: VPP functions defines moved into mfxconfig.h.in

IlyaSergeyev:
VPP usage flags moved into mfxconfig.h.in and ConfTargets.cmake

We'll got reducing of library size by switching off VPP at all. But, unfortunately, switching off separate functions don't give significant effect.


IlyaSergeyev:
default config: 10125K
fastboot config (denoise, composition and rotation are on) : 1358K
fastboot without vpp: 1034K

dvrogozh:
what is the MCTF impact? Can you, please, disable only it and provide a binary size?

IlyaSergeyev:
Almost have no effect to binary size

PR #579, opened 2018-08-28
gcc8 warning 'specified bound depends on the length of the source'

IlyaSergeyev:
Fixed gcc8 warning 'specified bound depends on the length of the source argument'
First - Why was strncpy (not memcpy) used here? We exactly knows number of bytes to copy ...
Second - code still dangerous, we rely on the fact that the callers understands what it does
Maybe it should be replaced with simple **snprintf(to, to_size, "%s", from)** ?

(And, I think, this c-style code can be removed 376a72044859cdb42ecd19f096d0f5f91936ef60 )

### DIFF
--- a/api/opensource/mfx_dispatch/include/mfx_dispatcher_defs.h
+++ b/api/opensource/mfx_dispatch/include/mfx_dispatcher_defs.h
@@ -21,6 +21,7 @@
 #pragma once
 #include "mfxdefs.h"
 #include <cstring>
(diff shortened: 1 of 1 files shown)


dvrogozh:
Aha, here is the reason. Thank you. Which OS are you using?

IlyaSergeyev:
Debian. I added one more commit here, that switches on treating warnings as errors.

PR #498, opened 2018-08-09
Fixed android warnings

Losev-Ivan:

PR #580, opened 2018-08-28
clang 6.0: fixed warning: 'operator delete' has a non-throwing exception specification

IlyaSergeyev:
THROWSEXCEPTION is empty and used only in umc_vc1_dec_task_store.h
Tested on gcc 8.2 and clang 6.0, not sure about icc.

### DIFF
--- a/_studio/shared/umc/codec/vc1_dec/include/umc_vc1_dec_task_store.h
+++ b/_studio/shared/umc/codec/vc1_dec/include/umc_vc1_dec_task_store.h
@@ -173,28 +173,6 @@ namespace UMC
         VC1TaskStore & operator = (const VC1TaskStore &);
 
     public:
-
-        void* operator new(size_t size, void* p)
-        {
-            if (!p)
-                throw VC1Exceptions::vc1_exception(VC1Exceptions::mem_allocation_er);
-            return new(p) uint8_t[size];
-        };
-
-        // external memory management. No need to delete memory
-        void operator delete(void *p) THROWSEXCEPTION
-        {
-            //Anyway its incorrect when we trying free null pointer
-            if (!p)
-                throw VC1Exceptions::vc1_exception(VC1Exceptions::mem_allocation_er);
-        };
-
-        void operator delete(void *, void *) THROWSEXCEPTION
-        {
-            // delete for system exceptions case
-            throw VC1Exceptions::vc1_exception(VC1Exceptions::mem_allocation_er);
-        };
-
         virtual bool     Init(uint32_t iConsumerNumber,
                       uint32_t iMaxFramesInParallel,
                       VC1VideoDecoder* pVC1Decoder);
--- a/_studio/shared/umc/codec/vc1_dec/include/umc_vc1_video_decoder_hw.h
+++ b/_studio/shared/umc/codec/vc1_dec/include/umc_vc1_video_decoder_hw.h
@@ -93,7 +93,6 @@ namespace UMC
                 // special header (with frame size) in case of .rcv format
                 SCoffset = -VC1FHSIZE;
 
-            m_pStore = m_pStore;
             Status umcRes = UMC_ERR_NOT_ENOUGH_DATA;
             try
             {
--- a/_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder_hw.cpp
+++ b/_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder_hw.cpp
@@ -74,8 +74,7 @@ Status VC1VideoDecoderHW::Init(BaseCodecParams *pInit)
     try
     // memory allocation and Init all env for frames/tasks store - VC1TaskStore object
     {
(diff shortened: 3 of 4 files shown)


Losev-Ivan:
This problem already solved in [PR498](https://github.com/Intel-Media-SDK/MediaSDK/pull/498) first commit, file _studio/shared/umc/core/umc/include/umc_defs.h

_studio/shared/umc/codec/vc1_dec/include/umc_vc1_dec_task_store.h
```diff
@@ -182,17 +182,16 @@ namespace UMC
         };
 
         // external memory management. No need to delete memory
-        void operator delete(void *p) THROWSEXCEPTION
+        void operator delete(void *p)
         {
             //Anyway its incorrect when we trying free null pointer
-            if (!p)
-                throw VC1Exceptions::vc1_exception(VC1Exceptions::mem_allocation_er);
+            VM_ASSERT(p);
```
Losev-Ivan: Please, add (void)p; after VM_ASSERT that not to generate new warnings

dvrogozh: yep.

onabiull: Btw, comment here contradict standard. Deleting null pointer is ok and defined.

_studio/shared/umc/codec/vc1_dec/include/umc_vc1_dec_task_store.h
```diff
@@ -182,17 +182,17 @@ namespace UMC
...
+            (void) p;
         };
 
-        void operator delete(void *, void *) THROWSEXCEPTION
+        void operator delete(void *, void *)
         {
             // delete for system exceptions case
-            throw VC1Exceptions::vc1_exception(VC1Exceptions::mem_allocation_er);
+            VM_ASSERT(0);
```
dvrogozh: I also like this operator: looks like we have an operator which should _never_ be called:). Can't that be handled on compilation level via compile errors?

dvrogozh: In this case we should declare and not define, like this:
```
// By intent we prohibit this type of delete
void operator delete(void *, void *);
```

In that case we will get linker level error and avoid catching this at runtime.

_studio/shared/umc/codec/vc1_dec/include/umc_vc1_dec_task_store.h
```diff
@@ -182,14 +182,11 @@ namespace UMC
...
-        void operator delete(void *p) THROWSEXCEPTION
+        void operator delete(void *)
         {
-            //Anyway its incorrect when we trying free null pointer
-            if (!p)
-                throw VC1Exceptions::vc1_exception(VC1Exceptions::mem_allocation_er);
         };
 
-        void operator delete(void *, void *) THROWSEXCEPTION
+        void operator delete(void *, void *) noexcept (false)
```
dvrogozh: We start to circle thru changes! Why this suddenly again?! In prev. comment you said that this operator is not actually used at all. I was expecting to see this line:
`void operator delete(void *, void *);`
I.e. declaration without implementation.

IlyaSergeyev:
@dvrogozh  The linker wants the "operator delete(void *, void *)" to be implemented, and causes linkage error if it is not, but do not calls it (I have checked it in gdb). I do not understand why it possible.

_studio/shared/umc/codec/vc1_dec/include/umc_vc1_dec_task_store.h
dvrogozh: @IlyaSergeyev : you actually need to write this:
`void operator delete(void *, void *) = delete;`
In this way linker will be happy. Sorry for misleading in prev. comment, I thought you will catch an idea.

IlyaSergeyev:
With "void operator delete(void *, void *) = delete;" I've get "attempt to use a deleted function" in any case of creating of VC1TaskStore object.

dvrogozh:
Hm. I think I tried only on gcc to compile this variant. Let me try clang. If it says this function is in use somewhere. Does compiler give any hint where? Please, post the full error/notes log here.

dvrogozh:
Ok, there is definitely difference between gcc and clang here.... I am trying to read description of new and delete operators: https://en.cppreference.com/w/cpp/memory/new/operator_new (see 9) and https://en.cppreference.com/w/cpp/memory/new/operator_delete (see 13) and I actually don't understand why we redefine all these operators at the first place? Can we simply delete all 3 including new and rely on standard library implementation?

dvrogozh:
@IlyaSergeyev : any progress on this one?

IlyaSergeyev:
@dvrogozh : I think, we need to use ifdef __clang__ here

dvrogozh:
I also checked what I have proposed and it worked for me. We are not in sync here. Please, schedule a f2f session and let's debug together.

PR #583, opened 2018-08-29
Correct values for Color420Only, YUV422ReconSupport, YUV444ReconSupport

dmenshov:
- we can init valid parameter
of caps from driver
Color420Only, YUV422ReconSupport, YUV444ReconSupport

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -792,7 +792,6 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     if (p.CodeName >= MFX_PLATFORM_SKYLAKE)
     {
-        m_caps.Color420Only       = 1;
         m_caps.BitDepth8Only      = 1;
         m_caps.MaxEncodedBitDepth = 0;
         m_caps.YUV422ReconSupport = 0;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -792,7 +792,6 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     if (p.CodeName >= MFX_PLATFORM_SKYLAKE)
     {
-        m_caps.Color420Only       = 1;
         m_caps.BitDepth8Only      = 1;
```
dmitryermilov: why does BitDepth8Only remain 1?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -833,9 +826,17 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     m_caps.UserMaxFrameSizeSupport = 1;
     m_caps.MBBRCSupport            = 1;
     m_caps.MbQpDataSupport         = 1;
-    m_caps.Color420Only            = 1; // FIXME in case VAAPI direct YUY2/RGB support added
     m_caps.TUSupport               = 73;
 
+    if (attrs[idx_map[VAConfigAttribRTFormat]].value == VA_RT_FORMAT_YUV420)
```
dmitryermilov: 1) Why here it's "==" while below "&"?
2) Can you please re-write it like below. It will be easier to read.
if (attrs[idx_map[VAConfigAttribRTFormat]].value & VA_RT_FORMAT_YUV420)
{
}
else if (attrs[idx_map[VAConfigAttribRTFormat]].value & VA_RT_FORMAT_YUV422)
{
}
else if (attrs[idx_map[VAConfigAttribRTFormat]].value & VA_RT_FORMAT_YUV444)
{
}

dmenshov: attrs[idx_map[VAConfigAttribRTFormat]].value is a mask. So if attrs[idx_map[VAConfigAttribRTFormat]].value equal VA_RT_FORMAT_YUV420 | VA_RT_FORMAT_YUV422 | VA_RT_FORMAT_YUV444 then we init Color420Only but YUV422ReconSupport and YUV444ReconSupport  will be zero.

PR #586, opened 2018-08-30
Delete hardcode value for RollingIntraRefresh

dmenshov:
- we can get correct value for RollingIntraRefresh
from driver

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -829,7 +829,8 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     m_caps.VCMBitRateControl =
         attrs[ idx_map[VAConfigAttribRateControl] ].value & VA_RC_VCM ? 1 : 0; //Video conference mode
-    m_caps.RollingIntraRefresh     = 0; /* (attrs[3].value & (~VA_ATTRIB_NOT_SUPPORTED))  ? 1 : 0*/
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -829,7 +829,8 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     m_caps.VCMBitRateControl =
         attrs[ idx_map[VAConfigAttribRateControl] ].value & VA_RC_VCM ? 1 : 0; //Video conference mode
-    m_caps.RollingIntraRefresh     = 0; /* (attrs[3].value & (~VA_ATTRIB_NOT_SUPPORTED))  ? 1 : 0*/
+    m_caps.RollingIntraRefresh =
```
dmitryermilov: This is not right. See https://github.com/intel/media-driver/blob/a13bda0ce112baa0175c455d9a80f63139c18b07/media_driver/linux/gen9/ddi/media_libva_caps_g9.cpp#L75

dvrogozh: Why this is not right? We will get the real status of the feature in the driver: it is _not_ supported:)! I.e. the change is correct and if you think there is a bug in the driver - go on and create it. Moreover,  the hardcoded cap looks that we don't support it in mediasdk as well...

dmitryermilov: VA_ENC_INTRA_REFRESH_NONE  is zero. No concerns here

PR #589, opened 2018-08-30
fastboot: Enable MCTF option moved to cmake files

IlyaSergeyev:
Binary size
with MFX_ENABLE_MCTF==ON : 1362K
with MFX_ENABLE_MCTF==OFF : 1305K


builder/ConfTargets.cmake
```diff
@@ -76,6 +76,14 @@ endif()
 
 option( MFX_ENABLE_KERNELS "Build with advanced media kernels support?" ON )
 
+if ( ${version} STRGREATER "1025" )
```
dvrogozh: Please, use ${API_VERSION} here instead of ${version}. And use standard version check in cmake like VERSION_GREATER. Mind that not all version checks are supported in cmake-2.8 which we follow now. I.e. your check should look like:
if (${API_VERSION} VERSION_GREATER 1.24)

builder/ConfTargets.cmake
```diff
@@ -76,6 +76,14 @@ endif()
...
+  set ( MFX_MCTF_OPTION_ALLOWED ON )
+else()
+  set ( MFX_MCTF_OPTION_ALLOWED OFF )
```
dvrogozh: I suggest to avoid MCTF in the name. We may have other options which appear from 1.25. I.e. better to name in this way:
`set( API_1_25_OPTIONS ON)`

And please, move this set _before_ defining any options...

builder/ConfTargets.cmake
```diff
@@ -76,6 +76,14 @@ endif()
...
+endif()
+
+option( MFX_ENABLE_MCTF "Build with MCTF support?"  ${MFX_MCTF_OPTION_ALLOWED} )
```
dvrogozh: move below after all codec options

builder/ConfTargets.cmake
```diff
@@ -76,6 +76,13 @@ endif()
 
 option( MFX_ENABLE_KERNELS "Build with advanced media kernels support?" ON )
 
+#if ( ${version} STRGREATER "1025" )
```
dvrogozh: remove comment

PR #591, opened 2018-08-30
[sample_fei] Improved description of DecodedOrder option

mgonchar:


samples/sample_fei/src/sample_fei.cpp
```diff
@@ -1067,9 +1069,13 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, AppConfig* pCon
 
     if (pConfig->DecodedOrder && pConfig->bDECODE && (pConfig->bPREENC || pConfig->bENCPAK || pConfig->bOnlyENC || pConfig->bOnlyPAK || pConfig->bENCODE)){
         if (bAlrShownHelp)
-            msdk_printf(MSDK_STRING("\nERROR: DecodedOrder turned on, all FEI interfaces in sample_fei expects frame in DisplayOrder!\n"));
+        {
```
lakulako: Strings concatenation like below would be more compact.

mgonchar: Yes, I agree with you. But we have too many code lines done in this way, so I won't fix it in this commit.

PR #592, opened 2018-08-30
[Android] Added ITT traces

DenWolf:


DenWolf:
@dmitrybrazhkin , could you please review this patch ?

Now Android build is broken and it's expected behavior, because ITT is not enabled to Android source tree.
Enabling ITT in progress and need to be done tomorrow

android/mfx_defs.mk
```diff
@@ -37,6 +37,13 @@ MFX_CFLAGS += \
   -Wformat -Wformat-security \
   -fexceptions -frtti
 
+# Enabled ITT traces by default
+MFX_ENABLE_ITT_TRACES := true
```
dvrogozh: While I am more or less ok to have itt support on Android, this line will not permit the user to enable/disable it on the command line, i.e.: `make MFX_ENABLE_ITT_TRACES=false`, while this definitely can and should be handled. Please, add a check whether the variable was already set, i.e. in command line, and setup default only if it was not set. Something like:
```
ifeq ($(MFX_ENABLE_ITT_TRACES),)
  MFX_ENABLE_ITT_TRACES := true
endif
```

DenWolf: Good point, agree

PR #593, opened 2018-08-30
Y410 allocator fix

vsidorow:
Set Y pointer in Y410 when Y410 surfaces is used,
as library heavily relies on Y pointer.

### DIFF
--- a/_studio/shared/src/libmfx_allocator.cpp
+++ b/_studio/shared/src/libmfx_allocator.cpp
@@ -415,8 +415,9 @@ mfxStatus mfxDefaultAllocator::LockFrame(mfxHDL pthis, mfxHDL mid, mfxFrameData 
     case MFX_FOURCC_Y410:
         ptr->PitchHigh = (mfxU16)((4 * ALIGN32(fs->info.Width)) / (1 << 16));
         ptr->PitchLow  = (mfxU16)((4 * ALIGN32(fs->info.Width)) % (1 << 16));
-        ptr->Y = ptr->U = ptr->V = ptr->A = 0;
         ptr->Y410 = (mfxY410*)sptr;
+        ptr->Y = (mfxU8*)(ptr->Y410);
+        ptr->U = ptr->V = ptr->A = nullptr;
         break;
 #endif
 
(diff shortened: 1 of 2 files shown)


_studio/shared/src/libmfx_allocator.cpp
```diff
@@ -415,8 +415,9 @@ mfxStatus mfxDefaultAllocator::LockFrame(mfxHDL pthis, mfxHDL mid, mfxFrameData
     case MFX_FOURCC_Y410:
         ptr->PitchHigh = (mfxU16)((4 * ALIGN32(fs->info.Width)) / (1 << 16));
         ptr->PitchLow  = (mfxU16)((4 * ALIGN32(fs->info.Width)) % (1 << 16));
-        ptr->Y = ptr->U = ptr->V = ptr->A = 0;
         ptr->Y410 = (mfxY410*)sptr;
+        ptr->Y = (mfxU8*)(ptr->Y410);
+        ptr->U = ptr->V = ptr->A = 0;
```
mgonchar: use nullptr for pointers

_studio/shared/src/libmfx_allocator.cpp
```diff
@@ -415,8 +415,9 @@ mfxStatus mfxDefaultAllocator::LockFrame(mfxHDL pthis, mfxHDL mid, mfxFrameData
     case MFX_FOURCC_Y410:
         ptr->PitchHigh = (mfxU16)((4 * ALIGN32(fs->info.Width)) / (1 << 16));
         ptr->PitchLow  = (mfxU16)((4 * ALIGN32(fs->info.Width)) % (1 << 16));
-        ptr->Y = ptr->U = ptr->V = ptr->A = 0;
         ptr->Y410 = (mfxY410*)sptr;
+        ptr->Y = (mfxU8*)(ptr->Y410);
```
alariono: Disagree…Good picture for Y410 description:
https://docs.microsoft.com/en-us/windows/desktop/medfound/10-bit-and-16-bit-yuv-video-formats
So, there is no “Y” data pointer in case of Y410 format.
If “library heavily relies on Y pointer” fix these places in library for correct processing of Y410 format

PR #594, opened 2018-08-31
Fix hardcode param BitDepth8Only, Color420Only, YUV422ReconSupport, YUV444ReconSupport

dmenshov:
- we can init valid parameter
of caps from driver
Color420Only, BitDepth8Only, YUV422ReconSupport, YUV444ReconSupport


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -792,23 +792,12 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     if (p.CodeName >= MFX_PLATFORM_SKYLAKE)
```
dmitryermilov: Perhaps, just 
m_caps.MaxEncodedBitDepth = p.CodeName >= MFX_PLATFORM_KABYLAKE ? 1 : 0;

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -834,9 +823,12 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     m_caps.UserMaxFrameSizeSupport = 1;
     m_caps.MBBRCSupport            = 1;
     m_caps.MbQpDataSupport         = 1;
-    m_caps.Color420Only            = 1; // FIXME in case VAAPI direct YUY2/RGB support added
     m_caps.TUSupport               = 73;
-
+    m_caps.Color420Only = !(attrs[idx_map[VAConfigAttribRTFormat]].value & (VA_RT_FORMAT_YUV422 | VA_RT_FORMAT_YUV444));
```
mgonchar: well, let's be consistent, use ?: instead of "!" as you did below

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
mgonchar: agree

dmenshov: I think it is not correct because m_caps.MaxEncodedBitDepth may be 0,1,2,3

dmitryermilov: in your implementation it accepts only two values:
    if (p.CodeName >= MFX_PLATFORM_SKYLAKE)
    {
        m_caps.MaxEncodedBitDepth = 0;
    }
    if (p.CodeName >= MFX_PLATFORM_KABYLAKE)
    {
        m_caps.MaxEncodedBitDepth = 1;
    }

PR #605, opened 2018-09-06
Colorfill on APL: BXT defines replaced with run-time APL detection

IlyaSergeyev:
Another one place where PLATFORM_BXT is used


dvrogozh:
And another comment which I have is: it seems we are trying to workaround something which should be handled in the driver... Why we are doing this instead of working with the driver team on the fix?

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1303,10 +1303,11 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
         }
     }
 
-#if defined(VPP_NO_COLORFILL)
-    vaSts = vaDestroyBuffer(m_vaDisplay, outputParamBuf);
-    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
-#endif
+    if (VppNoColorfill && outputParamBuf != VA_INVALID_ID)
```
dvrogozh: same as above

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1231,35 +1231,37 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
...
 
-#if defined(LINUX_TARGET_PLATFORM_BXTMIN) || defined(LINUX_TARGET_PLATFORM_BXT)
-// It looks like only BXT supports this at the moment
-#define VPP_NO_COLORFILL
-#endif
+    bool VppNoColorfill =
+            // It looks like only BXT supports this at the moment
+            (MFX_HW_APL == hwType)
+            // Do not disable colorfill if letterboxing is used
+            && (0 == output_region.x && 0 == output_region.y);
```
dvrogozh: @alariono :
* Why w/h is not checked here?
* Do we somehow program a colorfill color at the first place? if yes, do we expose user API for this?

alariono: @dvrogozh 
>Why w/h is not checked here?
Because this is not need for letterboxing filter detection.
>Do we somehow program a colorfill color at the first place? if yes, do we expose user API for this?
No, except for VPP composition. MSDK have hard cored value for background.
See link https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/src/mfx_vpp_vaapi.cpp#L1057

alariono: @dvrogozh 
>Why w/h is not checked here?

Because this is not need for letterboxing filter detection.

>Do we somehow program a colorfill color at the first place? if yes, do we expose user API for this?

No, except for VPP composition. MSDK have hard cored value for background.
See link https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/src/mfx_vpp_vaapi.cpp#L1057

dvrogozh: I don't like how no color fill support is implemented. And I still think that condition is wrong. Yes, we will support letter boxing with the color fill, but why we permanently disable it for other cases? What if user wants to render some picture in the middle of the surface and have everything else filled with red?

Finally, I see this change in media-driver: https://github.com/intel/media-driver/pull/324. If I understands correctly, it disables color fill if color value is 0. This seems to be much more reasonable solution than what we are trying to achieve with this patch.

I don't approve this patch and vote for the try out of the driver's PR I have mentioned + make similar changes in mediasdk if it works.

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1056,7 +1056,17 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
     output_region.width  = outInfo->CropW;
     m_pipelineParam[0].output_region = &output_region;
 
-    m_pipelineParam[0].output_background_color = 0xff000000; // black for ARGB
+    m_pipelineParam[0].output_background_color = pParams->iColorfillColor;
+
+    if ( pParams->bDisableColorfill ||
```
dvrogozh: This is not in this patch: where bDisableColorfill is being set? do we have some exposed mediasdk uAPI for that?

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1056,7 +1056,16 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
     output_region.width  = outInfo->CropW;
     m_pipelineParam[0].output_region = &output_region;
 
-    m_pipelineParam[0].output_background_color = 0xff000000; // black for ARGB
+    m_pipelineParam[0].output_background_color = 0xff000000;
+
+    // It looks like only BXT supports this at the moment
+    if (MFX_HW_APL == hwType &&
```
dvrogozh: @onabiull : do we have an uAPI in mediasdk to setup background color? do we have an uAPI to switch off background color entirely?

alariono: @dvrogozh, 
Answers:
(1): no MSDK do not have uAPI for setup background color.
(2): MSDK do not have uAPI to switch off background color entirely

alariono: @dvrogozh 
If you don't have concern any more I would like to merge it.

IlyaSergeyev:
@dvrogozh please, take a look to colorfill API draft

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5111,73 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
         ;
 };
 
+mfxU64 GetBackgroundColor(const mfxVideoParam &videoParam)
+{
+    mfxU64 iBackgroundColor;
+    mfxU64 def_back_color = 0xffff000000000000;
```
dvrogozh: In which color format background color is programmed? Below you deal with different color formats, but use the same function and assign the same value. I guess background color is in some certain color format like RGB. If that's the case, then better to emphasize that in variable name like <color-format>_back_color.

Why it is 64-bit by the way?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5111,73 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
...
+            #if (MFX_VERSION >= 1027)
+            videoParam.vpp.Out.FourCC == MFX_FOURCC_Y210 ||
+            videoParam.vpp.Out.FourCC == MFX_FOURCC_Y410 ||
+            #endif
+            videoParam.vpp.Out.FourCC == MFX_FOURCC_P210)
+    {
+        def_back_color = make_def_back_color_yuv(10);
+    }
+
+    iBackgroundColor = def_back_color;
```
dvrogozh: I don't understand why you need new iBackgroundColor. You can continue to use def_back_color.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5111,73 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
         ;
 };
 
+mfxU64 GetBackgroundColor(const mfxVideoParam &videoParam)
```
dvrogozh: static...

api/include/mfxstructures.h
```diff
@@ -1929,9 +1929,20 @@ typedef struct {
 
 typedef struct {
     mfxExtBuffer Header;
-
     mfxU16 Enable;        /* tri-state option */
-    mfxU16 reserved[11];
+    union {
```
dvrogozh: So... We DID have an uAPI for enable/disable color fill, but we could not program which color to use (I guess we had some default, right?) and I am not sure whether we use the enable/disable uAPI at all. Considering all that I believe we need to break this change into step-by step:
1. I think you have some code cleanup which precedes your further changes - this goes to separate commit
2. Looks like you have or can provide enable/disable color fill support (without programming the color) within existing uAPI - that's the second commit
3. You have new uAPI added to program color to use in color fill - that's the third commit.

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1056,7 +1066,12 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
     output_region.width  = outInfo->CropW;
     m_pipelineParam[0].output_region = &output_region;
 
-    m_pipelineParam[0].output_background_color = 0xff000000; // black for ARGB
+    m_pipelineParam[0].output_background_color = RGBA64TO32(pParams->iBackgroundColor);
+
+    // Do not disable colorfill if letterboxing is used
```
dvrogozh: PLEASE, no more of this CRAP!!

You have uAPI to enable/disable color fill. I wish to see these 3 lines of code GONE forever!

api/include/mfxstructures.h
```diff
@@ -1929,9 +1929,20 @@ typedef struct {
...
+        mfxU16   Y;
+        mfxU16   R;
+    };
+    union {
+        mfxU16   U;
+        mfxU16   G;
+    };
+    union {
```
dvrogozh: As a new uAPI this needs to be wrapped in VERSION_NEXT here and everywhere you use it.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5111,46 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
...
+    {
+        def_back_color = make_def_back_color_yuv(10);
+    }
+
+    for (mfxU32 i = 0; i < videoParam.NumExtParam; i++) {
+        if (videoParam.ExtParam[i]->BufferId != MFX_EXTBUFF_VPP_COLORFILL)
+            break;
+
+        mfxExtVPPColorFill *extCF = reinterpret_cast<mfxExtVPPColorFill *>(videoParam.ExtParam[i]);
+        mfxU32 targetFourCC = videoParam.vpp.Out.FourCC;
```
dvrogozh: targetFourCC is unused variable. Remove it (it will appear in the next commit).

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
dvrogozh: Address this comment, please: rename def_back_color to something which highlight output color format, for example, argb_back_color.

samples/sample_vpp/src/sample_vpp_parser.cpp
```diff
@@ -252,6 +252,11 @@ msdk_printf(MSDK_STRING("   [-mirror (mode)]      - mirror image using specified
 
 msdk_printf(MSDK_STRING("   [-n frames] - number of frames to VPP process\n\n"));
 
+#if MFX_VERSION >= MFX_VERSION_NEXT
+msdk_printf(MSDK_STRING("   [-cf_red R]           - enable colorfill (red component).\n"));
```
dvrogozh: That's an overkill for the user. Let's have a simple option:
`[-cf #RRGGBB] - use specified color for the colorfill (in 8-bit rgb format)`

Also, you miss a code which will convert the colorfill RGB value to YUV color space - copy it from the library if it is currently missed in the sample.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2044,6 +2044,13 @@ mfxStatus VideoVPPHW::GetVideoParams(mfxVideoParam *par) const
             MFX_CHECK_NULL_PTR1(bufFRC);
             bufFRC->Algorithm = m_executeParams.frcModeOrig;
         }
+
+        else if (MFX_EXTBUFF_VPP_COLORFILL == bufferId)
+        {
+            mfxExtVPPColorFill *bufColorfill = reinterpret_cast<mfxExtVPPColorFill *>(par->ExtParam[i]);
+            MFX_CHECK_NULL_PTR1(bufColorfill);
```
mgonchar: why do you need a cast if don't really use this pointer?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5111,71 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
...
+            videoParam.vpp.Out.FourCC == MFX_FOURCC_Y410 ||
+            #endif
+            videoParam.vpp.Out.FourCC == MFX_FOURCC_P210)
+    {
+        argb_back_color = make_def_back_color_yuv(10);
+    }
+
+    for (mfxU32 i = 0; i < videoParam.NumExtParam; i++) {
+        if (videoParam.ExtParam[i]->BufferId != MFX_EXTBUFF_VPP_COLORFILL)
+            break;
```
mgonchar: As I understand, it should be ```continue```

samples/sample_vpp/include/sample_vpp_config.h
```diff
@@ -118,6 +118,17 @@ struct sMirroringParam: public mfxExtVPPMirroring
     };
 } ;
 
+struct sColorFillParam: public mfxExtVPPColorFill
+{
+    FilterConfig mode;
+    sColorFillParam():
```
mgonchar: add zeroing of ```mfxExtVPPColorFill```:
```
sColorFillParam()
: mfxExtVPPColorFill()
, mode(VPP_FILTER_DISABLED)
{
...
}
```

PR #606, opened 2018-09-06
Fixed possible segmentation fault

DenWolf:

### DIFF
--- a/_studio/shared/src/libmfx_core.cpp
+++ b/_studio/shared/src/libmfx_core.cpp
@@ -1680,8 +1680,10 @@ bool CommonCORE::CheckOpaqueRequest(mfxFrameAllocRequest *request,
                                     mfxU32 NumOpaqueSurface,
                                     bool ExtendedSearch)
 {
-    if (pOpaqueSurface &&
-        request->NumFrameMin != NumOpaqueSurface)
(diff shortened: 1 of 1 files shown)


_studio/shared/src/libmfx_core.cpp
```diff
@@ -1680,6 +1680,9 @@ bool CommonCORE::CheckOpaqueRequest(mfxFrameAllocRequest *request,
                                     mfxU32 NumOpaqueSurface,
                                     bool ExtendedSearch)
 {
+    if (!pOpaqueSurface || !request)
+        return false;
+
     if (pOpaqueSurface &&
```
xuguangxin: 1683 will make sure pOpaqueSurface is not null, so pOpaqueSurface always true here. Better remove it from this if

_studio/shared/src/libmfx_core.cpp
```diff
@@ -1723,6 +1726,9 @@ bool CommonCORE::IsOpaqSurfacesAlreadyMapped(mfxFrameSurface1 **pOpaqueSurface,
     {
         UMC::AutomaticUMCMutex guard(m_guard);
 
+        if (!pOpaqueSurface || !response)
```
xuguangxin: just check the input. no need protect with mutex

PR #611, opened 2018-09-06
Fixed warnings in _studio/shared

Losev-Ivan:


_studio/shared/asc/src/iofunctions.cpp
```diff
@@ -23,42 +23,31 @@
...
     ASC_PRINTF("%s %0.3f ms.\n", message, timeval);
     return timeval;
 }
 
-mfxF64 CatchTime(ASCTime *timer, int index, const char* message) {
-    (void)timer;
-    (void)index;
-    (void)message;
-
+mfxF64 CatchTime(ASCTime */*timer*/, int /*index*/, const char* /*message*/) {
```
dvrogozh: `*/*` looks ugly. Can you, please write `ASCTime*`

_studio/shared/asc/src/iofunctions.cpp
```diff
@@ -23,42 +23,31 @@
...
     return 0.0;
 }
 
-mfxF64 CatchTime(ASCTime *timer, int indexInit, int indexEnd, const char* message) {
-    (void)timer;
-    (void)indexInit;
-    (void)indexEnd;
-    (void)message;
-
+mfxF64 CatchTime(ASCTime */*timer*/, int /*indexInit*/, int /*indexEnd*/, const char* /*message*/) {
```
dvrogozh: same as above

_studio/shared/umc/codec/h264_dec/include/umc_h264_frame_info.h
```diff
@@ -116,11 +116,10 @@ class H264DecoderFrameInfo : public H264DecoderLayer
 {
 public:
 
-    H264DecoderFrameInfo(H264DecoderFrame * pFrame,  H264_Heap_Objects * pObjHeap)
+    H264DecoderFrameInfo(H264DecoderFrame * pFrame,  H264_Heap_Objects * /*pObjHeap*/ = nullptr)
```
dvrogozh: If I understand correctly, after pre-processor we will get: `H264_Heap_Objects * = nullptr`. And I simply don't understand what's that?!

If you don't need 2nd parameter for H264DecoderFrameInfo, then delete it and fix those places where we call it.

dvrogozh:
@onabiull : rebase is done. Please, consider to merge.

PR #612, opened 2018-09-06
Fixed android warnings

Losev-Ivan:
Warnings: Wunused-function, Wexceptions, Wunneeded-internal-declaration

### DIFF
--- a/_studio/shared/asc/include/asc_common_impl.h
+++ b/_studio/shared/asc/include/asc_common_impl.h
@@ -174,11 +174,7 @@ static inline __m256i LoadPartialYmm(unsigned char *pSrc, mfxI32 len)
 }
 #endif //defined(__AVX2__)
 
-#ifdef _MSVC_LANG
-#pragma warning(disable:4505)
-#endif
-
-static void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
+static inline void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
     mfxI32 i, j;
     mfxU8 *pS = pSrc;
     mfxU8 *pS2 = pSrc + pitch;
@@ -200,9 +196,5 @@ static void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
     *CS += Cs >> 4;
     *RS += Rs >> 4;
 }
-
-#ifdef _MSVC_LANG
-#pragma warning(default:4505)
-#endif
 
 #endif //_ASC_COMMON_IMPL_H_
--- a/_studio/shared/asc/include/cpu_detect.h
+++ b/_studio/shared/asc/include/cpu_detect.h
@@ -28,25 +28,17 @@
 // 2) each stub configures a function pointer on first call
 //
 
-#ifdef _MSVC_LANG
-#pragma warning(disable:4505)
-#endif
-
-static mfxI32 CpuFeature_SSE41() {
+static inline mfxI32 CpuFeature_SSE41() {
     return((__builtin_cpu_supports("sse4.1")));
 }
 
-static mfxI32 CpuFeature_AVX2() {
+static inline mfxI32 CpuFeature_AVX2() {
 #if defined(__AVX2__)
-        return((__builtin_cpu_supports("avx2")));
+    return((__builtin_cpu_supports("avx2")));
 #else
     return 0;
 #endif //defined(__AVX2__)
 }
-
-#ifdef _MSVC_LANG
-#pragma warning(default:4505)
-#endif
 
 //
 // end Dispatcher
--- a/_studio/shared/asc/include/iofunctions.h
+++ b/_studio/shared/asc/include/iofunctions.h
@@ -24,17 +24,9 @@
 #include "asc.h"
 using namespace ns_asc;
 
-#ifdef _MSVC_LANG
-#pragma warning(disable:4505)
-#endif
-
-static mfxF64 TimeMeasurement(LARGE_INTEGER start, LARGE_INTEGER stop, LARGE_INTEGER frequency) {
+static inline mfxF64 TimeMeasurement(LARGE_INTEGER start, LARGE_INTEGER stop, LARGE_INTEGER frequency) {
     return((stop.QuadPart - start.QuadPart) * mfxF64(1000.0) / frequency.QuadPart);
 }
-
-#ifdef _MSVC_LANG
-#pragma warning(default:4505)
-#endif
 
 void TimeStart(ASCTime* timer);
 void TimeStart(ASCTime* timer, int index);
--- a/_studio/shared/include/mfxvideo++int.h
+++ b/_studio/shared/include/mfxvideo++int.h
@@ -55,7 +55,7 @@ typedef struct {
     unsigned char  Data4[8];
 } GUID;
 
-static int operator==(const GUID & guidOne, const GUID & guidOther)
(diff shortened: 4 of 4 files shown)


_studio/shared/include/mfxvideo++int.h
```diff
@@ -55,7 +55,7 @@ typedef struct {
     unsigned char  Data4[8];
 } GUID;
 
-static int operator==(const GUID & guidOne, const GUID & guidOther)
+static inline int operator==(const GUID & guidOne, const GUID & guidOther)
```
Losev-Ivan: According to compiler warning: 'static' function 'operator==' declared in header file should be declared 'static inline' [-Wunneeded-internal-declaration]

_studio/shared/asc/include/iofunctions.h
```diff
@@ -28,7 +28,7 @@ using namespace ns_asc;
 #pragma warning(disable:4505)
 #endif
 
-static mfxF64 TimeMeasurement(LARGE_INTEGER start, LARGE_INTEGER stop, LARGE_INTEGER frequency) {
+static inline mfxF64 TimeMeasurement(LARGE_INTEGER start, LARGE_INTEGER stop, LARGE_INTEGER frequency) {
```
Losev-Ivan: Fix warning "Wunused-function". This is a one line function which use only 1 time in _studio/shared/asc/src/iofunctions.cpp:43.

_studio/shared/asc/include/cpu_detect.h
```diff
@@ -32,11 +32,11 @@
 #pragma warning(disable:4505)
 #endif
 
-static mfxI32 CpuFeature_SSE41() {
+static inline mfxI32 CpuFeature_SSE41() {
     return((__builtin_cpu_supports("sse4.1")));
 }
 
-static mfxI32 CpuFeature_AVX2() {
+static inline mfxI32 CpuFeature_AVX2() {
```
Losev-Ivan: Fix warning "Wunused-function". Both functions are one-line-functions which use 3 times in _studio/shared/asc/src/asc.cpp

_studio/shared/asc/include/asc_common_impl.h
```diff
@@ -178,7 +178,7 @@ static inline __m256i LoadPartialYmm(unsigned char *pSrc, mfxI32 len)
 #pragma warning(disable:4505)
 #endif
 
-static void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
+static inline void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
```
Losev-Ivan: Fix warning "Wunused-function". This function use 2 times in _studio/shared/asc/src/asc_sse4_impl.cpp and in _studio/shared/asc/src/asc_c_impl.cpp

_studio/mfx_lib/decode/vp8/include/mfx_vp8_dec_decode_hw.h
```diff
@@ -166,7 +166,7 @@ class MFX_VP8_BoolDecoder
     }
 };
 
-static mfxStatus MFX_CDECL VP8DECODERoutine(void *p_state, void *pp_param, mfxU32 thread_number, mfxU32);
+static inline mfxStatus MFX_CDECL VP8DECODERoutine(void *p_state, void *pp_param, mfxU32 thread_number, mfxU32);
```
Losev-Ivan: Fix warning "Wunused-function". Since "inline" not a strict order to compiller and this function called only by address, compiler will not make it inline.

_studio/shared/include/mfxvideo++int.h
dvrogozh: ok

_studio/shared/asc/include/asc_common_impl.h
Losev-Ivan: This function use in _studio/shared/asc/src/asc_c_impl.cpp:218. Moreover, all other functions in header file, like LoadPartialYmm, defined as "static inline"

Losev-Ivan: In header file this function is not wrapped in any macros

_studio/shared/asc/include/cpu_detect.h
Losev-Ivan: In header file this functions do not wrapped in any macros and use in _studio/shared/asc/src/asc.cpp.

_studio/shared/asc/include/cpu_detect.h
```diff
@@ -32,11 +32,11 @@
 #pragma warning(disable:4505)
```
dvrogozh: to be removed

_studio/shared/asc/include/cpu_detect.h
```diff
@@ -32,11 +32,11 @@
...
 #endif
 
-static mfxI32 CpuFeature_SSE41() {
+static inline mfxI32 CpuFeature_SSE41() {
```
dvrogozh: ok, valid

_studio/shared/asc/include/cpu_detect.h
dvrogozh: ok, valid

_studio/shared/asc/include/iofunctions.h
```diff
@@ -28,7 +28,7 @@ using namespace ns_asc;
 #pragma warning(disable:4505)
```
dvrogozh: to be removed

_studio/shared/asc/include/iofunctions.h
dvrogozh: ok, valid

_studio/shared/asc/include/asc_common_impl.h
```diff
@@ -178,7 +178,7 @@ static inline __m256i LoadPartialYmm(unsigned char *pSrc, mfxI32 len)
 #pragma warning(disable:4505)
```
dvrogozh: If I read https://msdn.microsoft.com/en-us/library/z85eyax0.aspx correctly, your are fixing exactly this warning. Thus, I believe that you need to check and remove these pragmas as well...

_studio/shared/asc/include/asc_common_impl.h
dvrogozh: This place and file in general looks STRANGE (because of _missing_ inline and presence of the warning disable above). If author did not want to make this function inline, then he should have defined it in the .cpp file. If he wanted make it inline, then why disable a warning?

Actually the whole this file looks strange to me. Functions defined in it are not that small actually. I am not sure they should be made inline at the first place. @spossos: can you, please, comment. @onabiull: what do you think?

_studio/mfx_lib/decode/vp8/include/mfx_vp8_dec_decode_hw.h
dvrogozh: no. this is incorrect way to fix this. Actually you need to do the following:
1. Completely remove this declaration and friend declaration from line 217 below
2. Introduce an VideoDECODEVP8_HW class uAPI method, let's say VideoDECODEVP8_HW::TaskRoutine and rewrite P8DECODERoutine as using this new method. You can refer to mpeg2 decoder code:
```
static mfxStatus __CDECL MPEG2TaskRoutine(void *pState, void *pParam, mfxU32 /*threadNumber*/, mfxU32 /*callNumber*/)
{
    VideoDECODEMPEG2InternalBase *decoder = (VideoDECODEMPEG2InternalBase*)pState;

    MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_API, "VideoDECODEMPEG2::TaskRoutine");
    mfxStatus sts = decoder->TaskRoutine(pParam);
    return sts;
}
```

PR #619, opened 2018-09-08
Fix MFXVideoUSER_Unregister

akokoshn:
-Should returns MFX_ERR_NONE if no one plugins were registred.
 It's important in case of using HW HEVCe as plugin

### DIFF
--- a/_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
+++ b/_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
@@ -312,8 +312,10 @@ mfxStatus MFXVideoUSER_Unregister(mfxSession session, mfxU32 type)
     {
         SessionPtr sessionPtr(session, type);
         std::unique_ptr<VideoCodecUSER> & registeredPlg = sessionPtr.plugin();
+        // Do nothing if no registred plugins, because plugins from NativePlugins[] can't be registred - MFXVideoUSER_Register() does fake registration for them
+        // It's required for keep backward comatibility
         if (NULL == registeredPlg.get())
-            return MFX_ERR_NOT_INITIALIZED;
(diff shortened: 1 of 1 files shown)


dmitryermilov:
I talked to Alexey. He explained this case. @dvrogozh , @dmitry-gurulev , @onabiull ,perhaps we should store all fake register() calls in some new member of session so that in unregistered() we can distinguish real user plug-ins from MSDK plug-ins which has recently moved to MSDK back?

akokoshn:
@dvrogozh , same change already in the [master branch](https://github.com/Intel-Media-SDK/MediaSDK/blob/83c9359699a29d63f99b91199912c08dff94b495/_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp#L323) - is part of commit from @dmitry-gurulev - "_Return HEVC, VP8, VP9 HW decoders back to library_" (**1f568f**)  (wasn't ported to the mss2018_r2).
Also if explanation in source code OK, I can add it to commit message.

_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
```diff
@@ -312,8 +312,10 @@ mfxStatus MFXVideoUSER_Unregister(mfxSession session, mfxU32 type)
     {
         SessionPtr sessionPtr(session, type);
         std::unique_ptr<VideoCodecUSER> & registeredPlg = sessionPtr.plugin();
+        // Do nothing if no registred plugins, because plugins from NativePlugins[] can't be registred - MFXVideoUSER_Register() does fake registration for them
+        // It's required for keep backward comatibility
```
dmitryermilov: comatibility -> compatibility

akokoshn:
Shouldn't be include to hot fix scope.

PR #622, opened 2018-09-10
LINUX_TARGET_PLATFORM_BXT was changed to runtime detection

ilapichev:
Issue: MDP-49870


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -941,29 +941,30 @@ mfxStatus CheckAndFixRoi(MfxVideoParam  const & par, ENCODE_CAPS_HEVC const & ca
...
-    {
-        if (caps.ROIDeltaQPSupport == 0)
-            bROIViaMBQP = true;
-
-    }
-    else
-    {
-#if MFX_VERSION > 1021
-        if (ROI->ROIMode == MFX_ROI_MODE_QP_DELTA)
+    if (par.m_platform.CodeName != MFX_PLATFORM_APOLLOLAKE && par.m_platform.CodeName != MFX_PLATFORM_COFFEELAKE)
```
dvrogozh: If driver sets caps correctly you do _not_ need this condition at all. If it sets caps incorrectly, we need to file a bug. Could you, please, check the current status of driver caps?

ilapichev: It was found that driver sets caps correctly.
 So, can I remove this condition?

dvrogozh: of course!

PR #624, opened 2018-09-10
Using of platform definitions removed from copyright strings

IlyaSergeyev:
Using of platform definitions removed from copyright strings. Copyright strings can be defined in cmake (like it was made before for plugins)

### DIFF
--- a/_studio/mfx_lib/shared/src/libmfxsw.cpp
+++ b/_studio/mfx_lib/shared/src/libmfxsw.cpp
@@ -46,37 +46,12 @@ void* g_hModule = NULL; // DLL handle received in DllMain
 #define MFX_PRODUCT_VERSION "0.0.000.0000"
 #endif
 
-// Copyright strings
 #if defined(mfxhw64_EXPORTS) || defined(mfxhw32_EXPORTS) || defined(mfxsw64_EXPORTS) || defined(mfxsw32_EXPORTS)
-
-#if defined(LINUX_TARGET_PLATFORM_BDW)
 const char* g_MfxProductName = "mediasdk_product_name: Intel(R) Media SDK";
-#elif defined(LINUX_TARGET_PLATFORM_BXT) || defined (LINUX_TARGET_PLATFORM_BXTMIN)
-const char* g_MfxProductName = "mediasdk_product_name: Intel(R) Media SDK for Embedded Linux";
-#else
-const char* g_MfxProductName = "mediasdk_product_name: Intel(R) Media SDK";
-#endif
-
-const char* g_MfxCopyright = "mediasdk_copyright: Copyright(c) 2018 Intel Corporation";
+const char* g_MfxCopyright = "mediasdk_copyright: Copyright(c) 2011-2018 Intel Corporation";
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/libmfxsw.cpp
```diff
@@ -46,38 +46,6 @@ void* g_hModule = NULL; // DLL handle received in DllMain
...
-#endif
-
-const char* g_MfxCopyright = "mediasdk_copyright: Copyright(c) 2014-2018 Intel Corporation";
-const char* g_MfxFileVersion = "mediasdk_file_version: " MFX_FILE_VERSION;
-const char* g_MfxProductVersion = "mediasdk_product_version: " MFX_PRODUCT_VERSION;
-
-#endif // mfxaudioswXX_EXPORTS
-
-
 mfxStatus MFXInit(mfxIMPL implParam, mfxVersion *ver, mfxSession *session)
```
dvrogozh: Well. We _need_ to left some copyright and version information. This should be changed, not fully removed.

builder/ConfTargets.cmake
```diff
@@ -121,5 +123,7 @@ if (DEFINED MFX_CONFIG_FILE)
     include(${MFX_CONFIG_FILE})
 endif()
 
+add_definitions( " -DMFX_PRODUCT_NAME='\"${MFX_PRODUCT_NAME}\"'" )
```
dvrogozh: No. You don't need this. Instead add the following in the mfxconfig.h.in:
`#define MFX_PRODUCT_NAME "@MFX_PRODUCT_NAME@"`

builder/ConfTargets.cmake
```diff
@@ -110,6 +110,8 @@ option( MFX_ENABLE_VP9_VIDEO_ENCODE "Enable VP9 encoder?" ON)
 
 option( MFX_ENABLE_MCTF "Build with MCTF support?"  ${MFX_1_25_OPTIONS_ALLOWED} )
 
+set( MFX_PRODUCT_NAME "Intel(R) Media SDK" )
```
dvrogozh: Rewrite as:
```
if (NOT DEFINED MFX_PRODUCT_NAME)
  set( MFX_PRODUCT_NAME "Intel(R) Media SDK" )
endif()
```
And move it below after `include(${MFX_CONFIG_FILE}) ` and before `configure_file(mfxconfig.h.in mfxconfig.h)`. Since this is not a cmake option(), we can do this and this will avoid redifinition of the variable.

builder/profiles/fastboot.cmake
```diff
@@ -24,6 +24,8 @@ else()
   set( MFX_LIBNAME "mfxhw32-fastboot" )
 endif()
 
+set( MFX_PRODUCT_NAME "Intel(R) Media SDK for Embedded Linux" )
```
dvrogozh: I am not sure I lake this as a product name. @onabiull : maybe "Intel(R) Media SDK Fastboot"?

alariono: @dvrogozh lets keep one name only.

PR #625, opened 2018-09-11
Fixed android warning Wunused-function in vp8_dec

Losev-Ivan:


Losev-Ivan:
It's impossible to create function TaskRoutine() (like in mpeg2 decode), because in vp8 we get p_state pointer to private structure VideoDECODEVP8_HW::VP8DECODERoutineData. I suggest to make function VP8DECODERoutine non-static (as in vp9 decode)

PR #626, opened 2018-09-11
[h264e] Replace GetExtBuffer() to GetExtBufferRef() for MfxVideoParam

DenWolf:
This change fix possible problem with nullptr buffer

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_cm.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_cm.cpp
@@ -1093,7 +1093,7 @@ mfxStatus CmContext::QueryVme(
 
 
     { MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_INTERNAL, "Convert mb data");
-        mfxExtPpsHeader const * extPps = GetExtBuffer(m_video);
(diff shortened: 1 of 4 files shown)


dmitry-gurulev:
Here is impl. of _mfxExtBufferRefProxy_ (mfx_h264_enc_common_hw.h)
```
template <typename T> operator T&()
 {
            mfxExtBuffer * p = MfxHwH264Encode::GetExtBuffer(
                m_extParam,
                m_numExtParam,
                ExtBufTypeToId<typename GetPointedType<T*>::Type>::id);
            assert(p);
            return *(reinterpret_cast<T*>(p));
 }
```
Please note that just _assert_ is used, no real pointer check in release.

stasefremov:
Intention of GetExtBufferRef is to get ext buffers from MfxVideoParam structure which always has full set of extension buffers (so "MfxHwH264Encode::GetExtBuffer" can't return zero). Need to add respective comment near definition of the function.

DenWolf:
@dmitryermilov , I updated commit message with clarification for this change.
Also, I created another pull request for adding clarification comment to source code:
https://github.com/Intel-Media-SDK/MediaSDK/pull/629

DenWolf:
I tested this change on Android for h264 hw encoder - by CTS EncoderTest and screenrecord application (coded video scream from the screen of Android device using h264e) - all worked OK, tests passed.

PR #627, opened 2018-09-11
fastboot: Optional build and use the asc lib

IlyaSergeyev:
binary size on debian x86_64/gcc_8.2:
with asc 1301K
without asc 1173K


_studio/mfx_lib/CMakeLists.txt
```diff
@@ -155,12 +155,18 @@ else()
   endif()
 endif()
 
-if( API_VERSION VERSION_GREATER 1.25 )
+if( ${MFX_ENABLE_MCTF} )
```
dvrogozh: good catch, but this is a separate change non related to the ASC:). And you also forget this place: https://github.com/Intel-Media-SDK/MediaSDK/blob/ed67603472e33ef03ce92a80f28a55e3e7b78d65/_studio/mfx_lib/genx/CMakeLists.txt#L79

Don't create separate PR for this though...

_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
```diff
@@ -991,7 +991,7 @@ namespace MfxHwVideoProcessing
 
         CmCopyWrapper *m_pCmCopy;
 
-#if defined(MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP)
+#if defined(MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP) && defined(MFX_ENABLE_ASC)
```
dvrogozh: Are there any difference between MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP and MFX_ENABLE_ASC? @onabiull, @alariono : do you know?

dvrogozh: This question still remains. @onabiull, @alariono : your comments?

alariono: @dvrogozh Look like this is very long history, ask @spossos for details 
asc_genx_bdw kernel - SCD component -> file mfx_scd.cpp
genx_scd_bdw kernel - ASC component-> file asc.cpp

So, for FB we need to disable both...

onabiull: I believe ASC is "advanced mode" of SCD, but regular mode looks deprecated a long ago. Still not sure if we have non-advanced mode functional after this change.

spossos: The last element that was using non advanced SCD was ADI, but it has been updated to ASC. Regular SCD needs to be removed.

dvrogozh: @spossos : as far as this patch is concerned, would that be correct to:
1. Substitute MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP with MFX_ENABLE_ASC everywhere this patch has touched
2. Add the following somewhere in config.h:
```
#if defined(MFX_ENABLE_ASC)
  #define MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP
#endif
```

Some other patches will need to cleanup the rest of MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP places?

IlyaSergeyev: @dvrogozh,  @spossos: what will be the final decision?
1. Substitute MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP with MFX_ENABLE_ASC 
2. Second patch with substitution MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP with MFX_ENABLE_ASC in the rest of code?

spossos: That change looks safe, I analyzed the code and I agree with it.

dvrogozh: We have discussed the topic with @spossos and we have few important changes to make. Here is a summary:
1. In open source code old implementation for scene change detection is completely wiped out and substituted by ASC. Secondly, ASC is used not only in VPP but also in AVC encoder. I suggest the following decision:
1.1. MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP should remain in VPP to cover the named feature (do not add MFX_ENABLE_ASC)
1.2. The following code to be added in the mfx_config.h:
```
#if defined(MFX_ENABLE_ASC)
    #define MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP
#endif
```
2. We should wrap with the define (either MFX_ENABLE_ASC or better MFX_AVC_ENABLE_SOME_FEATURE) the code in AVC encoder which depends on ASC. We did not see the build failure because our fastboot don't enable AVC encoder. I believe this can be done later on as a follow up after the patch. @spossos : will you take the task?
3. MCTF actually depends on ASC and can't work without it.

Alternative decision for 1) and 2) could be to use MFX_ENABLE_ASC all over the VPP and AVC encoder. This would be that MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP will be removed, and we will loss an opportunity to enable/disable these features in VPP/AVC separately. I am not sure that we will need this though. @onabiull : how do you think?

builder/ConfTargets.cmake
```diff
@@ -109,6 +109,7 @@ cmake_dependent_option(
 option( MFX_ENABLE_VP9_VIDEO_ENCODE "Enable VP9 encoder?" ON)
 
 option( MFX_ENABLE_MCTF "Build with MCTF support?"  ${MFX_1_25_OPTIONS_ALLOWED} )
+option( MFX_ENABLE_ASC "Build with ASC support?"  ON )
```
dvrogozh: MCTF should be transformed to dependent option from ASC and binary kernels, per @spossos it won't work without them:
```
cmake_dependent_option(
  MFX_ENABLE_MCTF "Enable MCTF?" ON
  "MFX_ENABLE_ASC;MFX_ENABLE_KERNELS" OFF)
```
And please, use word "Enable" rather than build: "Enable ASC?"

spossos: I will do task 2) wrapping of features that depened on ASC in AVC

_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
onabiull: While it would require more work, I believe first proposed approach would be better. Thus we'll reduce unexpected surprises.

builder/ConfTargets.cmake
```diff
@@ -109,6 +109,7 @@ cmake_dependent_option(
 option( MFX_ENABLE_VP9_VIDEO_ENCODE "Enable VP9 encoder?" ON)
 
 option( MFX_ENABLE_MCTF "Build with MCTF support?"  ${MFX_1_25_OPTIONS_ALLOWED} )
```
dvrogozh: I commented before: "MCTF actually depends on ASC and can't work without it." This means that you should rewrite as:
```
cmake_dependent_option(
  MFX_ENABLE_MCTF "Enable MCTF?" ${MFX_1_25_OPTIONS_ALLOWED}
  "MFX_ENABLE_ASC;MFX_ENABLE_KERNELS" OFF)
```
And eventually have option for ASC listed first.

_studio/shared/include/mfx_config.h
```diff
@@ -116,8 +116,8 @@
 #define MFX_ENABLE_VP8_VIDEO_DECODE_HW
 #endif
 
-#if defined(MFX_ENABLE_ASC)
-    #define MFX_ENABLE_SCENE_CHANGE_DETECTION_VPP
+#if !defined(MFX_ENABLE_ASC)
```
dvrogozh: What's this? Looks completely wrong.

builder/ConfTargets.cmake
```diff
@@ -108,9 +108,12 @@ cmake_dependent_option(
 
 option( MFX_ENABLE_VP9_VIDEO_ENCODE "Enable VP9 encoder?" ON)
 
-option( MFX_ENABLE_MCTF "Build with MCTF support?"  ${MFX_1_25_OPTIONS_ALLOWED} )
 option( MFX_ENABLE_ASC "Enable ASC support?"  ON )
 
+cmake_dependent_option(
```
dvrogozh: This should be done in a previous patch once you have introduced ASC.

dvrogozh:
@onabiull : can this be merged?

PR #628, opened 2018-09-11
[hevc_fei] Move hevc fei to library

artem-shaporenko:
Needed to switch HEVC to internal interfaces.
Currently it is backward compatible with plugin usage
but there is no way to enable HEVC FEI without LoadPlugin.
FEI files moved to HEVC Encode files


_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_plugin_hw.h
```diff
@@ -0,0 +1,161 @@
+//
```
onabiull: Please change license, open source MSDK use MIT.

_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
```diff
@@ -169,11 +169,14 @@ namespace
...
 {
-    MFX_PLUGINID_HEVCE_HW,
-    MFX_PLUGINID_VP9E_HW,
     MFX_PLUGINID_HEVCD_HW,
     MFX_PLUGINID_VP8D_HW,
-    MFX_PLUGINID_VP9D_HW
+    MFX_PLUGINID_VP9D_HW,
+    MFX_PLUGINID_HEVCE_HW,
+    MFX_PLUGINID_VP9E_HW,
+#if MFX_VERSION >= 1027
```
onabiull: No way to get this into 1.27, please use 1.28 or VERSION_NEXT

_studio/shared/include/libmfx_core_vaapi.h
```diff
@@ -209,6 +209,8 @@ class VAAPIVideoCORE : public CommonCORE
 #ifdef MFX_ENABLE_MFE
     ComPtrCore<MFEVAAPIEncoder> m_mfe;
 #endif
+    //required to WA FEI enabling after move it from plugin to library
```
onabiull: I don't like idea of having just HEVC FEI specific member in the core. I'd like to see more generic interface.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -88,7 +96,15 @@ VideoENCODE* _mfxSession::Create<VideoENCODE>(mfxVideoParam& par)
 
 #if defined(MFX_ENABLE_H265_VIDEO_ENCODE)
     case MFX_CODEC_HEVC:
-        pENCODE = new MfxHwH265Encode::MFXVideoENCODEH265_HW(&m_coreInt, &mfxRes);
+#if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
+            feiEnabled = (bool*)core->QueryCoreInterface(MFXIFEIEnabled_GUID);
```
onabiull: I don't like idea of having pointer to bool as interface.

_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
artem-shaporenko: in this case other also should go under API something, VP9 decode, etc?

_studio/shared/include/libmfx_core_vaapi.h
artem-shaporenko: I'd also like, but we don't have generic usage for this.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
artem-shaporenko: As well as me, but this is what it is.

_studio/mfx_lib/shared/src/libmfxsw_plugin.cpp
onabiull: Why did you put here preprocessor check in the first hand? I'd expect ether no check here or MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE.

_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_plugin_hw.h
```diff
@@ -0,0 +1,172 @@
...
+            *mfxPlg = tmp_pplg->m_adapter;
+            tmp_pplg->m_createdByDispatcher = true;
+
+            return MFX_ERR_NONE;
+        }
+
+        virtual mfxStatus PluginInit(mfxCoreInterface *core) override
+        {
+            MFX_TRACE_INIT();
+            MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_API, "H265FeiEncodePlugin::PluginInit");
```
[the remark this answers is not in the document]
dvrogozh: Actually this will be the first thing which I will remove in the follow up clean-up of the FEI plugin build:).

PR #631, opened 2018-09-13
Replace GetExtBuffer() for MfxVideoParam - mfx_lib/shared/

DenWolf:
Intention of GetExtBufferRef is to get ext buffers
from MfxVideoParam structure which always has full set
of extension buffers ("MfxHwH264Encode::GetExtBuffer" can't return zero).

So, for MfxVideoParam it's better to use function GetExtBufferRef()
instead GetExtBuffer(), we can wrap issues from static code analyze
to one place (body of function GetExtBufferRef()) instead several
places (every calling GetExtBuffer())

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -375,11 +375,10 @@ namespace
     // calculate possible minimum level for encoding of input stream
     mfxU16 GetMinLevelForAllParameters(MfxVideoParam const & par)
     {
-        mfxExtSpsHeader * extSps = GetExtBuffer(par);
-        assert(extSps != 0);
+        mfxExtSpsHeader const & extSps = GetExtBufferRef(par);
 
-        if (par.mfx.FrameInfo.Width         == 0 ||
-            par.mfx.FrameInfo.Height        == 0)
+        if (par.mfx.FrameInfo.Width  == 0 ||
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -7560,17 +7558,17 @@ void MfxVideoParam::SyncVideoToCalculableParam()
 
     if (IsMvcProfile(mfx.CodecProfile))
     {
-        mfxExtMVCSeqDesc * extMvc = GetExtBuffer(*this);
-        if (extMvc->NumView)
+        mfxExtMVCSeqDesc & extMvc = GetExtBufferRef(*this);
```
stasefremov: Need avoid calling "GetExtBufferRef()" inside methods of MfxVideoParam. Maybe right now (in current state of the code) this particular call is OK, but it could happen that in the future this method of MfxVideoParam will be called inside constructor when preparation of ext buffers will not be finished yet, so you'll shoot in your leg by this call.

DenWolf: Ok, understand, moved it back inside MfxVideoParam

DenWolf:
@onabiull @dmitryermilov @stasefremov 
I tested this change on Android for h264 hw encoder - by CTS EncoderTest and screenrecord application (coded video scream from the screen of Android device using h264e) - all worked OK, tests passed.

PR #643, opened 2018-09-17
[ASG-HEVC][TOOLS] make install option was added to the asg-hevc and hevc_fei_extractor tools.

romka0017:
Commit includes enabling make install option for HEVC FEI validation tools.

### DIFF
--- a/tools/asg-hevc/CMakeLists.txt
+++ b/tools/asg-hevc/CMakeLists.txt
@@ -5,10 +5,12 @@ include_directories (
   ${CMAKE_CURRENT_SOURCE_DIR}/include
 )
 
-list( APPEND LIBS sample_common)
+list( APPEND LIBS sample_common )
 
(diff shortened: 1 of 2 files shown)


onabiull:
Actually logic behind this was: if user explicitly requested tools, then he/she would expect it to appear somewhere next to other artifacts. Samples folder looks like misguiding, so why not MFX_APPS_DIR?

dvrogozh:
MFX_APPS_DIR would be mean directly /usr/share/mfx. I thought about MFX_APPS_DIR/tools or simply /usr/bin (that's probably BINDIR)?

dvrogozh:
We discussed with @onabiull : let's put tools into MFX_APPS_DIR and let's move samples to this folder as well (this will be another commit and another PR). If later on we will need to reorganize MFX_APPS_DIR structure we will do that.

PR #649, opened 2018-09-19
Fixed check max DPB size and NumRefFrame for HEVCe

iveselov:


lakulako:
Alternative is https://github.com/Intel-Media-SDK/MediaSDK/pull/652

lakulako:
In this patch maxDPB is reduced so that NumRefFrame was <=. While DBP contains reference frames + current frame it seems to be better to limit NumRefFrame directly to <= maxDPB-1. What is done in #652.

iveselov:
MAX_DPB_SIZE for encoder is already set, considering current frame. I think, it is less readable to set maxDPB magic number, which can't be achived

lakulako:
maxDPB is already computed and verified according to DPB described in spec. It includes reference frames and current frame. It is apparent to require maxNumRefFrame == maxDPB-1.
maxDPB was 16 and it would be incorrect to reduce it.
MAX_DPB_SIZE is our internal variable that represents size of array with reference frames. It is more magic and dangerous to use defines that have similar roots in name but different sense.
More accurate would be to rename MAX_DPB_SIZE to something like MAX_REF_FRAMES to avoid confuses.

onabiull:
Alternative version merged.

PR #650, opened 2018-09-19
[h264d_umc] Remove part of Get functions inside classes

DenWolf:
GetSliceRef(pos), GetLayerRef(pos), GetPayloadRef(pos).
These functions don't check input value "pos",
they will be used ONLY when 100% sure
that 0 <= pos < array (vector) size.

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2018 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -72,10 +72,10 @@ H264SEIPayLoad* SeiPayloadArray::FindPayload(SEI_TYPE type) const
 
 int32_t SeiPayloadArray::FindPayloadPos(SEI_TYPE type) const
 {
-    size_t count = GetPayloadCount();
+    size_t count = m_payloads.size();
     for (size_t i = 0; i < count; i++)
     {
-        H264SEIPayLoad* payload = GetPayload(i);
+        H264SEIPayLoad* payload = m_payloads[i];
         if (payload->payLoadType == type)
             return (int32_t)i;
     }
@@ -174,8 +174,8 @@ SetOfSlices& SetOfSlices::operator=(const SetOfSlices& set)
 
 H264Slice * SetOfSlices::GetSlice(size_t pos) const
 {
-    if (pos >= GetSliceCount())
-        return 0;
+    if (pos >= m_pSliceQueue.size())
+        return nullptr;
 
     return m_pSliceQueue[pos];
 }
@@ -192,10 +192,10 @@ void SetOfSlices::AddSlice(H264Slice * slice)
 
 void SetOfSlices::Release()
 {
-    size_t count = GetSliceCount();
+    size_t count = m_pSliceQueue.size();
     for (size_t sliceId = 0; sliceId < count; sliceId++)
     {
-        H264Slice * slice = GetSlice(sliceId);
+        H264Slice * slice = m_pSliceQueue[sliceId];
         slice->DecrementReference();
     }
 
@@ -222,14 +222,14 @@ void SetOfSlices::AddSet(const SetOfSlices *set)
 
 void SetOfSlices::CleanUseless()
 {
-    size_t count = GetSliceCount();
+    size_t count = m_pSliceQueue.size();
     for (size_t sliceId = 0; sliceId < count; sliceId++)
     {
-        H264Slice * curSlice = GetSlice(sliceId);
+        H264Slice * curSlice = m_pSliceQueue[sliceId];
         if (curSlice->m_bDecoded)
         {
             m_pSliceQueue.erase(m_pSliceQueue.begin() + sliceId); // remove
-            count = GetSliceCount();
+            count = m_pSliceQueue.size();
             --sliceId;
             curSlice->Release();
             curSlice->DecrementReference();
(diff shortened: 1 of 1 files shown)


DenWolf:
@dmitry-gurulev,
If I right understand, functions GetSlice(pos), GetPayload(pos), GetLayer(pos) may return nullptr only if input value "pos" is in out of array (vector) size. If 0 <= "pos" < array (vector) size - these functions always return not nullptr.
So, I replaces the functions only in the places where we checked "pos" before calling the functions - additional check is not needed. For this purpose I add special Get functions without extra check for input index.
In another places where we not 100% sure about input index "pos" we still use old Get functions with checking input index. We need to investigate these places we more carefully - if we don't 100% sure that index is valid -> function may return nullptr -> we need to check return value in this case -> skipping, reporting an error, throw an exception or some another activity before directly using and deference.

This patch is a "pilot" for resolving this problem. If all will be OK and we will get an agreement, I will continue this activity to replace reference functions for h264d and 265d and investigate possible nullptr problem which I described upper at all.

dmitry-gurulev:
@DenWolf,
>For this purpose I add special Get functions without extra check for input index.
So, is it some kind of optimization?

DenWolf:
@dmitry-gurulev ,

I think we need to avoid extra checking and another actions when it's not needed. Also, it may confuse developers, why we have additional check where it's not necessary or why don't we check the returning result from the function before using where it's possible to return nullptr.

I will continue investigation of this case, because it's really possible that in some another places we can get nullptr and we don't check it, so we open the door to run-to-run problem (and maybe these problems already exists now).

The problem was found by static code analyze - for now it's the first priority task for Media SDK Android team.
And as I write upper, I think it's good that we get this potential problem - it's time to think about it more carefully and resolve it at all.

dmitry-gurulev:
@DenWolf 

>I think we need to avoid extra checking and another actions when it's not needed.

[GetSlice and others] are public methods of classes, they are part of classes' interface and hereby are the part of public contract. They MUST check theirs preconditions.

>The problem was found by static code analyze - for now it's the first priority task for Media SDK Android team.

What problem? Potential using of nullptr? Your patch doesn't resolve it . You just move **explicit** potential nullptr de-referencing to **implicit** accessing to memory that might be out of range. The latter is much more difficult to debug.

dmitryermilov:
I agree with DmitryG. This just gives an impression of safety but in fact doesn't solve problems (if they exist. I don't know)

DenWolf:
> [GetSlice and others] are public methods of classes, they are part of classes' interface and hereby are the part of public contract. They MUST check theirs preconditions.

I agree that these Get functions need to check input parameter,  especially if they are pubic methods and may be used outside of classes. For this purpose I didn't touch these functions and added new special.

> What problem? Potential using of nullptr? Your patch doesn't resolve it . You just move **explicit** potential nullptr de-referencing to **implicit** accessing to memory that might be out of range. The latter is much more difficult to debug.

So, I changed functions ONLY for the places where we checked input index before calling the function and didn't touch another places (for now).

For example:

    size_t count = GetSliceCount();
    for (size_t sliceId = 0; sliceId < count; sliceId++)
    {
        H264Slice * curSlice = GetSlice(sliceId);
        .......
    }

In this case input parameter "sliceId" always in array range:
0 <= sliceId < count == Size of Slice Array (Vector).

So, in this case checking for input parameter is extra, and function doesn't return nullptr, only if some array value is nullptr, but:
1. this is a bug and it's need to be solved by another way
2. current patch doesn't change behavior for this case


This patch is a proposal to resolve possible problem which was found by static code analyze and improve source code at all.
In my opinion it's easy to discuss some code changes by looking at the patch and code directly, so I prepared this patch. We can abandon or remake this patch any time to improve the code by a good way.

dmitry-gurulev:
@DenWolf 

> I agree that these Get functions need to check input parameter, especially if they are pubic methods and may be used outside of classes. For this purpose I didn't touch these functions and added new special.

But your new special methods are also public.

> So, in this case checking for input parameter is extra, and function doesn't return nullptr

So, you patch is just some kind of optimization, right?

> This patch is a proposal to resolve possible problem which was found by static code analyze and improve source code at all.

I still don't know what problem 'found by static code analyze' u're trying to resolve. I'm guessing that it's about nullptr, but the patch doesn't fix it.

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -359,7 +359,7 @@ SetOfSlices * AccessUnit::GetLastLayer()
     if (!GetLayersCount())
```
dmitry-gurulev: 'm_layers.empty()' ?

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -359,7 +359,7 @@ SetOfSlices * AccessUnit::GetLastLayer()
...
         return 0;
 
-    return GetLayer(GetLayersCount() - 1);
+    return &m_layers[GetLayersCount() - 1];
 }
```
dmitry-gurulev: 'm_layers.back()' ?

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -375,11 +375,11 @@ void AccessUnit::CombineSets()
     if (!m_layers.size())
```
dmitry-gurulev: m_layers.empty() ?

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -375,11 +375,11 @@ void AccessUnit::CombineSets()
...
         return;
 
-    SetOfSlices * setOfSlices = GetLayer(0);
-    size_t count = GetLayersCount();
+    SetOfSlices * setOfSlices = &m_layers[0];
```
dmitry-gurulev: -1, u'll get a problem here if 'm_layers' is empty.

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -442,10 +442,10 @@ bool AccessUnit::AddSlice(H264Slice * slice)
 
 void AccessUnit::Release()
```
dmitry-gurulev: I'd rather re-write it as
```
std::for_each(std::begin(m_layers), std::end(m_layers),
    [](SetOfSlices* set) { set->Release(); }
);
```

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -454,10 +454,10 @@ void AccessUnit::Release()
 
 void AccessUnit::Reset()
```
dmitry-gurulev: The same as for 'Release'

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -468,10 +468,10 @@ void AccessUnit::Reset()
 
 void AccessUnit::SortforASO()
```
dmitry-gurulev: The same as for 'Release'

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -327,26 +327,26 @@ size_t AccessUnit::GetLayersCount() const
 
 void AccessUnit::CleanUseless()
```
dmitry-gurulev: 'std::remove_if' ?

DenWolf:
> BTW, [SetOfSlices::Release] is also candidate for using [std::for_each]
Agree, but I suggest to prepare another special pull request for changing "for cycles" inside SetOfSlices class to [std::for_each], when current pull request will be approved and merged:)

DenWolf:
@onabiull , @dmitry-gurulev ,

Thank you for review, done all needed changes

PR #652, opened 2018-09-19
[HEVCE] Fix possible DPB overflow

lakulako:
Dpb index was checked against NumRefFrame instead of array size
On ICL where NumRefFrame can be > DPB_SIZE led to assertion fail.

Issue: MDP-45453

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
@@ -2833,7 +2833,7 @@ void UpdateDPB(
         }
         else
         {
-            for (st0 = 0; dpb[st0].m_ltr && st0 < end; st0 ++); // excess?
+            //for (st0 = 0; st0 < end && dpb[st0].m_ltr; st0 ++); // st0 is first !ltr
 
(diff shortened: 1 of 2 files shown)


lakulako:
Exploring failed checks, pls don't merge.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -2830,7 +2830,7 @@ void UpdateDPB(
     MFX_SORT_STRUCT((dpb + st0), mfxU16(end - st0), m_poc, >);
 
     // sliding window over STRs
-    if (end && end == par.mfx.NumRefFrame)
+    if (end && end == MAX_DPB_SIZE)
```
eistarov: You're breaking sliding window here.
HEVC encoder DPB should not contain more refs than NumRefFrame.
If NumRefFrame > MAX_DPB_SIZE consider it parameters validation issue.

lakulako:
Updated. Now limits NumRefFrame to maxDPB-1, i.e. dpb without current frame

lakulako:
Shouldn't be merged together with https://github.com/Intel-Media-SDK/MediaSDK/pull/649.
Both fix same issue different ways.

PR #661, opened 2018-09-21
[mpeg2d] Improved robustness for using Opaque Memory

DenWolf:

### DIFF
--- a/_studio/mfx_lib/decode/mpeg2/include/mfx_mpeg2_decode.h
+++ b/_studio/mfx_lib/decode/mpeg2/include/mfx_mpeg2_decode.h
@@ -147,6 +147,8 @@ public:
 
     virtual mfxStatus CompleteTasks(void *pParam) = 0;
 
+    bool IsOpaqueMemory() const { return m_isOpaqueMemory; }
+
 public:
     int32_t display_frame_count;
     int32_t cashed_frame_count;
@@ -178,7 +180,6 @@ public:
     std::unique_ptr<UMC::MPEG2VideoDecoderBase> m_implUmc;
 
     bool m_isSWBuf;
-    bool m_isOpaqueMemory;
     mfxVideoParam m_vPar;
     UMC::VideoDecoderParams m_vdPar;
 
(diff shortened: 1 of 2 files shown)


DenWolf:
@dmitryermilov ,

Code analyzer found possible nullptr dereference for pOpqExt (please see the picture below) :
mpeg2d_problem

As you wrote previously, field m_isOpaqueMemory set to false by default in constructor, and it can be true (go to if branch before dereference - 3) ONLY if there is OPAQUE memory - we go to if branch (before 2) and set m_isOpaqueMemory to true directly - in this case pOpqExt always not null.

I think it's a good solution to set m_isOpaqueMemory to false directly in "else" branch (2) - we can avoid misunderstanding for developers and code analyzer too.

What do you think?

dmitryermilov:
I believe there is no real problem here. Thus,  it doesn't make sense to introduce additional assignment to false if  m_isOpaqueMemory is set false by ctor.

DenWolf:
@dmitryermilov ,

OK, I agree. Field m_isOpaqueMemory sets to false in 2 places - constructor and Close() method, and sets to true only in one place (for now) - inside AllocFrames (line 1804).

But I was confused that field  m_isOpaqueMemory   is a "public", and it's possible to change this field outside of the class. So, this was the first reason why I created this patch.

For now we don't change value of the m_isOpaqueMemory anywhere except 3 places which I wrote previously, outside of the class this field is used read-only.
But it's possible to get some problems in the future and I think it's a time to avoid them.

I agree that current fix (set m_isOpaqueMemory  to true) is not good, and I suggest to add checking pOpqExt for nullptr and reporting an error MFX_ERR_UNDEFINED_BEHAVIOR before dereference.

Also, I think that it's better to hide m_isOpaqueMemory  to protected and create special method for getting this value (for example, IsOpaqueMemory()).

What do you think about that? If we will get an agreement I can do both of these changes.

dmitry-gurulev:
I don't see a real problem here (agree w/ @dmitryermilov) and at the same time I don't think this is a big deal to set [m_isOpaqueMemory = false] here explicitly to make analyzer happy.

@DenWolf 
> it's better to hide m_isOpaqueMemory to protected
Agrre, it makes sense. Maybe even private.

dmitryermilov:
If so, at line 1824 I'd replace m_isOpaqueMemory  with pOpqExt

PR #665, opened 2018-09-24
[h265d_umc] Use m_pSliceQueue directly w/o GetSlice

DenWolf:
Replace GetSlice() to directly using m_pSliceQueue
inside class H265DecoderFrameInfo where it's possible

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_frame_info.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_frame_info.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2018 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


DenWolf:
@dmitry-gurulev ,

Yes, m_pSliceQueue is a private field of class H265DecoderFrameInfo:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/h265_dec/include/umc_h265_frame_info.h#L211

I replace using GetSlice to m_pSliceQueue directly only inside class H265DecoderFrameInfo - functions 
H265DecoderFrameInfo::EliminateErrors() and void H265DecoderFrameInfo::EliminateASO()

DenWolf:
P.S.: @dmitry-gurulev ,@dmitryermilov , @onabiull ,

I noticed that we have mix of implementation and declaration in header of class H265DecoderFrameInfo :
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/h265_dec/include/umc_h265_frame_info.h

I think that it's not good and better to move all impementation to ".cpp" file and keep only declaration in ".h" file. What to you think about that?
If you agree with me I can prepare this change.

dmitry-gurulev:
@DenWolf 

> I think that it's not good and better to move all impementation to ".cpp" file and keep only declaration in ".h" file. What to you think about that?

I doubt that this is good idea. Defining functions in [.cpp] you forbids many compiler optimizations, even if LTO is turned on.

DenWolf:
> I doubt that this is good idea. Defining functions in [.cpp] you forbids many compiler optimizations, even if LTO is turned on.

Ok, agree with you, thank you for offline explanation in details about this question. Lets keep it.

PR #668, opened 2018-09-25
Add Media SDK Tracer code

onabiull:
(cherry picked from commit bdd11b75d1245d04d4170056c2a44313d288eb90)


semelianova:
I checked tracer work with new dynamic dispatcher. It works correctly by following instructions. But when I used LD_PRELOAD trick some variables from LoaderCtx class were uninitialized what caused errors.

PR #675, opened 2018-09-26
[h264e] Removed extra check nativeHandle for nullptr

DenWolf:
We can loose more useful error status from function
GetFrameHDL() upper, also nullptr can be processed
in the upper code.

Aligned the same function GetNativeHandleToRawSurface
with h265e


DenWolf:
Aligned GetNativeHandleToRawSurface() with h265e - the same change already merged with this patch:
Switch HEVC to use internal Core interfaces - 
https://github.com/Intel-Media-SDK/MediaSDK/commit/20f0cdc2b89843f24398fb1ec7893816b314ad93#diff-564c6b03a14ecab6938c086da7c12b2dL565

PR #676, opened 2018-09-26
[VP9] Fix PackPicParams for streams starting w/ inter frame

vshampor:
A VP9 stream starting w/ inter frame has it intra-coded without
references. Prior code led to error since the auxiliary ref frame array
was filled with VA_INVALID_SURFACE ids, which cannot be passed
to the GetSurfaceID functions.

### DIFF
--- a/_studio/shared/umc/codec/vp9_dec/include/umc_vp9_utils.h
+++ b/_studio/shared/umc/codec/vp9_dec/include/umc_vp9_utils.h
@@ -28,12 +28,15 @@
 #ifndef __UMC_VP9_UTILS_H_
 #define __UMC_VP9_UTILS_H_
 
-#define ALIGN_POWER_OF_TWO(value, n) \
-    (((value) + ((1 << (n)) - 1)) & ~((1 << (n)) - 1))
-
 namespace UMC_VP9_DECODER
 {
+    constexpr auto VP9_INVALID_REF_FRAME = -1;
     class VP9DecoderFrame;
+
+    inline mfxU32 AlignPowerOfTwo(mfxU32 value, mfxU32 n)
+    {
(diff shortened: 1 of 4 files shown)


_studio/shared/umc/codec/vp9_dec/src/umc_vp9_va_packer.cpp
```diff
@@ -126,7 +126,16 @@ void PackerVA::PackPicParams(VADecPictureParameterBufferVP9* picParam, VP9Decode
     else
     {
         for (mfxU8 ref = 0; ref < NUM_REF_FRAMES; ++ref)
-            picParam->reference_frames[ref] = m_va->GetSurfaceID(info->ref_frame_map[ref]);
+        {
+            if ((VASurfaceID) info->ref_frame_map[ref] != VA_INVALID_SURFACE)
```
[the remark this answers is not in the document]
vshampor: Done

_studio/shared/umc/codec/vp9_dec/include/umc_vp9_utils.h
```diff
@@ -31,6 +31,8 @@
 #define ALIGN_POWER_OF_TWO(value, n) \
     (((value) + ((1 << (n)) - 1)) & ~((1 << (n)) - 1))
 
+#define VP9_INVALID_REF_FRAME -1
```
onabiull: It's 2018, and C++ 11. Please move this into namespace and change define to constexpr auto.

vshampor: Consistency trumps fashion. Still, will fix it as you propose.

PR #682, opened 2018-09-27
Need to remove workaround for check

iveselov:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1786,16 +1786,17 @@ mfxStatus VAAPIEncoder::QueryStatus(Task & task)
...
-        MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+        m_feedbackCache.erase(m_feedbackCache.begin() + indxSurf);
+        guard.Unlock();
 
-        if (VASurfaceReady == surfSts)
         {
-            m_feedbackCache.erase(m_feedbackCache.begin() + indxSurf);
-            guard.Unlock();
+            MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaSyncSurface");
+            vaSts = vaSyncSurface(m_vaDisplay, waitSurface);
```
artem-shaporenko: We should not use vaSyncSurface with input surface for encoder synchronization, but vaMapBuffer with bitstream

dvrogozh: I partially disagree. As of now this is a designated uAPI to sync tasks. Yes, we move to use vaMapBuffer instead for the encoders, but this is not official way yet. Thus, I would like to have a commit which uses vaSyncSurface in a history. Current change is definitely better than vaQuerySurfaceStatus (how we managed to have that by the way?!). I suggest to merge and then switch to vaMapBuffer.

dvrogozh:
@onabiull : why #551 was ever merged?! it is totally wrong to pursue a minute benefit sacrificing correct code workable on a range of platforms.

Being said that the current change is wrong as well. I expect to see an honest revert of the #551 change, not a commit pretending that this is a new code and a fix.

Secondly, I believe that upcoming Q3 release can _not_ happen without this change.

dmitryermilov:
@iveselov , if this PR is not actual any more (it seems the original commit was reverted) please remove it

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
artem-shaporenko: Agree

PR #690, opened 2018-09-28
Adding MJPEG SW fallback

Alexandr-Konovalov:


contrib/ipp/src/asm_intel64/pjarithm7as.s
```diff
@@ -0,0 +1,384 @@
...
+ 
+ 
+ 
+ 
+ 
+ 
+ 
+ 
+ 
+
```
dvrogozh: looks weird. Please, remove all these empty lines.

contrib/ipp/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+set( MFX_ROOT "${CMAKE_HOME_DIRECTORY}/_studio/shared" )
+
+include_directories (
+  ${CMAKE_CURRENT_SOURCE_DIR}/include
+  ${MFX_ROOT}/include
```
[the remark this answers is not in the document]
Alexandr-Konovalov: This is for `mfxSize` that is `IppiSize` in closed source. Changing it back to `IppiSize` in open source is strange, defining `mfxSize` in contrib/ipp/include/ippdefs.h violates hierarchical levels. And certainty cross-dependence is bad. So, I don't know what is less evil.

contrib/ipp/include/ippdefs.h
```diff
@@ -0,0 +1,495 @@
...
+//              Common Types and Macro Definitions
+// 
+// 
+*/
+
+
+#ifndef __IPPDEFS_H__
+#define __IPPDEFS_H__
+
+#include "umc_defs.h" // for mfxSize
```
dvrogozh: No, we need to do something with this. Let's discuss later.

_studio/mfx_lib/encode/mjpeg/include/mfx_mjpeg_encode.h
```diff
@@ -0,0 +1,160 @@
...
+
+    std::unique_ptr<UMC::MJPEGVideoEncoder> m_pMJPEGVideoEncoder;
+
+    mfxStatus EncodePiece(const mfxU32 threadNumber);
+
+protected:
+    // Close the object, release all resources
+    void      Close(void);
+
+    UMC::Mutex m_guard;
```
mgonchar: std::mutex

contrib/ipp/src/pccycbcr44deintw7cn.c
```diff
@@ -83,7 +83,7 @@ extern void  mfxmyNV12ToYUY2_8u_P2C2R(const Ipp8u* pSrcY, int srcYStep,const Ipp
       dst  += 32;
       w += 16;
     }
-    for( w = w; w < width; w += 2 )
+    for( /*w = w*/; w < width; w += 2 )
```
dvrogozh: don't comment - remove! This assignment don't make any sense.

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
```diff
@@ -33,6 +33,9 @@
 
 #include "umc_mutex.h"
```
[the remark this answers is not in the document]
dvrogozh: Yes, this is highly desirable. We are in a process to do that across mediasdk code. You will help if you will avoid introducing new self-written mutex wrappers or refactor existing code.

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
```diff
@@ -33,6 +33,9 @@
...
 
+#ifdef MFX_ENABLE_SW_FALLBACK
```
dvrogozh: Does not make sense to protect standard header. I suggest to just include unconditionally.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -2076,5 +2083,255 @@ mfxStatus VideoDECODEMJPEGBase_HW::CompleteTask(void *pParam, mfxStatus )
...
+mfxStatus VideoDECODEMJPEGBase_SW::CheckTaskAvailability(mfxU32 maxTaskNumber)
+{
+    if (m_freeTasks.empty())
+    {
+        if (m_tasksCount >= maxTaskNumber)
+        {
+            return MFX_WRN_DEVICE_BUSY;
+        }
+
+        std::auto_ptr<CJpegTask> pTask(new CJpegTask());
```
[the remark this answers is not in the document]
dvrogozh: Yes,  no auto pointers, please. Make sure you don't use them with grep.

_studio/shared/umc/codec/color_space_converter/src/umc_deinterlacing.cpp
```diff
@@ -138,8 +138,8 @@ static void DeinterlacingEdgeDetect(uint8_t *psrc,
 
   mfxiCopy_8u_C1R(psrc, 2*iSrcPitch, pdst, 2*iDstPitch, roi);
 
-  memcpy_s(pdst + iDstPitch, iSrcPitch, psrc + iSrcPitch, iSrcPitch);
-  memcpy_s(pdst + (h - 1)*iDstPitch, w, psrc + (h - 2) * iSrcPitch, w);
+  std::copy(psrc + iSrcPitch, psrc + 2*iSrcPitch, pdst + iDstPitch);
```
dvrogozh: You bring a big code to open source. I believe you need to bring it "fixed". So, please:
1. Take this patch offline to the closed source version you are open sourcing
2. Bring fixed code initially. In a way this patch should be squashed to one of the previous.

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
```diff
@@ -143,19 +143,19 @@ class VideoDECODEMJPEGBase_SW : public VideoDECODEMJPEGBase
...
-    virtual mfxStatus GetVideoParam(mfxVideoParam *par);
-    virtual mfxStatus RunThread(void *pParam, mfxU32 threadNumber, mfxU32 callNumber);
-    virtual mfxStatus CompleteTask(void *pParam, mfxStatus taskRes);
-    virtual mfxStatus CheckTaskAvailability(mfxU32 maxTaskNumber);
-    virtual mfxStatus ReserveUMCDecoder(UMC::MJPEGVideoDecoderBaseMFX* &pMJPEGVideoDecoder, mfxFrameSurface1 *surf, bool isOpaq);
-    virtual void ReleaseReservedTask();
-    virtual mfxStatus AddPicture(UMC::MediaDataEx *pSrcData, mfxU32 & numPic);
-    virtual mfxStatus AllocateFrameData(UMC::FrameData *&data);
-    virtual mfxStatus FillEntryPoint(MFX_ENTRY_POINT *pEntryPoint, mfxFrameSurface1 *surface_work, mfxFrameSurface1 *surface_out);
+    virtual mfxStatus Init(mfxVideoParam *decPar, mfxFrameAllocRequest *request, mfxFrameAllocResponse *response, mfxFrameAllocRequest *request_internal, bool isUseExternalFrames, VideoCORE *core) override;
```
dvrogozh: Again, take this offline. Bring "fixed" code in the initial commit.

_studio/mfx_lib/encode/mjpeg/include/mfx_mjpeg_encode.h
dvrogozh: Yep.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -241,6 +243,76 @@ mfxStatus FallbackProxyENCODE::Init(mfxVideoParam *par)
 }
 
 
+struct Handlers {
+    typedef mfxStatus (*QueryFunc)(mfxSession s, mfxVideoParam *in, mfxVideoParam *out);
```
dvrogozh: `decltype(MFXVideoENCODE_Query)`? @mgonchar : how do you think?

dvrogozh: Yes, I thought about removing typedef. But indeed for that we should not have overloads. Anyway, that's minor. Let's form the solution first, then decide about this place.

contrib/ipp/src/pccycbcr44deintw7cn.c
Alexandr-Konovalov: I believe an IPP code author added this for documenting proposes, i.e. to stress that we continue from a current position, so I keep it as a comment.  Sure, it can be deleted.

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
Alexandr-Konovalov: Standard headers are huge and affect compilation time. The plan was to include only required stuff.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -2073,5 +2036,235 @@ mfxStatus VideoDECODEMJPEGBase_HW::CompleteTask(void *pParam, mfxStatus )
...
+    ConvertMFXParamsToUMC(decPar, &umcVideoParams);
+    umcVideoParams.numThreads = m_vPar.mfx.NumThread;
+    return MFX_ERR_NONE;
+}
+
+mfxStatus VideoDECODEMJPEGBase_SW::Reset(mfxVideoParam *par)
+{
+    m_tasksCount = 0;
+    pLastTask = NULL;
+    while(!m_freeTasks.empty())
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Just in case, this is std::queue, and one lacks begin()/end()/clear(). Should I switch to std::deque that is used underneath?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -460,34 +490,30 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
...
         {
-            auto it = codecId2Handlers.find(out->mfx.CodecId);
-            mfxRes = it == codecId2Handlers.end() ? MFX_ERR_UNSUPPORTED
-            : (it->second.queryFunc)(session, in, out);
+            auto handlers = codecId2Handlers.find(out->mfx.CodecId);
+            mfxRes = handlers == codecId2Handlers.end() ? MFX_ERR_UNSUPPORTED
+            : (handlers->second.queryFunc)(session, in, out);
 #if defined(MFX_ENABLE_H264_VIDEO_ENCODE) && !defined(MFX_ENABLE_H264_VIDEO_ENCODE_HW)
-            if (out->mfx.CodecId == MFX_CODEC_AVC)
+            if (out->mfx.CodecId != MFX_CODEC_AVC)
```
dvrogozh: We don't think these if() is really needed.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -562,34 +586,30 @@ mfxStatus MFXVideoENCODE_QueryIOSurf(mfxSession session, mfxVideoParam *par, mfx
...
         {
-            auto it = codecId2Handlers.find(par->mfx.CodecId);
-            mfxRes = it == codecId2Handlers.end() ? MFX_ERR_UNSUPPORTED
-                : (it->second.queryIOSurfFunc)(session, par, request);
+            auto handlers = codecId2Handlers.find(par->mfx.CodecId);
+            mfxRes = handlers == codecId2Handlers.end() ? MFX_ERR_UNSUPPORTED
+                : (handlers->second.queryIOSurfFunc)(session, par, request);
 #if defined(MFX_ENABLE_H264_VIDEO_ENCODE) && !defined(MFX_ENABLE_H264_VIDEO_ENCODE_HW)
-            if (out->mfx.CodecId == MFX_CODEC_AVC)
+            if (out->mfx.CodecId != MFX_CODEC_AVC)
```
dvrogozh: same

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -322,12 +332,26 @@ static const CodecId2Handlers codecId2Handlers =
             // queryFunc
             [](mfxSession s, mfxVideoParam *in, mfxVideoParam *out)
             { return MFXVideoENCODEMJPEG_HW::Query(s->m_pCORE.get(), in, out); },
+            // queryFallbackFunc
+#if defined(MFX_ENABLE_SW_FALLBACK)
+            [](mfxSession, mfxVideoParam *in, mfxVideoParam *out)
+            { return MFXVideoENCODEMJPEG::Query(in, out); },
+#else
```
dvrogozh: you don't need this

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -250,8 +250,10 @@ struct Handlers {
                                      mfxCoreInterface* coreInt, mfxStatus *mfxRes);
 
     CtorFunc        ctorFunc;
-    QueryFunc       queryFunc;
-    QueryIOSurfFunc queryIOSurfFunc;
+    QueryFunc       queryFunc,
+                    queryFallbackFunc;
```
dvrogozh: this one should be under ifdef (MFX_ENABLE_SW_FALLBACK)

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -279,6 +281,8 @@ static const CodecId2Handlers codecId2Handlers =
                 return MFXVideoENCODEH264::Query(in, out);
 #endif
             },
+            // queryFallbackFunc
```
dvrogozh:
```
#ifdef MFX_ENABLE_H264_VIDEO_ENCODE
     { MFX_CODEC_AVC,
#ifdef MFX_ENABLE_H264_VIDEO_ENCODE_HW
      ctror, query, queryIO
#endif
#ifdef MFX_ENABLE_SW_FALLBACK
      ctror, query, queryIO
#endif
#endif
```

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -138,7 +269,8 @@ static const CodecId2Handlers codecId2Handlers =
             [](VideoCORE* core, mfxCoreInterface*, mfxStatus* mfxRes)
             -> VideoENCODE*
             {
-                return new MFXVideoENCODEMJPEG_HW(core, mfxRes);
+                *mfxRes = MFX_ERR_NONE;
+                return new FallbackProxyENCODE(core, MFX_CODEC_JPEG);
```
dvrogozh: don't do this in this patch. And you need to add ctors for HW/SW encoders. Create proxy class checking if both ctors present.

_studio/shared/umc/codec/jpeg_enc/include/umc_mjpeg_video_encoder.h
```diff
@@ -0,0 +1,224 @@
...
+        app0_units       = 0;
+        app0_xdensity    = 1;
+        app0_ydensity    = 1;
+        threading_mode   = 0;
+        buf_size         = 0;
+        interleaved      = 0;
+        chroma_format    = 0;
+    }
+
+//    Status ReadParamFile(const vm_char *ParFileName) { return UMC_ERR_NOT_IMPLEMENTED; }
```
dvrogozh: remove it

_studio/shared/umc/codec/jpeg_enc/src/umc_mjpeg_video_encoder.cpp
```diff
@@ -0,0 +1,690 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include "umc_defs.h"
+#if defined (MFX_ENABLE_MJPEG_VIDEO_ENCODE)
+
+#include <string.h>
+#include "vm_debug.h"
```
dvrogozh: remove

_studio/shared/umc/codec/jpeg_enc/src/umc_mjpeg_video_encoder.cpp
```diff
@@ -0,0 +1,690 @@
...
+#include "umc_video_data.h"
+#include "umc_mjpeg_video_encoder.h"
+#include <umc_automatic_mutex.h>
```
dvrogozh: replace with std::

_studio/shared/umc/codec/jpeg_enc/src/umc_mjpeg_video_encoder.cpp
Alexandr-Konovalov: I have a question. In `umc_mjpeg_video_encoder.cpp` `vm_debug.h` is not needed, so it's clear. But it used in `jpegdec.cpp`. Should I change `VM_ASSERT` to `assert`? What to do with `vm_debug_trace1`?

dvrogozh: > Should I change VM_ASSERT to assert?
Yes.
> What to do with vm_debug_trace1?
Depends on what are they doing. If there is nothing important - better to remove. If there is something important replace with MFX_TRACE_*.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -79,7 +81,17 @@ static const CodecId2Handlers codecId2Handlers =
...
-        {   // queryFunc
+      {     // ctorFunc
+#if defined(MFX_ENABLE_H264_VIDEO_ENCODE_HW)
+            [](VideoCORE* core, mfxCoreInterface*, mfxStatus* mfxRes)
+            -> VideoENCODE*
+            {
+                return new MFXHWVideoENCODEH264(core, mfxRes);
+            },
+#else
+            nullptr,
```
Alexandr-Konovalov: This is behaviour from legacy, but is this correct?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -113,7 +120,13 @@ static const CodecId2Handlers codecId2Handlers =
 #else
                 return MFXVideoENCODEH264::QueryIOSurf(par, request);
 #endif
-            }
+            },
```
Alexandr-Konovalov: Is behavior for MFX_CODEC_AVC correct? Now it mimics legacy.

_studio/shared/umc/codec/color_space_converter/include/umc_video_resizing.h
```diff
@@ -0,0 +1,65 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include "umc_defs.h"
+
+#if defined (UMC_ENABLE_VPP_RESIZE)
```
dvrogozh: I do not see anywhere that you have defined UMC_ENABLE_VPP_RESIZE. If so, the code under it is _dead_ and should be dropped...

Questions:
1. Have you actually checked how SW jpeg decoder is running in open source? Did you run any tests? What is passing and what is not?
2. Could we try to break SW jpeg enabling support into 2 steps based on this UMC_ENABLE_VPP_RESIZE?

I very much don't like having UMC_ENABLE_VPP_RESIZE at all. From my perspective our SW_FALLBACK should be enough.

_studio/shared/umc/codec/color_space_converter/src/umc_color_space_conversion.cpp
```diff
@@ -0,0 +1,1007 @@
...
+// SOFTWARE.
+
+#include "umc_color_space_conversion.h"
+#include "umc_video_data.h"
+#include "ippi.h"
+#include "ippcc.h"
+#include "ippvc.h"
+
+#if defined(__GNUC__)
+    #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
dvrogozh: This will lower quality of our compile-level checks! I guess this comes from IPP declarations. If so - these declarations should be fixed.

_studio/shared/umc/codec/jpeg_dec/src/jpegdec.cpp
```diff
@@ -0,0 +1,4476 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include "umc_defs.h"
+#if defined (MFX_ENABLE_MJPEG_VIDEO_DECODE) && defined(MFX_ENABLE_SW_FALLBACK)
+#if defined(__GNUC__)
+    #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
dvrogozh: same

_studio/shared/umc/codec/color_space_converter/include/umc_video_resizing.h
Alexandr-Konovalov: `UMC_ENABLE_VPP_RESIZE` is not defined in this patch. But resize topic is interesting.

Manually I run
jpegd_conf jpegd_difresol jpegd_multi jpegd_multiscan jpegd_reli jpegd_resize jpegd_threadsafety mjpegd_conf mjpegd_double_rotate mjpegd_multi mjpegd_reli mjpegd_reposition mjpegd_resize mjpegd_threadsafety v30b5_jpegd_rotate v30b5_mjpegd_rotate jpege_func jpege_multi jpege_reli jpege_resize jpege_robust jpege_threadsafety mjpege_func mjpege_reli mjpege_resize mjpege_robust mjpege_threadsafety

Except for some flaky failures, new failures are jpegd_resize 0155 0159. That functionality is depends on `MFX_ENABLE_HW_ONLY_VPP` which is presented in closed-source only. This was discussed verbally and current status is "let's postpone it".

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -81,67 +83,100 @@ static const CodecId2Handlers codecId2Handlers =
...
 #ifdef MFX_ENABLE_H264_VIDEO_ENCODE
     { MFX_CODEC_AVC,
-      {
-          { [](mfxSession s, mfxVideoParam *in, mfxVideoParam *out)
-            {
-                return (!s->m_pENCODE.get())?
-                MFXHWVideoENCODEH264::Query(s->m_pCORE.get(), in, out)
-                : MFXHWVideoENCODEH264::Query(s->m_pCORE.get(), in, out, s->m_pENCODE.get());
+        {
+            {   // queryFunc
```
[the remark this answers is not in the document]
Alexandr-Konovalov: It would be more readable, but designator-based initialization is part of C99, it's impossible in C++.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -76,7 +76,12 @@ struct Handlers {
         QueryIOSurfFunc queryIOSurf;
     };
 
+    struct Fallback {
+        CtorFunc        ctor;
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Unfortunately, this works only since C++14, as for C++11 [aggregate initialization can't be used if a class has member-initializers](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3605.html). I'll try to use std::function instead.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
dvrogozh: Yeah, they will be part of C++20 (https://en.cppreference.com/w/cpp/language/aggregate_initialization). @onabiull : when did you say we switch to it?

What you can do now is have comments like:
```
{
// .queryFunc =
[](...) ...,
// .queryIOSurfFunc =
[](...) ...
}
```

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -81,67 +83,100 @@ static const CodecId2Handlers codecId2Handlers =
...
-          { [](mfxSession s, mfxVideoParam *in, mfxVideoParam *out)
-            {
-                mfxStatus mfxRes;
+        {
+            {   // queryFunc
+                [](mfxSession s, mfxVideoParam *in, mfxVideoParam *out)
+                {
+                    mfxStatus mfxRes;
+#if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
+                    bool * feiEnabled = (bool*)s->m_pCORE->QueryCoreInterface(MFXIFEIEnabled_GUID);//required to check FEI plugin registration.
```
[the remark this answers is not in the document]
Alexandr-Konovalov: For the approach with codecid/fei key I can't invent how to keep all codec features in one place. I.e., for some codecs I must call `(bool*)s->m_pCORE->QueryCoreInterface(MFXIFEIEnabled_GUID)` but not for others. So we are adding yet another map `codecId` -> `need fei`, aren't we? Is it inevitable?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
Alexandr-Konovalov: This is supported since GCC 8.1 and clang 3.5, but -std=c++14 required.

_studio/shared/umc/codec/color_space_converter/include/umc_video_resizing.h
dvrogozh: Hm. Why we need the code which is under UMC_ENABLE_VPP_RESIZE then? If we don't exercise it within Jpeg, then it should go away.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
dvrogozh: Just call it for all codecs. For those which don't have FEI support you should get false. If you don't get it - fix implementation. No need to in 2 maps.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
dvrogozh: I prefer to trust standard. Yes, compilers may implement some features in advance.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
Alexandr-Konovalov: Cool, thanks.

_studio/shared/umc/codec/jpeg_enc/src/jpegenc.cpp
```diff
@@ -0,0 +1,5025 @@
...
+    return jerr;
+
+  return JPEG_OK;
+} // CJPEGEncoder::WriteAPP14()
+
+
+JERRCODE CJPEGEncoder::WriteCOM(
+  char*  comment)
+{
+  std::string buf;
```
Alexandr-Konovalov: I switched to std::string here, hope this is OK. Code `sz = (int)MFX_MIN(strnlen_s(comment, 127),127);` from previous version looks suspicious.

_studio/shared/umc/codec/color_space_converter/src/umc_color_space_conversion.cpp
Alexandr-Konovalov: Done.

_studio/shared/umc/codec/jpeg_dec/src/jpegdec.cpp
Alexandr-Konovalov: Done.

_studio/shared/umc/codec/color_space_converter/include/umc_video_resizing.h
Alexandr-Konovalov: So, removed.

_studio/shared/umc/codec/color_space_converter/src/umc_color_space_conversion.cpp
```diff
@@ -0,0 +1,1003 @@
...
+    pSrcStep[2] *= 2;
+    srcFormat = YUV420;
+  }
+  const uint8_t *pYVU[3] = {pSrc[0], pSrc[2], pSrc[1]};
+  int32_t pYVUStep[3] = {pSrcStep[0], pSrcStep[2], pSrcStep[1]};
+  uint8_t *pDstYVU[3] = {pDst[0], pDst[2], pDst[1]};
+  int32_t pDstStepYVU[3] = {pDstStep[0], pDstStep[2], pDstStep[1]};
+  int status;
+
+#if !UMC_ALL_COLOR_CONVERSIONS_DEFINED
```
dvrogozh: This also looks to be a macro which was never defined.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -61,6 +64,128 @@
...
 #endif
 
+
+struct CodecKey {
+    const mfxU32 codecId;
+    const bool   fei;
+
+    CodecKey(mfxU32 codecId, bool fei) : codecId(codecId), fei(fei) {}
+
+    friend bool operator<(CodecKey l, CodecKey r)
```
dvrogozh: Maybe stupid question, but why this is 'friend' and not a struct member? @mgonchar : is that ok?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -388,53 +415,17 @@ mfxStatus MFXVideoENCODE_QueryIOSurf(mfxSession session, mfxVideoParam *par, mfx
...
-#if defined(MFX_ENABLE_H265_VIDEO_ENCODE)
-            case MFX_CODEC_HEVC:
-            {
 #if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
-                bool * feiEnabled = (bool*)session->m_pCORE->QueryCoreInterface(MFXIFEIEnabled_GUID);//required to check FEI plugin registration.
-                MFX_CHECK_NULL_PTR1(feiEnabled);
-                if(*feiEnabled)
-                    mfxRes = MfxHwH265FeiEncode::H265FeiEncode_HW::QueryIOSurf(session->m_pCORE.get(), par, request);
-                else
+            // required to check FEI plugin registration
```
dvrogozh: Remove `#ifdef`. Make sure MFXIFEIEnabled_GUID is not covered by any macro to be able to request it always.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -169,57 +294,17 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
...
-#if defined(MFX_ENABLE_H265_VIDEO_ENCODE)
-            case MFX_CODEC_HEVC:
-            {
 #if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
-                bool * feiEnabled = (bool*)session->m_pCORE->QueryCoreInterface(MFXIFEIEnabled_GUID);//required to check FEI plugin registration.
-                MFX_CHECK_NULL_PTR1(feiEnabled);
-                if(*feiEnabled)
-                    mfxRes = MfxHwH265FeiEncode::H265FeiEncode_HW::Query(session->m_pCORE.get(), in, out);
-                else
+            // required to check FEI plugin registration
```
dvrogozh: Remove `#ifdef`. Make sure MFXIFEIEnabled_GUID is not covered by any macro to be able to request it always.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -169,57 +294,17 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
...
-#if defined(MFX_ENABLE_VP9_VIDEO_ENCODE)
-            case MFX_CODEC_VP9:
-                mfxRes = MfxHwVP9Encode::MFXVideoENCODEVP9_HW::Query(&session->m_coreInt, in, out);
-                break;
-#endif //MFX_ENABLE_VP9_VIDEO_ENCODE
-
-            default:
-                mfxRes = MFX_ERR_UNSUPPORTED;
-            } // switch (out->mfx.CodecId)
+            auto handlers = codecId2Handlers.find(CodecKey(out->mfx.CodecId, fei));
```
dvrogozh: I think you can try to dispose of `#ifdef MFX_ENABLE_USER_ENCODE` section in the same way you did with FEI. I.e. add one more entry for the key:
```
struct CodecKey {
    const mfxU32 codecId;
    const bool   fei;
    const bool plugin;
}
```

And define
```
static const CodecId2Handlers codecId2Handlers =
{
 #ifdef MFX_ENABLE_USER_ENCODE
   {
       { .plugin=true, },
       { [](mfxSession s){ s->m_plgEnc->Query(session->m_pCORE.get(), in, out); } }
   }
#endif
}

```

You **must make sure** however that plugin entry is always the first to be found in a map, i.e. if `codecId2Handlers.find(CodecKey(out->mfx.CodecId, fei, isPlugin=true))`  should fall to plugin, not to codec. (user may register its own implementation of AVC as a plugin and it should have a preference).

@mgonchar : how do you feel about this idea?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -157,7 +168,9 @@ static const CodecId2Handlers codecId2Handlers =
             false
         },
         {
-            {   // .ctor =
+            // .fallback =
```
dvrogozh: This looks to be a mistake. You removed HW implementation entirely, called it fallback and did not define fallback at all:).

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -80,13 +80,20 @@ struct CodecKey {
 };
 
 struct Handlers {
+    typedef std::function<VideoENCODE*(VideoCORE* core, mfxCoreInterface* coreInt, mfxStatus *mfxRes)> CtorFunc;
```
dvrogozh: Better to define such typedefs right away in prev. patches rather than modify something you just recently introduced.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -269,6 +425,13 @@ VideoENCODE* _mfxSession::Create<VideoENCODE>(mfxVideoParam& par)
     {
         return nullptr;
     }
+    // if ctorFallback present, create FallbackProxyENCODE
+    if (handlers->second.fallback.ctor)
+    {
+        return new FallbackProxyENCODE(core, &m_coreInt, CodecId, MFX_PLATFORM_HARDWARE == m_currentPlatform, fei);
```
dvrogozh: That's over complex. Just use the handle you found to initialize FallbackProxyENCODE: `new FallbackProxyENCODE(handlers)` You will avoid second search in Fallback class and use codec definition from the handlers table right away. I.e. store reference to the handler in fallback class.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -216,67 +249,36 @@ static const CodecId2Handlers codecId2Handlers =
...
-            bool * feiEnabled = (bool*)core->QueryCoreInterface(MFXIFEIEnabled_GUID);//required to check FEI plugin registration.
-            if (feiEnabled == nullptr)
-                return nullptr;
-            if(*feiEnabled)
-                pENCODE = new MfxHwH265FeiEncode::H265FeiEncode_HW(core, &mfxRes);
-            else
-#endif
-                pENCODE = new MfxHwH265Encode::MFXVideoENCODEH265_HW(core, &mfxRes);
-        break;
+    bool *feiEnabled = (bool*)m_pCORE.get()->QueryCoreInterface(MFXIFEIEnabled_GUID);
```
dvrogozh: same: remove #if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -169,57 +294,17 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
...
-                mfxRes = MfxHwVP9Encode::MFXVideoENCODEVP9_HW::Query(&session->m_coreInt, in, out);
-                break;
-#endif //MFX_ENABLE_VP9_VIDEO_ENCODE
-
-            default:
-                mfxRes = MFX_ERR_UNSUPPORTED;
-            } // switch (out->mfx.CodecId)
+            auto handlers = codecId2Handlers.find(CodecKey(out->mfx.CodecId, fei));
+            mfxRes = handlers == codecId2Handlers.end() ? MFX_ERR_UNSUPPORTED
+                : (handlers->second.primary.query)(session, in, out);
```
dvrogozh: `auto handlers` is wrong name, it should be `auto handle` - yous search **single** handle.

@mgonchar : should it be `auto& handle`?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -80,20 +83,16 @@ struct CodecKey {
 };
 
 struct Handlers {
-    typedef std::function<VideoENCODE*(VideoCORE* core, mfxCoreInterface* coreInt, mfxStatus *mfxRes)> CtorFunc;
-
     struct Funcs {
-        CtorFunc ctor;
+        std::function<VideoENCODE*(VideoCORE* core, mfxCoreInterface* coreInt, mfxStatus *mfxRes)> ctor;
```
dvrogozh: You have some strange playing in prev. patches around this: please, remove and use single variant right away.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -182,7 +181,24 @@ static const CodecId2Handlers codecId2Handlers =
                 // .queryIOSurf =
                 [](mfxSession s, mfxVideoParam *par, mfxFrameAllocRequest *request)
                 { return MFXVideoENCODEMJPEG_HW::QueryIOSurf(s->m_pCORE.get(), par, request); }
+            },
+#if defined(MFX_ENABLE_SW_FALLBACK)
+            // .fallback =
```
dvrogozh: you already have a section commented as `// .fallback` :)

_studio/shared/umc/codec/jpeg_enc/src/jpegenc.cpp
dvrogozh: much better, thank you. Don't forget to move this patch to internal repo.

_studio/shared/umc/codec/jpeg_enc/src/jpegenc.cpp
```diff
@@ -0,0 +1,5025 @@
...
+  JERRCODE jerr;
+
+  TRC0("-> WriteCOM");
+
+  {
+    std::stringstream str;
+    str << "Intel(R) MSDK JPEG encoder [" << MFX_VERSION_MAJOR <<"."<< MFX_VERSION_MINOR <<"]";
```
dvrogozh: @onabiull : check the name, please.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: I disagree: in each `Handlers` structure we have bunch of handlers.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: Thank you, this is typo in the comment.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: I see that both `query` and `queryIOSurf` for plugin are codec-agnostic. So, can we use single universal handles for every plugin? And map it to some forever-unused codec ID. The code is unified, and plugin preference is provided automatically. One is added as a separate commit.

dvrogozh: Yes, you should have single entry for the plugins like {.isplugin=true, .isfei=false, .codecid=0 }

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dvrogozh: You search for the single entry in the codecIdHandlers table. As a result handlers is misleading to me.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dvrogozh: Just reuse the whole Funcs structure for fallbacks right away. I don't see the need to create 2 structures. Yes, you will not use some members in fallbacks, but I don't see this as a problem.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -2074,5 +2033,239 @@ mfxStatus VideoDECODEMJPEGBase_HW::CompleteTask(void *pParam, mfxStatus )
...
+        {
+            delete m_freeTasks.front();
+            m_freeTasks.pop();
+        }
+    }
+
+    memset(&m_stat, 0, sizeof(mfxDecodeStat));
+    m_vPar = *par;
+
+    MFX_CHECK(m_FrameAllocator->Reset() == UMC::UMC_OK, MFX_ERR_MEMORY_ALLOC);
```
mgonchar: Please don't put functions call to macro. This one would have additional assert at debug build, which will actually cause first argument to appear twice in macro expansion, i.e. Reset will be called twice. I suggest to save result to variable and then check it

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -2074,5 +2033,239 @@ mfxStatus VideoDECODEMJPEGBase_HW::CompleteTask(void *pParam, mfxStatus )
...
+
+    return MFX_ERR_NONE;
+}
+
+mfxStatus VideoDECODEMJPEGBase_SW::Close()
+{
+    m_FrameAllocator->Close();
```
mgonchar: I think here some status returned as well, so we need to check it

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -2074,5 +2033,239 @@ mfxStatus VideoDECODEMJPEGBase_HW::CompleteTask(void *pParam, mfxStatus )
...
+}
+
+mfxStatus VideoDECODEMJPEGBase_SW::FillEntryPoint(MFX_ENTRY_POINT *pEntryPoint, mfxFrameSurface1 *surface_work, mfxFrameSurface1 *surface_out)
+{
+    pLastTask = m_freeTasks.front();
+
+    pLastTask->surface_work = surface_work;
+    pLastTask->surface_out = surface_out;
+
+    pEntryPoint->requiredNumThreads = MFX_MIN(pLastTask->m_pMJPEGVideoDecoder->NumDecodersAllocated(),
```
mgonchar: please use std::min

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
```diff
@@ -131,6 +131,37 @@ class VideoDECODEMJPEGBase_HW : public VideoDECODEMJPEGBase
...
+    mfxStatus ReserveUMCDecoder(UMC::MJPEGVideoDecoderBaseMFX* &pMJPEGVideoDecoder, mfxFrameSurface1 *surf, bool isOpaq) override;
+    void ReleaseReservedTask() override;
+    mfxStatus AddPicture(UMC::MediaDataEx *pSrcData, mfxU32 & numPic) override;
+    mfxStatus AllocateFrameData(UMC::FrameData *&data) override;
+    mfxStatus FillEntryPoint(MFX_ENTRY_POINT *pEntryPoint, mfxFrameSurface1 *surface_work, mfxFrameSurface1 *surface_out) override;
+
+protected:
+    CJpegTask *pLastTask;
+    // Free tasks queue (if SW is used)
+    std::queue<CJpegTask *> m_freeTasks;
```
mgonchar: maybe ```std::queue<std::unique_ptr<CJpegTask>>```

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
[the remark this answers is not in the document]
mgonchar: hm, it depends on how this container is really used. If random access to elements is harmful, then - no. If it doesn't matter - deque always better

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
mgonchar: JFYI, I know that driver is strictly fixed to use c++ not higher than c++11, so maybe we will be stuck with it too. I also hope to see more advanced standards to be allowed

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
mgonchar: yes, it will work... but I also didn't get the point of friendship. @Alexandr-Konovalov could you comment?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -54,77 +60,446 @@
...
+#endif // MFX_ENABLE_SW_FALLBACK
+
+template<>
+VideoENCODE* _mfxSession::Create<VideoENCODE>(mfxVideoParam& par)
+{
+    VideoCORE* core = m_pCORE.get();
+    mfxU32 CodecId = par.mfx.CodecId;
+
+    // create a codec instance
+    bool *feiEnabled = (bool*)m_pCORE.get()->QueryCoreInterface(MFXIFEIEnabled_GUID);
```
mgonchar: no need in .get()

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -152,113 +528,42 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
...
-            mfxRes = session->m_plgEnc->Query(session->m_pCORE.get(), in, out);
-            if (mfxRes >= MFX_ERR_NONE &&
-              mfxRes != MFX_WRN_PARTIAL_ACCELERATION)
-              bIsHWENCSupport = true;
-            if (MFX_WRN_PARTIAL_ACCELERATION == mfxRes)
-            {
-                mfxRes = MFX_ERR_UNSUPPORTED;
-            }
+            handler = codecId2Handlers.find(CodecKey(MFX_CODEC_DUMMY_FOR_PLUGIN, false));
+            assert(handler != codecId2Handlers.end());
```
mgonchar: I believe it won't perform any check in release

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -54,77 +60,446 @@
...
+
+static const CodecId2Handlers codecId2Handlers =
 {
-    VideoENCODE* pENCODE = nullptr;
-    VideoCORE* core = m_pCORE.get();
-    mfxStatus mfxRes = MFX_ERR_MEMORY_ALLOC;
-    mfxU32 CodecId = par.mfx.CodecId;
-    // create a codec instance
-    switch (CodecId)
+#if defined(MFX_ENABLE_USER_ENCODE)
```
mgonchar: leave a comment to warn everyone that this entry should be first

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
mgonchar: @dvrogozh why reference, it is an iterator with cheap copy. Moreover as function doesn't return reference the legal is only assignment to const reference

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: I'm lost. Could you please clarify why it should be? To me, order is not important, I never relay on it.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: `friend` is to define a non-member function (overloaded operator in our case) in the class.
And this is to keep a class and a related operator together.

If you work group prefers making `operator<` a class member, ok, let's do it this way,
it's matter of taste.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: This is exactly logic I have. If `session->m_plgEnc.get()`, we use special `MFX_CODEC_DUMMY_FOR_PLUGIN`, and guarantee to have right handler (this handler is codec-independent). If not, other appropriate handlers are called. Why should we order something? We always know what to call.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -479,7 +523,13 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
 
             if (MFX_WRN_PARTIAL_ACCELERATION == mfxRes)
             {
+                assert(handler != codecId2Handlers.end());
+#if defined(MFX_ENABLE_SW_FALLBACK)
```
dvrogozh: You don't need ifdef here. If MFX_ENABLE_SW_FALLBACK is not defined handler->second.fallback.query will be true and you correctly return MFX_ERR_UNSUPPORTED via runtime check.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -407,11 +448,13 @@ VideoENCODE* _mfxSession::Create<VideoENCODE>(mfxVideoParam& par)
     {
         return nullptr;
     }
+#if defined(MFX_ENABLE_SW_FALLBACK)
```
dvrogozh: Hmm... Well, we can drop this as well and always define FallbackProxyENCODE class, but probably let's keep the check for now. Will safe a little bit on skipping class definition when we don't need it.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -569,13 +623,19 @@ mfxStatus MFXVideoENCODE_QueryIOSurf(mfxSession session, mfxVideoParam *par, mfx
 
             if (MFX_WRN_PARTIAL_ACCELERATION == mfxRes)
             {
+                assert(handler != codecId2Handlers.end());
+#if defined(MFX_ENABLE_SW_FALLBACK)
```
dvrogozh: not needed.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -79,6 +79,7 @@ struct CodecKey {
 
 struct Handlers {
     struct Funcs {
```
dmitry-gurulev: Why do we need this [Funcs] struct? Might all that [ctor\query] be defined at [Handlers] directly?

Alexandr-Konovalov: We need primary handlers and same set of handlers to support fallback. Certainty, we may repeat all 6 in rows, but to me having separate cases is more structured.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -97,12 +98,18 @@ static const CodecId2Handlers codecId2Handlers =
             false
         },
         {
-            { [](mfxSession s, mfxVideoParam *in, mfxVideoParam *out)
-              {
-                  return (!s->m_pENCODE.get())?
-                  MFXHWVideoENCODEH264::Query(s->m_pCORE.get(), in, out)
-                  : MFXHWVideoENCODEH264::Query(s->m_pCORE.get(), in, out, s->m_pENCODE.get());
-              }
+            {   // .query =
```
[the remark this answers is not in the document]
Alexandr-Konovalov: The idea was "we have a single field, so it's clear what is initializing; now we have 2 fields, let's make this explicit".

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -85,7 +88,9 @@ struct Handlers {
     };
 
     Funcs primary;
+#if defined(MFX_ENABLE_SW_FALLBACK)
```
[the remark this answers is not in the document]
Alexandr-Konovalov: clang under Android is called with `-W`, that implies `-Wmissing-field-initializers`, so unfortunately we get either this or multiple `nullptr` in `codecId2Handlers`.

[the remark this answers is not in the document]
dvrogozh: I don't understand how that can be. You use std::function which has constructor. So you don't need to initialize it. Have you actually tried? Can you post exact error here? Maybe it wants just `{}` for the `.fallback`?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -72,6 +72,12 @@ struct CodecKey {
 
     CodecKey(mfxU32 codecId, bool fei) : codecId(codecId), fei(fei) {}
 
+    enum {
+        // special value for codecId to denote plugin, it must be
+        // different from other MFX_CODEC_* used in codecId2Handlers
+        MFX_CODEC_DUMMY_FOR_PLUGIN = 0
+    };
+
```
dvrogozh: "Exact ordering rule is unsignificant" - this becomes untrue after this commit!! change the comment - see our discussion below.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
Alexandr-Konovalov: Please find an isolated test case. Thanks, `{}` do the trick. I do not understand why existence of empty ctor for std::function makes the compiler happy, not for Funcs it improves nothing.

```
#include <functional>
#include <map>

struct Handlers {
    struct Funcs {
        std::function<int(void)> ctor;
        Funcs() {}
        Funcs(std::function<int(void)> ctor) : ctor(ctor) {}
    };

    Funcs primary;
    Funcs fallback;
};

std::map<int, Handlers> m = {
    {
        1,
        {
            {
                []() -> int  { return 1; },
            },
#if 0
            {
            }
#endif
        }
    }
};

konovx@NNN:~/small$ clang++ -W -c -std=c++11 -Wall missing.cpp 
missing.cpp:26:9: warning: missing field 'fallback' initializer
      [-Wmissing-field-initializers]
        }
        ^
1 warning generated.
```

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -97,6 +103,38 @@ typedef std::map<CodecKey, Handlers> CodecId2Handlers;
...
+                },
+                // .queryIOSurf =
+                [](mfxSession s, mfxVideoParam *par, mfxFrameAllocRequest *request)
+                {
+                    assert(s->m_plgEnc);
+                    return s->m_plgEnc->QueryIOSurf(s->m_pCORE.get(), par, request, 0);
+                }
+            },
+            // .fallback =
+            {
```
dvrogozh: Looks like compiler just wants a constructor... Can you, please, check whether declaring constructor in Funcs will fix the issue and avoid this `{}`:
```
struct Funcs {
  Funcs() = default;
}
```

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -54,13 +57,153 @@
...
 #endif
 
-#if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
 #if defined(MFX_VA)
 #include "mfx_h265_fei_encode_hw.h"
 #include <libmfx_core_interface.h>
 #endif
+
+
+struct CodecKey {
```
dmitry-gurulev: I'd rather move that custom [CodecKey] logic into a handler u're binding:
```
/*helper function to check if FEI is supported*/
inline
bool check_fei(mfxSession s, mfxU32 /*codec*/)
{
    auto enabled = static_cast<bool*>(session->m_pCORE->QueryCoreInterface(MFXIFEIEnabled_GUID)); 
    return enabled && *enabled;
}

//... codec2helpers table
[](mfxSession s, mfxVideoParam *in, mfxVideoParam *out) 
{
    return check_fei(s, out->mfx.CodecId) ?
        MfxHwH265FeiEncode::H265FeiEncode_HW::Query(s->m_pCORE.get(), in, out) :
        MfxHwH265Encode::MFXVideoENCODEH265_HW::Query(s->m_pCORE.get(), in, out)
    ;
}
```

Profit: 
1. U don't need to introduce a new class to manage FEI handlers, actually it's just one.
2. U don't need to have 2 handlers for each codec.

dvrogozh: That would be a good idea if we wouldn't have _different_ macro definitions covering FEI and non-FEI. As a result we will once more time have a nightmare disaster. Right now we at lease have straight forward work with the macroses.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -293,6 +311,11 @@ class FallbackProxyENCODE : public VideoENCODE
     {
         assert(m_impl);
         mfxStatus mfxRes = m_impl->Reset(par);
+        // SW fallback if EncodeGUID is absence
+        if (m_useSWFallback && MFX_ERR_NONE <= mfxRes)
+        {
+            mfxRes = MFX_WRN_PARTIAL_ACCELERATION;
```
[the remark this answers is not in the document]
dvrogozh: I thought that this was old behavior copied from the open source. But I agree - it seems strange. If component removed some warning we wipe it out. And if we remove whatever component returns, we have no way to understand whether we experience SW fallback or not. Dead loop:).

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -169,57 +310,14 @@ mfxStatus MFXVideoENCODE_Query(mfxSession session, mfxVideoParam *in, mfxVideoPa
...
-                else
-#endif
-                    mfxRes = MfxHwH265Encode::MFXVideoENCODEH265_HW::Query(session->m_pCORE.get(), in, out);
-                break;
-            }
-#endif // MFX_ENABLE_H265_VIDEO_ENCODE
+            // required to check FEI plugin registration
+            bool *feiEnabled = (bool*)session->m_pCORE->QueryCoreInterface(MFXIFEIEnabled_GUID);
+            MFX_CHECK_NULL_PTR1(feiEnabled);
+            const bool fei = *feiEnabled;
```
dvrogozh: But I like the idea with check_fei inline helper. It will simplify this and few other places to just:
`const bool fei = check_fei(session, codec)`

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dmitry-gurulev: Isn't this an option?
```
[](mfxSession s, mfxVideoParam *in, mfxVideoParam *out) 
{
    return check_fei(s, out->mfx.CodecId) ?
#if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE) 
        MfxHwH265FeiEncode::H265FeiEncode_HW::Query(s->m_pCORE.get(), in, out) :
#else
        MFX_ERR_UNDEFINED_BEHAVIOR
#endif
        MfxHwH265Encode::MFXVideoENCODEH265_HW::Query(s->m_pCORE.get(), in, out)
    ;
}
```

I really don't think that having dozen of stubs like
```
{
    MFX_CODEC_AVC, 
    // .fei = 
    false 
}
```
is more suitable.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -94,14 +237,12 @@ VideoENCODE* _mfxSession::Create<VideoENCODE>(mfxVideoParam& par)
 #if defined(MFX_ENABLE_H265_VIDEO_ENCODE)
     case MFX_CODEC_HEVC:
     {
-#if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
```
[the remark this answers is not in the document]
Alexandr-Konovalov: All usage of `MfxHwH265FeiEncode::H265FeiEncode_HW` is protected by `MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE`, please see giant `codecId2Handlers` table.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -329,13 +427,11 @@ mfxStatus MFXVideoENCODE_QueryIOSurf(mfxSession session, mfxVideoParam *par, mfx
 #if defined(MFX_ENABLE_H265_VIDEO_ENCODE)
             case MFX_CODEC_HEVC:
             {
-#if defined (MFX_ENABLE_HEVC_VIDEO_FEI_ENCODE)
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Could you please explain a little? This is `MFXVideoENCODE_QueryIOSurf()`, and in  _Move MFXHWVideoENCODE*::QueryIOSurf to a table_ patch I move this functionality to `codecId2Handlers` table, `MFXVideoENCODEMPEG2_HW::QueryIOSurf` call is protected by `MFX_ENABLE_MPEG2_VIDEO_ENCODE`. Why is this wrong and how is it related to ctors?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: There is no difference between empty and default constructors in this regard. The compiler still warns us.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -280,6 +297,7 @@ class FallbackProxyENCODE : public VideoENCODE
                         bool isPlatformHardware)
         : m_core(core), m_coreInt(coreInt), m_codecHandlers(handlers)
         , m_isPlatformHardware(isPlatformHardware)
+        , m_useSWFallback(false)
```
dmitry-gurulev: Can we use [core->GetPlatfromType] instead of [m_isPlatformHardware] flag?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -293,6 +311,11 @@ class FallbackProxyENCODE : public VideoENCODE
     {
         assert(m_impl);
         mfxStatus mfxRes = m_impl->Reset(par);
+        // SW fallback if EncodeGUID is absence
```
dmitry-gurulev: What does it mean 'EncodeGUID is absence' in that context? I see no 'EncodeGUID' around, comment is just confusing.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -398,11 +422,21 @@ mfxStatus FallbackProxyENCODE::Init(mfxVideoParam *par)
...
+        if (m_isPlatformHardware)
+        {
+            m_useSWFallback = true;
+        }
         m_impl.reset(m_codecHandlers.fallback.ctor(m_core, m_coreInt, &mfxRes));
         MFX_CHECK(m_impl.get(), MFX_ERR_NULL_PTR);
         MFX_CHECK(mfxRes >= MFX_ERR_NONE, mfxRes);
         mfxRes = m_impl->Init(par);
     }
+    // SW fallback if EncodeGUID is absence
```
dmitry-gurulev: See prev. comment above about this

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: Thanks!

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: There are 4 places with that, I have nothing to clarify, so we may want to just drop them.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
Alexandr-Konovalov: Unfortunately with error processing it may not fit in single line.

dmitry-gurulev: @Alexandr-Konovalov,

Do u mean that we need to return [MFX_ERR_NULL_PTR] if `QueryCoreInterface(MFXIFEIEnabled_GUID)` returns NULL?

Alexandr-Konovalov: Yes, for queries. And nullptr from Create().

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dvrogozh: hm. strange. The trick forks for std::function somehow...

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
dvrogozh: I believe we clarified offline that Reset should NOT return MFX_WRN_PARTIAL_ACCELERATION for sure. We still have discussion around other functions, but at least Reset is clear: remove status modification.

dvrogozh:
Please, rebase - there is a conflict. Have you resolved Reset() comment?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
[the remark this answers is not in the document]
dvrogozh: @Alexandr-Konovalov: this is still not addressed!!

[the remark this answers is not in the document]
Alexandr-Konovalov: Sorry, fixed.

PR #691, opened 2018-09-28
Return an error on warnings from SyncOperation

dmitryermilov:
Currently the timeout used in SMT and sample_fei is 1500 seconds.
In regular case, it's even too much for task synchronization.
However in case of driver/MSDK synchronization problems,
SyncOperation may return MFX_WRN_IN_EXECUTION which was ignored.
It leaded to undefined behavior hard to debug.
So it's better to catch synchronization issues at once.

### DIFF
--- a/samples/sample_fei/src/auxiliary_interfaces.cpp
+++ b/samples/sample_fei/src/auxiliary_interfaces.cpp
@@ -184,8 +184,8 @@ mfxStatus MFX_VppInterface::VPPoneFrame(mfxFrameSurface1* pSurf_in, mfxFrameSurf
         }
         else if (MFX_ERR_NONE < sts && m_SyncPoint)
         {
-            // Ignore warnings if output is available
             sts = m_pmfxSession->SyncOperation(m_SyncPoint, MSDK_WAIT_INTERVAL);
+            MSDK_CHECK_ERR_NONE_STATUS(sts, MFX_ERR_ABORTED, "VPP: SyncOperation failed");
             mdprintf(stderr, "VPP synced : %d\n", sts);
 
             break;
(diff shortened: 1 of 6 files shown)


dmitryermilov:
>>For every MSDK_CHECK_NOT_EQUAL we need to keep error message reporting functionality 
Error messages are present with  MSDK_CHECK_NOT_EQUAL . Even line number is printed.

samples/sample_fei/src/auxiliary_interfaces.cpp
```diff
@@ -186,6 +186,7 @@ mfxStatus MFX_VppInterface::VPPoneFrame(mfxFrameSurface1* pSurf_in, mfxFrameSurf
         {
             // Ignore warnings if output is available
```
otsirule: Seems such comments are not valid anymore

samples/sample_fei/src/fei_encode.cpp
```diff
@@ -739,7 +739,7 @@ mfxStatus FEI_EncodeInterface::EncodeOneFrame(iTask* eTask)
             {
                 // ignore warnings if output is available
                 sts = m_pmfxSession->SyncOperation(m_SyncPoint, MSDK_WAIT_INTERVAL);
-                MSDK_CHECK_STATUS(sts, "FEI ENCODE: SyncOperation failed");
+                MSDK_CHECK_ERR_NONE_STATUS(sts, MFX_ERR_ABORTED, "SyncOperation failed");
```
mgonchar: I liked previous message, when failed interface explicitly mentioned it is easier to read error messages. I know that this information still can be extracted from code line, but it adds a little bit more work, especially when you code base and error log mismatches

PR #692, opened 2018-09-28
Dispatcher4Windows

fzhar:


api/opensource/mfx_dispatch/src/mfx_dispatcher.cpp
```diff
@@ -25,8 +25,13 @@
 #include <assert.h>
 
 #include <string.h>
+#if defined(_WIN32) || defined(_WIN64)
+    #include <windows.h>
+    #pragma warning(disable:4355)
```
dvrogozh: please, try to remove and try without it. If warning still persist - try to fix.

api/opensource/mfx_dispatch/src/mfx_dispatcher.cpp
```diff
@@ -25,8 +25,13 @@
...
+#else
     #include <dlfcn.h>
     #include <iostream>
+#endif // defined(_WIN32) || defined(_WIN64)
```
dvrogozh: hm. is that the only place which miss Windows support? Did you check that Windows dll loading code is actually available?

dmitryermilov:
@fzhar , any progress ?

dvrogozh:
This patch is now hanging in the air. Old dispatcher implementation is gone and a new dispatcher is Linux only. I am strongly against bringing old dispatcher for Windows without its deep refactoring and rewrite. I think this PR should be closed and new one created once Windows part will be really ready.

fzhar:
I agree with closing this commit but we'll have to add Windows dispatcher as-is in the nearest time to be able to upload Windows Samples

PR #694, opened 2018-10-01
[VP9e-plugin] Segmentation corner cases processing

eshiryae:
Taking care of cases when segmentation is set but doesn't make sense
because nothing is changed or can be applied directly to frame params.
Managed on the MSDK side after agreement with the driver team as it
requires more efforts to do such checkings in the driver


dmitryermilov:
@stasefremov , can you please review?

PR #699, opened 2018-10-02
[hevce fei] Fix memleak caused by commit b327806

otsirule:


dmitryermilov:
@artem-shaporenko , please review

_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_plugin_hw.h
```diff
@@ -158,6 +158,12 @@ namespace MfxHwH265FeiEncode
         ~H265FeiEncodePlugin()
         {}
 
+        static H265FeiEncodePlugin* GetInstance()
+        {
+            static H265FeiEncodePlugin instance(false);
```
artem-shaporenko: Probably better to use m_adapter instead of static object creation?

_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_plugin_hw.h
```diff
@@ -49,7 +49,7 @@ namespace MfxHwH265FeiEncode
 
             try
             {
-                tmp_pplg = new H265FeiEncodePlugin(false);
+                tmp_pplg = GetInstance();
```
mgonchar: no operator new, no bad_alloc exception, so remove that handle

_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_plugin_hw.h
mgonchar: This is static function it can be called even without any real H265FeiEncodePlugin instance existing, so you can't use H265FeiEncodePlugin::m_adapter here

_studio/hevc_fei/h265_fei/include/mfx_h265_fei_encode_plugin_hw.h
artem-shaporenko: Agree

mgonchar:
@dmitryermilov I'm ok with current state

mgonchar:
@dmitryermilov I'm ok with current state

PR #702, opened 2018-10-03
Return an error enabling/disabling composition via Reset

dmitryermilov:
This is a workaround to prevent undefined behavior.
Enabling/disabling composition via Reset() doesn't work currently.


_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
```diff
@@ -1211,17 +1211,21 @@ mfxStatus VideoVPPBase::Reset(mfxVideoParam *par)
         }
     }// Opaque
 
+    bool isCompositionModeEnabledInNewParams = (MFX_ERR_NONE == GetCompositionEnabledStatus(par)) ? true : false;
```
mgonchar: why do you use ?: here to convert bool into bool?
operator == already returns bool

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
```diff
@@ -1211,17 +1211,21 @@ mfxStatus VideoVPPBase::Reset(mfxVideoParam *par)
...
+
+    // Enabling/disabling composition via Reset() doesn't work currently.
+    // This is a workaround to prevent undefined behavior.
+    if (m_errPrtctState.isCompositionModeEnabled != isCompositionModeEnabledInNewParams)
```
mgonchar: Use MFX_CHECK macro instead, it would allow us to see this check in log if MSDK traces would be enabled

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
dmitryermilov: check what GetCompositionEnabledStatus returns

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
dmitryermilov: makes sense. thannks

_studio/mfx_lib/vpp/src/mfx_vpp_sw_internal.cpp
```diff
@@ -327,24 +327,20 @@ mfxStatus GetExternalFramesCount(VideoCORE* core,
 } // mfxStatus GetExternalFramesCount(...)
 
 // all check must be done before call
-mfxStatus GetCompositionEnabledStatus(mfxVideoParam* pParam )
+bool IsCompositionMode(mfxVideoParam* pParam)
```
mgonchar: you haven't checked pParam for null

_studio/mfx_lib/vpp/src/mfx_vpp_sw_internal.cpp
```diff
@@ -327,24 +327,20 @@ mfxStatus GetExternalFramesCount(VideoCORE* core,
...
 {
-    mfxU32 bufferIndex;
-
-    if( pParam->ExtParam && pParam->NumExtParam > 0 )
+    if (pParam->ExtParam && pParam->NumExtParam > 0)
     {
-        for( bufferIndex = 0; bufferIndex < pParam->NumExtParam; bufferIndex++ )
+        for (mfxU32 i = 0; i < pParam->NumExtParam; ++i)
```
mgonchar: this is actually find_if (or in this particular case any_of):
```
MFX_CHECK(pParam && pParam->ExtParam && pParam->NumExtParam > 0, false);

return std::any_of(pParam->ExtParam, pParam->ExtParam + pParam->NumExtParam, [](const mfxExtBuffer* pBuffer){ return pBuffer && pBuffer->BufferId == mfxU32(MFX_EXTBUFF_VPP_COMPOSITE); })
```

But I don't really insist on such change, missing pointer check is real issue here

PR #707, opened 2018-10-04
Fix condition in VPP reset when checking composition enable/disable changes.

HermanMolinder:
Wrong condition on commit 8ee12dd7c5cbb48df992c83b572343c7a2bad914 when checking if composition has been enabled/disabled.


dmitryermilov:
Thanks @HermanMolinder !

PR #709, opened 2018-10-05
[VP9] Fix PackPicParams for streams starting w/ inter frame

vshampor:
A VP9 stream starting w/ inter frame has it intra-coded without
references. Prior code led to error since the auxiliary ref frame array
was filled with VA_INVALID_SURFACE ids, which cannot be passed
to the GetSurfaceID functions.


vshampor:
This is the same as 39c0ce52a38e74b1055385b5f31f8273831f357a merged to master.

vshampor:
Not approved for merging into the release branch.

PR #710, opened 2018-10-05
core: check library version before creating a session

dvrogozh:
There is no point to avoid checking library version right away
on MFXInit* calls. This will safe memory allocation/free if user
requested a wrong version for any reason.


_studio/mfx_lib/shared/src/libmfxsw.cpp
```diff
@@ -100,7 +100,14 @@ mfxStatus MFXInitEx(mfxInitParam par, mfxSession *session)
     MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_API, "MFXInit");
     MFX_LTRACE_1(MFX_TRACE_LEVEL_API, "^ModuleHandle^libmfx=", "%p", g_hModule);
 
-        // check error(s)
+    // check the library version
+    if ((MFX_VERSION_MAJOR != par.Version.Major) ||
+        (MFX_VERSION_MINOR < par.Version.Minor))
+    {
+        mfxRes = MFX_ERR_UNSUPPORTED;
```
dmitry-gurulev: Looks like we have to return w/ unsupported here rather than just assign [mfxRes]

dvrogozh: Yes. My mistake. Thank you.

PR #719, opened 2018-10-10
Remove vm cond

akokoshn:


_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -286,31 +278,18 @@ mfxStatus mfxSchedulerCore::Synchronize(mfxTaskHandle handle, mfxU32 timeToWait)
     }
     else
     {
-        UMC::AutomaticMutex guard(m_guard);
+        std::unique_lock<std::mutex> guard(m_guard);
```
dvrogozh: these 3 variables below are not needed anymore

_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
```diff
@@ -248,6 +240,11 @@ class mfxSchedulerCore : public MFXIScheduler2
     // external synchronization is required.
     //
 
+
```
dvrogozh: Please, avoid moving the code around if you don't need it.

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -460,6 +439,8 @@ mfxStatus mfxSchedulerCore::WaitForTaskCompletion(const void *pOwner)
 
 mfxStatus mfxSchedulerCore::ResetWaitingStatus(const void *pOwner)
 {
+    std::lock_guard<std::mutex> guard(m_guard);
```
dvrogozh: Please, guard only WakeUpThreads. ResetWaitingTasks can actually take a long time especially on multisession use cases.

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_thread.cpp
```diff
@@ -135,6 +134,7 @@ void mfxSchedulerCore::WakeupThreadProc()
 
             //MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_SCHED, "HW Event");
             IncrementHWEventCounter();
+            std::lock_guard<std::mutex> guard(m_guard);
```
dvrogozh: Please, add {} around lock and protected function. This emphasize what you actually protect.

_studio/shared/include/mfx_mfe_adapter.h
```diff
@@ -27,10 +27,11 @@
 #include <map>
 #include <list>
 #include <vector>
-#include "vm_mutex.h"
-#include "vm_cond.h"
 #include "vm_time.h"
 #include <mfxstructures.h>
+#undef max
+#undef min
```
dvrogozh: Why you need these undefs? I am _not_ ok with such a code. This is a hack which should be understood and eliminated.

_studio/shared/src/mfx_mfe_adapter.cpp
```diff
@@ -389,7 +356,7 @@ mfxStatus MFEVAAPIEncoder::Submit(VAContextID context, vm_tick timeToWait, bool
             }
             m_framesCollected -= m_contexts.size();
         }
-        vm_cond_broadcast(&m_mfe_wait);
+        m_mfe_wait.notify_all();
```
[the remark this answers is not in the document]
dvrogozh: We have multiple encoders calling this function, let's say 3. Only one of them is allowed to pass thru and submit a task to the driver. And it submits the combined task for all 3 encoders. Once he is done, he should wake up 2 remained encoders which are still waiting for them to continue work (they will see that their task is already submitted and simply exit the function). So, yes, notify_all() is a must here.

_studio/shared/src/mfx_mfe_adapter.cpp
```diff
@@ -329,28 +309,15 @@ mfxStatus MFEVAAPIEncoder::Submit(VAContextID context, vm_tick timeToWait, bool
 
     if(!framesToSubmit)
     {
-        vm_mutex_unlock(&m_mfe_guard);
         return MFX_ERR_UNDEFINED_BEHAVIOR;
     }
     vm_tick start_tick = vm_time_get_tick();
```
dvrogozh: I don't think you need start_tick anymore, right?

_studio/shared/include/mfx_mfe_adapter.h
akokoshn: The problem is conflict definition of macros "max" from chrono and cm_mem_copy.h which includes in mfx_vpp_hw.h.
Conflict definitions happens in mfx_vpp_hw.cpp: 
    mfx_vpp_hw.h <- cm_mem_copy.h
    linmfx_core_vaapi.h <- mfx_mfe_adapter.h <- condition_variable <- chrono

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -34,15 +34,15 @@
 #include "mfx_common_int.h"
 #include "mfx_enc_common.h" // UMC::FrameRate
 
+#ifdef MFX_VA_LINUX
```
dvrogozh: Any reason to change where this is defined?

_studio/shared/include/mfx_mfe_adapter.h
```diff
@@ -27,10 +27,9 @@
 #include <map>
 #include <list>
 #include <vector>
-#include "vm_mutex.h"
-#include "vm_cond.h"
 #include "vm_time.h"
 #include <mfxstructures.h>
+#include <condition_variable>
```
dvrogozh: Please, move cv header closer to list and vector. And check whether you really need vm_time.h, I believe you don't.

_studio/shared/src/mfx_mfe_adapter.cpp
```diff
@@ -200,45 +186,42 @@ mfxStatus MFEVAAPIEncoder::Join(VAContextID ctx, vm_tick timeout)
...
-    vm_mutex_unlock(&m_mfe_guard);
     return (VA_STATUS_SUCCESS == vaSts)? MFX_ERR_NONE: MFX_ERR_DEVICE_FAILED;
 }
 
 mfxStatus MFEVAAPIEncoder::Destroy()
 {
-    vm_mutex_lock(&m_mfe_guard);
+    VAStatus vaSts = VA_STATUS_SUCCESS;
+    {
+        std::lock_guard<std::mutex> guard(m_mfe_guard);
```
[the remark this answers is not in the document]
artem-shaporenko: It is protected

_studio/shared/src/mfx_mfe_adapter.cpp
```diff
@@ -329,28 +308,12 @@ mfxStatus MFEVAAPIEncoder::Submit(VAContextID context, vm_tick timeToWait, bool
...
-        {
-            spent_ticks = (vm_time_get_tick() - start_tick);
-        }
-        else
-        {
-            vm_mutex_unlock(&m_mfe_guard);
-            return MFX_ERR_UNKNOWN;
-        }
-    }
+    m_mfe_wait.wait_for(guard, std::chrono::milliseconds(timeToWait), [this, &framesToSubmit, &cur_stream] {
```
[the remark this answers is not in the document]
artem-shaporenko: Right, this will break latency

_studio/shared/src/mfx_mfe_adapter.cpp
```diff
@@ -329,28 +308,12 @@ mfxStatus MFEVAAPIEncoder::Submit(VAContextID context, vm_tick timeToWait, bool
 
     if(!framesToSubmit)
     {
-        vm_mutex_unlock(&m_mfe_guard);
         return MFX_ERR_UNDEFINED_BEHAVIOR;
     }
-    vm_tick start_tick = vm_time_get_tick();
-    vm_tick spent_ticks = 0;
-    timeout = timeout? timeout: cur_stream->timeout;
```
artem-shaporenko: Removed this one - broken functionality passing timeout from application

_studio/shared/src/mfx_mfe_adapter.cpp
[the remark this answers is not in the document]
artem-shaporenko: Right, we need to notify all threads about submission

_studio/shared/src/mfx_mfe_adapter.cpp
artem-shaporenko: or it was broken earlier..

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
akokoshn: for avoid conflict definition macros "max", "min" (in chrono and cm_mem_copy.h) cm_mem_copy.h should be defined after chrono.
cm_mem_copy.h:
#ifndef max
#define max (a, b) (((a) > (b)) ? (a) : (b))
#endif

_studio/shared/src/mfx_mfe_adapter.cpp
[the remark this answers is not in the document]
akokoshn: I just keep old implementation - early m_mfe_guard was unlocked and destroyed before clearing m_streams_pool/m_streamsMap.
@artem-shaporenko , what are you think about it, it's make sense to protect clearing operation?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
dvrogozh: Can you, please, prepare a change and drop self-define min/max in cm_mem_copy.h. Please, switch to std::min and std::max (use this notation when we call these primitives). This should be a pre-requisite commit. And this change moving headers around should be dropped.

_studio/shared/src/mfx_mfe_adapter.cpp
```diff
@@ -329,28 +305,12 @@ mfxStatus MFEVAAPIEncoder::Submit(VAContextID context, vm_tick timeToWait, bool
...
-        {
-            spent_ticks = (vm_time_get_tick() - start_tick);
-        }
-        else
-        {
-            vm_mutex_unlock(&m_mfe_guard);
-            return MFX_ERR_UNKNOWN;
-        }
-    }
+    m_mfe_wait.wait_for(guard, std::chrono::microseconds(timeToWait), [this, &framesToSubmit, &cur_stream] {
```
dvrogozh: Wait... In which values we have timeToWait? I thought it is in milliseconds. In this case std::chrono::milliseconds(timeToWait) should be used otherwise you will treat timeToWait as microseconds and eventually wait much shorter than you should. @artem-shaporenko : can you, please, comment?

_studio/shared/src/mfx_mfe_adapter.cpp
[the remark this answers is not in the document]
dmitry-gurulev: @artem-shaporenko, it's a public method, so - yes, it makes. But I'd rather make [Destroy] to be a private (protected) one.

_studio/shared/src/mfx_mfe_adapter.cpp
artem-shaporenko: if you look at initial interface there is a comment:
//time passed in vm_tick, so milliseconds to be multiplied by vm_frequency/1000

_studio/shared/include/cm_mem_copy.h
```diff
@@ -59,14 +59,6 @@ typedef mfxI32 cmStatus;
 #define CM_ALIGNED(PTR) (!((mfxU64(PTR))&0xf))
```
dvrogozh: "marcos" in commit message sound so... cool... :)

_studio/shared/src/mfx_mfe_adapter.cpp
dvrogozh: yep, I missed that. Thank you. @akokoshn : does that mean that validation we started is wrong for MFE?

akokoshn: @dvrogozh, to be honest I'm not sure, but  probably yes. Please correct me if I wrong, it's possible to rerun MFE cases, if validation already started. Any way I will update test_system.

mgonchar: @akokoshn why you captured framesToSubmit and cur_stream by reference? They are trivially copyable, so capturing by value fits better

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -286,31 +278,15 @@ mfxStatus mfxSchedulerCore::Synchronize(mfxTaskHandle handle, mfxU32 timeToWait)
...
-                return MFX_ERR_UNKNOWN;
-            }
-        }
+        std::unique_lock<std::mutex> guard(m_guard);
+
+        MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_PRIVATE, "Scheduler::Wait");
+        MFX_LTRACE_1(MFX_TRACE_LEVEL_SCHED, "^Depends^on", "%d", pTask->param.task.nParentId);
+        MFX_LTRACE_I(MFX_TRACE_LEVEL_SCHED, timeToWait);
+
+        pTask->done.wait_for(guard, std::chrono::milliseconds(timeToWait), [&pTask, &handle] {
```
dvrogozh: @mgonchar : what's your recommendation for this capture?

mgonchar: As I see mfxTaskHandle size is sizeof(size_t), so it definitely can be captured by value, as well as pTask

PR #720, opened 2018-10-11
[avce] Use the reordered PTS as DTS.

hongbolv:


lakulako:
Is incorrect for B-pyramid higher than minimal, if >1 backward references are used.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2465,6 +2468,9 @@ mfxStatus ImplementationAvc::AsyncRoutine(mfxBitstream * bs)
         DdiTaskIter task = FindFrameToStartEncode(m_video, m_lookaheadFinished.begin(), m_lookaheadFinished.end());
         assert(task != m_lookaheadFinished.end());
 
+        task->m_decodeTimeStamp = m_decodeTimeStamp.front();
```
stasefremov: Maybe I'm missing something (haven't been looking to this code for years). But with this implementation few first encoded frames will have DecodeTimeStamp = MFX_TIMESTAMP_UNKNOWN. 

Below is example for B-pyramid for 3 B-frames: I b B b P b B b P

Number of EncodeFrameAync call |  1 |   2 |   3 |   4 |   5 |   6 |   7 |   8 |   9
Input frame with PTS                      | I0 | b1 | B2 | b3 | P4 | b5 | B6 | b7 | P8
Encoded frame                               |     |      |       | I0 |  P4 | B2 | b1 | b3 | P8
DTS assigned to encoded frame    |      |     |        | U |   U |   0 |   1 |   2 |   3

It's my main concern. There are few other (mostly to look code nicer and avoid multiple implementations of same logic). But let's deal with this one first.

hongbolv: Yes, I set the few first encoded frames with unkown DTS. From frame's  PTS information, there is not a good way to get the correct DTS for the first few frames. And it is not necessary that all frames should have DTS. It is better to let few first frame's DTS as unknown DTS, so that decoder side can handle it itself. Do you have better idea here?

stasefremov: Actually I don't understand why we need this intentional delay filled with MFX_TIMESTAMP_UNKNOWN.  Why in the example above coded B2 should have DST = 0? E.g. in previous logic I0 would have DTS 0 and P4 would have DTS 1.

hongbolv: Your understanding is not the case. If I0 have DTS 0 and P4 have DTS 1, then b1 would have DTS 4, but b1 have PTS 1. It is not correct that b1's DTS is bigger than PTS. 
In fact, the intentional delay is to make every frame's DTS is smaller than PTS. In this B pyramid, we need delay two frames unit so that the b1 have DTS 1 and PTS 1.

stasefremov: I encoded few streams with different GOP structures, carefully wrote all numbers and see that you are right (there are no problems with intentional delay as I originally thought). E.g. below is table for 3B-frame B-pyramid:

xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  |xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx
------------------------------  | -- | -- | -- | -- | -- | -- | -- | -- | --
Number of EncodeFrameAync call  |  1 |  2 |  3 |  4 |  5 |  6 |  7 |  8 |  9
Input frame (with display order)| I0 | b1 | B2 | b3 | P4 | b5 | B6 | b7 | P8
PTS (linear)                    | 3K | 6K | 9K | 12K| 15K| 18K| 21K| 24K| 27K 
PTS (jump)                      | 3K | 6K | 90K| 93K| 96K| 99K|102K|105K|108K 
Encoded frame                   |    |    |    | I0 | P4 | B2 | b1 | b3 | P8
cpb_removal_delay               |    |    |    |  0 |  2 |  4 |  6 |  8 | 10 
dpb_output_delay                |    |    |    |  4 | 10 |  4 |  0 |  2 | 10 
dpb output time (HRD)               |    |    |    |  4 | 12 |  8 |  6 | 10 | 20 
DTS (linear) - current          |    |    |    | -3K|  0 | 3K | 6K | 9K | 12K
DTS (jump) - current            |    |    |    | -3K| 81K| 84K| 6K | 90K| 93K
DTS (linear) - Hongbo           |    |    |    | Unk| Unk| 3K | 6K | 9K | 12K
DTS (jump) - Hongbo             |    |    |    | Unk| Unk| 3K | 6K | 90K| 93K

It also shows that "DTS = reordered PTS" fixes "jittering" of DTS in case of PTS jumps.
So concern with algorithm itself is resolved. Will share more concerns about implementation and customer-side behavior in next comments.

_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
```diff
@@ -2044,6 +2048,7 @@ namespace MfxHwH264Encode
         std::list<DdiTask>  m_histRun;
         std::list<DdiTask>  m_histWait;
         std::list<DdiTask>  m_encoding;
+        std::list<mfxU64>   m_decodeTimeStamp;
```
stasefremov: Please rename (e.g. to m_decodeTimeStamps). Otherwise same name in DDITask class will cause problems during reading code/debugging.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -3253,6 +3259,32 @@ mfxStatus ImplementationAvc::QueryStatus(
     return MFX_ERR_NONE;
 }
 
+void ImplementationAvc::InitDTSList()
+{
+    mfxExtCodingOption2 * extOpt2 = GetExtBuffer(m_video);
+
+    // implemented the buffering through padding unknown DTS in the list head.
+    m_decodeTimeStamp.clear();
+    if (m_video.mfx.GopRefDist > 1)
```
stasefremov: I suggest to call function "MfxHwH264Encode::GetNumReorderFrames(..)" to calculate "padding" size (to not duplicate implementation of same logic in several places in the code).

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -1492,6 +1493,7 @@ mfxStatus ImplementationAvc::Reset(mfxVideoParam *par)
     }
     m_video = newPar;
 
+    InitDTSList();
```
stasefremov: Reset function can be called for on-the-fly parameters change (e.g. bitrate, framerate, number of slices, etc.). In this case internal encoder state isn't reset. I.e. counters are not dropped, all the calculations (GOP generation, HRD parameters generation, etc.) are smoothly continued for next frames after Reset. This relates to DTS either. Please move call of "InitDTSList()" to the code under "if (isIdrRequired || IsOn(extResetOpt->StartNewSequence))" - this code does deep encoder reset.

stasefremov:
And I'm still concerned with return of "MFX_TIMESTAMP_UNKNOWN" for first frames. It's change of "observable" behavior. There can be applications which treat "MFX_TIMESTAMP_UNKNOWN" for first frame as broken DTS reporting. If proper DTS reporting is mandatory for such app, backward compatibility will be hardly broken for such use case of MSDK.
I don't have good universal solution/algorithm which can fix the situation. But maybe it's good enough decision to keep legacy method(call CalcDTSFromPTS) in case when "m_decodeTimeStamp.front()" contains "MFX_TIMESTAMP_UNKNOWN".

hongbolv:
The "MFX_TIMESTAMP_UNKNOW" padding is only at the begin of sequence following the API calling (Init/Reset). The application will encounter this kind of "unkown dts" when meeting a new sequence or inserted IDR.  It should be not treated as broken DTS at all.
Keep the legacy method in case when "m_decodeTimeStamp.front()" contains "MFX_TIMESTAMP_UNKNOWN". It will bring the side effect for the case which application set some frames' PTS = MFX_TIMESTAMP_UNKOWN. 
So some DTS is from the new method, some DTS is from legacy method, it will cause new jitter.

Maybe we can just keep the legacy method for the first few frames as below:
if (m_encOrder < numberReorderFrames)
{
    DTS = CalcDTSFromPTS;
}
else
{
    DTS = reorderedPTS;
}

Is m_encOrder safe here?

lakulako:
We can improve a bit for the case when first PTS0 > 0. The dts buffer can be filled with {PTS0-N, ... PTS0-1} while resulting DTS is >=0.
In current implementation we can check dts buffer tail before push_back(PTSnext). If tail is TS_UNKNOWN, then it can be updated to PTSnext-1. And continue in reverse order while >=0.
It will work as well after reset.
Later (if not yet) input frames PTS0 can be adjusted if reorder is used.

hongbolv:
Improved a bit. Calculated the first NumReorderFrames DTS from the first time stamp. 
BTW, we don't make sure that DTS is >=0, because DTS can be wrapped back.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2465,6 +2468,19 @@ mfxStatus ImplementationAvc::AsyncRoutine(mfxBitstream * bs)
         DdiTaskIter task = FindFrameToStartEncode(m_video, m_lookaheadFinished.begin(), m_lookaheadFinished.end());
         assert(task != m_lookaheadFinished.end());
 
+        mfxU8 numReorderFrames = GetNumReorderFrames(m_video);
+        if (task->m_encOrder < numReorderFrames)
+        {
+            mfxU64 firstTimeStamp = m_decodeTimeStamps.front();
+            mfxU8 distToFirstTimeStamp = numReorderFrames-task->m_encOrder;
```
stasefremov: Finally, I'm OK with this implementation :) There are few cosmetic suggestions:
(1) It would be good to explain what happens here in comment.
(2) Also maybe it's good to add intermediate local variable like distInTickUnits = 2 * distToFirstTimeStamp to describe meaning of 2x multiplication.

fulinjie:
After applying this patch, some unexpected warnings occur when using CQP:

cmdline:
ffmpeg -v verbose -hwaccel qsv -c:v h264_qsv -i input.mp4 -c:v h264_qsv -flags qscale -global_quality 20 -y out_normal.mp4

`[mp4 @ 0x560d68b9f280] Non-monotonous DTS in output stream 0:0; previous: 172544                                                                                                                                                             , current: 172544; changing to 172545. This may result in incorrect timestamps i                                                                                                                                                             n the output file.
[mp4 @ 0x560d68b9f280] Invalid DTS: 175104 PTS: 174592 in output stream 0:0, rep                                                                                                                                                             lacing by guess
[mp4 @ 0x560d68b9f280] Non-monotonous DTS in output stream 0:0; previous: 174592                                                                                                                                                             , current: 174592; changing to 174593. This may result in incorrect timestamps i                                                                                                                                                             n the output file.
[mp4 @ 0x560d68b9f280] Invalid DTS: 177152 PTS: 176640 in output stream 0:0, rep                                                                                                                                                             lacing by guess
`

And version 2cc253409433 works well.

PR #721, opened 2018-10-11
Add support setting some cli options for I-,P- and B-frames separately

otsirule:
NumFramePartitions, ForceCtuSplit and FastIntraMode have a huge impact on quality/performance trade-off, so setting them for each frame type separately allow better experience in tuning HEVC FEI.


samples/sample_hevc_fei/include/sample_hevc_fei_defs.h
```diff
@@ -55,6 +55,20 @@ struct SourceFrameInfo
     }
 };
 
+struct PerFrameTypeCtrl
+{
+    mfxExtFeiHevcEncFrameCtrl   CtrlI;
+    mfxExtFeiHevcEncFrameCtrl   CtrlP; // also applicable for GPB frames
+    mfxExtFeiHevcEncFrameCtrl   CtrlB;
+
+    PerFrameTypeCtrl()
```
mgonchar: you can do it simpler:
```
PerFrameTypeCtrl()
: CtrlI()
, CtrlP()
, CtrlB()
{}
```

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -596,13 +650,26 @@ mfxStatus CheckOptions(const sInputParams& params, const msdk_char* appName)
         PrintHelp(appName, "Unsupported NumMvPredictorsL1 value (must be in range [0,4])");
         return MFX_ERR_UNSUPPORTED;
     }
-    if (!(params.encodeCtrl.NumFramePartitions == 1 ||
-          params.encodeCtrl.NumFramePartitions == 2 ||
-          params.encodeCtrl.NumFramePartitions == 4 ||
-          params.encodeCtrl.NumFramePartitions == 8 ||
-          params.encodeCtrl.NumFramePartitions == 16  ))
+    std::array<mfxU16, 4> nfp_values = {1, 2, 4, 8};
+    auto it = std::find_if(nfp_values.begin(), nfp_values.end(), [&params](const mfxU16 nfp_val)
```
mgonchar: this one is not find_if, but find, you don't need such complicated machinery to find value in array.
Anyway, see comment above

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -596,13 +650,26 @@ mfxStatus CheckOptions(const sInputParams& params, const msdk_char* appName)
         PrintHelp(appName, "Unsupported NumMvPredictorsL1 value (must be in range [0,4])");
         return MFX_ERR_UNSUPPORTED;
     }
-    if (!(params.encodeCtrl.NumFramePartitions == 1 ||
-          params.encodeCtrl.NumFramePartitions == 2 ||
-          params.encodeCtrl.NumFramePartitions == 4 ||
-          params.encodeCtrl.NumFramePartitions == 8 ||
-          params.encodeCtrl.NumFramePartitions == 16  ))
+    std::array<mfxU16, 4> nfp_values = {1, 2, 4, 8};
```
mgonchar: make it std::set and then use std::set::contains directly in each if

samples/sample_hevc_fei_abr/include/sample_hevc_fei_defs.h
```diff
@@ -54,6 +54,20 @@ struct SourceFrameInfo
     }
 };
 
+struct PerFrameTypeCtrl
+{
+    mfxExtFeiHevcEncFrameCtrl   CtrlI;
+    mfxExtFeiHevcEncFrameCtrl   CtrlP; // also applicable for GPB frames
+    mfxExtFeiHevcEncFrameCtrl   CtrlB;
+
+    PerFrameTypeCtrl()
```
mgonchar: same

samples/sample_hevc_fei_abr/src/cmd_processor.cpp
```diff
@@ -194,13 +200,26 @@ mfxStatus CheckOptions(const sInputParams & params)
             PrintHelp("Unsupported NumMvPredictorsL1 value (must be in range [0,4])");
             return MFX_ERR_UNSUPPORTED;
         }
-        if (!(params.encodeCtrl.NumFramePartitions == 1 ||
-                params.encodeCtrl.NumFramePartitions == 2 ||
-                params.encodeCtrl.NumFramePartitions == 4 ||
-                params.encodeCtrl.NumFramePartitions == 8 ||
-                params.encodeCtrl.NumFramePartitions == 16  ))
+        std::array<mfxU16, 4> nfp_values = {1, 2, 4, 8};
```
mgonchar: same

otsirule:
@mgonchar, thanks for your remarks. I've changed code according to them. By the way, std::set::contains is available since C++20 (https://en.cppreference.com/w/cpp/container/set/contains).

samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -596,13 +650,20 @@ mfxStatus CheckOptions(const sInputParams& params, const msdk_char* appName)
         PrintHelp(appName, "Unsupported NumMvPredictorsL1 value (must be in range [0,4])");
         return MFX_ERR_UNSUPPORTED;
     }
-    if (!(params.encodeCtrl.NumFramePartitions == 1 ||
-          params.encodeCtrl.NumFramePartitions == 2 ||
-          params.encodeCtrl.NumFramePartitions == 4 ||
-          params.encodeCtrl.NumFramePartitions == 8 ||
-          params.encodeCtrl.NumFramePartitions == 16  ))
+    std::list<mfxU16> nfp_values = {1, 2, 4, 8};
```
mgonchar: well, you actually don't need any stl container, just array would be fine. Since we don't have set::contains (which is sad), I believe best and simplest way would be:
```
mfxU16 nfp_values[] = {1,2,4,8}; // you could benefit from the fact that they are sorted
if (std::binary_search(std::begin(nfp_values), std::end(nfp_values), params.frameCtrl.CtrlI.NumFramePartitions) == std::end(nfp_values))
```

But I think now code is good enough to be merged

PR #725, opened 2018-10-11
 VP9d - use C++ primitives (memcpy => std::copy) 

tkuzne1x:


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -589,7 +589,7 @@ mfxStatus VideoDECODEVP9_HW::GetDecodeStat(mfxDecodeStat *pStat)
     m_stat.NumSkippedFrame = 0;
     m_stat.NumCachedFrame = 0;
 
-    memcpy_s(pStat, sizeof(m_stat), &m_stat, sizeof(m_stat));
+    std::copy(&m_stat, &m_stat + sizeof(m_stat), pStat);
```
mgonchar:
```
m_stat = *pStat;
```

PR #726, opened 2018-10-11
[core][HW] Add presence check for HW device selected by impl

vshampor:


_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#if defined(LINUX32) || defined(LINUX64)
+std::vector<PciDisplayDevice> GetDisplayPciDevices()
+{
+    std::vector<PciDisplayDevice> retval;
+    typedef int (*fsort)(const struct dirent**, const struct dirent**);
+    struct mfx_disp_adapters* adapters = NULL;
+    struct dirent** dir_entries = NULL;
+    mfxU32 entries_num = scandir(PCI_DIR.c_str(), &dir_entries, mfx_dir_filter, (fsort)alphasort);
+    for (mfxU32 i = 0; i < entries_num; ++i)
```
dvrogozh: add empty line, please, to separate variable declarations from the code.

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <dirent.h>
+#include <string>
+#include <list>
+#include <fstream>
+#include <iomanip>
+
+const std::string PCI_DIR = "/sys/bus/pci/devices";
```
dvrogozh: PCI_DIR is definitely Linux specifc, Class ID and Controller Class - not sure, but quite possible that on Windows there will be absolutely different method to get this info.

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,114 @@
...
+constexpr mfxU32 PCI_DISPLAY_CONTROLLER_CLASS = 0x03;
+constexpr mfxU32 CLASS_ID_OFFSET = 16;
+constexpr mfxU32 INTEL_VENDOR_ID = 0x8086;
+
+int mfx_dir_filter(const struct dirent* dir_ent)
```
dvrogozh: I doubt we will need this function for Windows. Can it be moved closer to Linux part, and optimally - made via lambda? or at least declared static?

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    if (!dir_ent) return 0;
+    if (!strcmp(dir_ent->d_name, ".")) return 0;
+    if (!strcmp(dir_ent->d_name, "..")) return 0;
+    return 1;
+}
+
+mfxU32 ReadFirstHexNumInFile(std::string file_name)
```
dvrogozh: Also would be used only on Linux... Either should be moved under define (and declared static) or done via lambda in a function where you really use it.

_studio/mfx_lib/shared/src/libmfxsw_query.cpp
```diff
@@ -44,6 +45,13 @@ mfxStatus MFXQueryIMPL(mfxSession session, mfxIMPL *impl)
     {
         currentImpl = (mfxIMPL) (MFX_IMPL_HARDWARE2 + (session->m_adapterNum - 1));
     }
+
+    if (currentImpl == MFX_IMPL_HARDWARE &&
+        !CheckForIntelHWDisplayDevicePresent(session->m_adapterNum, session->m_implInterface))
```
dvrogozh: This is also needed on MFXInit. And I suggest to cache gathered information to avoid repeated operations.

_studio/shared/src/libmfx_core_hw.cpp
vshampor: `dirent` has a C interface, therefore I don't think we can pass lambdas to it.

dmitryermilov:
This section is about Windows implementation.

_studio/shared/include/libmfx_core_hw.h
```diff
@@ -18,15 +18,54 @@
...
 
+#include "mfx_common.h"
 #include "umc_va_base.h"
+#include <memory>
+#include <mutex>
+#include <vector>
+
+struct PciDisplayDevice
+{
+    mfxU32 vendor_id;
```
mgonchar: looks like a good place for default initialization with invalid vendor & device id

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+#endif // #if defined(LINUX32) || defined(LINUX64)
+
+
+void HardwareInfoStore::LoadPciDisplayDevicesList()
+{
+
+#if defined(LINUX32) || defined(LINUX64)
+    typedef int (*fsort)(const struct dirent**, const struct dirent**);
+    struct mfx_disp_adapters* adapters = NULL;
```
mgonchar: nullptr

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+        class_id = ReadFirstHexNumInFile(full_pci_dev_path + "/class");
+
+        if (PCI_DISPLAY_CONTROLLER_CLASS == (class_id >> CLASS_ID_OFFSET))
+        {
+            vendor_id = ReadFirstHexNumInFile(full_pci_dev_path + "/vendor");
+            device_id = ReadFirstHexNumInFile(full_pci_dev_path + "/device");
+            if (vendor_id && device_id)
+            {
+                m_pci_disp_dev_list.emplace_back(PciDisplayDevice{vendor_id, device_id});
```
mgonchar: don't create temporary object, in that case you won't benefit from perfect forwarding. Pass arguments directly:
```
m_pci_disp_dev_list.emplace_back(vendor_id, device_id);
```

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
...
+        return false; // Only VAAPI is currently supported
+    }
+
+    const std::vector<PciDisplayDevice>& pci_disp_dev_list = HardwareInfoStore::GetInstance().GetPciDisplayDeviceList();
+
+    if (adapter_num >= pci_disp_dev_list.size())
+    {
+        return false;
+    }
+    PciDisplayDevice str;
```
mgonchar: unused

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+
+void HardwareInfoStore::LoadPciDisplayDevicesList()
+{
+
+#if defined(LINUX32) || defined(LINUX64)
+    typedef int (*fsort)(const struct dirent**, const struct dirent**);
+    struct mfx_disp_adapters* adapters = NULL;
+    struct dirent** dir_entries = NULL;
+    mfxU32 entries_num = scandir(PCI_DIR.c_str(), &dir_entries, mfx_dir_filter, (fsort)alphasort);
```
mgonchar: return value is signed, moreover values < 0 indicate errors

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
...
+const std::string PCI_DIR = "/sys/bus/pci/devices";
+int mfx_dir_filter(const struct dirent* dir_ent)
+{
+    if (!dir_ent) return 0;
+    if (!strcmp(dir_ent->d_name, ".")) return 0;
+    if (!strcmp(dir_ent->d_name, "..")) return 0;
+    return 1;
+}
+
+mfxU32 ReadFirstHexNumInFile(std::string file_name)
```
mgonchar: better to pass string by const ref here

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+constexpr mfxU32 PCI_DISPLAY_CONTROLLER_CLASS = 0x03;
+constexpr mfxU32 CLASS_ID_OFFSET = 16;
+constexpr mfxU32 INTEL_VENDOR_ID = 0x8086;
+
+std::unique_ptr<HardwareInfoStore> HardwareInfoStore::m_instance;
+std::mutex HardwareInfoStore::m_mtx;
+
+#if defined(LINUX32) || defined(LINUX64)
+const std::string PCI_DIR = "/sys/bus/pci/devices";
```
mgonchar: constexpr would be better

_studio/shared/src/libmfx_core_hw.cpp
```diff
@@ -0,0 +1,116 @@
+#include "libmfx_core_hw.h"
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <dirent.h>
+#include <iomanip>
```
mgonchar: do you use something from this header?

_studio/shared/include/libmfx_core_hw.h
```diff
@@ -18,15 +18,54 @@
...
+            if (m_pci_disp_dev_list.empty())
+            {
+                LoadPciDisplayDevicesList();
+            }
+        }
+        return m_pci_disp_dev_list;
+    };
+
+private:
+    HardwareInfoStore() = default;
```
mgonchar: I think you need to ban explicitly copy and move constructor

vshampor:
To be reopened via PR from fork

_studio/shared/include/libmfx_core_hw.h
```diff
@@ -18,15 +18,54 @@
...
+    mfxU32 vendor_id;
+    mfxU32 device_id;
+};
+
+class HardwareInfoStore
+{
+public:
+    static HardwareInfoStore& GetInstance()
+    {
+        std::lock_guard<std::mutex> lock(m_mtx);
```
dmitry-gurulev: @vshampor 
I'd rather use
> static m_instance

in [GetInstance] method scope directly, instead of having it at class scope. No static instance of mutex is required here, initialization of block-scope variables with static storage duration is thread safe since C++11 (see details at [6.7 Declaration statement](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf), c.4).
Profit:
1. U don't need this mutex to be static =>
2. U don't need [libmfx_core_hw.cpp :: std::mutex HardwareInfoStore::m_mtx] definition =>
3. U don't mix mutex usage from static/non-static methods.

PR #724, opened 2018-10-11
Replaced memcpy with std::copy / operator = in AVC components

mgonchar:
Replaced memcpy with std::copy / operator =

PR #731, opened 2018-10-12
Process h264d - use C++ primitives: memcpy => std::copy

ilapichev:
Issue: MDP-50956

### DIFF
--- a/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
@@ -708,9 +708,9 @@ mfxStatus VideoDECODEH264::GetVideoParam(mfxVideoParam *par)
             return MFX_ERR_NOT_ENOUGH_BUFFER;
         }
 
-        memcpy_s(mvcSeqDesc->View, mvcSeqDesc->NumView * sizeof(mfxMVCViewDependency), mvcSeqDescInternal->View, mvcSeqDescInternal->NumView * sizeof(mfxMVCViewDependency));
-        memcpy_s(mvcSeqDesc->ViewId, mvcSeqDesc->NumViewId * sizeof(mfxU16), mvcSeqDescInternal->ViewId, mvcSeqDescInternal->NumViewId * sizeof(mfxU16));
-        memcpy_s(mvcSeqDesc->OP, mvcSeqDesc->NumOP * sizeof(mfxMVCOperationPoint), mvcSeqDescInternal->OP, mvcSeqDescInternal->NumOP * sizeof(mfxMVCOperationPoint));
+        std::copy(mvcSeqDesc->View, mvcSeqDesc->View + mvcSeqDesc->NumView * sizeof(mfxMVCViewDependency), mvcSeqDescInternal->View);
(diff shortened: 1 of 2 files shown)


ilapichev:

Original:
virtual void* GetDataPointer(void)   { return m_pDataPointer; }
Now:
virtual uint8_t* GetDataPointer(void)   { return m_pDataPointer; }

mgonchar:
It is already done in https://github.com/Intel-Media-SDK/MediaSDK/pull/724/

_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
```diff
@@ -1640,7 +1640,7 @@ mfxStatus VideoDECODEH264::GetUserData(mfxU8 *ud, mfxU32 *sz, mfxU64 *ts)
     if (!m_isInit)
         return MFX_ERR_NOT_INITIALIZED;
 
-    MFX_CHECK_NULL_PTR3(ud, sz, ts);
+    //MFX_CHECK_NULL_PTR3(ud, sz, ts);
```
mgonchar: why you decided to remove check?

_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
```diff
@@ -708,9 +708,9 @@ mfxStatus VideoDECODEH264::GetVideoParam(mfxVideoParam *par)
             return MFX_ERR_NOT_ENOUGH_BUFFER;
         }
 
-        memcpy_s(mvcSeqDesc->View, mvcSeqDesc->NumView * sizeof(mfxMVCViewDependency), mvcSeqDescInternal->View, mvcSeqDescInternal->NumView * sizeof(mfxMVCViewDependency));
-        memcpy_s(mvcSeqDesc->ViewId, mvcSeqDesc->NumViewId * sizeof(mfxU16), mvcSeqDescInternal->ViewId, mvcSeqDescInternal->NumViewId * sizeof(mfxU16));
-        memcpy_s(mvcSeqDesc->OP, mvcSeqDesc->NumOP * sizeof(mfxMVCOperationPoint), mvcSeqDescInternal->OP, mvcSeqDescInternal->NumOP * sizeof(mfxMVCOperationPoint));
+        std::copy(mvcSeqDesc->View, mvcSeqDesc->View + mvcSeqDesc->NumView * sizeof(mfxMVCViewDependency), mvcSeqDescInternal->View);
```
mgonchar: std::copy has different parameter order than memcpy:

```
errno_t memcpy_s(  
   void *dest, // <- dest
   size_t destSize,  
   const void *src, // <- src
   size_t count   
);
```
But std::copy has other
```
template< class InputIt, class OutputIt >
OutputIt copy( InputIt first, InputIt last, <- src
OutputIt d_first ); // <- dest
```

_studio/shared/umc/core/umc/include/umc_media_data.h
```diff
@@ -73,7 +73,7 @@ class MediaData
 
     // Get an address of the beginning of valid data.
     // This pointer could not be equal to the beginning of the buffer.
-    virtual void* GetDataPointer(void)   { return m_pDataPointer; }
+    virtual uint8_t* GetDataPointer(void)   { return m_pDataPointer; }
```
mgonchar: You could workaround it with cast in place

ilapichev:
already done

PR #732, opened 2018-10-12

ilapichev:
Issue: MDP-50957


ilapichev:

Original:
virtual void* GetDataPointer(void) { return m_pDataPointer; }
Now:
virtual uint8_t* GetDataPointer(void) { return m_pDataPointer; }

PR #733, opened 2018-10-12

ilapichev:
Issue: MDP-50957

### DIFF
--- a/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
@@ -534,8 +534,8 @@ mfxStatus VideoDECODEH265::GetVideoParam(mfxVideoParam *par)
         spsPps->SPSBufSize = spsPpsInternal->SPSBufSize;
         spsPps->PPSBufSize = spsPpsInternal->PPSBufSize;
 
-        memcpy_s(spsPps->SPSBuffer, spsPps->SPSBufSize, spsPpsInternal->SPSBuffer, spsPps->SPSBufSize);
-        memcpy_s(spsPps->PPSBuffer, spsPps->PPSBufSize, spsPpsInternal->PPSBuffer, spsPps->PPSBufSize);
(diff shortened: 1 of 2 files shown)


ilapichev:

Original:
virtual void* GetDataPointer(void) { return m_pDataPointer; }
Now:
virtual uint8_t* GetDataPointer(void) { return m_pDataPointer; }

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -534,8 +534,8 @@ mfxStatus VideoDECODEH265::GetVideoParam(mfxVideoParam *par)
         spsPps->SPSBufSize = spsPpsInternal->SPSBufSize;
         spsPps->PPSBufSize = spsPpsInternal->PPSBufSize;
 
-        memcpy_s(spsPps->SPSBuffer, spsPps->SPSBufSize, spsPpsInternal->SPSBuffer, spsPps->SPSBufSize);
-        memcpy_s(spsPps->PPSBuffer, spsPps->PPSBufSize, spsPpsInternal->PPSBuffer, spsPps->PPSBufSize);
+        std::copy(spsPps->SPSBuffer, spsPps->SPSBuffer + spsPps->SPSBufSize, spsPpsInternal->SPSBuffer);
```
mgonchar: std::copy has different parameter order than memcpy:
```
errno_t memcpy_s(  
   void *dest, // <- dest
   size_t destSize,  
   const void *src, // <- src
   size_t count   
);
```
But std::copy has other
```
template< class InputIt, class OutputIt >
OutputIt copy( InputIt first, InputIt last, <- src
OutputIt d_first ); // <- dest
```

_studio/shared/umc/core/umc/include/umc_media_data.h
```diff
@@ -73,7 +73,7 @@ class MediaData
 
     // Get an address of the beginning of valid data.
     // This pointer could not be equal to the beginning of the buffer.
-    virtual void* GetDataPointer(void)   { return m_pDataPointer; }
+    virtual uint8_t* GetDataPointer(void)   { return m_pDataPointer; }
```
mgonchar: I would recommend to use cast in place. Such change is harmful, in theory it could silently break inheritance (maybe already have done it)

PR #734, opened 2018-10-12
[VPP] Fix of background color normalization

ndomnina:
The make_back_color_yuv function takes 8-bit background color values as input
and checks them for color ranges of 10-bit/16-bit formats, which is incorrect,
because these values will later be converted to the required format and may be
out of range.

### DIFF
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -1352,7 +1352,7 @@ mfxStatus TaskManager::AssignTask(
         // m_MCTFSurfacesInQueue means number of frames including the current;
         // in case of 2 refs or 4 refs, its enough to have 2 or 3 to output.
         // Effectively, it is used to count number of forward refs + current;
-        // Semantic can be changed to track how many frames are in queue (it 
+        // Semantic can be changed to track how many frames are in queue (it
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5081,10 +5081,10 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
 
     return
-        ((mfxU64)                                 (max_val - 1) << 48) |
-        ((mfxU64)VPP_RANGE_CLIP(Y, (16 << shift), (235 << shift))     << 32) |
-        ((mfxU64)VPP_RANGE_CLIP(U, (16 << shift), (240 << shift))     << 16) |
-        ((mfxU64)VPP_RANGE_CLIP(V, (16 << shift), (240 << shift))     <<  0)
+        ((mfxU64)             (max_val - 1)  << 48)  |
+        ((mfxU64)VPP_RANGE_CLIP(Y, 16, 235)  << 32)  |
+        ((mfxU64)VPP_RANGE_CLIP(U, 16, 240)  << 16)  |
+        ((mfxU64)VPP_RANGE_CLIP(V, 16, 240)  <<  0)
```
lakulako: Here, for example 10 bit value is taken and clipped as it were 8 bit. It will be dark-green world!

PR #736, opened 2018-10-12
[BUILD] makes build possible without MFX_HOME env variable

IlyaSergeyev:
In some environments we can't set environment variable properly and build fails.

Environment variable MFX-HOME replaced with cmake one in FindMFX.cmake
NO_CMAKE_FIND_ROOT_PATH parameter added to cmake_find (file search fails without this parameter in some environments like buildroot).


CMakeLists.txt
```diff
@@ -37,8 +37,11 @@ if( NOT DEFINED CMAKE_MFX_HOME )
   set( CMAKE_MFX_HOME "${CMAKE_HOME_DIRECTORY}/api" )
 endif()
 
-if( "$ENV{MFX_HOME}" STREQUAL "" )
+if(NOT DEFINED $ENV{MFX_HOME} )
   set( ENV{MFX_HOME} ${CMAKE_HOME_DIRECTORY} )
```
dvrogozh: Do we really need to set `ENV{MFX_HOME}` ourselves? I suggest to drop it. What we should care about is just ${MFX_HOME}.

builder/FindMFX.cmake
```diff
@@ -35,10 +35,10 @@ endif()
 if( CMAKE_MFX_HOME )
   set( MFX_API_HOME ${CMAKE_MFX_HOME} )
 else()
-  set( MFX_API_HOME $ENV{MFX_HOME} )
+  set( MFX_API_HOME ${MFX_HOME} )
 endif()
 
-find_path( MFX_INCLUDE mfxdefs.h PATHS ${MFX_API_HOME}/include )
+find_path( MFX_INCLUDE mfxdefs.h PATHS ${MFX_API_HOME}/include NO_CMAKE_FIND_ROOT_PATH )
```
dvrogozh: Explain, please, why we need `NO_CMAKE_FIND_ROOT_PATH` here?

IlyaSergeyev: In certain cross-compiling environments such as buildroot find_path not works properly. Correct behaivour can be  also reached by passing CMAKE_FIND_ROOT_PATH_MODE_INCLUDE parameter to cmake, but parameter affect all cases of find_file, it is not needed here.

dvrogozh: ok.

CMakeLists.txt
dvrogozh: I still await for the response/fix.

dvrogozh:
@dmitryermilov : why this PR was closed? Is that a mistake? I don't see any clarifying message from you.

PR #738, opened 2018-10-12

ilapichev:
Issue: MDP-50957


_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -1366,7 +1366,8 @@ mfxStatus VideoDECODEH265::GetUserData(mfxU8 *ud, mfxU32 *sz, mfxU64 *ts)
 
     *sz = (mfxU32)data.GetDataSize();
     *ts = GetMfxTimeStamp(data.GetTime());
-    memcpy_s(ud, *sz, data.GetDataPointer(), data.GetDataSize());
+    uint8_t* dataPointer = reinterpret_cast <uint8_t*> (data.GetDataPointer());
```
lakulako: const

PR #739, opened 2018-10-12
VP9d - use C++ primitives (memcpy => std::copy)

tkuzne1x:


dvrogozh:
@dmitryermilov : this change is entirely with operator=, we are not blocked to merge this one.

PR #743, opened 2018-10-12
Dispatcher tests

vshampor:
Test case set to be expanded.


CMakeLists.txt
```diff
@@ -163,5 +168,6 @@ message("Build:")
 message("  BUILD_RUNTIME                           : ${BUILD_RUNTIME}")
 message("  BUILD_DISPATCHER                        : ${BUILD_DISPATCHER}")
 message("  BUILD_SAMPLES                           : ${BUILD_SAMPLES}")
+message("  BUILD_TESTS                             : ${BUILD_TESTS}")
```
dvrogozh: Change README.md, please, and add BUILD_TESTS in there.

tests/mfx_dispatch/linux/CMakeLists.txt
```diff
@@ -22,7 +22,12 @@
 # library implementation inside. A key thing for the trick to work is
 # avoiding non-default symbol versioning, i.e. -Wl,--default-symver which
 # is propagated by mediasdk test system should be avoided. 
-add_executable(mfx_dispatch_tests mfx_dispatch_tests.cpp)
+include_directories(include)
```
dvrogozh: maybe we will keep all files in a single folder? why we need 'include'?

tests/mfx_dispatch/linux/CMakeLists.txt
```diff
@@ -22,7 +22,12 @@
...
+add_executable(mfx_dispatch_tests 
+		mfx_dispatch_test_main.cpp
```
dvrogozh: Do you have tabs here? Keep just 2 spaces in the beginning of the line, please.

tests/mfx_dispatch/linux/mfx_dispatch_test_main.cpp
```diff
@@ -19,33 +19,36 @@
...
 
 #include "gtest/gtest.h"
-#include <mfxvideo.h>
+#include "mfx_dispatch_tests_main.h"
+
+DlopenHookType g_e_dlopen_hook_state = DlopenHookType::ALWAYS_NULL;
+DlsymHookType  g_e_dlsym_hook_state  = DlsymHookType::ALWAYS_NULL;
+
+std::map<DlopenHookType, DlopenHook> g_dlopen_hook_table {
+            {DlopenHookType::ALWAYS_NULL, TEST_DLOPEN_HOOKS::AlwaysNull},
```
dvrogozh: Remove tabs, use 2 spaces, please.

tests/mfx_dispatch/linux/mfx_dispatch_test_main.cpp
```diff
@@ -19,33 +19,36 @@
...
+    };
+
 
 extern "C"
 {
   void *dlopen(const char *filename, int flag)
   {
     printf("dlopen: filename=%s, flag=%d\n", filename, flag);
-    return NULL;
+    return g_dlopen_hook_table[g_e_dlopen_hook_state](filename, flag);
```
dvrogozh: I think you need to implement a range of tests to cover minimal variety of possible tests. This may cause dramatic change to your current idea of 2 maps for dlopen and dlsym. For example, one of the dispatcher tasks is to load the library of the lower version even if dispatcher was built with the higher version. Thus, you will have situation when you will need to return valid pointers in dlsym for 0..M uAPI functions and return NULL for M+1..N. Think how you will handle this.

tests/mfx_dispatch/linux/CMakeLists.txt
vshampor: Usually source and include files directories are kept separte.

dvrogozh: Include are sometimes kept separate if that's externally exposed uAPI. In other cases, for internal things they are usually kept together:
- https://github.com/FFmpeg/FFmpeg/tree/master/libavcodec
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/gpu/drm/i915

In this case, for the test, I think better to keep everything together. This will be simpler.

tests/mfx_dispatch/linux/include/mfx_dispatch_test_hooks.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <mfxvideo.h>
+#include <functional>
+
+struct HookInternalParams
+{
+    mfxVersion emulated_api_version = {{0, 0}};
+};
+
+typedef void* HookHandle;
```
dvrogozh: Have you thought about returning hook handle or some other pointer/structure which you have created inside the test as mfxSession from MFXInit(Ex)? The idea of the dispatcher is that it should call real API function and pass this handle intact (though it can wrap it in something else inside). I.e. that's something you can check:
MFXInitEx -> you catch dlopen, you catch dlsym and if test really makes init, you catch:
- MFXInitEx and return something in mfxSession
- MFXQueryVersion and can check/work what you have passed as mfxSession
- similar for MFxQueryIMPL

Actually returning different results/statuses from all 3 functions I mentioned above makes a test for MFXInit(Ex).

tests/mfx_dispatch/linux/mfx_dispatch_test_main.cpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return g_dlsym_hook(handle, symbol);
+    }
+
+    FILE * fopen(const char *filename, const char *opentype)
+    {
+        return g_fopen_hook(filename, opentype);
+    }
+
+    mfxStatus MFXInitEx(mfxInitParam par, mfxSession *session)
```
dvrogozh: wait... But with this you will catch MFXInitEx which application makes to the dispatcher... right? Why do you need this? I thought the idea is to return pointer to mock implementation to the dispatcher. In this way you don't need this mock implementation to be name MFXInitEx, you just need to return correct function pointer with MFXInitEx prototype when dispatcher calls dlsym with "MFXInitEx".

dmitry-gurulev:
@vshampor, why don't u use [gmock] instead of this hand-crafted  hooks?

dmitry-gurulev:
@vshampor, it might be reasonable to elaborate sources tree - make a common folder like [units] where tests for certain component/subsystem have to be placed. [MediaSDK/tests] is too common, putting all into this folder makes a mess. I suggest the following:
-MediaSDK/tests
--units
---hooks/mocks/whatever (some common/utils/support code)
---suites
----dispatcher
-----Linux
----allocators/session/core/whatever

tests/mfx_dispatch/linux/mfx_dispatch_test_hooks.cpp
```diff
@@ -19,14 +19,59 @@
...
+#include <algorithm>
+
 // dlopen hooks implementation:
 
 void* TEST_DLOPEN_HOOKS::AlwaysNull(const char *filename, int flag)
 {
     return nullptr;
 }
 
+void* TEST_DLOPEN_HOOKS::AlwaysNullNameCheck(const char *filename, int flag, HookInternalParams par)
```
dmitry-gurulev: This looks like different expectations mixed together for unknown reasons. IMO, it should be different cases, like
`CheckImplAuto`, `CheckImplHW`, `CheckImplSW`

tests/mfx_dispatch/linux/mfx_dispatch_test_main.cpp
vshampor: Good point.

vshampor:
@dmitry-gurulev 
>why don't u use [gmock] instead of this hand-crafted hooks?

In what manner specifically do you suggest to use gmock to test the dispatcher functionality?

> it might be reasonable to elaborate sources tree 

My changes are local to the tests/mfx_dispatch/linux/ folder. Inside it, there is no reason to impose elaborate source tree structures. I will not add any changes in the structure of source tree outside the tests/mfx_dispatch/linux/ as part of this PR.

> CheckImplAuto, CheckImplHW, CheckImplSW

What exactly should each case check for? We do not check implementations, we check that the correct library names are searched for each implementation.

> [tests/mfx_dispatch/linux/mfxvideo_functions.h] - why do we need this copy-paste?

For the same reason we need it in the mfx_dispatch code proper:
https://github.com/Intel-Media-SDK/MediaSDK/blob/29d17bd53d5b04a0ba29b2bfe09c48e3faec9599/api/mfx_dispatch/linux/mfxloader.cpp#L60

I require the same pattern here to emulate API versions: 
https://github.com/Intel-Media-SDK/MediaSDK/blob/624a48fab8b669238f8e1f33ecce61bf0363016a/tests/mfx_dispatch/linux/mfx_dispatch_test_hooks.cpp#L167-L169

dmitry-gurulev:
@vshampor,

>in what manner specifically do you suggest to use gmock to test the dispatcher functionality?

With using of gmock u may write
```
//dlobj - mocking obj. for dlopen
//we expect here that dlopen will be called w/ HW lib. name
EXPECT_CALL(dlobj, open(testing::StrEq("libmfxhw64.so"), _))
            .Times(1)
            .WillOnce(testing::Return(nullptr))
            ;

//...

auto sts =  MFXInit(MFX_IMPL_HARDWARE, &ver, &session);
```

vshampor:
@dmitry-gurulev 
Ok, but the current mfx_dispatch implementation calls dlopen/dlsym/fopen directly, not through an internal MediaSDK-defined interface. Had it been the other way around, there would have been no need to use hooks at all.

vshampor:
Although, I suppose, the gmock framework may be used here, providing a benefit even.

dmitry-gurulev:
@vshampor 

>current mfx_dispatch implementation calls dlopen/dlsym/fopen directly, not through an internal MediaSDK-defined interface.

That's not a problem, consider this approach -
```
//dynamic loading interface
struct dl
{
    virtual void* open(const char* filename, int flag) = 0;
    //close etc.
};

//Implementation - mock object
struct dlmock
{
    MOCK_METHOD2(open, void*(const char*, int));
} dlobj;

//mock external free functions
extern "C" 
{
    void *dlopen(const char *filename, int flag) 
    { return dlobj.open(filename, flag); }
}

//...
//here u may use EXPECT_CALL(dlobj, ....
```

dmitry-gurulev:
@vshampor 

>>[tests/mfx_dispatch/linux/mfxvideo_functions.h] - why do we need this copy-paste?

>For the same reason we need it in the mfx_dispatch code proper:

Why can't we use original [MediaSDK/api/mfx_dispatch/linux/mfxvideo_functions.h]?

dvrogozh:
@dmitry-gurulev  : please, avoid accidental closing of the pull request(s). It will be alive till we make it a good one.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_libs.cpp
```diff
@@ -19,32 +19,42 @@
...
 
 TEST_F(DispatcherTest, ShouldFailIfNoLibraryIsFound)
 {
+    MockCallObj& mock = *g_call_obj_ptr;
+    EXPECT_CALL(mock, dlopen).Times(AtLeast(1));
     mfxStatus sts = MFXInit(impl, &ver, &session);
     ASSERT_EQ(sts, MFX_ERR_UNSUPPORTED);
 }
 
 TEST_F(DispatcherTest, ShouldEnumerateCorrectLibNames)
```
dmitry-gurulev: It would be better to have separate test cases instead of using 'for' loop, otherwise u'll get too generic test failure message like '[ FAILED ] DispatcherTest.ShouldEnumerateCorrectLibNames' and u need to parse log to understand which 'impl_case' really failed MFX_IMPL_AUTO, MFX_IMPL_SOFTWARE/whatever. If u separate this big test into several ones - u'll get clear message like '[ FAILED ] DispatcherTest.CheckCorrectLibNamesForAuto'.

To avoid code duplication, [gtest] 'Typed Test' approach might be used. Something like that:
```
template <int T>
using impl_type = std::integral_constant<int, T>;

using impl_case_list = testing::Types<
    impl_type<MFX_IMPL_SOFTWARE>,
    impl_type<MFX_IMPL_HARDWARE>
    //... other impl
>;

//DispatcherTest has to be template
TYPED_TEST_CASE(DispatcherTest, impl_case_list);
TYPED_TEST(DispatcherTest, CheckCorrectLibNames)
{
    int constexpr impl = TypeParam::value;
    //... here the same code as in 'for (auto impl_case : impl_cases_list) ' loop
```

If some certain case failed (let it be MFX_IMPL_HARDWARE == 2), u'd get the following message:

> [  FAILED  ] DispatcherTest/1.CheckCorrectLibNames, where TypeParam = struct std::integral_constant<int,2>

That's also is not a perfect message but u get a hint what happened which allows u to avoid gazing into the whole log.
As an additional benefit, this allows u to have finer grained test filters -
`--gtest_filter=DispatcherTest/<#case>.CheckCorrectLibNames`

dmitry-gurulev:
@vshampor,

>Using the same MediaSDK/api/mfx_dispatch/linux/mfxvideo_functions.h is problematic because then we have to conserve the relative directory location between mfx_dispatch and mfx_dispatch test.

Why can't u add this directory location to your [mfx_dispatch_tests] target via 'target_include_directories' and then use this [mfxvideo_functions.h] header in the same way as any other API header?

dvrogozh:
> @vshampor,
> 
> > Using the same MediaSDK/api/mfx_dispatch/linux/mfxvideo_functions.h is problematic because then we have to conserve the relative directory location between mfx_dispatch and mfx_dispatch test.
> 
> Why can't u add this directory location to your [mfx_dispatch_tests] target via 'target_include_directories' and then use this [mfxvideo_functions.h] header in the same way as any other API header?

I would prefer to decouple component and test code. If component will need to be updated we will avoid rewriting of the test. So, I am strongly against to use internal component staff for internal part of the test.

README.md
```diff
@@ -94,6 +94,7 @@ Media SDK depends on a number of packages which are identified and checked for t
 | BUILD_ALL | ON\|OFF | Build all the BUILD_* targets below (default: OFF) |
 | BUILD_RUNTIME | ON\|OFF | Build mediasdk runtime, library and plugins (default: ON) |
 | BUILD_SAMPLES | ON\|OFF | Build samples (default: ON) |
+| BUILD_TESTS | ON\|OFF | Build unit tests (default: OFF) |
```
dvrogozh: This belongs to prev. commit.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_libs.cpp
```diff
@@ -0,0 +1,282 @@
...
+    EXPECT_CALL(mock, MFXInitEx).Times(AtLeast(1)).WillRepeatedly(DoAll(SetArgPointee<1>(MOCK_SESSION_HANDLE), Return(MFX_ERR_NONE)));
+    EXPECT_CALL(mock, MFXQueryIMPL).Times(AtLeast(1)).WillRepeatedly(Return(MFX_ERR_NONE));
+    EXPECT_CALL(mock, MFXQueryVersion).Times(AtLeast(1)).WillRepeatedly(Invoke(&mock, &MockCallObj::ReturnEmulatedVersion));
+    EXPECT_CALL(mock, dlclose).Times(AtLeast(1));
+
+    sts = MFXInit(impl, &ver, &session);
+    EXPECT_EQ(sts, MFX_ERR_NONE);
+}
+
+
```
dvrogozh: Other tests:
- Does MFXInitEx implemented in dispatcher pass mfxInitParam* untouched?
- Does dispatcher pass mfxSession* allocated inside the library to other library calls untouched?

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
```diff
@@ -0,0 +1,79 @@
...
+    EXPECT_CALL(mock, CreatePlugin(plugin.PluginUID,_)).WillRepeatedly(Invoke(&mock, &MockCallObj::ReturnMockPlugin));
+    EXPECT_CALL(mock, GetPluginParam(&mock.m_mock_plugin,_)).WillRepeatedly(Return(MFX_ERR_NONE));
+    EXPECT_CALL(mock, MFXVideoUSER_Register(session,plugin.Type, &mock.m_mock_plugin)).WillRepeatedly(Return(MFX_ERR_NONE));
+
+    mfxStatus sts = MFXVideoUSER_Load(session, &plugin.PluginUID, 0);
+}
+
+INSTANTIATE_TEST_CASE_P(EnumeratingPlugins, DispatcherPluginsTestParametrized, ::testing::Range((size_t) 0, TEST_PLUGIN_COUNT));
+
+
```
dvrogozh: Other tests:
- Plugins work in dispatcher is protected by the mutex. Do we step into problems here?

dvrogozh:
In general I think we need to stop for a moment and do the following:
- Avoid adding new tests
- Clean up the current code, address review comments
- I suggest to keep 1st commit separate and squash 2+ into one
- Merge the unittest
- Address any failing test cases
- Integrate the test into CI

Once that will be done let's add new tests.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_libs.cpp
```diff
@@ -0,0 +1,282 @@
...
+    EXPECT_CALL(mock, MFXQueryVersion).Times(AtLeast(1)).WillRepeatedly(Invoke(&mock, &MockCallObj::ReturnEmulatedVersion));
+
+    mfxStatus sts = MFXInit(impl, &ver, &session);
+    ASSERT_EQ(sts, MFX_ERR_NONE);
+}
+
+TEST_F(DispatcherLibsTest, ShouldFailIfNoLibraryIsFound)
+{
+    MockCallObj& mock = *g_call_obj_ptr;
+    EXPECT_CALL(mock, dlopen).Times(AtLeast(1));
```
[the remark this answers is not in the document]
vshampor: The proper syntax is Times(0)

README.md
vshampor: Done

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_libs.cpp
vshampor: Should create an issue about this once the current PR is merged.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
vshampor: Same as above.

CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ cmake_dependent_option(BUILD_SAMPLES "Build samples?" ON "BUILD_DISPATCHER" OFF
 # Tools depend on samples (sample_common) and can't be built without it. The
 # following BUILD_TOOLS option declaration assures that.
 cmake_dependent_option(BUILD_TOOLS "Build tools?" "${BUILD_ALL}" "BUILD_SAMPLES" OFF)
+option(BUILD_TESTS "Build tests?" "${BUILD_ALL}")
```
azhogov: "Build unit tests?"

dvrogozh: this can be not only unit tests going forward. I think generic "Build tests?" is enough for now.

azhogov: ok, but now it will mean build and run unit tests. Probably "build tests" should just build all tests.

dvrogozh: Building and running of the tests is decoupled from each other. We don't have any other as of now.

PR #745, opened 2018-10-15
Mpeg2e - use C++ primitives (memcpy => std::copy)

ilapichev:
Issue: MDP-50975


dvrogozh:
@ilapichev: you are based on outdated forked copy, that's why tests are failing. Please, rebase your patch!

PR #750, opened 2018-10-15
[AVCe] Fix conversion of the slice structure from VAAPI to MFX

romka0017:
Current commit includes removing of the
VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE from slice structure
reported by driver. It should be done for correct work of the MSDK
on Linux platform.

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -1439,8 +1439,12 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     if (attrs[idx_map[VAConfigAttribEncSliceStructure]].value != VA_ATTRIB_NOT_SUPPORTED)
     {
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -87,7 +87,8 @@ VAProfile ConvertProfileTypeMFX2VAAPI(mfxU32 type)
 
 mfxU8 ConvertSliceStructureVAAPIToMFX(mfxU8 structure)
 {
-    switch (structure)
+    // Current implementation of Linux doesn't support SliceLevelRateCtrl
```
stasefremov: The comment looks ambiguous. This function maps "VAConfigAttribEncSliceStructure" to "m_caps.SliceStructure".  It must always discard bit "VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE". Doesn't matter if max slice size (aka SliceLevelRateCtrl) supported by driver or not.

PR #753, opened 2018-10-15
[mfx_trace] Refactored file parsing, removed dependency from SafeStrings lib

mgonchar:
Used c++11 approach to parse file, get rid of dependency from SafeStrings lib

### DIFF
--- a/_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
+++ b/_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
@@ -21,26 +21,58 @@
 
 #include "mfx_trace_utils.h"
 #include <sstream>
-#include <iomanip>
+#include <cstring>
 
 #ifdef MFX_TRACE_ENABLE
 #include "vm_sys_info.h"
 
-#if defined(ANDROID)
-#include "snprintf_s.h"
-#endif
+inline std::string& trim_string(std::string&& s)
+{
+    // Remove leading spaces, tabulation, = character
+    size_t pos = s.find_first_not_of(" =\t");
+    if (pos != std::string::npos)
+        s.erase(0, pos);
+
+    // Remove trailing spaces, tabulation, newline symbol, spaces
+    pos = s.find_last_not_of(" \t\n\r\f\v");
+    if (pos != std::string::npos)
+        s.erase(pos + 1);
+
+    return s;
+}
+
+std::string mfx_trace_get_value(FILE* file, const char* pName)
+{
+    if (!file || !pName)
+        return std::string("");
+
+    char line[MAX_PATH] = {0}, *s = nullptr;
+    std::string str, str_name(pName);
+
+    while ((s = fgets(line, MAX_PATH,  file)))
+    {
+        str = s;
+
+        size_t pos = str.find(str_name);
+        if (pos != std::string::npos)
+        {
+            pos += str_name.size();
+            return trim_string(str.substr(pos));
+        }
+    }
+
+    return std::string("");
+}
 
 extern "C"
 {
 
 #include <stdlib.h>
 
-/*------------------------------------------------------------------------------*/
-
 FILE* mfx_trace_open_conf_file(const char* name)
 {
-    FILE* file = NULL;
-    char file_name[MAX_PATH] = {0};
+    FILE* file = nullptr;
+    std::stringstream ss;
 
 #if defined(ANDROID)
     const char* home = "/data/data/com.intel.vtune/mediasdk";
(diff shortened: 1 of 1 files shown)


_studio/shared/mfx_trace/CMakeLists.txt
```diff
@@ -28,5 +28,5 @@ file( GLOB_RECURSE srcs "${MSDK_STUDIO_ROOT}/shared/mfx_trace/src/*.cpp" )
 list( APPEND sources ${srcs})
 
 set( defs "" )
-make_library( mfx_trace none static )
+make_library( mfx_trace none static "nosafestring")
```
dvrogozh: Don't do this, please. I guess this don't affect your change. And we in any case still have a dependency from this library for the rest of mediasdk library.

_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
```diff
@@ -50,46 +46,49 @@ FILE* mfx_trace_open_conf_file(const char* name)
...
-mfxTraceU32 mfx_trace_get_value_pos(FILE* file,
-                                    const char* pName,
-                                    char* line, mfxTraceU32 line_size,
-                                    char** value_pos)
+inline std::string& trim_string(std::string&& s)
+{
+    // Remove leading spaces, tabulation, = character
+    s.erase(0, s.find_first_not_of(" =\t"));
+    // Remove trailing spaces, tabulation, newline symbol, spaces
+    s.erase(s.find_last_not_of(" \t\n\r\f\v") + 1);
```
dvrogozh: What will happen if find will return npos?

_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
```diff
@@ -50,46 +46,49 @@ FILE* mfx_trace_open_conf_file(const char* name)
...
 /*------------------------------------------------------------------------------*/
 
-mfxTraceU32 mfx_trace_get_value_pos(FILE* file,
-                                    const char* pName,
-                                    char* line, mfxTraceU32 line_size,
-                                    char** value_pos)
+inline std::string& trim_string(std::string&& s)
+{
+    // Remove leading spaces, tabulation, = character
+    s.erase(0, s.find_first_not_of(" =\t"));
```
dvrogozh: what will happen if find will return npos?

dvrogozh:
@mgonchar : and eventually Android build went wrong...

_studio/shared/mfx_trace/CMakeLists.txt
mgonchar: I thought this will affect only mfx_trace target, which no longer needs SafeStrings library, isn't it?

_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
mgonchar: It will throw out_range exception. Will fix it

_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
mgonchar: it will erase entire string. definitely not what I wanted. will fix

_studio/shared/mfx_trace/CMakeLists.txt
dvrogozh: it will, still not needed.

mgonchar:
@dvrogozh yes, I see. Build failed because of C-linkage... do we really need it in this component?

_studio/shared/mfx_trace/src/mfx_trace_utils_linux.cpp
```diff
@@ -21,26 +21,58 @@
...
+{
+    // Remove leading spaces, tabulation, = character
+    size_t pos = s.find_first_not_of(" =\t");
+    if (pos != std::string::npos)
+        s.erase(0, pos);
+
+    // Remove trailing spaces, tabulation, newline symbol, spaces
+    pos = s.find_last_not_of(" \t\n\r\f\v");
+    if (pos != std::string::npos)
+        s.erase(pos + 1);
```
dvrogozh: should not that be `s.erase(pos+1, s.end())`? And what will happen if `pos+1=s.end()`.

mgonchar: no, I'm calling
```
erase(size_t index = 0, size_t count = npos)
```
so pos is not an iterator but index. In my case it will just remove all characters in string started from first of any of " \t\n\r\f\v" occurrence  (in our simple par-file format it is ok).

This function works fine if pos == str.size() it just do nothing. it throws out_of_range exception if pos > str.size() but it shouldn't happen in this code

dvrogozh: I am ok then.

PR #760, opened 2018-10-16
[mfx_dispatch] Fix relation operator for mfxVersion

vshampor:
Courtesy of mfx_dispatch_test unit tests still in development.

### DIFF
--- a/api/mfx_dispatch/linux/mfxloader.h
+++ b/api/mfx_dispatch/linux/mfxloader.h
@@ -41,12 +41,13 @@ inline bool operator != (const mfxPluginUID &lhs, const mfxPluginUID & rhs)
 
 inline bool operator < (const mfxVersion &lhs, const mfxVersion & rhs)
 {
-    return (lhs.Major == rhs.Major) && (lhs.Minor < rhs.Minor);
+    return (lhs.Major < rhs.Major ||
+            (lhs.Major == rhs.Major && lhs.Minor < rhs.Minor));
 }
 
 inline bool operator <= (const mfxVersion &lhs, const mfxVersion & rhs)
 {
-    return (lhs.Major == rhs.Major) && (lhs.Minor <= rhs.Minor);
+    return (lhs < rhs || (lhs.Major == rhs.Major && lhs.Minor == rhs.Minor));
 }
 
 namespace MFX {


api/mfx_dispatch/linux/mfxloader.h
```diff
@@ -41,12 +41,13 @@ inline bool operator != (const mfxPluginUID &lhs, const mfxPluginUID & rhs)
...
 {
-    return (lhs.Major == rhs.Major) && (lhs.Minor < rhs.Minor);
+    return (lhs.Major < rhs.Major ||
+            (lhs.Major == rhs.Major && lhs.Minor < rhs.Minor));
 }
 
 inline bool operator <= (const mfxVersion &lhs, const mfxVersion & rhs)
 {
-    return (lhs.Major == rhs.Major) && (lhs.Minor <= rhs.Minor);
+    return (operator<(lhs, rhs) || (lhs.Major == rhs.Major && lhs.Minor == rhs.Minor));
```
mgonchar: do not over complicate syntax, just lhs < rhs

api/mfx_dispatch/linux/mfxloader.h
```diff
@@ -41,12 +41,13 @@ inline bool operator != (const mfxPluginUID &lhs, const mfxPluginUID & rhs)
 
 inline bool operator < (const mfxVersion &lhs, const mfxVersion & rhs)
 {
-    return (lhs.Major == rhs.Major) && (lhs.Minor < rhs.Minor);
+    return (lhs.Major < rhs.Major ||
```
dvrogozh: Formally you are right, but changing of major version would mean backward incompatible change in mediasdk API. I am not sure that dispatcher will ever handle this. More likely it will work for single version of the major API. In a way this check was on-purpose...

vshampor: Currently, for example, version 0.0 does not evaluate '<' than 1.28. Surely the dispatcher should cut off such blatantly wrong reported versions at lines:
https://github.com/Intel-Media-SDK/MediaSDK/blob/090318ce4667375f1410690a28b84f89c5431315/api/mfx_dispatch/linux/mfxloader.cpp#L226-L229

Otherwise, if the lib reports a version of 0.0 (or a version of 0.x, for that matter) but everything else is fine, the MFXInit returns MFX_ERR_NONE, which is probably not we want.

PR #761, opened 2018-10-16
h265e- use C++ primitives (memcpy => std::copy)

ilapichev:
Issue: MDP-50973

### DIFF
--- a/samples/sample_hevc_fei/include/hevc_fei_encode.h
+++ b/samples/sample_hevc_fei/include/hevc_fei_encode.h
@@ -31,7 +31,8 @@ class FEI_Encode
 {
 public:
     FEI_Encode(MFXVideoSession* session, mfxHDL hdl,
-        MfxVideoParamsWrapper& encode_pars, const mfxExtFeiHevcEncFrameCtrl& encodeCtrl,
+        MfxVideoParamsWrapper& encode_pars,
+        const mfxExtFeiHevcEncFrameCtrl& frameCtrl, const PerFrameTypeCtrl& frametypeCtrl,
         const msdk_char* strDstFile, const msdk_char* mvpInFile,
         const msdk_char* repackctrlFile, const msdk_char* repackstatFile,
         PredictorsRepaking* repacker);
@@ -72,7 +73,8 @@ private:
 
     std::auto_ptr<PredictorsRepaking> m_repacker;
 
-    mfxExtFeiHevcEncFrameCtrl m_defFrameCtrl; // contain default per-frame options including user-specified
+    mfxExtFeiHevcEncFrameCtrl m_defFrameCtrl;     // contain default and user-specified options per frame
+    PerFrameTypeCtrl          m_ctrlPerFrameType; // contain default and user-specified options per frame type
 
     mfxU32  m_processedFrames;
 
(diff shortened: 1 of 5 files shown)


_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h
```diff
@@ -77,7 +77,7 @@ template<class T> inline void Zero(T * first, size_t cnt)     { memset(first, 0,
 template<class T, class U> inline void Copy(T & dst, U const & src)
 {
     static_assert(sizeof(T) == sizeof(U), "copy_objects_of_different_size");
-    memcpy_s(&dst, sizeof(dst), &src, sizeof(dst));
+    std::copy(std::begin(src), std::end(src), std::begin(dst));
```
mgonchar: no, I disagree. Passed arguments might have any type. It could even be non-POD, so previous code was super dangerous. Remove this template function at all and fix copying in place everywhere it used. I don't see other way

_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h
```diff
@@ -600,7 +600,7 @@ namespace ExtBuffer
         bUnsuppoted = (memcmp(&buf_ref, &buf, sizeof(T))!=0);
         if (bUnsuppoted && bFix)
         {
-            memcpy_s(&buf, sizeof(T), &buf_ref, sizeof(T));
+            buf_ref = buf;
```
mgonchar: no, the previous code did other assignment: ```buf = buf_ref;```

_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h
```diff
@@ -77,7 +77,7 @@ template<class T> inline void Zero(T * first, size_t cnt)     { memset(first, 0,
 template<class T, class U> inline void Copy(T & dst, U const & src)
 {
     static_assert(sizeof(T) == sizeof(U), "copy_objects_of_different_size");
-    memcpy_s(&dst, sizeof(dst), &src, sizeof(dst));
+    std::copy(std::begin(src), std::end(src), std::begin(dst));
 }
 template<class T> inline void CopyN(T* dst, const T* src, size_t N)
```
mgonchar: this function also has memcpy inside and should also be removed

_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h
dvrogozh: I wonder, do we have the same function declared in AVC and VP9e utils?

mgonchar: This file has more dangerous templates, like Zero and Fill. + many other functions which are candidates for clean up (max/min/abs, even sort on macro)

I agree that we need to check other files

dvrogozh: Yep. I remember that there is an awful duplication of utility code between our encoders. That's really a chance to clean up this mess. Need to plan the activity though. No need to mix everything together. If the idea is to remove self-written wrappers in a preference of std functions - this will be a good step forward by itself.

PR #763, opened 2018-10-16
safec: remove the library in a preference of glibc/libstd

dvrogozh:
@maintainers: #753 from @mgonchar should be merged first. I will rebase afterwards. - DONE


chivakker:
contrib/Android.mk should be removed as well.

dvrogozh:
@chivakker : it is deleted. I think you managed to take the code between my updates. In an update I fixed Android build and build for some gcc compilers.

PR #764, opened 2018-10-17
VP9e - use C++ primitives (memcpy => std::copy)

ilapichev:
Issue: MDP-50976


_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_ddi.h
```diff
@@ -194,13 +194,14 @@ typedef struct tagENCODE_CAPS_VP9
         mfxU32 bufferSize)
     {
         mfxU32   ivf_file_header[8] = { 0x46494B44, 0x00200000, 0x30395056, width + (height << 16), FrameRateN, FrameRateD, numFramesInFile, 0x00000000 };
-        memcpy_s(pBitstream, bufferSize, ivf_file_header, sizeof (ivf_file_header));
+        std::copy(std::begin(ivf_file_header),std::end(ivf_file_header), pBitstream);
     };
 
     inline void AddPictureHeader(mfxU8* pBitstream, mfxU32 bufferSize)
     {
         mfxU32 ivf_frame_header[3] = { 0x00000000, 0x00000000, 0x00000000 };
```
mgonchar: now bufSize is not used at all. maybe it is good idea to add some check + exception? @dvrogozh what do you think?

dvrogozh: I think that we should take care about arguments we pass to function calls. This place eventually miss a check for the buffer size to be sufficient. Maybe check presents where these functions are called, but this is not visible from the patch.

Check is a minimal thing to be done here. But you need to pass error from this function which returns void. Thus, @ilapichev, please, look where function is called, modify prototype to return an error and check the return status.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw.cpp
```diff
@@ -918,7 +918,7 @@ mfxStatus MFXVideoENCODEVP9_HW::GetVideoParam(mfxVideoParam *par)
 inline void UpdatePictureHeader(mfxU32 frameLen, mfxU32 frameNum, mfxU8* pPictureHeader, mfxU32 bufferSize)
 {
     mfxU32 ivf_frame_header[3] = {frameLen, frameNum << 1, 0x00000000};
-    memcpy_s(pPictureHeader, bufferSize, ivf_frame_header, sizeof (ivf_frame_header));
+    std::copy(std::begin(ivf_frame_header),std::end(ivf_frame_header), pPictureHeader);
```
dvrogozh: same as above

PR #769, opened 2018-10-17
clang 6.0: result of comparison is always false warning fixed

IlyaSergeyev:
Possible it was a bug.
short signed int mv[i][ref].x will never equal 0x8000, even it contains this binary value


samples/sample_hevc_fei/src/fei_predictors_repaking.cpp
```diff
@@ -442,7 +442,8 @@ void SelectFromMV(const mfxI16Pair(* mv)[2], mfxI32 count, mfxI16Pair (&res)[2])
         mfxI32 found = 0, xsum = 0, ysum = 0;
         for (int i = 0; i < count; i++)
         {
-            if (mv[i][ref].x == 0x8000) // ignore intra
+            int tx = mv[i][ref].x & 0xFFFF;
+            if (tx == 0x8000) // ignore intra
```
mgonchar: too complicated, just make literal negative:
```
if (mv[i][ref].x == -0x8000)
```

PR #772, opened 2018-10-17
h265e- use C++ primitives (memcpy => std::copy)

ilapichev:
Issue: MDP-50973

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
@@ -385,7 +385,10 @@ mfxStatus ImplementationAvc::Query(
         sts = ReadSpsPpsHeaders(tmp);
         MFX_CHECK_STS(sts);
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h
```diff
@@ -473,7 +462,7 @@ namespace ExtBuffer
 
     #define _CopyPar(dst, src, PAR) dst.PAR = src.PAR;
     #define _CopyPar1(PAR) _CopyPar(buf_dst, buf_src, PAR);
-    #define _CopyStruct1(PAR) Copy(buf_dst.PAR, buf_src.PAR);
+    #define _CopyStruct1(PAR) std::copy(std::begin(buf_src.PAR), std::end(buf_src.PAR), std::begin(buf_dst.PAR));
```
mgonchar: this code assumes that PAR is a container. if it always true, you could rename it to _CopyArray or something like this

PR #774, opened 2018-10-17
[MJPEGd/UMC]Avoid using address of local variable as member of CBitSt…

akokoshn:
…reamInput

-Time live of instance CBitStreamInput longer then local variable
 so move responsibility for memory management CBaseStreamInput* m_in
 to class  CBitStreamInput

### DIFF
--- a/_studio/shared/umc/codec/jpeg_dec/include/jpegdec_base.h
+++ b/_studio/shared/umc/codec/jpeg_dec/include/jpegdec_base.h
@@ -27,6 +27,7 @@
 #include "decqtbl.h"
 #include "dechtbl.h"
 #include "colorcomp.h"
+#include "membuffin.h"
 #include "bitstreamin.h"
 
 class CJPEGDecoderBase
@@ -38,8 +39,8 @@ public:
 
   virtual void Reset(void);
 
-  JERRCODE SetSource(
-    CBaseStreamInput* pStreamIn);
+  JERRCODE SetSource(const uint8_t* pBuf, size_t buflen);
+  JERRCODE Seek(long offset, int origin);
 
   virtual JERRCODE ReadHeader(
     int*     width,
@@ -115,6 +116,7 @@ public:
   int      m_al;
   int      m_ah;
   JMARKER  m_marker;
+  CMemBuffInput m_stream_in;
 
   int      m_nblock;
 
--- a/_studio/shared/umc/codec/jpeg_dec/src/jpegdec_base.cpp
+++ b/_studio/shared/umc/codec/jpeg_dec/src/jpegdec_base.cpp
@@ -110,6 +110,8 @@ void CJPEGDecoderBase::Reset(void)
   m_marker                 = JM_NONE;
 
   m_nblock                 = 0;
+
+  m_stream_in.Close();
   return;
 } // CJPEGDecoderBase::Reset(void)
 
@@ -157,17 +159,23 @@ JERRCODE CJPEGDecoderBase::Clean(void)
 
 #define BS_BUFLEN 16384
 
-JERRCODE CJPEGDecoderBase::SetSource(
-  CBaseStreamInput* pInStream)
+JERRCODE CJPEGDecoderBase::SetSource(const uint8_t* pBuf, size_t buflen)
 {
-  JERRCODE jerr;
+  JERRCODE jerr = m_stream_in.Open(pBuf, buflen);
+  if(JPEG_OK != jerr)
+    return jerr;
 
-  jerr = m_BitStreamIn.Attach(pInStream);
+  jerr = m_BitStreamIn.Attach(&m_stream_in);
   if(JPEG_OK != jerr)
     return jerr;
 
   return m_BitStreamIn.Init(BS_BUFLEN);
 } // CJPEGDecoderBase::SetSource()
+
+JERRCODE CJPEGDecoderBase::Seek(long offset, int origin)
+{
+  return m_stream_in.Seek(offset, origin);
+}// CJPEGDecoderBase::Seek
 
 JERRCODE CJPEGDecoderBase::DetectSampling(void)
 {
--- a/_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
+++ b/_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
@@ -324,15 +324,12 @@ Status MJPEGVideoDecoderBaseMFX::_GetFrameInfo(const uint8_t* pBitStream, size_t
     int32_t   precision;
     JSS      sampling;
     JCOLOR   color;
-    CMemBuffInput in;
     JERRCODE jerr;
 
     if (!m_IsInit)
         return UMC_ERR_NOT_INITIALIZED;
 
-    in.Open(pBitStream,nSize);
-
-    jerr = m_decBase->SetSource(&in);
+    jerr = m_decBase->SetSource(pBitStream,nSize);
     if(JPEG_OK != jerr)
         return UMC_ERR_FAILED;
 
@@ -370,15 +367,12 @@ void MJPEGVideoDecoderBaseMFX::SetFrameAllocator(FrameAllocator * frameAllocator
 
 Status MJPEGVideoDecoderBaseMFX::FindStartOfImage(MediaData * in)
 {
-    CMemBuffInput source;
     JERRCODE jerr;
 
     if (!m_IsInit)
         return UMC_ERR_NOT_INITIALIZED;
 
-    source.Open((uint8_t*) in->GetDataPointer(), in->GetDataSize());
-
-    jerr = m_decBase->SetSource(&source);
+    jerr = m_decBase->SetSource((uint8_t*) in->GetDataPointer(), in->GetDataSize());
     if(JPEG_OK != jerr)
         return UMC_ERR_FAILED;
 
--- a/_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp
+++ b/_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp
@@ -27,7 +27,6 @@
 
 #include <string.h>
 #include "vm_debug.h"
-#include "membuffin.h"
 #include <algorithm>
 #include <limits.h>
 
@@ -226,17 +225,13 @@ mfxStatus MJPEGVideoDecoderMFX_HW::CheckStatusReportNumber(uint32_t statusReport
     return MFX_TASK_DONE;
 }
 
-Status MJPEGVideoDecoderMFX_HW::_DecodeHeader(CBaseStreamInput* in, int32_t* cnt)
+Status MJPEGVideoDecoderMFX_HW::_DecodeHeader(int32_t* cnt)
 {
     JSS      sampling;
     JERRCODE jerr;
 
     if (!m_IsInit)
         return UMC_ERR_NOT_INITIALIZED;
-
-    jerr = m_decBase->SetSource(in);
-    if(JPEG_OK != jerr)
-        return UMC_ERR_FAILED;
 
     mfxSize size = {};
 
@@ -254,7 +249,7 @@ Status MJPEGVideoDecoderMFX_HW::_DecodeHeader(CBaseStreamInput* in, int32_t* cnt
 
     if ((m_frameSampling != (int)sampling) || (m_frameDims.width && sizeHaveChanged))
     {
-        in->Seek(-m_decBase->GetNumDecodedBytes(),UIC_SEEK_CUR);
+        m_decBase->Seek(-m_decBase->GetNumDecodedBytes(),UIC_SEEK_CUR);
         *cnt = 0;
         return UMC_ERR_NOT_ENOUGH_DATA;
     }
@@ -267,10 +262,10 @@ Status MJPEGVideoDecoderMFX_HW::_DecodeField(MediaDataEx* in)
 {
     int32_t     nUsedBytes = 0;
 
-    CMemBuffInput stream;
-    stream.Open((uint8_t*)in->GetDataPointer(), (int32_t)in->GetDataSize());
+    if(JPEG_OK != m_decBase->SetSource((uint8_t*)in->GetDataPointer(), (int32_t)in->GetDataSize()))
+        return UMC_ERR_FAILED;
 
-    Status status = _DecodeHeader(&stream, &nUsedBytes);
+    Status status = _DecodeHeader(&nUsedBytes);
     if (status > 0)
     {
         in->MoveDataPointer(nUsedBytes);
@@ -414,10 +409,7 @@ Status MJPEGVideoDecoderMFX_HW::GetFrameHW(MediaDataEx* in)
                         (pAuxData->offsets[nextNotRSTMarkerPos] - pAuxData->offsets[i]) :
                         (srcSize - pAuxData->offsets[i]);
 
-            CMemBuffInput stream;
-            stream.Open((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
-
-            jerr = m_decBase->SetSource(&stream);
+            jerr = m_decBase->SetSource((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
             if(JPEG_OK != jerr)
                 return UMC_ERR_FAILED;
 
@@ -458,10 +450,7 @@ Status MJPEGVideoDecoderMFX_HW::GetFrameHW(MediaDataEx* in)
         }
         else if (JM_DRI == marker)
         {
-            CMemBuffInput stream;
-            stream.Open((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
-
-            jerr = m_decBase->SetSource(&stream);
+            jerr = m_decBase->SetSource((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
             if(JPEG_OK != jerr)
                 return UMC_ERR_FAILED;
 
@@ -476,10 +465,7 @@ Status MJPEGVideoDecoderMFX_HW::GetFrameHW(MediaDataEx* in)
         }
         else if (JM_DQT == marker)
         {
-            CMemBuffInput stream;
-            stream.Open((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
-
-            jerr = m_decBase->SetSource(&stream);
+            jerr = m_decBase->SetSource((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
             if(JPEG_OK != jerr)
                 return UMC_ERR_FAILED;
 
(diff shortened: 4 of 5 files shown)


_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -58,15 +58,18 @@ JERRCODE CBitStreamInput::Attach(CBaseStreamInput* in)
 
 JERRCODE CBitStreamInput::Detach(void)
 {
-  if(0 != m_pData)
+  if(nullptr != m_pData)
```
mgonchar: no sense in this check. delete works fine for nullptr (it just do nothing)

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp
```diff
@@ -398,6 +398,13 @@ Status MJPEGVideoDecoderMFX_HW::GetFrameHW(MediaDataEx* in)
             foundSOS = true;
         }
 
+        CMemBuffInput* stream = new CMemBuffInput();
```
dvrogozh: This looks to be a separate change.

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -58,15 +58,13 @@ JERRCODE CBitStreamInput::Attach(CBaseStreamInput* in)
...
-  {
-    // deallocate internal memory
-    delete[] m_pData;
-    m_pData = 0;
-  }
+  // deallocate internal memory
+  delete[] m_pData;
+  m_pData = nullptr;
+
+  delete m_in;
```
dvrogozh: this is a key change which is lost on other changes.

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -80,10 +78,7 @@ JERRCODE CBitStreamInput::Init(int bufSize)
 {
   m_DataLen = (int)bufSize;
 
-  if (m_pData)
-  {
-    delete[] m_pData;
-  }
+  delete[] m_pData;
```
dvrogozh: this is not a key change: separate patch

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -65,8 +65,9 @@ JERRCODE CBitStreamInput::Detach(void)
     m_pData = 0;
   }
 
-  m_in         = 0;
-  m_pData      = 0;
+  delete m_in;
+  m_in = nullptr;
```
dvrogozh: hm. We did not have `delete m_in` before. Why? We had a memory leak? Can you, please, provide valgrind test results for the command line you are using to verify that change don't break anything. I hope to see sample decode or encode command line:
`valgrind --tool=memcheck --leak-check=full --log-file=vlg.log <cmdline>`

I am looking for the last section with the memleaks only (omit errors). Please, provide 3 results for each commit you have.

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -324,15 +324,15 @@ Status MJPEGVideoDecoderBaseMFX::_GetFrameInfo(const uint8_t* pBitStream, size_t
     int32_t   precision;
     JSS      sampling;
     JCOLOR   color;
-    CMemBuffInput in;
+    CMemBuffInput* in = new CMemBuffInput();
```
dvrogozh: Is that new corresponding to delete in the prev. patch? If yes, prev. patch looks wrong to me!

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp
```diff
@@ -398,6 +398,13 @@ Status MJPEGVideoDecoderMFX_HW::GetFrameHW(MediaDataEx* in)
...
+        stream->Open((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
+
+        jerr = m_decBase->SetSource(stream);
+        if(JPEG_OK != jerr)
+            return UMC_ERR_FAILED;
+
```
dvrogozh: This change looks code refactoring, please (removal of code duplication), extract it to a separate change!

_studio/shared/umc/codec/jpeg_common/include/bitstreamin.h
```diff
@@ -58,7 +58,7 @@ class CBitStreamInput
 protected:
   CBaseStreamInput* m_in;
 
-  uint8_t*            m_pData;
+  uint8_t*          m_pData;
```
dvrogozh: I would prefer to see this change as a first commit in the series, not the last one.

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
akokoshn: no leaks before: m_in - pointer to local variable (on stack) which will be deleted when leave function's body

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -324,15 +324,15 @@ Status MJPEGVideoDecoderBaseMFX::_GetFrameInfo(const uint8_t* pBitStream, size_t
     int32_t   precision;
     JSS      sampling;
     JCOLOR   color;
-    CMemBuffInput in;
+    std::shared_ptr<CBaseStreamInput> in(new CMemBuffInput);
```
dvrogozh: std::make_shared

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -370,15 +370,15 @@ void MJPEGVideoDecoderBaseMFX::SetFrameAllocator(FrameAllocator * frameAllocator
 
 Status MJPEGVideoDecoderBaseMFX::FindStartOfImage(MediaData * in)
 {
-    CMemBuffInput source;
+    std::shared_ptr<CBaseStreamInput> source(new CMemBuffInput);
```
dvrogozh: std::make_shared

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp
```diff
@@ -267,10 +267,10 @@ Status MJPEGVideoDecoderMFX_HW::_DecodeField(MediaDataEx* in)
 {
     int32_t     nUsedBytes = 0;
 
-    CMemBuffInput stream;
-    stream.Open((uint8_t*)in->GetDataPointer(), (int32_t)in->GetDataSize());
+    std::shared_ptr<CBaseStreamInput> stream(new CMemBuffInput);
```
dvrogozh: std::make_shared

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp
```diff
@@ -400,10 +400,10 @@ Status MJPEGVideoDecoderMFX_HW::GetFrameHW(MediaDataEx* in)
         if ((JM_RST0 <= marker && JM_RST7 >= marker) || (JM_EOI == marker))
             continue;
 
-        CMemBuffInput stream;
-        stream.Open((uint8_t*)in->GetDataPointer() + pAuxData->offsets[i] + 2, (int32_t)in->GetDataSize() - pAuxData->offsets[i] - 2);
+        std::shared_ptr<CBaseStreamInput> stream(new CMemBuffInput);
```
dvrogozh: std::make_shared

_studio/shared/umc/codec/jpeg_common/include/bitstreamin.h
```diff
@@ -36,7 +37,7 @@ class CBitStreamInput
   CBitStreamInput(void);
   virtual ~CBitStreamInput(void);
 
-  JERRCODE Attach(CBaseStreamInput* in);
+  JERRCODE Attach(std::shared_ptr<CBaseStreamInput> in);
```
mgonchar: this will call heavy shared_ptr copy constructor. I don't think that this is desired

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -113,7 +106,7 @@ JERRCODE CBitStreamInput::FillBuffer(int nMinBytes)
     m_currPos = 0;
   }
 
-  if(!m_eod)
+  if(!m_eod && nullptr != m_in.get())
```
mgonchar: shared_ptr / unique_ptr have operator bool (), which you can use to check if it isn't empty:
```
if(!m_eod && m_in)
```

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -151,7 +144,7 @@ JERRCODE CBitStreamInput::Seek(long offset, int origin)
         m_currPos     = _offset;
         m_nUsedBytes += offset;
       }
-      else
+      else if (nullptr != m_in.get())
```
mgonchar: same

_studio/shared/umc/codec/jpeg_dec/include/jpegdec_base.h
```diff
@@ -39,7 +39,7 @@ class CJPEGDecoderBase
   virtual void Reset(void);
 
   JERRCODE SetSource(
-    CBaseStreamInput* pStreamIn);
+    std::shared_ptr<CBaseStreamInput> pStreamIn);
```
mgonchar: same about copy

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -324,15 +324,15 @@ Status MJPEGVideoDecoderBaseMFX::_GetFrameInfo(const uint8_t* pBitStream, size_t
...
     JCOLOR   color;
-    CMemBuffInput in;
+    std::shared_ptr<CBaseStreamInput> in = std::make_shared<CMemBuffInput>();
     JERRCODE jerr;
 
     if (!m_IsInit)
         return UMC_ERR_NOT_INITIALIZED;
 
-    in.Open(pBitStream,nSize);
+    ((CMemBuffInput*)in.get())->Open(pBitStream,nSize);
```
mgonchar: shared_pointer is a proxy of pointer, so think about it as a pointer, just use ->
```
in->Open(pBitStream,nSize);
```

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -370,15 +370,15 @@ void MJPEGVideoDecoderBaseMFX::SetFrameAllocator(FrameAllocator * frameAllocator
...
 {
-    CMemBuffInput source;
+    std::shared_ptr<CBaseStreamInput> source = std::make_shared<CMemBuffInput>();
     JERRCODE jerr;
 
     if (!m_IsInit)
         return UMC_ERR_NOT_INITIALIZED;
 
-    source.Open((uint8_t*) in->GetDataPointer(), in->GetDataSize());
+    ((CMemBuffInput*)source.get())->Open((uint8_t*) in->GetDataPointer(), in->GetDataSize());
```
mgonchar: same

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -324,15 +324,15 @@ Status MJPEGVideoDecoderBaseMFX::_GetFrameInfo(const uint8_t* pBitStream, size_t
...
     JERRCODE jerr;
 
     if (!m_IsInit)
         return UMC_ERR_NOT_INITIALIZED;
 
-    in.Open(pBitStream,nSize);
+    ((CMemBuffInput*)in.get())->Open(pBitStream,nSize);
 
-    jerr = m_decBase->SetSource(&in);
+    jerr = m_decBase->SetSource(in);
```
mgonchar: I believe what you actually want  to do is to move(!) this pointer to another pointer

_studio/shared/umc/codec/jpeg_common/include/bitstreamin.h
```diff
@@ -56,9 +57,9 @@ class CBitStreamInput
   JERRCODE ReadDword(int* dword);
 
 protected:
-  CBaseStreamInput* m_in;
+  std::shared_ptr<CBaseStreamInput> m_in;
```
mgonchar: from context below I believe you can use unique_ptr, not shared

akokoshn: @mgonchar , seems we can't use unique_ptr - object under pointer should be changed in the function body after sent pointer to m_decBase. Look at https://github.com/Intel-Media-SDK/MediaSDK/blob/3d35d1c0be7453cffa62612961233801c033e36b/_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_hw.cpp#L257

mgonchar: @akokoshn, didn't get why exactly we can't use unique_ptr in that code?

akokoshn: @mgonchar , how we can call in->Seek(), after send in to SetSource() (move ownership), if in - unique_ptr?

mgonchar: @akokoshn I see... ok, in general you are right, we can't do so. But this code breaks responsibility principle anyway - since we transferred ownership of CBaseStreamInput (even if it is a raw pointer when we setting source we mean that since that moment m_decBase is responsible for all operations with this input stream (btw thats why unique_ptr fits this model so well)), but after than we manipulate this pointer from outside for some reason. I see that there is a method CBitStreamInput::Seek, which supposed to be used for file pointer seeking, but for some reason it wasn't used. CBitStreamInput::Seek call is not equivalent to existing code, so it looks strange why it wasn't used. Could you check if we really need to operate with CBaseStreamInput objeect directly or we should call CBitStreamInput::Seek instead?

akokoshn: @mgonchar , we haven't interface for call CBitStreamInput::Seek through CJPEGDecoderBase. and CBitStreamInput::Seek is not same as CMemBuffInput::Seek()

_studio/shared/umc/codec/jpeg_common/src/bitstreamin.cpp
```diff
@@ -28,8 +28,8 @@
 
 CBitStreamInput::CBitStreamInput(void)
 {
-  m_in         = 0;
-  m_pData      = 0;
+  m_in.reset();
```
mgonchar: no need to do this

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode_base.cpp
```diff
@@ -324,15 +324,15 @@ Status MJPEGVideoDecoderBaseMFX::_GetFrameInfo(const uint8_t* pBitStream, size_t
     int32_t   precision;
     JSS      sampling;
     JCOLOR   color;
-    CMemBuffInput in;
+    std::shared_ptr<CBaseStreamInput> in = std::make_shared<CMemBuffInput>();
```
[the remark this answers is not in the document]
akokoshn: but pBitStream comes from MJPEGVideoDecoderBaseMFX::DecodeHeader(MediaData* in) <= VideoDECODEMJPEG::DecodeHeader(... mfxBitstream *bs ...), where bs - comes from application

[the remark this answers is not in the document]
dmitry-gurulev: @akokoshn 

> reset state CBitStreamInput m_BitStreamIn

If u did it in that way, sending pointer to local variable wouldn't be an issue.

_studio/shared/umc/codec/jpeg_dec/src/jpegdec_base.cpp
```diff
@@ -110,6 +110,8 @@ void CJPEGDecoderBase::Reset(void)
   m_marker                 = JM_NONE;
 
   m_nblock                 = 0;
+
+  m_stream_in.Close();
```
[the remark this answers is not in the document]
akokoshn: not sure for SetSource(), if pBuf = 0 we should set m_stream_in = 0 or  keep old value (in this case SetSource() will returns error)?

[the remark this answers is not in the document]
mgonchar: If I were a designer of this class, I would call Close first inside Open. But for minimizing changes let's keep current implementation. those functions actually do almost nothing, so I don't think it would be a problem

_studio/shared/umc/codec/jpeg_dec/src/jpegdec_base.cpp
```diff
@@ -157,18 +159,22 @@ JERRCODE CJPEGDecoderBase::Clean(void)
...
     return jerr;
 
+  m_BitStreamIn.Attach(&m_stream_in);
   return m_BitStreamIn.Init(BS_BUFLEN);
 } // CJPEGDecoderBase::SetSource()
 
+JERRCODE CJPEGDecoderBase::Seek(long offset, int origin)
+{
+  return m_stream_in.Seek(offset, origin);
+}// CJPEGDecoderBase::Seek
```
mgonchar: useless comment. We already know what function had just finished. It makes some sense for large functions which occupies several screens, but still not much, modern IDEs/text editors easily find end of scope brackets

_studio/shared/umc/codec/jpeg_dec/src/jpegdec_base.cpp
```diff
@@ -157,18 +159,22 @@ JERRCODE CJPEGDecoderBase::Clean(void)
...
+JERRCODE CJPEGDecoderBase::SetSource(const uint8_t* pBuf, size_t buflen)
 {
-  JERRCODE jerr;
+  JERRCODE jerr = m_stream_in.Open(pBuf, buflen);
 
-  jerr = m_BitStreamIn.Attach(pInStream);
   if(JPEG_OK != jerr)
     return jerr;
 
+  m_BitStreamIn.Attach(&m_stream_in);
```
mgonchar: you forgot to check attach return status

akokoshn: It not make sense, Attach() can return only  JPEG_OK

JERRCODE CBitStreamInput::Attach(CBaseStreamInput* in)
{
  Detach();

  m_in = in;

  return JPEG_OK;
} // CBitStreamInput::Attach()

_studio/shared/umc/codec/jpeg_dec/src/jpegdec_base.cpp
akokoshn: I just keep local code style
If remove this comment, need remove all another - changes in many lines (which not relate to PR).
But if it really need, I can make separate commit in this PR for remove useless comment in jpegdec_base.h/cpp (actually, what about another files?)

mgonchar: Yes, I understand, but this function was introduced by you, so you could not copy bad practice within it. Definitely it is really minor issue, I don't want to waste your time to fix it here/in another files. It is just remark from me :)

PR #768, opened 2018-10-17
clang 6.0: zero initialization warnings in fei

IlyaSergeyev:
zero initialization, that not causes clang 6.0 warnings

PR #776, opened 2018-10-17
Fixed last android warnings and add werror flag for libmfx sw/hw 32/64

Losev-Ivan:


Losev-Ivan:
@dvrogozh this commit add Werror flag for all targets except samples therefore it do not influence on #769. The #768 fix warnings in mfx_h264_la.cpp. mfx_h264_la.h includes only if defined MFX_ENABLE_LA_H264_VIDEO_HW (in libmfxsw_enc.cpp) but this identifier is always undef in case of defined ANDROID. (_studio/shared/include/mfx_config.h:46   _studio/shared/include/mfx_config.h:61)

Losev-Ivan:
rebase done

PR #779, opened 2018-10-17
Remove low level copies

mgonchar:


_studio/shared/include/libmfx_core_interface.h
```diff
@@ -170,20 +171,22 @@ class EncodeHWCaps
...
         if (!m_caps)
         {
             m_encode_guid = encode_guid;
             m_size = array_size;
             m_caps = malloc(sizeof(CAPS)*array_size);
             if (!m_caps)
                 return MFX_ERR_MEMORY_ALLOC;
-            memcpy_s(m_caps, sizeof(CAPS)*array_size, hwCaps, sizeof(CAPS)*array_size);
+
+            std::copy(hwCaps, hwCaps + array_size, dst);
```
dvrogozh: that's wrong. You assigned dst=m_caps before you actually allocated m_caps on line 179. Why you don't want to use reinterpret_cast<CAPS*>(m_caps) directly in std::copy?

mgonchar: you are right. tried to not increase code line width too much, several reinterpret_casts occupy significant space and reduce readability. will fix

PR #782, opened 2018-10-18
h264e - use C++ primitives (memcpy => std::copy)

ilapichev:
Issie: MDP-50970


ilapichev:
Hi guys, I completely removed macro definitions from this code.
Is it fine with this?

PR #800, opened 2018-10-19
[hevce] Replaced memmove in utility functions with std primitives (co…

mgonchar:
…py / fill)


mgonchar:
Need more manual testing

PR #802, opened 2018-10-22
[hevce] Replaced memmove in utility functions with std primitives (co…

mgonchar:
…py / fill)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -3586,7 +3606,8 @@ void ConfigureTask(
 
     //construct ref lists
     Zero(task.m_numRefActive);
```
dvrogozh: Why you did not change this Zero?

mgonchar: I planned to get rid of Zero functions in later change

PR #804, opened 2018-10-23
[RFC] Expand minTargetKbps for lower bit rate.

wangyan42164:
Current formula may be based on experience.
It could be expanded for lower bit rate when debugging the following
issues:
https://github.com/intel/media-driver/issues/325

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -3394,7 +3394,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
             {
                 mfxF64 rawDataBitrate = 12.0 * par.mfx.FrameInfo.Width * par.mfx.FrameInfo.Height *
                     par.mfx.FrameInfo.FrameRateExtN / par.mfx.FrameInfo.FrameRateExtD;
-                mfxU32 minTargetKbps = mfxU32(MFX_MIN(0xffffffff, rawDataBitrate / 1000 / 700));
+                mfxU32 minTargetKbps = mfxU32(MFX_MIN(0xffffffff, rawDataBitrate / 1000 / 800));
 
                 if (par.calcParam.targetKbps < minTargetKbps)
                 {


dmitryermilov:
@wangyan42164 , if you need to disable MSDK logic which sets minimal bitrate you may use mfxExtCodingOption2::BitrateLimit:

BitrateLimit | Turn off this flag to remove   bitrate limitations imposed by the SDK encoder. This flag is intended for   special usage models and usually the application should not set it. Setting   this flag may lead to violation of HRD conformance and severe visual   artifacts. See the CodingOptionValue enumerator for values of this   option. The default value is ON, i.e. bitrate is limitted. This parameter is valid   only during initialization.
-- | --

wangyan42164:
@lakulako @dmitryermilov Thanks for your comments. I just found it when I investigated the submitted issue in media driver repo. I can try your solution. 
Thanks.
Yan Wang

wangyan42164:
@lakulako @dmitryermilov Sorry for slow response because of my another task. I tried to set BitrateLimit to OFF by the following FFMPEG command (-bitrate_limit:v 0) but no effect:

ffmpeg -hwaccel qsv -init_hw_device qsv=hw -filter_hw_device hw -v verbose -f rawvideo -video_size 352x288 -pix_fmt yuv420p -i ./MOBILE_352x288_30_orig_01.yuv -an -vf hwupload=extra_hw_frames=64,format=qsv -c:v h264_qsv -g 30 -bf 3 -r 30 -b:v 30k -maxrate:v 30k -bitrate_limit:v 0 -preset 4 -vframes 100 -y ./h264e_cbr_352x288_gop30_refdist3_bitrate100k.h264

I think MSDK may have some issues for BitrateLimit or FFMPEG may use it wrong.
I did some initial debugging:
in init_video_param() of libavcodec/qsvenc.c of FFMPEG, q->extco2.BitrateLimit is set rightly.
But in fact MfxHwH264Encode::CheckVideoParamQueryLike() will be called multiply times.
So the calculation of par.calcParam.targetKbps may have some problems.
Could you please give me some comments?
Thanks.
Yan Wang

stasefremov:
BitrateLimit is 3-state option: 0x00 - default (MSDK choice), 0x10 - ON (aka MFX_CODINGOPTION_ON), 0x20 - OFF (aka MFX_CODINGOPTION_OFF). Default is OFF. Is BitrateLimit set properly?

wangyan42164:
@stasefremov In init_video_param() of libavcodec/qsvenc.c of FFMPEG,
if (q->bitrate_limit >= 0)
   q->extco2.BitrateLimit = q->bitrate_limit ? MFX_CODINGOPTION_ON : MFX_CODINGOPTION_OFF;

I think it is right.

PR #808, opened 2018-10-23
[build] Add RUN_TESTS option to build config

vshampor:


CMakeLists.txt
```diff
@@ -109,6 +111,10 @@ if (BUILD_TESTS AND GTEST_FOUND)
   add_subdirectory(tests)
 endif()
 
+if (RUN_TESTS)
+  set(BUILD_TESTS ON)
```
dvrogozh: You don't need this. option() should set variable to OFF or to whatever you have specified manually.

tests/mfx_dispatch/linux/CMakeLists.txt
```diff
@@ -30,19 +30,25 @@ endif( )
 add_definitions( -DMFX_MODULES_DIR="${MFX_MODULES_DIR}" )
 add_definitions( -DMFX_PLUGINS_CONF_DIR="${MFX_PLUGINS_CONF_DIR}" )
 
+set( CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BIN_DIR}/${CMAKE_BUILD_TYPE})
```
dvrogozh: Try to avoid redefinition CMAKE_RUNTIME_OUTPUT_DIRECTORY or at least check whether it is defined by the user. This is cmake standard variable  https://cmake.org/cmake/help/v3.0/variable/CMAKE_RUNTIME_OUTPUT_DIRECTORY.html?highlight=cmake_runtime_output_directory

CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ cmake_dependent_option(BUILD_SAMPLES "Build samples?" ON "BUILD_DISPATCHER" OFF
 cmake_dependent_option(BUILD_TOOLS "Build tools?" "${BUILD_ALL}" "BUILD_SAMPLES" OFF)
 option(BUILD_TESTS "Build tests?" "${BUILD_ALL}")
 
+option(RUN_TESTS "Run tests during build?" OFF )
```
dvrogozh: You need cmake_dependent_option since you can't run tests if your don't build them:
`cmake_dependent_option(RUN_TESTS "Run tests?" OFF "BUILD_TESTS" OFF)`

tests/mfx_dispatch/linux/CMakeLists.txt
```diff
@@ -30,19 +30,25 @@ endif( )
...
 
-set_target_properties(mfx_dispatch_test PROPERTIES
-  RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BIN_DIR}/${CMAKE_BUILD_TYPE})
-
 target_link_libraries( mfx_dispatch_test mfx gtest gmock pthread )
 
 target_include_directories( mfx_dispatch_test PUBLIC ../../../api/mfx_dispatch/linux )
 
+if ( RUN_TESTS )
+  add_custom_target(run_mfx_dispatch_test ALL
```
dvrogozh: I would try to go with ctest if possible. Need to find out how to use it.

CMakeLists.txt
```diff
@@ -79,6 +79,13 @@ cmake_dependent_option(BUILD_SAMPLES "Build samples?" ON "BUILD_DISPATCHER" OFF
 cmake_dependent_option(BUILD_TOOLS "Build tools?" "${BUILD_ALL}" "BUILD_SAMPLES" OFF)
 option(BUILD_TESTS "Build tests?" "${BUILD_ALL}")
 
+set(IS_CMAKE3 OFF)
```
dvrogozh: Why do you need this? I see that both add_test() and enable_testing() present in cmake-2.8.5: https://cmake.org/cmake/help/v2.8.5/cmake.html. I think you can try cmake-2.8 on CentOS.... They have both packages cmake and cmake3...

tests/mfx_dispatch/linux/CMakeLists.txt
```diff
@@ -37,12 +37,18 @@ add_executable(mfx_dispatch_test
...
 
+target_link_libraries( mfx_dispatch_test mfx gtest gmock pthread )
+
+target_include_directories( mfx_dispatch_test PUBLIC ../../../api/mfx_dispatch/linux )
+
 set_target_properties(mfx_dispatch_test PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BIN_DIR}/${CMAKE_BUILD_TYPE})
 
-target_link_libraries( mfx_dispatch_test mfx gtest gmock pthread )
+add_test(NAME run_mfx_dispatch_test
```
dvrogozh: should this block be protected w/ if(RUN_TESTS)?

vshampor: This will only have effect if enable_testing() had been invoked, which is guarded w/ if (RUN_TESTS)

CMakeLists.txt
vshampor: Indeed, the whole setup works with cmake 2.8.12.2 at least.

tests/mfx_dispatch/linux/CMakeLists.txt
dvrogozh: ok.

CMakeLists.txt
dvrogozh: ok, update, please

azhogov: CMake 3 is needed for compiling the latest Googletest on CentOS (that is needed for the UT). See https://github.com/Intel-Media-SDK/MediaSDK/wiki/Build-Media-SDK-on-CentOS (Packages needed for Googletest)

dvrogozh: But as of now we don't build googletests within building mediasdk. So, this check is really not needed. Google test framework could be either build or installed on some systems - that's responsibility of the user.

CMakeLists.txt
```diff
@@ -105,10 +107,15 @@ if (BUILD_DISPATCHER)
...
 
+if(RUN_TESTS)
+    enable_testing()
+endif()
+
 if (BUILD_TESTS AND GTEST_FOUND)
   add_subdirectory(tests)
 endif()
 
+
```
dvrogozh: minor, but: please, remove this new line

PR #813, opened 2018-10-24
[smt] fixed hang in threads waiting mechanism

vilichev:
Old mechanism used pthread_cond_timedwait()/pthread_cond_signal() to wait till end
of all threads executions. pthread_cond_signal() issued only once at the end of each
thread execution but the signal has no effect if the main thread is not blocked on
corresponding pthread_cond_timedwait. As a result smt could hang in the infinite loop.

### DIFF
--- a/samples/sample_multi_transcode/include/pipeline_transcode.h
+++ b/samples/sample_multi_transcode/include/pipeline_transcode.h
@@ -25,14 +25,13 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "mfxplugin.h"
 #include "mfxplugin++.h"
 
-#include "vm/atomic_defs.h"
-#include "vm/thread_defs.h"
-
 #include <memory>
 #include <vector>
 #include <list>
 #include <ctime>
 #include <map>
+#include <future>
+#include <chrono>
 
 #include "sample_defs.h"
 #include "sample_utils.h"
@@ -88,8 +87,6 @@ const mfxF64  MCTF_LOSSLESS_BPP = 0.0;
 
 namespace TranscodingSample
 {
-    extern mfxU32 MFX_STDCALL TranscodeRoutine(void   *pObj);
-
     enum PipelineMode
     {
         Native = 0,        // means that pipeline is based depends on the cmd parameters (decode/encode/transcode)
@@ -910,28 +907,38 @@ namespace TranscodingSample
         // Pointer to the session's pipeline
         std::unique_ptr<CTranscodingPipeline> pPipeline;
         // Pointer to bitstream handling object
-        FileBitstreamProcessor *pBSProcessor;
+        FileBitstreamProcessor *pBSProcessor = nullptr;
         // Session implementation type
-        mfxIMPL implType;
+        mfxIMPL implType = MFX_IMPL_AUTO;
 
         // Session's starting status
-        mfxStatus startStatus;
+        mfxStatus startStatus = MFX_ERR_NONE;
         // Session's working time
-        mfxF64 working_time;
+        mfxF64 working_time = 0;
 
         // Number of processed frames
-        mfxU32 numTransFrames;
+        mfxU32 numTransFrames = 0;
         // Status of the finished session
-        mfxStatus transcodingSts;
+        mfxStatus transcodingSts = MFX_ERR_NONE;
 
-        ThreadTranscodeContext()
+        // Thread handle
+        std::future<void> handle;
+
+        void TranscodeRoutine()
         {
-            pBSProcessor = NULL;
-            implType = MFX_IMPL_AUTO;
-            startStatus = MFX_ERR_NONE;
-            working_time = 0;
-            numTransFrames = 0;
+            using namespace std::chrono;
+            MSDK_CHECK_POINTER_NO_RET(pPipeline);
             transcodingSts = MFX_ERR_NONE;
+
+            auto start_time = system_clock::now();
+            while (MFX_ERR_NONE == transcodingSts)
+            {
+                transcodingSts = pPipeline->Run();
+            }
+            working_time = duration_cast<microseconds>(system_clock::now() - start_time).count();
+
+            MSDK_IGNORE_MFX_STS(transcodingSts, MFX_WRN_VALUE_NOT_CHANGED);
+            numTransFrames = pPipeline->GetProcessFrames();
         }
     };
 }
--- a/samples/sample_multi_transcode/include/sample_multi_transcode.h
+++ b/samples/sample_multi_transcode/include/sample_multi_transcode.h
@@ -63,11 +63,8 @@ namespace TranscodingSample
 
         // command line parser
         CmdProcessor m_parser;
-        // sessions to process playlist
-        std::vector<ThreadTranscodeContext*> m_pSessionArray;
-        // handles
-        typedef std::vector<MSDKThread*>::iterator MSDKThreadsIterator;
-        std::vector<MSDKThread*>             m_HDLArray;
+        // threads contexts to process playlist
+        std::vector<ThreadTranscodeContext*> m_pThreadContextArray;
         // allocator for each session
         std::vector<GeneralAllocator*>       m_pAllocArray;
         // input parameters for each session
--- a/samples/sample_multi_transcode/src/pipeline_transcode.cpp
+++ b/samples/sample_multi_transcode/src/pipeline_transcode.cpp
@@ -46,28 +46,6 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 
 using namespace TranscodingSample;
 
-mfxU32 MFX_STDCALL TranscodingSample::TranscodeRoutine(void   *pObj)
-{
-    mfxU64 start = TranscodingSample::GetTick();
-    ThreadTranscodeContext *pContext = (ThreadTranscodeContext*)pObj;
-    pContext->transcodingSts = MFX_ERR_NONE;
-    for(;;)
-    {
-        while (MFX_ERR_NONE == pContext->transcodingSts)
-        {
-            pContext->transcodingSts = pContext->pPipeline->Run();
-        }
-        break; // exit loop
-    }
-
-    MSDK_IGNORE_MFX_STS(pContext->transcodingSts, MFX_WRN_VALUE_NOT_CHANGED);
-
-    pContext->working_time = TranscodingSample::GetTime(start);
-    pContext->numTransFrames = pContext->pPipeline->GetProcessFrames();
-
-    return 0;
-} // mfxU32 __stdcall TranscodeRoutine(void   *pObj)
-
 #ifdef ENABLE_MCTF
 namespace TranscodingSample
 {
--- a/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
+++ b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
@@ -29,6 +29,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #error MFX_VERSION not defined
 #endif
 
+#include <future>
 using namespace std;
 using namespace TranscodingSample;
 
@@ -305,26 +306,26 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
         pThreadPipeline->startStatus = MFX_WRN_DEVICE_BUSY;
         // set other session's parameters
         pThreadPipeline->implType = m_InputParamsArray[i].libType;
-        m_pSessionArray.push_back(pThreadPipeline.release());
+        m_pThreadContextArray.push_back(pThreadPipeline.release());
 
         mfxVersion ver = {{0, 0}};
-        sts = m_pSessionArray[i]->pPipeline->QueryMFXVersion(&ver);
-        MSDK_CHECK_STATUS(sts, "m_pSessionArray[i]->pPipeline->QueryMFXVersion failed");
+        sts = m_pThreadContextArray[i]->pPipeline->QueryMFXVersion(&ver);
+        MSDK_CHECK_STATUS(sts, "m_pThreadContextArray[i]->pPipeline->QueryMFXVersion failed");
 
         PrintInfo(i, &m_InputParamsArray[i], &ver);
     }
 
     for (i = 0; i < m_InputParamsArray.size(); i++)
     {
-        sts = m_pSessionArray[i]->pPipeline->CompleteInit();
-        MSDK_CHECK_STATUS(sts, "m_pSessionArray[i]->pPipeline->CompleteInit failed");
+        sts = m_pThreadContextArray[i]->pPipeline->CompleteInit();
+        MSDK_CHECK_STATUS(sts, "m_pThreadContextArray[i]->pPipeline->CompleteInit failed");
 
-        if (m_pSessionArray[i]->pPipeline->GetJoiningFlag())
+        if (m_pThreadContextArray[i]->pPipeline->GetJoiningFlag())
             msdk_printf(MSDK_STRING("Session %d was joined with other sessions\n"), i);
         else
             msdk_printf(MSDK_STRING("Session %d was NOT joined with other sessions\n"), i);
 
-        m_pSessionArray[i]->pPipeline->SetPipelineID(i);
+        m_pThreadContextArray[i]->pPipeline->SetPipelineID(i);
     }
 
     msdk_printf(MSDK_STRING("\n"));
@@ -341,7 +342,7 @@ void Launcher::Run()
     m_StartTime = GetTick();
 
     // Robust flag is applied to every seession if enabled in one
-    if (m_pSessionArray[0]->pPipeline->GetRobustFlag())
+    if (m_pThreadContextArray[0]->pPipeline->GetRobustFlag())
     {
         DoRobustTranscoding();
     }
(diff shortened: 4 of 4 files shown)


samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,36 +359,28 @@ void Launcher::DoTranscoding()
...
     mfxStatus sts = MFX_ERR_NONE;
 
-    // get parallel sessions parameters
-    mfxU32 totalSessions = (mfxU32)m_pSessionArray.size();
-    MSDKThread* pthread = 0;
-
-    for (mfxU32 i = 0; i < totalSessions; i++)
-    {
-        pthread = new MSDKThread(sts, TranscodeRoutine, (void *)m_pSessionArray[i]);
-        m_HDLArray.push_back(pthread);
```
mgonchar: remove m_HDLArray completely, since you don't use it anymore

vilichev: Done

samples/sample_multi_transcode/include/sample_multi_transcode.h
```diff
@@ -67,7 +67,6 @@ namespace TranscodingSample
         std::vector<ThreadTranscodeContext*> m_pSessionArray;
         // handles
         typedef std::vector<MSDKThread*>::iterator MSDKThreadsIterator;
```
alariono: Does this still required?

vilichev: Agree. Fixed it

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,36 +359,28 @@ void Launcher::DoTranscoding()
...
     // Transcoding threads waiting cycle
+    size_t totalSessions = m_pSessionArray.size();
     size_t aliveSessionsCount = totalSessions;
     while (aliveSessionsCount)
     {
         aliveSessionsCount = 0;
-        for (mfxU32 i = 0; i < totalSessions; i++)
+        size_t aliveOverlaySessionsCount = 0;
+
+        for (size_t i = 0; i < totalSessions && thrHandles[i].valid(); i++)
```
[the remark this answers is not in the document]
vilichev: Agree

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -409,39 +406,39 @@ void Launcher::DoTranscoding()
...
             }
         }
 
         // If all sessions are already stopped (no matter with error or not) - we need to forcibly stop all overlay sessions
-        if (aliveSessionsCount <= nOverlayThreads + 1)
+        if (aliveOverlaySessionsCount && aliveSessionsCount <= aliveOverlaySessionsCount + 1)
         {
             // Sending stop message
-            for (size_t i = 0; i < totalSessions; i++)
+            for (size_t i = 0; i < totalSessions && thrHandles[i].valid(); i++)
```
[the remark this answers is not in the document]
vilichev: Agree

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,36 +359,32 @@ void Launcher::DoTranscoding()
...
-        m_HDLArray.push_back(pthread);
-    }
-
-    // Need to determine overlay threads count first
-    size_t nOverlayThreads = 0;
-    for (size_t i = 0; i < m_pSessionArray.size(); i++)
-    {
-        if (m_pSessionArray[i]->pPipeline->IsOverlayUsed())
-            nOverlayThreads++;
+        thrHandles.emplace_back(std::async(std::launch::async,
```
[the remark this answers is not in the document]
mgonchar: Yes, I agree. If we decided to simplify this function, I would point some other things below

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,36 +359,32 @@ void Launcher::DoTranscoding()
...
+                                TranscodeRoutine, (void *)session));
     }
 
     // Transcoding threads waiting cycle
+    size_t totalSessions = m_pSessionArray.size();
```
mgonchar: I don't think we need another alias for size of vector, it just overcomplicates the loop below. I suggest to use ```m_pSessionArray.size()``` instead

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,36 +359,32 @@ void Launcher::DoTranscoding()
...
-            {
-                aliveSessionsCount++;
+            if (!thrHandles[i].valid())
+                continue;
 
-                sts = m_HDLArray[i]->TimedWait(1);
+            auto waitSts = thrHandles[i].wait_for(std::chrono::milliseconds(1));
+            if (waitSts == std::future_status::ready)
+            {
+                auto sts = thrHandles[i].get();
```
mgonchar: this one would be zero anyway... see return of
```
mfxU32 MFX_STDCALL TranscodingSample::TranscodeRoutine(void *pObj)
```
also, returned status of thread routine here doesn't make any sense, because we track errors by transcodingSts. I suggest to make ```TranscodeRoutine``` return ```void```

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -409,21 +410,29 @@ void Launcher::DoTranscoding()
                             }
                         }
                     }
-
-                    // Now clear its handle and thread info
-                    MSDK_SAFE_DELETE(m_HDLArray[i]);
-                    m_HDLArray[i] = NULL;
+                }
```
mgonchar: I believe here we need
```
else if (m_pSessionArray[i]->transcodingSts > MFX_ERR_NONE)
{
    msdk_stringstream ss;
    ss << MSDK_STRING("\n\n session ") << i << MSDK_STRING(" [")
    << m_pSessionArray[i]->pPipeline->GetSessionText()
    << MSDK_STRING("] returned warning status ")
    << StatusToString(m_pSessionArray[i]->transcodingSts)
    << std::endl << std::endl;
    msdk_printf(MSDK_STRING("%s"), ss.str().c_str());
}
```

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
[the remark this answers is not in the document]
vilichev: Done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
vilichev: Done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
vilichev: Done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -432,16 +441,13 @@ void Launcher::DoTranscoding()
             // Waiting for them to be stopped
             for (size_t i = 0; i < totalSessions; i++)
             {
-                if (m_HDLArray[i])
-                {
-                    m_HDLArray[i]->Wait();
-                    MSDK_SAFE_DELETE(m_HDLArray[i]);
-                    m_HDLArray[i]=NULL;
-                }
+                if (!thrHandles[i].valid())
```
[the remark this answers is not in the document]
vilichev: Done. Except cases where the index is required

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
vilichev: Done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,90 +359,112 @@ void Launcher::DoTranscoding()
...
-    MSDKThread* pthread = 0;
-
-    for (mfxU32 i = 0; i < totalSessions; i++)
+    // Helper structure to manage transcoding threads/sessions
+    struct Handles
     {
-        pthread = new MSDKThread(sts, TranscodeRoutine, (void *)m_pSessionArray[i]);
-        m_HDLArray.push_back(pthread);
-    }
+        std::future<void>       thrh;
```
dmitry-gurulev: [thrh] - what does this name means?

vilichev: 'thread handle' as @dvrogozh proposed. I can rename 'thrh' -> 'thrHandle'. Will be it OK?

dmitry-gurulev: But this is [future] not thread handle.

vilichev: The future handles the return from the thread. It is common naming (it is used in samples for std::async). But I can rename it

samples/sample_multi_transcode/include/pipeline_transcode.h
```diff
@@ -88,7 +85,7 @@ const mfxF64  MCTF_LOSSLESS_BPP = 0.0;
 
 namespace TranscodingSample
 {
-    extern mfxU32 MFX_STDCALL TranscodeRoutine(void   *pObj);
+    extern void MFX_STDCALL TranscodeRoutine(void   *pObj);
```
dvrogozh: 1. Do we really need this extern declaration here?
2. I don't think we need MFX_STDCALL anymore.

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -46,7 +46,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 
 using namespace TranscodingSample;
 
-mfxU32 MFX_STDCALL TranscodingSample::TranscodeRoutine(void   *pObj)
+void MFX_STDCALL TranscodingSample::TranscodeRoutine(void   *pObj)
```
dvrogozh: I believe we should be able to change this function from C-style to C++ member of ThreadTranscodeContext and remove some explicit casts....

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,90 +359,112 @@ void Launcher::DoTranscoding()
...
     // Transcoding threads waiting cycle
-    size_t aliveSessionsCount = totalSessions;
+    size_t aliveSessionsCount = handles.size();
     while (aliveSessionsCount)
     {
         aliveSessionsCount = 0;
-        for (mfxU32 i = 0; i < totalSessions; i++)
+        size_t aliveOverlaySessionsCount = 0;
+
+        for (size_t i = 0; i < handles.size(); i++)
```
dvrogozh: I think the idea with handles was to have `for (auto& handle: handles)` here. We can't do that for the reason I miss?

vilichev: Session index is used to print the session number in case of error/warnings (l.407-412, l.424-429)

samples/sample_multi_transcode/include/pipeline_transcode.h
vilichev: Agree. extern and MFX_STDCALL are useless here

samples/sample_multi_transcode/src/pipeline_transcode.cpp
vilichev: Fixed explicit cast. Not sure about to make this function as a member (in this case it needs to make the function to be a static)

dvrogozh: No, you should be able to bypass static via lambdas and `this` capturing. That's not exactly what you need, but here is an idea:
`[this](void* p){ TranscodeRoutine(p); }`
or something similar should make a trick. I think @mgonchar will be able to guide.

vilichev: OK, seems I got it. Done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,90 +359,98 @@ void Launcher::DoTranscoding()
...
         {
-            if (m_HDLArray[i])
-            {
-                aliveSessionsCount++;
+            if (!m_pThreadContextArray[i]->handle.valid())
+                continue;
 
-                sts = m_HDLArray[i]->TimedWait(1);
-                if (sts <= 0)
+            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(1));
```
dvrogozh: I am starting to feel that this should be an Wait() API of ThreadTranscodeContext().

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,90 +359,98 @@ void Launcher::DoTranscoding()
...
         // If all sessions are already stopped (no matter with error or not) - we need to forcibly stop all overlay sessions
-        if (aliveSessionsCount <= nOverlayThreads + 1)
+        if (aliveOverlaySessionsCount && aliveSessionsCount <= aliveOverlaySessionsCount + 1)
         {
             // Sending stop message
-            for (size_t i = 0; i < totalSessions; i++)
+            for (auto context : m_pThreadContextArray)
             {
-                if (m_HDLArray[i] && m_pSessionArray[i]->pPipeline->IsOverlayUsed())
+                if (!context->handle.valid())
```
dvrogozh: Separate Stop() API for ThreadTranscodeContext?

samples/sample_multi_transcode/include/pipeline_transcode.h
```diff
@@ -910,28 +907,44 @@ namespace TranscodingSample
...
+        mfxU32 numTransFrames = 0;
         // Status of the finished session
-        mfxStatus transcodingSts;
+        mfxStatus transcodingSts = MFX_ERR_NONE;
+
+        // Thread handle
+        std::future<void> handle;
 
-        ThreadTranscodeContext()
+        void TranscodeRoutine()
```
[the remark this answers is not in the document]
mgonchar: @dvrogozh you mean that we can get rid of ```RunTranscodeRoutine()``` in a way like:
```
// Run transcoding threads
        auto RunTranscodeRoutine = [](ThreadTranscodeContext* context)
        {
            context->handle = std::async(std::launch::async, [context](){ context->TranscodeRoutine(); });
        }

    for (auto context : m_pThreadContextArray)
    {
        MSDK_CHECK_POINTER_NO_RET(context);
        RunTranscodeRoutine(context);
    }
```

or something different?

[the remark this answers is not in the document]
dvrogozh: Well, I meant that we need to separate class public and private API, but I think I like your variant.

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
mgonchar: It would be tiny wrapper, it should accept amount in ms, so in fact it will decrease
```auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(1));```
to
```auto waitSts = m_pThreadContextArray[i]->Wait(std::chrono::milliseconds(1));```

So it's nice to have but hardly worth it because there should be additional checks inside to be sure that future is not in detach state and so on

File: samples/sample_multi_transcode/src/sample_multi_transcode.cpp (hunk 15 above)
Comment (mgonchar) [reply samples/sample_multi_transcode/src/sample_multi_transcode.cpp]: StopSession is already a Stop API for ThreadTranscodeContext and it looks pretty nice (see lines 401-404). This one is special case for overlay sessions. It's not clear for me if we need it in API

**Comment by mgonchar** [comment]
@dvrogozh as for me I still don't like that this loop is so complicated. It does pretty simple thing and should be simple. I'm not sure that further simplification should go through API change for ThreadContext. I will think about how we can simplify it by playing with loop conditions and so on. Maybe we don't need alive session counters

File: samples/sample_multi_transcode/src/sample_multi_transcode.cpp [hunk 17]
```diff
@@ -358,90 +359,98 @@ void Launcher::DoTranscoding()
 {
     mfxStatus sts = MFX_ERR_NONE;
```
mgonchar: unused variable

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,90 +359,98 @@ void Launcher::DoTranscoding()
...
 
-    // get parallel sessions parameters
-    mfxU32 totalSessions = (mfxU32)m_pSessionArray.size();
-    MSDKThread* pthread = 0;
-
-    for (mfxU32 i = 0; i < totalSessions; i++)
+    // Run transcoding threads
```
mgonchar: this one is obvious comment, we don't need it. function name is telling whats happening

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -358,90 +359,98 @@ void Launcher::DoTranscoding()
...
-                if (m_HDLArray[i] && m_pSessionArray[i]->pPipeline->IsOverlayUsed())
+                if (!context->handle.valid())
+                    continue;
+
+                auto waitSts = context->handle.wait_for(std::chrono::milliseconds(1));
+                if (waitSts != std::future_status::ready &&
+                    context->pPipeline->IsOverlayUsed())
                 {
-                    m_pSessionArray[i]->pPipeline->StopSession();
+                    context->pPipeline->StopSession();
```
mgonchar: @vilichev could you check the following thing, we already sent Stop to all sessions at line 403 (no matter overlay or not). Why do we need to do it once more? Looks like there is no need for it

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
vilichev: Fixed

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
vilichev: As we agreed, we need to forcibly stop of overlay sessions since they are never stopped by themselves. Thanks for your help @mgonchar !

samples/sample_multi_transcode/include/pipeline_transcode.h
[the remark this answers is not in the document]
vilichev: Done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -341,7 +342,7 @@ void Launcher::Run()
     m_StartTime = GetTick();
 
     // Robust flag is applied to every seession if enabled in one
-    if (m_pSessionArray[0]->pPipeline->GetRobustFlag())
+    if (m_pThreadContextArray[0]->pPipeline->GetRobustFlag())
```
mgonchar: I believe here should be actually a result of std::any_of, but this should be another change

mgonchar:
@dvrogozh @dmitry-gurulev please look at the latest patch

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -829,31 +844,31 @@ mfxStatus Launcher::CreateSafetyBuffers()
 
 void Launcher::Close()
 {
-    while(m_pSessionArray.size())
+    while(m_pThreadContextArray.size())
```
dvrogozh: I don't like all this code in a Close(). I think that all that should be handled something like:
```
std::vector<std::unique_ptr<ThreadTranscodeContext>> m_pThreadContextArray;
std::vector<std::unique_ptr<GeneralAllocator>>       m_pAllocArray;

Close()
{
  m_pThreadContextArray.clear();
  m_pAllocArray.clear();
  ...
}
```
But that's definitely out of scope of this commit and this PR. If you guys agree I suggest @vilichev to implement the change after this PR will be merged.

dvrogozh:
@dmitry-gurulev : I am good with this change. Your turn.

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
mgonchar: this code releases memory in reverse order... not sure if there is any sense in this

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -829,31 +844,31 @@ mfxStatus Launcher::CreateSafetyBuffers()
...
     {
-        delete m_pSessionArray[m_pSessionArray.size()-1];
-        m_pSessionArray[m_pSessionArray.size() - 1] = NULL;
-        m_pSessionArray.pop_back();
+        delete m_pThreadContextArray[m_pThreadContextArray.size()-1];
+        m_pThreadContextArray[m_pThreadContextArray.size() - 1] = nullptr;
```
dmitry-gurulev: 1 .Whether it makes sense when it's followed by
`m_pThreadContextArray.pop_back(); `
?
2. I'd rather use
`m_pThreadContextArray.back()`
instead of 
`m_pThreadContextArray[m_pThreadContextArray.size()-1]`
and
`while (!m_pThreadContextArray.empty())`
instead of
`while(m_pThreadContextArray.size())`

BTW, functional approach might be even better -
```
std::for_each(std::rbegin(m_pThreadContextArray), std::rend(m_pThreadContextArray), std::default_delete<ThreadTranscodeContext>{});
m_pThreadContextArray.clear();
```

mgonchar: absolutely agree

PR #827, opened 2018-10-26
Minor clean up

mgonchar:
Some general code improvements:

- commented code removed
- code alignment improved
- replaced element-wise array copy with std::copy

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
@@ -688,7 +688,7 @@ void PackerVA::PackProcessingInfo(H264DecoderFrameInfo * sliceInfo)
         throw h264_exception(UMC_ERR_FAILED);
 
     UMCVACompBuffer *pipelineVABuf;
-    VAProcPipelineParameterBuffer* pipelineBuf = (VAProcPipelineParameterBuffer*)m_va->GetCompBuffer(VAProcPipelineParameterBufferType, &pipelineVABuf, sizeof(VAProcPipelineParameterBuffer));
+    auto* pipelineBuf = reinterpret_cast<VAProcPipelineParameterBuffer *>(m_va->GetCompBuffer(VAProcPipelineParameterBufferType, &pipelineVABuf, sizeof(VAProcPipelineParameterBuffer)));
     if (!pipelineBuf)
         throw h264_exception(UMC_ERR_FAILED);
     pipelineVABuf->SetDataSize(sizeof(VAProcPipelineParameterBuffer));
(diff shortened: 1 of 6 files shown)


_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_dec.cpp
```diff
@@ -649,61 +649,61 @@ MPEG2VideoDecoderBase::~MPEG2VideoDecoderBase()
...
-    sequenceHeader.is_decoded = false;
-    sequenceHeader.first_i_occure  = 0;
-    sequenceHeader.first_p_occure  = 0;
-    sequenceHeader.broken_link = 0;
-    sequenceHeader.closed_gop  = 0;
-    sequenceHeader.time_code.gop_picture = 0;
-    sequenceHeader.time_code.gop_seconds = 0;
-    sequenceHeader.time_code.gop_minutes = 0;
-    sequenceHeader.time_code.gop_hours   = 0;
+    sequenceHeader.is_decoded                    = false;
```
dvrogozh: These kind of change, while beautiful, significantly complicate cherry picking across branches. I would not do them at all and focused on real changes. You have few of them below.

So, I strongly suggest to do beautification changes only across +/- 2-3 lines around the place you really changed.

Also, I personally don't like this particular change with aligning =. I prefer to have just single space around it otherwise you will have troubles adding/deleting members - you will be forced to realign the padding.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_dec_slice_hw.cpp
```diff
@@ -386,10 +386,9 @@ Status MPEG2VideoDecoderHW::PostProcessFrame(int display_index, int task_num)
                            + pack_w.pSliceInfo[-1].slice_data_size;
             if (pack_w.bs_size >= pack_w.bs_size_getting)
             {
-                VASliceParameterBufferMPEG2  s_info;
+                VASliceParameterBufferMPEG2 s_info = pack_w.pSliceInfo[-1];
```
dvrogozh: that's real one.

_studio/shared/umc/codec/vp9_dec/src/umc_vp9_va_packer.cpp
```diff
@@ -73,7 +73,7 @@ void PackerVA::PackAU(VP9Bitstream* bs, VP9DecoderFrame const* info)
     if (!bs || !info)
         throw vp9_exception(MFX_ERR_NULL_PTR);
 
-    UMC::UMCVACompBuffer* pCompBuf = NULL;
+    UMC::UMCVACompBuffer* pCompBuf = nullptr;
```
dvrogozh: and these...

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_dec.cpp
mgonchar: @dvrogozh ok, I will remove the unecessary difference

mgonchar:
@dvrogozh I think I addressed your requests

_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
```diff
@@ -702,27 +702,21 @@ void PackerVA::PackProcessingInfo(H264DecoderFrameInfo * sliceInfo)
...
-    for(j = 0; j < 6; j++)
+    int32_t j;
+
+    for(j = 0; j < 6; ++j)
     {
-        for(i = 0; i < 16; i++)
-        {
-             pQmatrix_H264->ScalingList4x4[j][i] = scaling->ScalingLists4x4[j].ScalingListCoeffs[i];
-        }
+        std::copy(std::begin(scaling->ScalingLists4x4[j].ScalingListCoeffs), std::end(scaling->ScalingLists4x4[j].ScalingListCoeffs), std::begin(pQmatrix_H264->ScalingList4x4[j]));
```
dvrogozh: I hope you ran some tests after this change...

lakulako: Better don't change here. It is different structures. And code is increased.

dvrogozh: @lakulako: I am not sure what do you mean... You accepted changes and code _is_ changed...

mgonchar: @dvrogozh I did manual unit testing for this change

lakulako: It is correct change for now. Now type of ScalingListCoeffs matches ScalingLists4x4[j]. But what if one of them is changed, for example to U**16**[]? Will it generate any warning?

mgonchar: if there would be any change that cause conversion with possible loss of data it will generate warning, of cource

PR #825, opened 2018-10-26
[VPP] black tiled video wall issue

IlyaSergeyev:
Issue #803 

Driver should not use colorfill for tiled video wall (works on APL)

Upd:
        m_pipelineParam[i].pipeline_flags |= VA_PROC_PIPELINE_SUBPICTURES;
        m_pipelineParam[i].filter_flags |= VA_FILTER_SCALING_HQ;

works for both: APL, CFL and SKL

PR #831, opened 2018-10-29
[VPP] runtime hardware detection for Execute_composition

IlyaSergeyev:
Separated from #825

Removing compile-time APL platform detection


IlyaSergeyev:
Needs to be merged with #825

PR #832, opened 2018-10-29
Fix LumaIsNull for y410/non-y410 case

vsidorow:


_studio/shared/include/mfx_utils.h
```diff
@@ -99,9 +99,16 @@ mfxU64 GetMfxTimeStamp(mfxF64 ts)
 }
 
 inline
-bool LumaIsNull(mfxFrameSurface1 * surf)
+bool LumaIsNull(const mfxFrameSurface1 * surf)
 {
-    return !surf->Data.Y && !surf->Data.Y410;
+    if (surf->Info.FourCC == MFX_FOURCC_Y410)
+    {
+        return !surf->Data.Y && !surf->Data.Y410;
```
dmitry-gurulev: [Data.Y] is allowed to be null for Y410

_studio/shared/include/mfx_utils.h
```diff
@@ -99,9 +99,16 @@ mfxU64 GetMfxTimeStamp(mfxF64 ts)
 }
 
 inline
-bool LumaIsNull(mfxFrameSurface1 * surf)
+bool LumaIsNull(const mfxFrameSurface1 * surf)
```
mgonchar: actually, here you also need a pointer check. But signature of this function doesn't allow to handle it natively, so I suggest to change signature of function to accept const reference to surface

dmitry-gurulev:
@vsidorow, consider using of [_studio/mfx_lib/shared/include/mfx_common_int.h :: GetFramePointer]

PR #835, opened 2018-10-29
Fix build-time Dispatcher discovery

onabiull:
Issue: MDP-51381
Test: build

### DIFF
--- a/builder/FindMFX.cmake
+++ b/builder/FindMFX.cmake
@@ -39,16 +39,10 @@ else()
 endif()
 
 find_path( MFX_INCLUDE mfxdefs.h PATHS ${MFX_API_HOME}/include NO_CMAKE_FIND_ROOT_PATH )
-find_library( MFX_LIBRARY libmfx.a PATHS ${MFX_API_HOME}/lib PATH_SUFFIXES ${os_arch} )
 
 if( NOT MFX_INCLUDE MATCHES NOTFOUND )
   include_directories( ${MFX_API_HOME}/mediasdk_structures )
   include_directories( ${MFX_INCLUDE} )
-endif()
-
-if( NOT MFX_LIBRARY MATCHES NOTFOUND )
-  get_filename_component( MFX_LIBRARY_PATH ${MFX_LIBRARY} PATH )
-  link_directories( ${MFX_LIBRARY_PATH} )
 endif()
 
 if( NOT MFX_INCLUDE MATCHES NOTFOUND )
@@ -60,11 +54,6 @@ if( NOT DEFINED MFX_FOUND )
   message( FATAL_ERROR "Intel(R) Media SDK was not found (required)! Set/check MFX_HOME environment variable!")
 else()
   message( STATUS "Intel(R) Media SDK was found here ${MFX_HOME}")
-endif()
-
-if( NOT MFX_LIBRARY MATCHES NOTFOUND )
-  get_filename_component(MFX_LIBRARY_PATH ${MFX_LIBRARY} PATH )
-  link_directories( ${MFX_LIBRARY_PATH} )
 endif()
 
 # Potential source of confusion here. Environment $MFX_VERSION translates to product name (strings libmfxhw64.so | grep mediasdk),
--- a/builder/FindPackages.cmake
+++ b/builder/FindPackages.cmake
@@ -145,7 +145,10 @@ function(configure_target target cflags libdirs)
 endfunction()
 
 function(configure_libmfx_target target)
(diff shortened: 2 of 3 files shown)


builder/FindPackages.cmake
```diff
@@ -144,8 +144,13 @@ function(configure_target target cflags libdirs)
   endforeach()
 endfunction()
 
-function(configure_libmfx_target target)
-  configure_target(${ARGV0} "${LIBMFX_CFLAGS}" "${LIBMFX_LIBRARY_DIRS}")
+function(configure_libmfx_target atarget)
```
dvrogozh: do you really need s/target/atarget/?

builder/FindPackages.cmake
```diff
@@ -144,8 +144,13 @@ function(configure_target target cflags libdirs)
...
+    # prefer local Dispatcher sources to preinstalled
```
dvrogozh: what's wrong with the alignment?

builder/FindPackages.cmake
```diff
@@ -488,6 +493,7 @@ if( Linux )
   # optional:
   pkg_check_modules( PKG_X11       x11 )
   pkg_check_modules( PKG_LIBVA_X11 libva-x11>=0.33 )
+  pkg_check_modules( PKG_MFX       mfx>=1.28 )
```
dvrogozh: better to use "libmfx" instead of "mfx" since the later on is a symlink, i.e. legacy back compatibility...

builder/FindPackages.cmake
```diff
@@ -144,8 +144,13 @@ function(configure_target target cflags libdirs)
   endforeach()
 endfunction()
 
-function(configure_libmfx_target target)
-  configure_target(${ARGV0} "${LIBMFX_CFLAGS}" "${LIBMFX_LIBRARY_DIRS}")
+function(configure_libmfx_target atarget)
+    # prefer local Dispatcher sources to preinstalled
+  if (TARGET mfx)
+     configure_target(${ARGV0} "${LIBMFX_CFLAGS}" "${LIBMFX_LIBRARY_DIRS}")
```
dvrogozh: I don't see that we setup these variables anywhere:
```
[dvrogozh@dvrscl MediaSDK]$ fgrep -rsn LIBMFX_CFLAGS
builder/FindPackages.cmake:148:  configure_target(${ARGV0} "${LIBMFX_CFLAGS}" "${LIBMFX_LIBRARY_DIRS}")
[dvrogozh@dvrscl MediaSDK]$ fgrep -rsn LIBMFX_LIBRARY_DIRS
builder/FindPackages.cmake:148:  configure_target(${ARGV0} "${LIBMFX_CFLAGS}" "${LIBMFX_LIBRARY_DIRS}")
[dvrogozh@dvrscl MediaSDK]$
```

Should not the code be just
if (NOT TARGET mfx)
  configure_target(${ARGV0} "${PKG_MFX_CFLAGS}" "${PKG_MFX_LIBRARY_DIRS}")
endif()

builder/FindMFX.cmake
```diff
@@ -39,18 +39,12 @@ else()
 endif()
 
 find_path( MFX_INCLUDE mfxdefs.h PATHS ${MFX_API_HOME}/include NO_CMAKE_FIND_ROOT_PATH )
```
dvrogozh: It seems that there should be (or could be?) a part in this file which is not needed if we take dispatcher from the sources... I.e. the part protected by `if (PKG_MFX_FOUND)`. Maybe that's not for this change, but I would feel more comfortable having this part clearly separated.

builder/FindPackages.cmake
dvrogozh: ah, that's the reason. Ok then.

builder/FindPackages.cmake
onabiull: will fix

builder/FindPackages.cmake
onabiull: ok

builder/FindPackages.cmake
onabiull: I was sure that add_library(mfx) sets them. But in practice - looks they always empty. So ok.

builder/FindPackages.cmake
dvrogozh: so, you will fix that?

builder/FindPackages.cmake
dvrogozh: I am not sure I understood: you will drop this?

PR #838, opened 2018-10-30
[vp9e] Fix std::copy applications

vshampor:
The source and destination static arrays must have the same type


mgonchar:
@vshampor do not forget to double it to closed source

PR #843, opened 2018-10-30
[build] Remove RUN_TESTS option

vshampor:
The option was deemed redundant - BUILD_TESTS does the job already.


CMakeLists.txt
```diff
@@ -110,11 +108,8 @@ if (BUILD_RUNTIME)
   add_subdirectory(${CMAKE_HOME_DIRECTORY}/_studio)
 endif()
 
-if(RUN_TESTS)
-    enable_testing()
-endif()
-
-if (BUILD_TESTS AND GTEST_FOUND)
+if (BUILD_TESTS)
```
dvrogozh: That's correct. Do _not_ add GTEST_FOUND here. If someone configures build with BUILD_TESTS=ON he should get configuration error  much earlier (from FindGTest.cmake at the moment). Maybe that's not exactly so at the moment, but that's where we are heading to.

PR #845, opened 2018-10-30
[AVCe, HEVCe] Fix for default MaxQPI, MaxQPP, MaxQPB

ogolikov:
Issue: MDP-50776
Test:  manually (behavior is needed)

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
@@ -2908,6 +2908,23 @@ void SetDefaults(
     if (IsOff(CO3.EnableQPOffset))
         Zero(CO3.QPOffset);
 
+    if (CO2.MinQPI || CO2.MinQPP || CO2.MinQPB ||
+        CO2.MaxQPI || CO2.MaxQPP || CO2.MaxQPB)
+    {
+        if (!CO2.MinQPI)
+            CO2.MinQPI = (mfxU8)minQP;
+        if (!CO2.MinQPP)
+            CO2.MinQPP = (mfxU8)minQP;
+        if (!CO2.MinQPB)
+            CO2.MinQPB = (mfxU8)minQP;
+
+        if (!CO2.MaxQPI)
+            CO2.MaxQPI = (mfxU8)maxQP;
+        if (!CO2.MaxQPP)
+            CO2.MaxQPP = (mfxU8)maxQP;
+        if (!CO2.MaxQPB)
+            CO2.MaxQPB = (mfxU8)maxQP;
+    }
 
 
 #if (MFX_VERSION >= 1026)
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -5703,6 +5703,16 @@ void MfxHwH264Encode::SetDefaults(
     if (extOpt2->RepeatPPS == MFX_CODINGOPTION_UNKNOWN)
         extOpt2->RepeatPPS = MFX_CODINGOPTION_ON;
 
+    if (extOpt2->MinQPI || extOpt2->MinQPP || extOpt2->MinQPB ||
+        extOpt2->MaxQPI || extOpt2->MaxQPP || extOpt2->MaxQPB)
+    {
+        if (!extOpt2->MaxQPI)
```
mgonchar: why don't you change mins as well?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -2908,6 +2908,23 @@ void SetDefaults(
     if (IsOff(CO3.EnableQPOffset))
         Zero(CO3.QPOffset);
 
+    if (CO2.MinQPI || CO2.MinQPP || CO2.MinQPB ||
```
mgonchar: why do we need this condition?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
ogolikov: question isn't clear for me

mgonchar: I mean don't we need following code
```
if (!extOpt2->MinQPI)
    extOpt2->MinQPI = 1;
...
```

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
lakulako: Don't refer internal tools.
Most likely "if any is initialized, other should be initialized too".

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -5703,6 +5703,16 @@ void MfxHwH264Encode::SetDefaults(
...
+           extOpt2->MaxQPI = 51;
```
lakulako: Will it work for >8bit samples?

ogolikov: there is no >8bits support in h264 encoder

mgonchar:
@ogolikov the question is still there, why don't you set minimum QP to 1 inside ```void MfxHwH264Encode::SetDefaults``` ? You did it for hevc

mgonchar:
ok, I checked driver code, so they are clamping QPs to range [1, 51], see https://github.com/intel/media-driver/blob/bc5190beb1f538f239b00ba3b4aa222fe7f4ee7d/media_driver/agnostic/common/codec/hal/codechal_vdenc_avc.cpp#L3212

PR #846, opened 2018-10-30
[AVCe] fix for frameType in DisplayOrder

ogolikov:
Issue: MDP-49070
Test:  manually

adding IDR or REF attribute
if it is needed.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
@@ -2130,6 +2130,24 @@ struct FindNonSkip
 
     mfxU16 m_skipFrameMode;
 };
+mfxStatus FixForcedFrameType (DdiTask & newTask, mfxU32 frameOrder)
+{
+    if (frameOrder == 0 &&
+        !(newTask.m_type[0] & MFX_FRAMETYPE_IDR) &&
+        !(newTask.m_type[1] & MFX_FRAMETYPE_IDR))
+    {
+        if (newTask.m_type[0] & MFX_FRAMETYPE_I)
+            newTask.m_type[0] |= MFX_FRAMETYPE_IDR;
+        else if (newTask.m_type[1] & MFX_FRAMETYPE_I)
+            newTask.m_type[1] |= MFX_FRAMETYPE_IDR;
+        else
+            return MFX_ERR_UNDEFINED_BEHAVIOR;
+    }
+    newTask.m_type[0] |= MFX_FRAMETYPE_REF;
+    newTask.m_type[1] |= MFX_FRAMETYPE_REF;
+
+    return MFX_ERR_NONE;
+}
 
 void ImplementationAvc::AssignFrameTypes(DdiTask & newTask)
 {
@@ -2250,6 +2268,12 @@ mfxStatus ImplementationAvc::AsyncRoutine(mfxBitstream * bs)
             {
                 if (newTask.m_type[0] == 0)
                     newTask.m_type = GetFrameType(m_video, m_frameOrder - m_frameOrderIdrInDisplayOrder);
+                else
+                {
+                    mfxStatus sts = FixForcedFrameType(newTask, m_frameOrder - m_frameOrderIdrInDisplayOrder);
+                    if (sts != MFX_ERR_NONE)
+                        return Error(sts);
+                }
 
                 newTask.m_frameOrder = m_frameOrder;
                 AssignFrameTypes(newTask);


_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2130,6 +2130,26 @@ struct FindNonSkip
...
+        !(newTask.m_type[1] & MFX_FRAMETYPE_IDR))
+    {
+        if (newTask.m_type[0] & MFX_FRAMETYPE_I)
+            newTask.m_type[0] = newTask.m_type[0] | MFX_FRAMETYPE_IDR;
+        else if (newTask.m_type[1] & MFX_FRAMETYPE_I)
+            newTask.m_type[1] = newTask.m_type[1] | MFX_FRAMETYPE_IDR;
+        else
+            return MFX_ERR_UNDEFINED_BEHAVIOR;
+    }
+    if (!(newTask.m_type[0] & MFX_FRAMETYPE_REF))
```
mgonchar: looks strange to me, why we should force reference to every frame?

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2130,6 +2130,26 @@ struct FindNonSkip
 
     mfxU16 m_skipFrameMode;
 };
+mfxStatus FixForcedFrameType (DdiTask & newTask, mfxU32 frameOrder)
```
mgonchar: passing ```DdiTask``` is too much for this function, I think it should be ```Pair<mfxU8>& type``` instead

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
ogolikov: because only key frames can be forced in display order

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
ogolikov: where are a lot similar functions in encoder. Let's it be in one style.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2130,6 +2130,26 @@ struct FindNonSkip
...
+{
+    if (frameOrder == 0 &&
+        !(newTask.m_type[0] & MFX_FRAMETYPE_IDR) &&
+        !(newTask.m_type[1] & MFX_FRAMETYPE_IDR))
+    {
+        if (newTask.m_type[0] & MFX_FRAMETYPE_I)
+            newTask.m_type[0] = newTask.m_type[0] | MFX_FRAMETYPE_IDR;
```
lakulako: Use |=

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2130,6 +2130,26 @@ struct FindNonSkip
...
+        else if (newTask.m_type[1] & MFX_FRAMETYPE_I)
+            newTask.m_type[1] = newTask.m_type[1] | MFX_FRAMETYPE_IDR;
+        else
+            return MFX_ERR_UNDEFINED_BEHAVIOR;
+    }
+    if (!(newTask.m_type[0] & MFX_FRAMETYPE_REF))
+        newTask.m_type[0] = newTask.m_type[0] | MFX_FRAMETYPE_REF;
+    if (!(newTask.m_type[1] & MFX_FRAMETYPE_REF))
```
lakulako: No need to check condition

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
lakulako: 1. How it is guaranteed?
2. Can't one field be not reference field?

ogolikov: 1) Function CheckEncodeFrameParam
2) not supported in current implementation

mgonchar: ok, if I got it right, for 1. you mean following code
```
        if (ctrl != 0 && ctrl->FrameType)
        {
            // check FrameType for forced key-frame generation
            mfxU16 type = ctrl->FrameType & (MFX_FRAMETYPE_IPB | MFX_FRAMETYPE_xIPB);
            MFX_CHECK(
                type   == (MFX_FRAMETYPE_I)                     ||
                type   == (MFX_FRAMETYPE_I | MFX_FRAMETYPE_xI)  ||
                type   == (MFX_FRAMETYPE_I | MFX_FRAMETYPE_xP)  ||
                ((type == (MFX_FRAMETYPE_P | MFX_FRAMETYPE_xI)) && isHwSupportArbRef),
                MFX_ERR_INVALID_VIDEO_PARAM);
}
```
- Why don't we fix the frametype there?
- So we've checked that I frames are there, so we don't need error status here. I suggest to simplify this function to bare minimum to not confuse developers who are unfamiliar with this approach
```
void FixForcedFrameType (DdiTask & newTask, mfxU32 frameOrder)
{
    if (frameOrder == 0)
    {
        if (newTask.m_type[0] & MFX_FRAMETYPE_I)
            newTask.m_type[0] = MFX_FRAMETYPE_IDR;
        else if (newTask.m_type[1] & MFX_FRAMETYPE_I)
            newTask.m_type[1] |= MFX_FRAMETYPE_IDR;
    }

    newTask.m_type[0] |= MFX_FRAMETYPE_REF; // As I remember at index 0 we have type for progressive, so this entry should be reference anyway

    if (newTask.m_type[1]) // <- this is not progressive case
        newTask.m_type[1] |= MFX_FRAMETYPE_REF;
}
```

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2231,6 +2251,12 @@ mfxStatus ImplementationAvc::AsyncRoutine(mfxBitstream * bs)
             {
                 if (newTask.m_type[0] == 0)
                     newTask.m_type = GetFrameType(m_video, m_frameOrder - m_frameOrderIdrInDisplayOrder);
+                else
+                {
+                    mfxStatus sts = FixForcedFrameType(newTask, m_frameOrder - m_frameOrderIdrInDisplayOrder);
```
mgonchar: I think this is not a place for fixing frametypes, I suggest to move it to ```CheckEncodeFrameParam```

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
mgonchar: Olga's point that we can't move this function closer to parameter check, since we don't know there the IDR interval, so we couldn't correct I to IDR there

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
ogolikov: formally condition is not needed, m_type[0] and m_type[1] should be reference for key frames. It is needed to avoid non necessary  operations = for fields of big structure. It helps while debug.

lakulako: Unconditional assignment is definitely better than condition checking followed either by assignment or by jump.
For debugging conditional breakpoints can be used instead.

lakulako: Your code is "if(a!=b) a=b;" I propose to change it to "a=b;".
It is cheaper, faster, simpler and less code.

mgonchar: moreover, what will happen in progressive case? We will add MFX_FRAMETYPE_REF to type of non-existing field, as I see

dvrogozh:
@ogolikov : this one also needs rebase on top of the latest master to pass CI.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
mgonchar: @ogolikov , I think following implementation will address all concerns about this function:
```
// Only reference key-frames can be forced
mfxStatus FixForcedFrameType (DdiTask & newTask, mfxU32 FrameOrderInIDRsamples)
{
    if (FrameOrderInIDRsamples == 0)
    {
        if (newTask.m_type[0] & MFX_FRAMETYPE_I)
            newTask.m_type[0] |= MFX_FRAMETYPE_IDR;
        else if (newTask.m_type[1] & MFX_FRAMETYPE_I)
            newTask.m_type[1] |= MFX_FRAMETYPE_IDR;
        else
    }

    newTask.m_type[0] |= MFX_FRAMETYPE_REF;
    newTask.m_type[1] |= MFX_FRAMETYPE_REF;

    return MFX_ERR_NONE;
}
```

why this one is better?
- less unnecessary checks which were hiding the logic
- operator |= is less error-prone
- frameOrder was unclear name. From first look on this function user might think that only first frame could be IDR, but this is not what you meant

Now the following question, what will happen if user passes I & I type in BFF mode (it is allowed according to check) while currently should be IDR? I guess that this function will force top field to be IDR. Thats why we need to return part of the check and update the logic:
```
// Only reference key-frames can be forced
mfxStatus FixForcedFrameType (DdiTask & newTask, mfxU32 FrameOrderInIDRsamples)
{
    if ((FrameOrderInIDRsamples == 0)
         && !(newTask.m_type[0] & MFX_FRAMETYPE_IDR)
         && !(newTask.m_type[1] & MFX_FRAMETYPE_IDR))
    {
        mfxU32 ffid = newTask.m_fid[0], sfid = newTask.m_fid[1];

        if (newTask.m_type[ffid] & MFX_FRAMETYPE_I)
            newTask.m_type[ffid] |= MFX_FRAMETYPE_IDR;
        else if (newTask.m_type[sfid] & MFX_FRAMETYPE_I)
            newTask.m_type[sfid] |= MFX_FRAMETYPE_IDR;
        else
    }

    newTask.m_type[0] |= MFX_FRAMETYPE_REF;
    newTask.m_type[1] |= MFX_FRAMETYPE_REF;

    return MFX_ERR_NONE;
}
```

In this code ```newTask.m_type[newTask.m_fid[0]]``` means take first field. It is universal and supports both BFF and TFF, so in case of I I frame it will correctly force IDR to bottom field, and still allows to use inverted pairs (P I fields in TFF case)

mgonchar: After offline discussion we found that fields will be swapped in following code, the access to fields is correct in current implementation

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
dvrogozh: use |= here, please.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2130,6 +2130,26 @@ struct FindNonSkip
...
+mfxStatus FixForcedFrameType (DdiTask & newTask, mfxU32 frameOrder)
+{
+    if (frameOrder == 0 &&
+        !(newTask.m_type[0] & MFX_FRAMETYPE_IDR) &&
+        !(newTask.m_type[1] & MFX_FRAMETYPE_IDR))
+    {
+        if (newTask.m_type[0] & MFX_FRAMETYPE_I)
+            newTask.m_type[0] = newTask.m_type[0] | MFX_FRAMETYPE_IDR;
+        else if (newTask.m_type[1] & MFX_FRAMETYPE_I)
+            newTask.m_type[1] = newTask.m_type[1] | MFX_FRAMETYPE_IDR;
```
dvrogozh: use |= here, please.

PR #848, opened 2018-10-31
Delete bad resolution check in mjpeg decode

aulitin:
Delete bad resolution check in MFX_JPEG_Utility::IsNeedPartialAcceleration function in _studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp.
Check of resolution implemented in IsGuidSupported function.
The call functions comes from MFX_JPEG_Utility::GetPlatform.

### DIFF
--- a/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
@@ -1039,50 +1039,40 @@ bool MFX_JPEG_Utility::IsNeedPartialAcceleration(VideoCORE * core, mfxVideoParam
         || core->GetHWType() == MFX_HW_CFL
        )
     {
-        if (par->mfx.FrameInfo.Width > 4096 || par->mfx.FrameInfo.Height > 4096)
+        switch (par->mfx.FrameInfo.FourCC)
         {
-            return true;
-        }
-        else
-        {
-            switch (par->mfx.FrameInfo.FourCC)
-            {
-                case MFX_FOURCC_NV12:
-                    if (par->mfx.JPEGColorFormat == MFX_JPEG_COLORFORMAT_YCbCr &&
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1095,8 +1085,7 @@ bool MFX_JPEG_Utility::IsNeedPartialAcceleration(VideoCORE * core, mfxVideoParam
 
         if (par->mfx.JPEGColorFormat == MFX_JPEG_COLORFORMAT_YCbCr &&
             par->mfx.JPEGChromaFormat == MFX_CHROMAFORMAT_YUV420 &&
-            par->mfx.FrameInfo.FourCC == MFX_FOURCC_RGB4 &&
-            (par->mfx.FrameInfo.Width > 4096 || par->mfx.FrameInfo.Height > 4096))
+            par->mfx.FrameInfo.FourCC == MFX_FOURCC_RGB4)
```
otsirule: Does driver really report about such configuration?

aulitin:
Driver should report 16K resolution on SKL and upper platforms (check on CFL, SKL). This is normal, because resolution support is 16K in readme to media-driver.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
aulitin: @otsirule , driver should report 16K resolution on SKL and upper platforms (check on CFL, SKL). This is normal, because resolution support is 16K in readme to media-driver.

otsirule: But this condition not only about resolution but also about FourCC. Why resolution support should influence complex conditions like that?

aulitin: @otsirule , I corrected

PR #853, opened 2018-11-01
[VP9e] Fix the order of parameter buffer submission

vshampor:
Temporal layer information should be supplied to HW before
frame rate/rate control information


vshampor:
Waiting for an opportunity to verify.

PR #857, opened 2018-11-01
[sample_hevc_fei] Fix regression caused by 29d17bd

otsirule:
For display order (e.g. pipeline w/o preenc) reordering is
performed in library and SetCtrlParams is not called, so
there is no way to use per frame type controls for setting
ForceCtuSplit and FastIntraMode. Change restores setting them
as before commit 29d17bd.

### DIFF
--- a/samples/sample_hevc_fei/src/sample_hevc_fei.cpp
+++ b/samples/sample_hevc_fei/src/sample_hevc_fei.cpp
@@ -424,7 +424,8 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU32 nArgNum, sInputParams& 
         }
         else if (0 == msdk_strcmp(strInput[i], MSDK_STRING("-ForceCtuSplit")))
         {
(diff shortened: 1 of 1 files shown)


samples/sample_hevc_fei/src/sample_hevc_fei.cpp
```diff
@@ -424,6 +424,7 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU32 nArgNum, sInputParams&
         }
         else if (0 == msdk_strcmp(strInput[i], MSDK_STRING("-ForceCtuSplit")))
         {
+            params.encodeCtrl.ForceCtuSplit = 1; // for display order mode (w/o preenc)
```
mgonchar: what is wrong with encoded order (I think nothing)? And what will happen in case of PreENC + ENCODE?
I don't see any additional checks in this file

otsirule: @mgonchar, removed a confusing comment and added checks.

Issue #674, opened 2018-09-25
sample_decode freezes system when running X11 mode


chivakker:
This might be particular to the configuration given.  I am enabling wayland and x11 at the same time.  This is the configuration output:

I have not hooked up GDB yet.

********************** Media SDK Configuration Summary **********************
MediaSDK:
  API (request to use this version)       :
  API_VERSION                             : 1.27
  API_FLAGS                               :
  MFX_CONFIG_FILE                         :
  See detailed per-component mediasdk configuration in the auto-generated file:
    /builddir/build/BUILD/MediaSDK-intel-mediasdk-18.3.pre2/clr-build/mfxconfig.h
Cmake:
  CMAKE_VERSION                           : 3.12.2
  CMAKE_GENERATOR                         : Unix Makefiles
  CMAKE_BUILD_TOOL                        : /usr/bin/gmake
Target:
  CMAKE_SYSTEM_NAME                       : Linux
  CMAKE_SYSTEM_VERSION                    : 4.18.9-635.native
  CMAKE_SYSTEM_PROCESSOR                  : x86_64
General:
  CMAKE_BUILD_TYPE                        : Release
  CMAKE_TOOLCHAIN_FILE                    :
  CMAKE_C_COMPILER                        : /usr/lib64/ccache/bin/gcc
  CMAKE_CXX_COMPILER                      : /usr/lib64/ccache/bin/g++
  Building architecture                   : 64-bit
Flags:
  CMAKE_C_FLAGS                           : -msse4.2 -m64 -g -O2 -fPIC -g -feliminate-unused-debug-types -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=32 -Wformat -Wformat-security -Wl,--copy-dt-needed-entries,-z,now,-z,relro,-z,max-page-size=0x1000,-z,separate-code -Wno-error -ftree-vectorize -Wl,--enable-new-dtags -m64 -march=westmere -mtune=haswell -fasynchronous-unwind-tables -fno-omit-frame-pointer -Wp,-D_REENTRANT -pipe -fPIC -Wall -Wformat -Wformat-security -Wno-deprecated-declarations -Wno-unknown-pragmas -Wno-unused
[...]

PR #861, opened 2018-11-02
Pass to libdrm render node, not primary node

Alexandr-Konovalov:
Fix failure on CentOS 7.5: primary drm node was created for libdri,
but it lacks permissions required for ioctl(.., DRM_IOCTL_I915_GETPARAM)
to run.

### DIFF
--- a/samples/sample_common/src/vaapi_device.cpp
+++ b/samples/sample_common/src/vaapi_device.cpp
@@ -95,7 +95,9 @@ mfxStatus CVAAPIDeviceX11::Init(mfxHDL hWindow, mfxU16 nViews, mfxU32 nAdapterNu
     }
     m_xcbconn = x11xcblib.XGetXCBConnection(m_dpy);
 
-    m_dri_fd = open("/dev/dri/card0", O_RDWR);
+    // it's enough to pass render node, because we only request
+    // information from kernel via m_dri_fd
+    m_dri_fd = open("/dev/dri/renderD128", O_RDWR);
     if (m_dri_fd < 0) {
         msdk_printf(MSDK_STRING("Failed to open dri device\n"));
         return MFX_ERR_NOT_INITIALIZED;
--- a/samples/sample_common/src/vaapi_device.cpp
+++ b/samples/sample_common/src/vaapi_device.cpp
@@ -184,7 +184,7 @@ mfxStatus CVAAPIDeviceX11::RenderFrame(mfxFrameSurface1 * pSurface, mfxFrameAllo
     if (MFX_ERR_NONE == mfx_res)
     {
         VADisplay dpy = m_X11LibVA.GetVADisplay();
-        VADisplay rnddpy = m_X11LibVA.GetVADisplay(true);
+        VADisplay rnddpy = m_X11LibVA.GetVADisplay();
         VASurfaceID rndsrf;
         void* ctx;
 
--- a/samples/sample_common/src/vaapi_device.cpp
+++ b/samples/sample_common/src/vaapi_device.cpp
@@ -88,12 +88,7 @@ mfxStatus CVAAPIDeviceX11::Init(mfxHDL hWindow, mfxU16 nViews, mfxU32 nAdapterNu
     MfxLoader::XLib_Proxy & x11lib = m_X11LibVA.GetX11();
     MfxLoader::X11_Xcb_Proxy & x11xcblib = m_X11LibVA.GetX11XcbX11();
 
-    m_dpy = x11lib.XOpenDisplay(NULL);
-    if (m_dpy == NULL){
-        msdk_printf(MSDK_STRING("Failed to open display\n"));
-        return MFX_ERR_NOT_INITIALIZED;
-    }
-    m_xcbconn = x11xcblib.XGetXCBConnection(m_dpy);
+    m_xcbconn = x11xcblib.XGetXCBConnection(VAAPI_GET_X_DISPLAY(m_X11LibVA.GetXDisplay()));
 
     // it's enough to pass render node, because we only request
     // information from kernel via m_dri_fd
@@ -131,11 +126,6 @@ void CVAAPIDeviceX11::Close(void)
     if (m_dri_fd)
     {
         close(m_dri_fd);
-    }
-    if (m_dpy)
-    {
-        MfxLoader::XLib_Proxy & x11lib = m_X11LibVA.GetX11();
-        x11lib.XCloseDisplay(m_dpy);
     }
 #endif
 }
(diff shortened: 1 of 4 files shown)


alariono:
@Alexandr-Konovalov,
add link on issue https://github.com/Intel-Media-SDK/MediaSDK/issues/674 into commit message|
GitHub will close issue automatically once the commit is merged into the default branch.

samples/sample_common/src/vaapi_device.cpp
```diff
@@ -95,7 +95,9 @@ mfxStatus CVAAPIDeviceX11::Init(mfxHDL hWindow, mfxU16 nViews, mfxU32 nAdapterNu
     }
     m_xcbconn = x11xcblib.XGetXCBConnection(m_dpy);
 
-    m_dri_fd = open("/dev/dri/card0", O_RDWR);
+    // it's enough to pass render node, because we only request
+    // information from kernel via m_dri_fd
+    m_dri_fd = open("/dev/dri/renderD128", O_RDWR);
```
dvrogozh: this looks ok, I think.

However, the rest of the function does not make sense to me. For example, on line 67 we call XCreateSimpleWindow, thus we should already have X display opened, but we open it on the line 91 actually. Just grepping the code:
```
samples/sample_common/src/vaapi_device.cpp:91:    m_dpy = x11lib.XOpenDisplay(NULL);
samples/sample_common/src/vaapi_utils_x11.cpp:49:            m_display = m_x11lib.XOpenDisplay(currentDisplay);
samples/sample_common/src/vaapi_utils_x11.cpp:51:            m_display = m_x11lib.XOpenDisplay(VAAPI_X_DEFAULT_DISPLAY);
samples/sample_common/src/vaapi_utils_x11.cpp:136:            m_display = m_x11lib.XOpenDisplay(currentDisplay);
samples/sample_common/src/vaapi_utils_x11.cpp:138:            m_display = m_x11lib.XOpenDisplay(VAAPI_X_DEFAULT_DISPLAY);
```
Do we really need all 3 functions?

samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -51,114 +50,99 @@ X11LibVA::X11LibVA(void)
...
-            va_res = m_libva.vaInitialize(m_va_dpy, &major_version, &minor_version);
-            sts = va_to_mfx_status(va_res);
-            if (MFX_ERR_NONE != sts) {
-                m_x11lib.XCloseDisplay(m_display);
-                }
+                va_res = m_libva.vaInitialize(m_va_dpy, &major_version, &minor_version);
+                sts = va_to_mfx_status(va_res);
                 unsetenv("LIBVA_DRIVER_NAME");
             } else {
                 sts = MFX_ERR_NOT_INITIALIZED;
```
Alexandr-Konovalov: I treated lack of m_x11lib.XCloseDisplay(m_display) on that branch as a bug. Is it correct or there was some plan?

samples/sample_common/src/vaapi_device.cpp
Alexandr-Konovalov: It seems CVAAPIDeviceX11 can use connection from m_X11LibVA it includes.

samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -51,114 +50,99 @@ X11LibVA::X11LibVA(void)
...
-            {
-                m_libva.vaTerminate(m_va_dpy);
-                m_x11lib.XCloseDisplay(m_display);
-                sts = MFX_ERR_NULL_PTR;
-            }
-            if (!setenv("LIBVA_DRIVER_NAME", RENDERING_DRIVER_NAME, 1)) {
-                va_res = m_libva.vaInitialize(m_va_dpy_render, &major_version, &minor_version);
-                sts = va_to_mfx_status(va_res);
-                if (MFX_ERR_NONE != sts)
+            if (!msdk_strcmp(PROCESSING_DRIVER_NAME, RENDERING_DRIVER_NAME)) {
```
dvrogozh: I believe we need to completely drop this processing/rendering drivers stuff. So, the behind the scenes story was that iHD did not support efficient on-screen rendering via X11, but i965 did. So, to optimized that we have used 2 drivers approach. However, now we are going to support iHD driver only, so I believe that all this dual driver approach could be dropped. @alariono : do you agree?

This should be a preceding patch I believe.

alariono: @dvrogozh, Yes, agree lets drop dual driver approach

Alexandr-Konovalov: Dropped, please check.

samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -29,8 +29,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 
 #define VAAPI_X_DEFAULT_DISPLAY ":0.0"
 
-const char* PROCESSING_DRIVER_NAME="iHD";
-const char* RENDERING_DRIVER_NAME="iHD";
+const char* DRIVER_NAME="iHD";
```
dvrogozh: drop.

samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -64,11 +63,11 @@ X11LibVA::X11LibVA(void)
         }
         if (MFX_ERR_NONE == sts)
         {
-            if (!setenv("LIBVA_DRIVER_NAME", PROCESSING_DRIVER_NAME, 1)) {
-            va_res = m_libva.vaInitialize(m_va_dpy, &major_version, &minor_version);
-            sts = va_to_mfx_status(va_res);
-            if (MFX_ERR_NONE != sts) {
-                m_x11lib.XCloseDisplay(m_display);
+            if (!setenv("LIBVA_DRIVER_NAME", DRIVER_NAME, 1)) {
```
dvrogozh: nope, you don't need to set _any_ driver name. If needed user will do that in a shell. Just call vaInitialize().

samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -169,14 +147,11 @@ X11LibVA::X11LibVA(void)
...
+    //release context
+    if (m_contextID != VA_INVALID_ID)
     {
-        //release context
-        if (m_contextID != VA_INVALID_ID)
-        {
-            m_libva.vaDestroyContext(m_va_dpy_render, m_contextID);
-            m_contextID = VA_INVALID_ID;
-        }
+        m_libva.vaDestroyContext(m_va_dpy_render, m_contextID);
```
dvrogozh: What is m_va_dpy_render if you did not create it? I believe you need to delete this variable as well and use m_va_dpy everywhere.

samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -50,14 +49,23 @@ X11LibVA::X11LibVA(void)
...
+
+            if (!m_va_dpy)
+            {
+                sts = MFX_ERR_NULL_PTR;
+            }
+        }
+#if !defined(X11_DRI3_SUPPORT)
+        if (MFX_ERR_NONE == sts)
+        {
             m_va_dpy_render = m_vax11lib.vaGetDisplay(m_display);
```
dvrogozh: remove and use m_va_dpy

samples/sample_common/src/vaapi_utils_x11.cpp
Alexandr-Konovalov: Done, thank you.

dvrogozh:
@alariono : looks good to me, your turn.

PR #865, opened 2018-11-05
h264e: change ICQ quality factor range to [1, 51].

ChenXiaomin36:
Change the ICQ quality factor range to [1, 51]
which aligns with the definition in VAAPI.

Change-Id: I582753964de57d0b479c1eebee42b05700f516d1

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -3854,8 +3854,11 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         changed = true;
     }
 
-    if (par.mfx.RateControlMethod == MFX_RATECONTROL_ICQ ||
-        par.mfx.RateControlMethod == MFX_RATECONTROL_LA_ICQ)
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_ICQ)
+    {
+        if (!CheckRange(par.mfx.ICQQuality, 1, 51)) changed = true;
+    }
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_LA_ICQ)
     {
         if (!CheckRange(par.mfx.ICQQuality, 0, 51)) changed = true;
     }


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -2304,6 +2304,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
             par.mfx.RateControlMethod != MFX_RATECONTROL_AVBR &&
             par.mfx.RateControlMethod != MFX_RATECONTROL_QVBR &&
             par.mfx.RateControlMethod != MFX_RATECONTROL_VCM &&
+            par.mfx.RateControlMethod != MFX_RATECONTROL_ICQ &&
```
lizhong1008: So, ICQ mode can be supported by Lowpower mode  too?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -3854,8 +3855,11 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         changed = true;
     }
 
-    if (par.mfx.RateControlMethod == MFX_RATECONTROL_ICQ ||
-        par.mfx.RateControlMethod == MFX_RATECONTROL_LA_ICQ)
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_ICQ)
+    {
+        if (!CheckRange(par.mfx.ICQQuality, 1, 51)) changed = true;
+    }
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_LA_ICQ)
```
lizhong1008: Why we need such a change?

dvrogozh:
pr #866 will take care of the enabling icq in sample_encode...

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
ChenXiaomin36: ICQ mode can't be supported by Lowpower mode now, so I remove this code. @lizhong1008

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
ChenXiaomin36: In VAAPI, the comment for ICQ_quality_factor is that "Initial quality factor used in ICQ mode." and "This value must be between 1 and 51.". So I think in MSDK, the value check for it should align with the definition in VAAPI. @lizhong1008

ChenXiaomin36:
The purpose of this change is to enable ICQ for AVC in sample_encode. I have removed the codes for ICQ support for Lowpower mode. @dvrogozh

ChenXiaomin36:
Just see you submit the same PR. @dvrogozh

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
ChenXiaomin36: If MSDK still checks the value in the range of [0, 51], and doesn't have process for 0, then 0 can be transferred to media driver. At present, the valid range for ICQ quality factor in media driver is [1, 51], if 0 is set, then media driver will report error. Currently, I haven't seen there is process for 0. If I missed this process, please correct me. Thanks! @dvrogozh

ChenXiaomin36:
If MSDK still checks the value in the range of [0, 51], and doesn't have process for 0, then 0 can be transferred to media driver. At present, the valid range for ICQ quality factor in media driver is [1, 51], if 0 is set, then media driver will report error. Currently, I haven't seen there is process for 0. If I missed this process, please correct me. Thanks! @dvrogozh

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
stasefremov: @ogolikov, @takulako, please look.

ogolikov:
0 - is default value for ICQQuality
MediaSDK sets ICQQuality to 26 in this case in SetDefaults function.

lizhong1008:
> 0 - is default value for ICQQuality
> MediaSDK sets ICQQuality to 26 in this case in SetDefaults function.
Could you show me the code where set it to be 26?

lizhong1008:
> 0 - is default value for ICQQuality
> MediaSDK sets ICQQuality to 26 in this case in SetDefaults function.

So I prefer iHD driver can take 0 as 26 too.  And application can have same setting for ICQ and ICQ_LA

ogolikov:
mfx_h264_enc_common_hw.cpp: 5395   
 if (par.mfx.RateControlMethod == MFX_RATECONTROL_ICQ || par.mfx.RateControlMethod == MFX_RATECONTROL_LA_ICQ)
    {
        if (par.mfx.ICQQuality == 0)
            par.mfx.ICQQuality = 26;
    }

ChenXiaomin36:
Since there is separate processing for "ICQQuality==0", cancel this PR.

PR #867, opened 2018-11-06
LA_ICQ: enable support in sample_encode

ChenXiaomin36:
./sample_encode h264 -i xxx.yuv -o xxx.h264 ...... -icq ICQQualityFactor -la
or
./sample_encode h264 -i xxx.yuv -o xxx.h264 ...... -icq ICQQualityFactor -la -lad LookAheadDepth
or
./sample_encode h264 -i xxx.yuv -o xxx.h264 ...... -icq ICQQualityFactor -lad LookAheadDepth
ICQQualityFactor should be in [0, 51]
LookAheadDepth should be in [10, 100]
 
Change-Id: I84cc422460dbfc9dc1b0023265aec0e37e17da01

### DIFF
--- a/samples/sample_encode/src/sample_encode.cpp
+++ b/samples/sample_encode/src/sample_encode.cpp
@@ -83,8 +83,10 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage, ...)
     msdk_printf(MSDK_STRING("   [-num_active_BL1 numRefs] - number of maximum allowed references for B frames in L1 (for HEVC only)\n"));
     msdk_printf(MSDK_STRING("   [-la] - use the look ahead bitrate control algorithm (LA BRC) (by default constant bitrate control method is used)\n"));
     msdk_printf(MSDK_STRING("           for H.264, H.265 encoder. Supported only with -hw option on 4th Generation Intel Core processors. \n"));
+    msdk_printf(MSDK_STRING("           if [-icq] option is also enabled simultaneously, then LA_ICQ bitrate control algotithm will be used. \n"));
     msdk_printf(MSDK_STRING("   [-lad depth] - depth parameter for the LA BRC, the number of frames to be analyzed before encoding. In range [10,100].\n"));
     msdk_printf(MSDK_STRING("            may be 1 in the case when -mss option is specified \n"));
+    msdk_printf(MSDK_STRING("            if [-icq] option is also enabled simultaneously, then LA_ICQ bitrate control algotithm will be used. \n"));
     msdk_printf(MSDK_STRING("   [-dstw width] - destination picture width, invokes VPP resizing\n"));
     msdk_printf(MSDK_STRING("   [-dsth height] - destination picture height, invokes VPP resizing\n"));
     msdk_printf(MSDK_STRING("   [-hw] - use platform specific SDK implementation (default)\n"));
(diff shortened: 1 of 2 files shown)


samples/sample_encode/src/sample_encode.cpp
```diff
@@ -335,7 +337,10 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
         else if (0 == msdk_strcmp(strInput[i], MSDK_STRING("-lad")))
         {
             VAL_CHECK(i+1 >= nArgNum, i, strInput[i]);
-            pParams->nRateControlMethod = MFX_RATECONTROL_LA;
+            if (!pParams->nRateControlMethod)
```
lakulako: If nRateControlMethod is already set, then it is not changed to MFX_RATECONTROL_LA. But nLADepth is set.
Thus parameter is ignored and no error is reported.
Now every new RC method in command line replaces previous. Not good to change it.

ChenXiaomin36:
I have changed the title and codes have been rebased.

samples/sample_encode/src/sample_encode.cpp
```diff
@@ -370,7 +372,10 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
...
+            {
+                pParams->nRateControlMethod = MFX_RATECONTROL_LA;
```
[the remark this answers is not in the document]
ChenXiaomin36: since there is a dedicated option "-la" for  LA mode, I think "-laicq" is also OK for LA_ICQ mode. @dvrogozh @artem-shaporenko 
Thanks.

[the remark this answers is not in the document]
ChenXiaomin36: What's your opinion for my reply? @dvrogozh 
Thanks.

ChenXiaomin36:
I have updated the codes. Please help to review it. Thanks! @dvrogozh @onabiull

samples/sample_encode/src/sample_encode.cpp
```diff
@@ -365,12 +365,37 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
...
+            }
+            else if (pParams->nRateControlMethod == MFX_RATECONTROL_ICQ)
+            {
+                pParams->nRateControlMethod = MFX_RATECONTROL_LA_ICQ;
+            }
+            else
+            {
+                PrintHelp(strInput[0], MSDK_STRING("More than one BRC modes assigned, and another BRC mode isn't compatible with LA."));
```
dvrogozh: LA tests failed after your recent modifications. That's probably because you don't handle -la and -lad options in the same command line. I.e. once parser will meet -la it will set RC=LA and when it will meet -lad it will simply error out here.

Question you probably need to answer from the code: can -lad option present in the command line without -la? If it can't then remove processing of RC from here at all. If it can, then you need to add handling for LA and LA_ICQ modes in both branches.

samples/sample_encode/src/sample_encode.cpp
```diff
@@ -83,8 +83,10 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage, ...)
     msdk_printf(MSDK_STRING("   [-num_active_BL1 numRefs] - number of maximum allowed references for B frames in L1 (for HEVC only)\n"));
     msdk_printf(MSDK_STRING("   [-la] - use the look ahead bitrate control algorithm (LA BRC) (by default constant bitrate control method is used)\n"));
     msdk_printf(MSDK_STRING("           for H.264, H.265 encoder. Supported only with -hw option on 4th Generation Intel Core processors. \n"));
+    msdk_printf(MSDK_STRING("           if [-icq] options is also enabled simultaneously, then LA_ICQ bitrate control algotithm will be used. \n"));
```
dvrogozh: "if [-icq] option" - not options

samples/sample_encode/src/sample_encode.cpp
```diff
@@ -83,8 +83,10 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage, ...)
...
     msdk_printf(MSDK_STRING("   [-lad depth] - depth parameter for the LA BRC, the number of frames to be analyzed before encoding. In range [10,100].\n"));
     msdk_printf(MSDK_STRING("            may be 1 in the case when -mss option is specified \n"));
+    msdk_printf(MSDK_STRING("            if [-icq] options is also enabled simultaneously, then LA_ICQ bitrate control algotithm will be used. \n"));
```
dvrogozh: the same

samples/sample_encode/src/sample_encode.cpp
ChenXiaomin36: updated.

samples/sample_encode/src/sample_encode.cpp
ChenXiaomin36: updated.

ChenXiaomin36:
I have had the codes rebased.

PR #868, opened 2018-11-06
h264e: change the quality factor range to [1, 51] for QVBR

ChenXiaomin36:
Although there is no explicit declaration in VAAPI
for the range of "quality_factor",
for AVC, the range should be [1, 51],
like "ICQ_quality_factor" in VAAPI.

Change-Id: Idd5877d45bf7185ff8d189c9af2cbcc7c1ab5d73

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -3862,7 +3862,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
 
     if (par.mfx.RateControlMethod == MFX_RATECONTROL_QVBR)
     {
-        if (!CheckRange(extOpt3->QVBRQuality, 0, 51)) changed = true;
+        if (!CheckRange(extOpt3->QVBRQuality, 1, 51)) changed = true;
     }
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -3862,7 +3862,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
 
     if (par.mfx.RateControlMethod == MFX_RATECONTROL_QVBR)
     {
-        if (!CheckRange(extOpt3->QVBRQuality, 0, 51)) changed = true;
+        if (!CheckRange(extOpt3->QVBRQuality, 1, 51)) changed = true;
```
stasefremov: @ogolikov, @takulako, please look.

ogolikov: It is usual for MediaSDK API that 0 is default value. There is no error if application sets parameter to 0. It means that MediaSDK  sets this parameter itself (function SetDefaults defines default parameters, it is 26 for QVBRQuality).  
I checked documentation, there is no info about default value for QVBRQuality. But, I think, it should be corrected.

ChenXiaomin36:
Since there is separate processing for "QVBRQuality==0", cancel this PR.

PR #872, opened 2018-11-07
[ASG HEVC] Replaced unsafe sscanf function

semelianova:
Unsafe sscanf caused a build error on windows, it was replaced by using istringstream object for further parsing.

### DIFF
--- a/tools/asg-hevc/src/test_processor.cpp
+++ b/tools/asg-hevc/src/test_processor.cpp
@@ -218,7 +218,8 @@ void TestProcessor::RunRepackVerify()
         fpRepackStat.getline(repackStat, sizeof(repackStat));
         if (!fpRepackStat.good())
             throw string("ERROR: Repack stat file read failed");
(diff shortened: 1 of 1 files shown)


tools/asg-hevc/src/test_processor.cpp
```diff
@@ -218,7 +218,8 @@ void TestProcessor::RunRepackVerify()
         fpRepackStat.getline(repackStat, sizeof(repackStat));
         if (!fpRepackStat.good())
             throw string("ERROR: Repack stat file read failed");
-        sscanf(repackStat, "%d NumPasses\n", &activeNumPasses);
+        istringstream repackStatStream(string(repackStat, strlen(repackStat)));
```
mgonchar: actually you don't need this constructor, ```string (const char* s, size_t n);``` is used to construct std::string owning substring of C-string. In your case ```string (const char* s);``` would be enough. But thats not a real issue, so I don't insist on code change

PR #875, opened 2018-11-07
h264e: Replace GetExtBuffer on GetExtBufferRef

Losev-Ivan:

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -1463,14 +1463,14 @@ mfxStatus MfxHwH264Encode::ReadSpsPpsHeaders(MfxVideoParam & par)
         if (extBits.SPSBuffer)
         {
             InputBitstream reader(extBits.SPSBuffer, extBits.SPSBufSize);
-            mfxExtSpsHeader * extSps = GetExtBuffer(par);
-            ReadSpsHeader(reader, *extSps);
+            mfxExtSpsHeader & extSps = GetExtBufferRef(par);
+            ReadSpsHeader(reader, extSps);
 
             if (extBits.PPSBuffer)
             {
                 InputBitstream pps_reader(extBits.PPSBuffer, extBits.PPSBufSize);
-                mfxExtPpsHeader * extPps = GetExtBuffer(par);
-                ReadPpsHeader(pps_reader, *extSps, *extPps);
+                mfxExtPpsHeader & extPps = GetExtBufferRef(par);
+                ReadPpsHeader(pps_reader, extSps, extPps);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -7516,17 +7516,17 @@ void MfxVideoParam::SyncVideoToCalculableParam()
 
     if (IsMvcProfile(mfx.CodecProfile))
     {
-        mfxExtMVCSeqDesc * extMvc = GetExtBuffer(*this);
-        if (extMvc->NumView)
+        mfxExtMVCSeqDesc & extMvc = GetExtBufferRef(*this);
```
DenWolf: It's not a good idea to use this method inside class MfxVideoParam - there is no 100% guarantee that needed buffer is existed, so there is a way to undefined behavior.
As result, if pointer extMvc can be nullptr, it's better to check it directly, for example in the "if" condition at line 7520

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -1463,14 +1463,14 @@ mfxStatus MfxHwH264Encode::ReadSpsPpsHeaders(MfxVideoParam & par)
         if (extBits.SPSBuffer)
         {
             InputBitstream reader(extBits.SPSBuffer, extBits.SPSBufSize);
-            mfxExtSpsHeader * extSps = GetExtBuffer(par);
-            ReadSpsHeader(reader, *extSps);
+            mfxExtSpsHeader & extSps = GetExtBufferRef(par);
```
mgonchar: ok

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -1463,14 +1463,14 @@ mfxStatus MfxHwH264Encode::ReadSpsPpsHeaders(MfxVideoParam & par)
...
+            ReadSpsHeader(reader, extSps);
 
             if (extBits.PPSBuffer)
             {
                 InputBitstream pps_reader(extBits.PPSBuffer, extBits.PPSBufSize);
-                mfxExtPpsHeader * extPps = GetExtBuffer(par);
-                ReadPpsHeader(pps_reader, *extSps, *extPps);
+                mfxExtPpsHeader & extPps = GetExtBufferRef(par);
```
mgonchar: ok

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
mgonchar: Here starts the list of buffers which can be safely extracted with GetExtBufferRef: https://github.com/Intel-Media-SDK/MediaSDK/blob/dd25daaa1ff50d8ee41f4c14b8599db3e11d2331/_studio/mfx_lib/shared/include/mfx_h264_enc_common_hw.h#L547

mfxExtMVCSeqDesc is not in this list so you can't use GetExtBufferRef for it

PR #876, opened 2018-11-07
[JPEGE] BGR4 in vaapi_allocator

IlyaSergeyev:
BGR4 was added in libmfx_allocator_vaapi.cpp (shared/src), but does not in vaapi_allocator (sample_common)

Issue: MDP-47046

### DIFF
--- a/samples/sample_common/src/vaapi_allocator.cpp
+++ b/samples/sample_common/src/vaapi_allocator.cpp
@@ -52,6 +52,8 @@ unsigned int ConvertMfxFourccToVAFormat(mfxU32 fourcc)
 #endif
     case MFX_FOURCC_RGB4:
         return VA_FOURCC_ARGB;
+    case MFX_FOURCC_BGR4:
+        return VA_FOURCC_ABGR;
     case MFX_FOURCC_RGBP:
         return VA_FOURCC_RGBP;
     case MFX_FOURCC_P8:
@@ -166,6 +168,7 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
                        (VA_FOURCC_R5G6B5 != va_fourcc) &&
 #endif
                        (VA_FOURCC_ARGB   != va_fourcc) &&
+                       (VA_FOURCC_ABGR   != va_fourcc) &&
                        (VA_FOURCC_RGBP   != va_fourcc) &&
                        (VA_FOURCC_P208   != va_fourcc) &&
                        (VA_FOURCC_P010   != va_fourcc)))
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -503,6 +506,17 @@ mfxStatus vaapiFrameAllocator::LockFrame(mfxMemId mid, mfxFrameData *ptr)
                 }
                 else mfx_res = MFX_ERR_LOCK_MEMORY;
                 break;
+            case VA_FOURCC_ABGR:
+                if (mfx_fourcc == MFX_FOURCC_BGR4)
+                {
+                    ptr->Pitch = (mfxU16)vaapi_mid->m_image.pitches[0];
+                    ptr->R = pBuffer + vaapi_mid->m_image.offsets[0];
+                    ptr->G = ptr->R + 1;
```
dvrogozh: I would prefer to see the code which get use of pBuffer + image.offsets[i]

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -503,6 +506,17 @@ mfxStatus vaapiFrameAllocator::LockFrame(mfxMemId mid, mfxFrameData *ptr)
...
                 else mfx_res = MFX_ERR_LOCK_MEMORY;
                 break;
+            case VA_FOURCC_ABGR:
+                if (mfx_fourcc == MFX_FOURCC_BGR4)
+                {
+                    ptr->Pitch = (mfxU16)vaapi_mid->m_image.pitches[0];
+                    ptr->R = pBuffer + vaapi_mid->m_image.offsets[0];
+                    ptr->G = pBuffer + vaapi_mid->m_image.offsets[1];
+                    ptr->B = pBuffer + vaapi_mid->m_image.offsets[2];
+                    ptr->A = ptr->R + 3;
```
dvrogozh: Wow. Interesting that VAImage supports only 3 planes as a maximum. This potentially can be a problem if someone will try to play with the offsets. @XinfengZhang: pay attention on this. A candidate for the VAAPI 3.x series:).

artem-shaporenko: ARGB is single plain, 4 samples format e.g. all the colors/Alpha channel are interleaved.
RGBP is 3 plain format where each color placed in separate plain. There are no format for 4 plains.

artem-shaporenko: I believe here you are mixing plains vs samples(colors)

dvrogozh: Yep, I can mix that up. Though only one of 2 ways could have been valid and programmed in the driver. @IlyaSergeyev : what iHD driver returns for:
- VAImage::num_planes ?
- VAImage::offsets[n], where n=0 .. num_planes-1 ?

Per Artem this should be VAImage::num_planes=1 with only 1st valid offsets. I tend to agree with Artem that this is what we should expect meaning that what you have written initially should be correct. However, I would like to understand driver behavior.

PR #877, opened 2018-11-07
[hevcd] hevcd_afl_broken-plugin_hevcdhw segfault

IlyaSergeyev:
check slice ptr before acces

Issue: MDP-51108

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_frame_info.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_frame_info.cpp
@@ -103,10 +103,8 @@ void H265DecoderFrameInfo::EliminateErrors()
     if (!GetSlice(0))
         return;
 
-    uint32_t count = m_SliceCount;
-
     // Remove dependent slices without a corresponding independent slice
-    for (uint32_t sliceId = 0; sliceId < count; sliceId++)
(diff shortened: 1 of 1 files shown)


_studio/shared/umc/codec/h265_dec/src/umc_h265_frame_info.cpp
```diff
@@ -127,7 +127,12 @@ void H265DecoderFrameInfo::EliminateErrors()
         bool bIndepSliceMissing = false;
         for (uint32_t sliceId = 1; sliceId < count; sliceId++)
         {
-            H265SliceHeader *sliceHeader = GetSlice(sliceId)->GetSliceHeader();
+            H265Slice *slice = GetSlice(sliceId);
```
dmitry-gurulev: State when we have [count] slices but have no slice for this idx looks like invariant violation. It seems the thing is 
`uint32_t count = m_SliceCount;`
- we save number of slices and then use this number. But we may remove some slices (line #115). We need either adjust [count] or use [GetSliceCount] for this loop.

IlyaSergeyev: Why we can't use m_SliceCount directly in loops?

PR #880, opened 2018-11-08
[VPP][ICL] Add correct condition checking for AYUV supporting

romka0017:


_studio/mfx_lib/vpp/src/mfx_vpp_utils.cpp
```diff
@@ -1327,7 +1327,7 @@ mfxStatus CheckFrameInfo(mfxFrameInfo* info, mfxU32 request, eMFXHWType platform
         case MFX_FOURCC_YUY2:
             break;
         case MFX_FOURCC_AYUV:
-#if !defined(_WIN32) || !defined(_WIN64)
+#if !defined(_WIN32) && !defined(_WIN64)
```
dmitryermilov: please remove #if

PR #885, opened 2018-11-09
scheduler: create threads per-component requests

dvrogozh:
This PR is not yet fully completed. Pay attention and provide comments:
1. How many threads each component needs. Pay attention on VPP, encoders and FEI - these components create 2 tasks (Submit & Query) - this part was moved away from this pr to #911.
2. Scheduler internal API altered - are we ok with that?

First 5 cleanup commits extracted to separate pr #903, please, review them there.

### DIFF
--- a/_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
+++ b/_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
@@ -24,6 +24,7 @@
 #include <mfxdefs.h>
 
 #include <functional>
+#include <future>
 #include <memory>
 #include <thread>
 #include <condition_variable>
@@ -37,10 +38,13 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
 
     MFX_SCHEDULER_THREAD_CONTEXT() = delete;
 
-    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id, SchedulerThreadFunc func)
+    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id)
       : state(State::Waiting)
       , threadNum(id)
-      , threadHandle([this, func](){ func(this); })
+      , threadHandle([this]() {
+          start.get_future().wait();
+          if (threadFunc) threadFunc(this);
+      })
       , workTime(0)
       , sleepTime(0)
     {
@@ -54,7 +58,17 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
     ~MFX_SCHEDULER_THREAD_CONTEXT()
     {
         if (threadHandle.joinable())
+        {
+            if (!threadFunc) // meaning Start() was not called
+                start.set_value();
             threadHandle.join();
+        }
+    }
+
+    void Start(SchedulerThreadFunc func)
+    {
+        threadFunc = std::move(func);
+        start.set_value();
     }
 
     void Stop(std::lock_guard<std::mutex>& /*lock*/)
@@ -71,8 +85,10 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
 
     State state;                       // thread state, waiting or running
     mfxU32 threadNum;                  // thread number assigned by the core
-    std::thread threadHandle;          // thread handle
     std::condition_variable taskAdded; // cond. variable to signal new tasks
+    std::promise<void> start;
+    SchedulerThreadFunc threadFunc;
+    std::thread threadHandle;          // thread handle
 
     mfxU64 workTime;                   // integral working time
     mfxU64 sleepTime;                  // integral sleeping time
--- a/_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
+++ b/_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
@@ -23,6 +23,8 @@
 
 #include <mfxdefs.h>
 
+#include <functional>
+#include <memory>
 #include <thread>
 #include <condition_variable>
 
@@ -31,14 +33,35 @@ class mfxSchedulerCore;
 
 struct MFX_SCHEDULER_THREAD_CONTEXT
 {
-    MFX_SCHEDULER_THREAD_CONTEXT()
+    typedef std::function<void(MFX_SCHEDULER_THREAD_CONTEXT*)> SchedulerThreadFunc;
+
+    MFX_SCHEDULER_THREAD_CONTEXT() = delete;
+
+    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id, SchedulerThreadFunc func)
       : state(State::Waiting)
-      , pSchedulerCore(NULL)
-      , threadNum(0)
-      , threadHandle()
+      , threadNum(id)
+      , threadHandle([this, func](){ func(this); })
       , workTime(0)
       , sleepTime(0)
-    {}
+    {
+    }
+
+    MFX_SCHEDULER_THREAD_CONTEXT(const MFX_SCHEDULER_THREAD_CONTEXT&) = delete;
+    MFX_SCHEDULER_THREAD_CONTEXT(const MFX_SCHEDULER_THREAD_CONTEXT&&) = delete;
+    MFX_SCHEDULER_THREAD_CONTEXT& operator=(const MFX_SCHEDULER_THREAD_CONTEXT& other) = delete;
+    MFX_SCHEDULER_THREAD_CONTEXT& operator=(MFX_SCHEDULER_THREAD_CONTEXT&& other) = delete;
+
+    ~MFX_SCHEDULER_THREAD_CONTEXT()
+    {
+        if (threadHandle.joinable())
+            threadHandle.join();
+    }
+
+    void Stop(std::lock_guard<std::mutex>& /*lock*/)
+    {
+        state = MFX_SCHEDULER_THREAD_CONTEXT::Stopping;
+        taskAdded.notify_one();
+    }
 
     enum State {
         Waiting,  // thread is waiting for incoming tasks
@@ -47,7 +70,6 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
     };
 
     State state;                       // thread state, waiting or running
-    mfxSchedulerCore *pSchedulerCore;  // pointer to the owning core
     mfxU32 threadNum;                  // thread number assigned by the core
     std::thread threadHandle;          // thread handle
     std::condition_variable taskAdded; // cond. variable to signal new tasks
--- a/_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
+++ b/_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
@@ -41,8 +41,9 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
     {}
 
     enum State {
-        Waiting, // thread is waiting for incoming tasks
-        Running  // thread is executing a task
+        Waiting,  // thread is waiting for incoming tasks
+        Running,  // thread is executing a task
+        Stopping, // thread is going to stop
     };
 
     State state;                       // thread state, waiting or running
--- a/_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
+++ b/_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
@@ -44,11 +44,6 @@ mfxStatus mfxSchedulerCore::Initialize(const MFX_SCHEDULER_PARAM *pParam)
         MFX_SCHEDULER_PARAM* casted_param2 = &param2;
         *casted_param2 = *pParam;
     }
-    if (!param2.numberOfThreads) {
-        // that's just in case: core, which calls scheduler, is doing
-        // exactly the same what we are doing below
-        param2.numberOfThreads = vm_sys_info_get_cpu_num();
-    }
     return Initialize2(&param2);
 }
 
@@ -80,14 +75,8 @@ mfxStatus mfxSchedulerCore::Initialize2(const MFX_SCHEDULER_PARAM2 *pParam)
 
     if (MFX_SINGLE_THREAD != m_param.flags)
     {
-        if (m_param.numberOfThreads && m_param.params.NumThread) {
-            // use user-overwritten number of threads
-            m_param.numberOfThreads = m_param.params.NumThread;
-        }
-        if (!m_param.numberOfThreads) {
-            return MFX_ERR_UNSUPPORTED;
-        }
-        if (m_param.numberOfThreads == 1) {
+        // use user-overwritten number of threads
+        if (m_param.params.NumThread == 1) {
             // we need at least 2 threads to avoid dead locks
             return MFX_ERR_UNSUPPORTED;
         }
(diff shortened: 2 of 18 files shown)


_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
```diff
@@ -417,25 +417,34 @@ mfxStatus MFXVideoDECODE_DecodeHeader(mfxSession session, mfxBitstream *bs, mfxV
...
         {
             // create a new instance
             session->m_pDECODE.reset(session->Create<VideoDECODE>(*par));
             MFX_CHECK(session->m_pDECODE.get(), MFX_ERR_INVALID_VIDEO_PARAM);
         }
 
-    mfxRes = session->m_pDECODE->Init(par);
+        mfxRes = session->m_pDECODE->Init(par);
+    }
+    catch(MFX_CORE_CATCH_TYPE)
```
dmitry-gurulev: IMO, having _catch(...)_ is much more readable than _MFX_CORE_CATCH_TYPE_ Do we really need the latter one?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -116,15 +105,17 @@ mfxStatus mfxSchedulerCore::Initialize2(const MFX_SCHEDULER_PARAM2 *pParam)
 
             auto thread_proc = [this](MFX_SCHEDULER_THREAD_CONTEXT* ctx) { ThreadProc(ctx); };
             // start threads
-            for (i = 0; i < m_param.numberOfThreads; ++i)
+            for (i = 0; i < m_param.params.NumThread; ++i)
             {
```
dmitry-gurulev: It's stated that sessions are started without any threads but here we're creating at least 2 threads.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -37,16 +38,25 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
 
     MFX_SCHEDULER_THREAD_CONTEXT() = delete;
 
-    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id, SchedulerThreadFunc func)
+    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id)
       : state(State::Waiting)
       , threadNum(id)
-      , threadHandle([this, func](){ func(this); })
+      , threadHandle([this, start_future=start.get_future()]() {
+          start_future.wait();
```
dmitry-gurulev: if _SetScheduling_ fails we will have dangling thread here (ctx dtor will wait infinite at _threadHandle.join()_)

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dvrogozh: no, we don't. params.NumThread is mediasdk public uAPI given to the user to specify on MFXInitEx how many threads he wants to have. If he used the API we will create this number of threads and ignore everything which components said to us. However, by default (when user did not use the uAPI) we will get param.NumThread=0 here and create nothing.

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -142,6 +133,73 @@ mfxStatus mfxSchedulerCore::Initialize2(const MFX_SCHEDULER_PARAM2 *pParam)
     return MFX_ERR_NONE;
 }
 
+mfxStatus mfxSchedulerCore::SetThreadNum(const void *pComponent, mfxU32 threadNum)
+{
+    MFX_CHECK(pComponent, MFX_ERR_INVALID_HANDLE);
+
+    // If user requested certain number of threads we follow his choice
+    // doing nothing in this function.
+    if (m_param.params.NumThread)
```
dvrogozh: that's why we will not listen for the components if user requested certain threads number.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
dvrogozh: I will check, thank you.

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
dvrogozh: yep. If everyone agrees I will completely remove MFX_CORE_CATCH_TYPE from all mediasdk. Cry out loudly _now_ if you have anything against.

_studio/mfx_lib/shared/include/mfx_interface_scheduler.h
```diff
@@ -120,7 +120,8 @@ class MFXIScheduler : public MFXIUnknown
     virtual
     mfxStatus AdjustPerformance(const mfxSchedulerMessage message) = 0;
 
-
+    virtual
+    mfxStatus SetThreadNum(const void *pComponent, mfxU32 threadNum) = 0;
```
dvrogozh: @onabiull : that's the major question as of now:
1. Can we add method here?
2. Or we need MFXIScheduler3?
3. Or we can add to MFXIScheduler2?

Let's nail this down...

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
dvrogozh: done

_studio/mfx_lib/shared/include/mfx_interface_scheduler.h
onabiull: Right now and Linux only we can. But considering future changes I'd stick to option 2 - adding new interface.

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
onabiull: I believe it is right thing to do.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
dmitry-gurulev: @dvrogozh, Using of identifier initializer at lambda's capture list:
`start_future=start.get_future()`
requires C++14. AFAIR we're restricted to C++11, aren't we?

dvrogozh: Yes, that's why I changed this place. See in the updated code.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -37,15 +38,24 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
 
     MFX_SCHEDULER_THREAD_CONTEXT() = delete;
 
-    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id, SchedulerThreadFunc func)
+    MFX_SCHEDULER_THREAD_CONTEXT(mfxU32 id)
       : state(State::Waiting)
       , threadNum(id)
-      , threadHandle([this, func](){ func(this); })
+      , threadHandle([this]() {
+          start.get_future().wait();
```
dvrogozh: @dmitry-gurulev : here it is changed to avoid c++14

_studio/shared/include/mfxvideo++int.h
```diff
@@ -499,6 +529,12 @@ class VideoUSER
     // Get the plugin's threading policy
     virtual
     mfxTaskThreadingPolicy GetThreadingPolicy(void) {return MFX_TASK_THREADING_DEFAULT;}
+    virtual
+    mfxStatus GetThreadNum(mfxU32& threadNum) {
+        threadNum = std::thread::hardware_concurrency();
```
mgonchar: actually std::thread::hardware_concurrency could return 0, so I believe it should be
```
// some mediasdk components require 2 threads as minimum
threadNum = std::max(std::thread::hardware_concurrency(), 2u);
```

Why we returning status here? Maybe better to return actual number of threads?

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -31,25 +34,56 @@ class mfxSchedulerCore;
...
     State state;                       // thread state, waiting or running
-    mfxSchedulerCore *pSchedulerCore;  // pointer to the owning core
     mfxU32 threadNum;                  // thread number assigned by the core
-    std::thread threadHandle;          // thread handle
     std::condition_variable taskAdded; // cond. variable to signal new tasks
+    std::promise<void> start;
+    SchedulerThreadFunc threadFunc;
+    std::thread threadHandle;          // thread handle
 
     mfxU64 workTime;                   // integral working time
```
mgonchar: shouldn't it be switched to std::chrono in future?

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -31,25 +34,56 @@ class mfxSchedulerCore;
...
+          start.get_future().wait();
+          if (threadFunc) threadFunc(this);
+      })
       , workTime(0)
       , sleepTime(0)
-    {}
+    {
+    }
+
+    void Start(SchedulerThreadFunc func)
```
mgonchar: tricky thing about should it be reference or not. Potentially copying of std::function may cause deep copy involving heap allocation (but as I see you are using lambdas with small capture to pass here, so some internal optimization for small objects may be used which result in cheap copy).

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -31,25 +34,56 @@ class mfxSchedulerCore;
...
+    {
+        threadFunc = func;
```
mgonchar: it definitely should be move-assignment

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -149,8 +130,74 @@ mfxStatus mfxSchedulerCore::Initialize2(const MFX_SCHEDULER_PARAM2 *pParam)
...
+    else
+        m_pThreadsMap.erase(pComponent);
+
+    threadNum = 0; // reuse it now to count number of total threads
+    for (auto it: m_pThreadsMap) {
+        threadNum += it.second;
+    }
+
+    mfxU32 maxParallelism = std::thread::hardware_concurrency();
+    if (threadNum > maxParallelism) threadNum = maxParallelism;
```
mgonchar: As stated in https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency
"Number of concurrent threads supported. If the value is not well defined or not computable, returns"

So maybe add a check for zero in following way:
```
mfxU32 maxParallelism = std::thread::hardware_concurrency();
if (maxParallelism && threadNum > maxParallelism)
     threadNum = maxParallelism;
```

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
dvrogozh: probably yes. But that's definitely out of scope of this PR

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
dvrogozh: I will follow your guidance:).... Since you are not sure I will remain it as is for now.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
dvrogozh: std::move then, right?

_studio/shared/include/mfxvideo++int.h
dvrogozh: thank's for suggestion for std::max

and you saw the answer to the second question in some other patch: we may not always determine number of threads for the legacy components. for these cases we will return an error.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
mgonchar: yep,
```
threadFunc = std::move(func);
```
it should remove one copy

_studio/shared/include/mfxvideo++int.h
mgonchar: yes, I got it

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
mgonchar: yes, let's remain

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -41,8 +41,9 @@ struct MFX_SCHEDULER_THREAD_CONTEXT
     {}
 
     enum State {
-        Waiting, // thread is waiting for incoming tasks
-        Running  // thread is executing a task
+        Waiting,  // thread is waiting for incoming tasks
+        Running,  // thread is executing a task
+        Stopping, // thread is going to stop
```
dmitry-gurulev: IMO, [Stopping] is quite weird state for thread. If [bQuit] is really unacceptable there is yet another common trick - submit special 'stop task' that will signal a thread to stop.

dvrogozh: I know. And I feel your pain. Right now in current scheduler implementation quality that's the best I can suggest. Right now we don't have a reasonable way to send a special task.

dvrogozh:
@mgonchar, @dmitry-gurulev : can you, please, take a look once more time?

@onabiull : can we, please, schedule some validation for this PR? We need functional one at the first place.

_studio/mfx_lib/scheduler/include/mfx_scheduler_core_thread.h
```diff
@@ -31,25 +34,56 @@ class mfxSchedulerCore;
...
+        start.set_value();
+    }
+
+    void Stop(std::lock_guard<std::mutex>& /*lock*/)
+    {
+        state = MFX_SCHEDULER_THREAD_CONTEXT::Stopping;
+        taskAdded.notify_one();
+    }
+
+    ~MFX_SCHEDULER_THREAD_CONTEXT()
```
mgonchar: let's follow rule of five and declare explicitly following:
- copy ctor
- move ctor
- copy =
- move =

dvrogozh: done

_studio/shared/include/mfxvideo++int.h
```diff
@@ -299,6 +307,12 @@ class VideoPAK
     mfxStatus Close(void) = 0;
     virtual
     mfxTaskThreadingPolicy GetThreadingPolicy(void) {return MFX_TASK_THREADING_DEFAULT;}
+    virtual
+    mfxStatus GetThreadNum(mfxU32& threadNum) {
+        // some mediasdk components require 2 threads as a minimum
+        threadNum = std::max(std::thread::hardware_concurrency(), 2u);
```
dmitry-gurulev: I believe resulting thread number should depend on threading policy, otherwise we have an ambiguity - default thread policy [MFX_TASK_THREADING_DEFAULT] states that we doesn't support parallel task execution, all task is executed by thread #0 only and, at the same time we request at least two threads (or number of hardware available ones)

PR #886, opened 2018-11-09
[AVCe] Fix for Paff SingleFieldMode

kelvinhu325:
Signed-off-by: Hu Kelvin <kelvin.hu@intel.com>

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
@@ -2664,7 +2664,8 @@ mfxStatus ImplementationAvc::AsyncRoutine(mfxBitstream * bs)
             }
         }
 
-        task->m_singleFieldMode = IsOn(extFeiParams->SingleFieldProcessing);
+        // In case of progressive frames in PAFF mode need to switch the flag off to prevent m_fieldCounter changes
+        task->m_singleFieldMode = (task->m_fieldPicFlag != 0) && IsOn(extFeiParams->SingleFieldProcessing);
 
 
         for (mfxU32 f = 0; f <= task->m_fieldPicFlag; f++)


_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2664,7 +2664,7 @@ mfxStatus ImplementationAvc::AsyncRoutine(mfxBitstream * bs)
             }
         }
 
-        task->m_singleFieldMode = IsOn(extFeiParams->SingleFieldProcessing);
+        task->m_singleFieldMode = (task->m_fieldPicFlag == 0) ? false : IsOn(extFeiParams->SingleFieldProcessing);
```
mgonchar: @kelvinhu325 where the problem is really arising in case of progressive frame + this flag set to true? I'm not sure that the fix should be in reforcing this flag. I see code in common part for PreENC and ENCODE which has handling of this situation: https://github.com/Intel-Media-SDK/MediaSDK/blob/97de8773fa0f66e7a2d0f9fd9de878fb698a5fbd/_studio/mfx_lib/fei/h264_common/fei_common.cpp#L827

kelvinhu325: @mgonchar yes, this change is for special PAFF case when encoding progressive frames with enabling SingleFieldProcessing, SingleFieldMode control is only for interlaced field case, it impacts ExBuffer selecting for first or second field in Execute() function and also impact if second Execute() is called in AsyncRoutine() function,   so task->m_singleFieldMode should be false for progressive frame here to avoid incorrect operation for fields, it does not need second Execute() at all for progressive frame.

mgonchar: got it, will check the code you mentioned first to fully understand the problem

mgonchar: @kelvinhu325 you are referencing to this code https://github.com/Intel-Media-SDK/MediaSDK/blob/f3aa42458778dffa1d6b1333e6379f139df98e71/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp#L2734 right?

So in case of progressive with this flag on, internal state will switch to expect another field and this would break all the subsequent processing

kelvinhu325: @mgonchar almost yes, my change makes internal state work as a progressive frame instead of trigger two fields encoding, and task->m_singleFieldMode also impacts m_ddi->Execute() behavior as my description in former comment

mgonchar: ok, I got your point. Looks like this is really the easiest way to get this functionality work right now.

Could you change this patch as following to remove unnecessary branching:
```
// In case of progressive frames in PAFF mode need to switch the flag off to prevent m_fieldCounter changes
task->m_singleFieldMode = (task->m_fieldPicFlag != 0) && IsOn(extFeiParams->SingleFieldProcessing);
```

then I would mark it as ready to merge

kelvinhu325: @mgonchar ok, I upstream my change again as your suggestion, pls check it, thanks:)

mgonchar:
@kelvinhu325 thank you for patch

PR #887, opened 2018-11-09
[VP9d] Fix of AspectRatio initialization

ndomnina:
To initialize the default values, need to use m_vInitPar instead of m_vPar.
Otherwise, the new values will not be used.


ndomnina:
The GetVideoParam should return the changed values after the Init. 
In the current version, if the initial value of the AspectRatio is 0, the GetVideoParam returns 0, although the value has already been changed.

ndomnina:
As I understand it, GetVideoParam() apply the changed value after the initialization phase.
After
```
par->mfx = m_vPar.mfx
```
follows
```
par->mfx.FrameInfo.AspectRatioH = m_vInitPar.mfx.FrameInfo.AspectRatioH;
par->mfx.FrameInfo.AspectRatioW = m_vInitPar.mfx.FrameInfo.AspectRatioW;
```

ndomnina:
Fixed. Now the default value is written in both m_vInitPar and m_vPar.

PR #889, opened 2018-11-09
h265d: add NumPicTotalCurr sanity check

dmitry-gurulev:
Issue: MDP-51203


dmitry-gurulev:
Accord. to ITU-T H.265 A.4.1 General tier and level limits the value of NumPicTotalCurr is restricted to be less than or equal to 8. We will treat a slice that violates this restriction as _invalid_ and skip it.

PR #890, opened 2018-11-09
[VP9e] Fix the order of parameter buffer submission

eshiryae:
Temporal layers buffer should be sent to driver before frame rate/rate control
buffers.


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -1110,28 +1118,30 @@ mfxStatus VAAPIEncoder::Execute(
 
         configBuffers[buffersCount++] = m_packedHeaderDataBufferId;
 
-//*********
-
         // 8. hrd parameters
         configBuffers[buffersCount++] = m_hrdBufferId;
-        // 9. RC parameters
+        
+        // 9. temporal layers
```
lakulako: Does buffers order matters? Don't introduce changes if it doesn't.

lakulako: Then OK, but insist them to document it. Because it looks out unnatural.

PR #891, opened 2018-11-09
[VPP] code cleanup for further background color functions

IlyaSergeyev:
Choosing of background color moved into separate function.

### DIFF
--- a/_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
+++ b/_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
@@ -1005,6 +1005,13 @@ namespace MfxHwVideoProcessing
         public:
             void SetCmDevice(CmDevice * device) { m_pCmDevice = device; }
     };
+
+    class VpUnsupportedError : public std::exception
+    {
+    public:
+        VpUnsupportedError() : std::exception() { assert(!"VpUnsupportedError"); }
+    };
+
 }; // namespace MfxHwVideoProcessing
 
 
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -5081,33 +5081,105 @@ template <class T> T get_total_area(std::vector< cRect<T> > &rects) {
 inline
 mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
 {
-    VM_ASSERT(bit_depth);
+    if (bit_depth<8)
+            throw MfxHwVideoProcessing::VpUnsupportedError();
 
     mfxU64 const shift = bit_depth - 8;
-    mfxU64 const max_val = 256 << shift;
+    mfxU64 const max_val = (1 << bit_depth) - 1;
 
     return
-        ((mfxU64)                                 (max_val - 1) << 48) |
-        ((mfxU64)VPP_RANGE_CLIP(Y, (16 << shift), (235 << shift))     << 32) |
-        ((mfxU64)VPP_RANGE_CLIP(U, (16 << shift), (240 << shift))     << 16) |
-        ((mfxU64)VPP_RANGE_CLIP(V, (16 << shift), (240 << shift))     <<  0)
-        ;
-};
+        ((mfxU64) max_val << 48) |
+        ((mfxU64) VPP_RANGE_CLIP(Y, (16 << shift), (235 << shift))     << 32) |
+        ((mfxU64) VPP_RANGE_CLIP(U, (16 << shift), (240 << shift))     << 16) |
+        ((mfxU64) VPP_RANGE_CLIP(V, (16 << shift), (240 << shift))     <<  0);
+}
+
+inline
+mfxU64 make_back_color_argb(mfxU16 bit_depth, mfxU16 R, mfxU16 G, mfxU16 B)
+{
+    if (bit_depth<8)
+            throw MfxHwVideoProcessing::VpUnsupportedError();
+
+    mfxU64 const max_val = (1 << bit_depth) - 1;
+
+    return ((mfxU64) max_val << 48) |
+           ((mfxU64) VPP_RANGE_CLIP(R, 0, max_val) << 32) |
+           ((mfxU64) VPP_RANGE_CLIP(G, 0, max_val) << 16) |
+           ((mfxU64) VPP_RANGE_CLIP(B, 0, max_val) << 0);
+}
 
 inline
 mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
 {
-    assert(bit_depth);
+    if (bit_depth<8)
+            throw MfxHwVideoProcessing::VpUnsupportedError();
 
     mfxU64 const shift = bit_depth - 8;
-    mfxU64 const min_val = 16 << shift, max_val = 256 << shift;
-    return
-        ((max_val - 1ULL) << 48) |
-        ( min_val      << 32) |
-        ( max_val / 2  << 16) |
-        ( max_val / 2  <<  0)
-        ;
-};
+    mfxU16 const min_val = 16 << shift, max_val = 256 << shift;
+    return make_back_color_yuv(bit_depth, min_val, max_val / 2, max_val / 2);
+}
+
+inline
+mfxU64 make_def_back_color_argb(mfxU16 bit_depth)
+{
+    return make_back_color_argb(bit_depth, 0, 0, 0);
+}
+
+static
+mfxU64 get_background_color(const mfxVideoParam & videoParam)
+{
+    for (mfxU32 i = 0; i < videoParam.NumExtParam; i++)
+    {
+        if (videoParam.ExtParam[i]->BufferId == MFX_EXTBUFF_VPP_COMPOSITE)
+        {
+            mfxExtVPPComposite *extComp = (mfxExtVPPComposite *) videoParam.ExtParam[i];
+            switch (videoParam.vpp.Out.FourCC) {
+                case MFX_FOURCC_NV12:
+                case MFX_FOURCC_YV12:
+                case MFX_FOURCC_NV16:
+                case MFX_FOURCC_YUY2:
+                case MFX_FOURCC_AYUV:
+                    return make_back_color_yuv(8, extComp->Y, extComp->U, extComp->V);
+                case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+                case MFX_FOURCC_Y210:
+                case MFX_FOURCC_Y410:
+#endif
+                case MFX_FOURCC_P210:
+                    return make_back_color_yuv(10, extComp->Y, extComp->U, extComp->V);
+                case MFX_FOURCC_RGB4:
+                case MFX_FOURCC_BGR4:
+                    return make_back_color_argb(8, extComp->R, extComp->G, extComp->B);
+                default:
+                    break;
+            }
+        }
+    }
+
+    switch (videoParam.vpp.Out.FourCC)
+    {
+        case MFX_FOURCC_NV12:
+        case MFX_FOURCC_YV12:
+        case MFX_FOURCC_NV16:
+        case MFX_FOURCC_YUY2:
+        case MFX_FOURCC_AYUV:
+            return make_def_back_color_yuv(8);
+        case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y210:
+        case MFX_FOURCC_Y410:
+#endif
+        case MFX_FOURCC_P210:
+            return make_def_back_color_yuv(10);
+        case MFX_FOURCC_RGB4:
+        case MFX_FOURCC_BGR4:
+            return make_def_back_color_argb(8);
+        default:
+            break;
+    }
+
+    throw MfxHwVideoProcessing::VpUnsupportedError();
+}
 
 //---------------------------------------------------------
 // Do internal configuration
@@ -5135,27 +5207,7 @@ mfxStatus ConfigureExecuteParams(
     config.m_surfCount[VPP_IN]  = 1;
     config.m_surfCount[VPP_OUT] = 1;
 
-    mfxU64 def_back_color = 0xffff000000000000;
-    if (videoParam.vpp.Out.FourCC == MFX_FOURCC_NV12 ||
-        videoParam.vpp.Out.FourCC == MFX_FOURCC_YV12 ||
-        videoParam.vpp.Out.FourCC == MFX_FOURCC_NV16 ||
-        videoParam.vpp.Out.FourCC == MFX_FOURCC_YUY2 ||
-        videoParam.vpp.Out.FourCC == MFX_FOURCC_AYUV )
-    {
-        def_back_color = make_def_back_color_yuv(8);
-    }
-    else if(videoParam.vpp.Out.FourCC == MFX_FOURCC_P010 ||
-#if (MFX_VERSION >= 1027)
-            videoParam.vpp.Out.FourCC == MFX_FOURCC_Y210 ||
-            videoParam.vpp.Out.FourCC == MFX_FOURCC_Y410 ||
-#endif
-            videoParam.vpp.Out.FourCC == MFX_FOURCC_P210)
-    {
-        def_back_color = make_def_back_color_yuv(10);
-    }
-
-    executeParams.iBackgroundColor = def_back_color;
-
+    executeParams.iBackgroundColor = get_background_color(videoParam);
 
     //-----------------------------------------------------
     for (mfxU32 j = 0; j < pipelineList.size(); j += 1)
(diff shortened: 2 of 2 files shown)


_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5104,49 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
         ;
 };
 
+static
+mfxU64 get_background_color(const mfxVideoParam & videoParam)
+{
+    mfxU64 argb_back_color = 0xffff000000000000;
+    mfxU32 targetFourCC = videoParam.vpp.Out.FourCC;
+
+    switch (targetFourCC)
```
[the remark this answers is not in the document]
IlyaSergeyev: No. We will need videoParam.ExtParam to set background in future commits.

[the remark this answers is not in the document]
mgonchar: I think Dmitry meant not to remove ```mfxVideoParam``` from input parameters, but to get rid of ```targetFourCC``` in following way:
```
switch (videoParam.vpp.Out.FourCC)
```

so just one tmp variable less

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5104,6 +5104,49 @@ mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
...
+        case MFX_FOURCC_BGR4:
+            for (mfxU32 i = 0; i < videoParam.NumExtParam; i++)
+            {
+                if (videoParam.ExtParam[i]->BufferId == MFX_EXTBUFF_VPP_COMPOSITE)
+                {
+                    mfxExtVPPComposite *extComp = (mfxExtVPPComposite *) videoParam.ExtParam[i];
+                    argb_back_color = ((mfxU64) 0xff << 48) |
+                                      ((mfxU64) VPP_RANGE_CLIP(extComp->R, 0, 255) << 32) |
+                                      ((mfxU64) VPP_RANGE_CLIP(extComp->G, 0, 255) << 16) |
+                                      ((mfxU64) VPP_RANGE_CLIP(extComp->B, 0, 255) << 0);
```
mgonchar: add ```break;``` to this loops body, since I believe there could be only one such buffer

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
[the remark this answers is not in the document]
IlyaSergeyev: Oops!

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
IlyaSergeyev: Added. There will be other buffer type with background color, but in next commits.

IlyaSergeyev:
@lakulako 
1. Good question. As I understand now, only 8 bit per component are used and additional space is reserved for Ice Lake 10 bit color schemes (A2R10G10B10). I keep old code that used only low byte of component, but maybe we need rewrite code for using 16 bit components in right way (using high byte of component instead low one for 64bit ARGB). @dvrogozh can you, please, comment this?

2. It's black (0xffff000000000000)

dvrogozh:
emmm... I don't know... @onabiull : can you comment?

onabiull:
I believe statement "I keep old code that used only low byte of component, but maybe we need rewrite code for using 16 bit components in right way (using high byte of component instead low one for 64bit ARGB)." cries to be written as comment in the code :)

lakulako:
@IlyaSergeyev 
> 2. It's black (0xffff000000000000)
0xffff here is **maximum** U16. Don't you mean that here maximum is black and minimum is white?
Usually black is minimum value (0).

IlyaSergeyev:
@lakulako 
A = 0xFFFF 
R = 0
G = 0
B = 0
It's a non transparent black

lakulako:
OK, I meant only A (transparency). Having it 0 by default would be simpler.
BTW, I see 2 issues in make_def_back_color_yuv(). It should assert(bit_depth >= 8); and use either studio swing (16, 235) or full swing (0, 255).

IlyaSergeyev:
0 means do not use colorfill for driver. Why MR closed?

lakulako:
Sorry, closed by error!

IlyaSergeyev:
@lakulako code updated, could you please, take a look

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5090,20 +5090,104 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+                    back_color = make_back_color_yuv(10, extComp->Y, extComp->U, extComp->V);
+                    break;
+                case MFX_FOURCC_RGB4:
+                case MFX_FOURCC_BGR4:
+                    back_color = make_back_color_argb(8, extComp->R, extComp->G, extComp->B);
+                    break;
+                default:
+                    break;
+            }
+        }
```
mgonchar: and again, I think you can break after you've found MFX_EXTBUFF_VPP_COMPOSITE :)

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5090,20 +5090,104 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
 
 inline
-mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
+mfxU64 make_back_color_argb(mfxU16 bit_depth, mfxU16 R, mfxU16 G, mfxU16 B)
 {
-    assert(bit_depth);
+    VM_ASSERT(bit_depth >=8);
+
+    mfxU64 const shift = bit_depth - 8;
+    mfxU64 const max_val = 256 << shift;
```
lakulako: A bit simpler:
mfxU64 const max_val = 1 << bit_depth;
also -1 here is possible

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5090,20 +5090,104 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
 
+    return ((mfxU64) (max_val - 1ULL) << 48) |
+           ((mfxU64) VPP_RANGE_CLIP(R, 0, 255) << 32) |
+           ((mfxU64) VPP_RANGE_CLIP(G, 0, 255) << 16) |
+           ((mfxU64) VPP_RANGE_CLIP(B, 0, 255) << 0);
```
lakulako: Why no dependency on bit_depth? Clipping range should increase with bit_depth.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5090,20 +5090,104 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+};
+
+inline
+mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
+{
+    // A = max, Y=min, U,V = max/2
+    assert(bit_depth >= 8);
```
lakulako: assert or VM_ASSERT as above?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5090,20 +5090,104 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
-mfxU64 make_def_back_color_yuv(mfxU16 bit_depth)
+mfxU64 make_back_color_argb(mfxU16 bit_depth, mfxU16 R, mfxU16 G, mfxU16 B)
 {
-    assert(bit_depth);
+    VM_ASSERT(bit_depth >=8);
+
+    mfxU64 const shift = bit_depth - 8;
+    mfxU64 const max_val = 256 << shift;
 
+    return ((mfxU64) (max_val - 1ULL) << 48) |
```
lakulako: No need in ULL, same below.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5090,20 +5090,104 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+    mfxU16 const min_val = 16 << shift, max_val = 256 << shift;
     return
-        ((max_val - 1ULL) << 48) |
-        ( min_val      << 32) |
-        ( max_val / 2  << 16) |
-        ( max_val / 2  <<  0)
+        ((mfxU64)(max_val - 1ULL) << 48) |
+        ((mfxU64)VPP_RANGE_CLIP(min_val    , (16 << shift), (235 << shift))     << 32) |
+        ((mfxU64)VPP_RANGE_CLIP(max_val / 2, (16 << shift), (240 << shift))     << 16) |
+        ((mfxU64)VPP_RANGE_CLIP(max_val / 2, (16 << shift), (240 << shift))     <<  0)
```
lakulako: No need in VPP_RANGE_CLIP

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
lakulako: Even more:
inside if{} in switch **return make_back_color** *.
First switch for default back_color move after the loop. And return each time immediately, don't use back_color at all.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5079,31 +5079,111 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+        default:
+            break;
+    }
+
+    for (mfxU32 i = 0; i < videoParam.NumExtParam; i++)
+    {
+        if (videoParam.ExtParam[i]->BufferId == MFX_EXTBUFF_VPP_COMPOSITE)
+        {
+            mfxExtVPPComposite *extComp = (mfxExtVPPComposite *) videoParam.ExtParam[i];
+            switch (videoParam.vpp.Out.FourCC) {
```
dvrogozh: This switch will never be reached for the known color formats since you simply return from the function handling the switch above at line 5137. Probably what you wanted to write was the reverse order: first check whether user tried to program background color via ext param and if he did not chose default color.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5079,31 +5079,111 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+        case MFX_FOURCC_YUY2:
+        case MFX_FOURCC_AYUV:
+            return make_def_back_color_yuv(8);
+        case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y210:
+        case MFX_FOURCC_Y410:
+#endif
+        case MFX_FOURCC_P210:
+            return make_def_back_color_yuv(10);
```
dvrogozh: you miss RGB4 and BGR4 cases here

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5079,31 +5079,111 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+
+    mfxU64 const max_val = 1 << bit_depth;
+
+    return ((mfxU64)(max_val - 1) << 48);
 };
 
+static
+mfxU64 get_background_color(const mfxVideoParam & videoParam)
+{
+    mfxU64 back_color = make_def_back_color_argb(8);
```
dvrogozh: Bad idea. With this code we will end up returning rgb back color for non-rgb formats. I think better to move this in the switch and assign honest 0 + add assert in the "default" switch case. Well, you also can change function prototype returning mfxStatus.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5079,31 +5079,111 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
     return
-        ((max_val - 1ULL) << 48) |
-        ( min_val      << 32) |
-        ( max_val / 2  << 16) |
-        ( max_val / 2  <<  0)
-        ;
+        ((mfxU64)(max_val - 1) << 48) |
+        ((mfxU64)min_val       << 32) |
+        ((mfxU64)max_val / 2   << 16) |
+        ((mfxU64)max_val / 2   <<  0);
```
dvrogozh: Call to make_back_color_yuv() with the default parameters would be more clear.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5079,31 +5079,111 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+
+inline
+mfxU64 make_def_back_color_argb(mfxU16 bit_depth)
+{
+    // A = max, R,G,B = 0
+    VM_ASSERT(bit_depth >= 8);
+
+    mfxU64 const max_val = 1 << bit_depth;
+
+    return ((mfxU64)(max_val - 1) << 48);
```
dvrogozh: same, call to make_back_color_argb(), please.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5084,30 +5084,102 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+inline
+mfxU64 make_def_back_color_argb(mfxU16 bit_depth)
+{
+    return make_back_color_argb(bit_depth, 0, 0, 0);
+}
+
+static
+mfxU64 get_background_color(const mfxVideoParam & videoParam)
+{
+    mfxU64 back_color = 0;
```
dvrogozh: you don't need this variable actually.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5084,30 +5084,100 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
...
+        case MFX_FOURCC_P210:
+            return make_def_back_color_yuv(10);
+        case MFX_FOURCC_RGB4:
+        case MFX_FOURCC_BGR4:
+            return make_def_back_color_argb(8);
+        default:
+            break;
+    }
+
+    VM_ASSERT(0);
```
dvrogozh: Ok. This patch looks almost good except errors handling. We will get unpredictable behavior on new color formats not handled here. Thus, please, change VM_ASSERTs to c++ throw of exceptions. We probably want to have runtime error (even very generic one) rather than silent behavior with wrong color format.

This applies to all VM_ASSERTs including bit_depth ones above.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5081,33 +5081,106 @@ template <class T> T get_total_area(std::vector< cRect<T> > &rects) {
...
-};
+}
+
+inline
+mfxU64 make_back_color_argb(mfxU16 bit_depth, mfxU16 R, mfxU16 G, mfxU16 B)
+{
+    if (bit_depth<8)
+            throw MfxHwVideoProcessing::VpUnsupprtedError();
+
+    mfxU64 const max_val = 1 << bit_depth;
```
lakulako: may be make it  max_val = (1 << bit_depth) - 1;

_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
```diff
@@ -1005,6 +1005,13 @@ namespace MfxHwVideoProcessing
         public:
             void SetCmDevice(CmDevice * device) { m_pCmDevice = device; }
     };
+
+    class VpUnsupprtedError : public std::exception
+    {
+    public:
+        VpUnsupprtedError() : std::exception() { assert(!"VpUnsupprtedError"); }
```
dvrogozh: @mgonchar : can you, please, comment?

dvrogozh:
@IlyaSergeyev: please, rebase (test don't pass - I think that's fake).

IlyaSergeyev:
@lakulako @dvrogozh any new comments?

PR #893, opened 2018-11-11
[dispatcher_test] Adjusted test cases of loading plugin

semelianova:
Added return status check in and check the number of function calls in test of loading plugin. Befor the changes in ShouldLoadFromGoodPluginsCfgCorrectly case only one function (dlopen) was called and returned error. The case was considered successful, but it is wrong. Also corrected some parameters in other cases.

### DIFF
--- a/tests/mfx_dispatch/linux/mfx_dispatch_test_mocks.h
+++ b/tests/mfx_dispatch/linux/mfx_dispatch_test_mocks.h
@@ -105,7 +105,7 @@ public:
     char* FeedEmulatedPluginsCfgLines(char * str, int num, FILE * stream);
     mfxStatus ReturnMockPlugin(mfxPluginUID uid, mfxPlugin* pPlugin)
     {
-        pPlugin = &m_mock_plugin;
+        *pPlugin = m_mock_plugin;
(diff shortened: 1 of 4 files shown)


tests/mfx_dispatch/linux/mfx_dispatch_test_mocks.h
```diff
@@ -105,7 +105,7 @@ class MockCallObj : public CallObj
     char* FeedEmulatedPluginsCfgLines(char * str, int num, FILE * stream);
     mfxStatus ReturnMockPlugin(mfxPluginUID uid, mfxPlugin* pPlugin)
     {
-        pPlugin = &m_mock_plugin;
+        *pPlugin = m_mock_plugin;
```
mgonchar: pointer check is required, pPlugin might be nullptr here. Other solution - change function signature to accept reference

semelianova: This function signature is determined by CreatePlugin function signature. I will add nullptr check for pPlugin. Changing of pPlugin pointer and further check are impossible.

PR #895, opened 2018-11-13
CORE: Release Scheduler before Core

akokoshn:
-Fix possible issue: during destroy instance of scheduler may be used m_pCore,
 so need release m_pCore only after scheduler


_studio/mfx_lib/shared/src/mfx_session.cpp
```diff
@@ -612,11 +612,13 @@ void _mfxSession::Cleanup(void)
     m_pVPP.reset();
     m_pDECODE.reset();
     m_pENCODE.reset();
-    m_pCORE.reset();
 
     // release m_pScheduler and m_pSchedulerAllocated
     ReleaseScheduler();
 
+    // release core
```
dmitry-gurulev: Please add comment (why we need destroy core after scheduler) here.

PR #896, opened 2018-11-13
Fixed possible nullptr dereference

Losev-Ivan:

### DIFF
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -1703,7 +1703,13 @@ void TaskManager::UpdatePTS_Mode30i60p(
     mfxU32 taskIndex,
     mfxStatus *intSts)
 {
-    if (input == NULL)
+    if (output == nullptr)
+    {
+        *intSts = MFX_ERR_MORE_SURFACE;
+        return;
+    }
+
+    if (input == nullptr)
     {
         output->Data.TimeStamp = (mfxU64)MFX_TIME_STAMP_INVALID;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -1703,7 +1703,13 @@ void TaskManager::UpdatePTS_Mode30i60p(
     mfxU32 taskIndex,
     mfxStatus *intSts)
 {
-    if (input == NULL)
+    if (output == nullptr)
+    {
+        *intSts = MFX_ERR_MORE_SURFACE;
```
DenWolf: I'm not sure that this is correct status in the current case.

I mean : if we didn't check this parameter for nullptr previously and dereference in below in several places - maybe in this case "output" always not nullptr? So, we if it can be null maybe we need to return status MFX_ERR_UNDEFINED_BEHAVIOR or MFX_ERR_NULL_PTR of throw an exception ?

Please, clarify it.

Losev-Ivan: Look at line 1391. In case of SimpleMode we check that input and output is not equal to null, and I think that in case of Mode30i60p we should do the same check. 

About return MFX_ERR_MORE_SURFACE. At line 1294: if mode 30i60p is not enabled then we may call DoCpuFRC_AndUpdatePTS where in case of output == null we return MFX_ERR_MORE_SURFACE. I suppose that in case of mode 30i60p we should return same status.

onabiull: @DenWolf are you ok with this change?

PR #898, opened 2018-11-13
[dispatcher_test] Strengthening of plugin test requirements

semelianova:
Added return status check and check the number of function calls in test of loading plugin. Before the changes in ShouldLoadFromGoodPluginsCfgCorrectly case only one function (dlopen) was called and returned error. The case was considered successful, but it is wrong.

### DIFF
--- a/tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
+++ b/tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
@@ -59,19 +59,25 @@ TEST_P(DispatcherPluginsTestParametrized, ShouldLoadFromGoodPluginsCfgCorrectly)
         // and the parsed plugin list is stored as a global variable in the
         // dispatcher lib. Therefore, the calls below will only be executed once.
         std::string path = std::string(MFX_PLUGINS_CONF_DIR) + "/plugins.cfg";
-        EXPECT_CALL(mock, fopen(StrEq(path.c_str()),_)).WillRepeatedly(Return(MOCK_FILE_DESCRIPTOR));
-        EXPECT_CALL(mock, fgets(_,_,MOCK_FILE_DESCRIPTOR)).WillRepeatedly(Invoke(&mock, &MockCallObj::FeedEmulatedPluginsCfgLines));
-        EXPECT_CALL(mock, fclose);
+        EXPECT_CALL(mock, fopen(StrEq(path.c_str()), _)).Times(1).WillRepeatedly(Return(MOCK_FILE_DESCRIPTOR));
+        EXPECT_CALL(mock, fgets(_, _, MOCK_FILE_DESCRIPTOR)).Times(AtLeast(1)).WillRepeatedly(Invoke(&mock, &MockCallObj::FeedEmulatedPluginsCfgLines));
+        EXPECT_CALL(mock, fclose).Times(1).WillRepeatedly(Return(MFX_ERR_NONE));
         successfully_parsed_at_least_once = true;
     }
 
-    EXPECT_CALL(mock, dlopen(StrEq(plugin.Path.c_str()),_));
-    EXPECT_CALL(mock, dlsym(_,StrEq("CreatePlugin"))).WillRepeatedly(Return(reinterpret_cast<void*>(CreatePluginWrap)));
-    EXPECT_CALL(mock, CreatePlugin(plugin.PluginUID,_)).WillRepeatedly(Invoke(&mock, &MockCallObj::ReturnMockPlugin));
-    EXPECT_CALL(mock, GetPluginParam(&mock.m_mock_plugin,_)).WillRepeatedly(Return(MFX_ERR_NONE));
-    EXPECT_CALL(mock, MFXVideoUSER_Register(session,plugin.Type, &mock.m_mock_plugin)).WillRepeatedly(Return(MFX_ERR_NONE));
+    EXPECT_CALL(mock, dlopen(StrEq(plugin.Path.c_str()), _)).Times(1).WillRepeatedly(Return(MOCK_DLOPEN_HANDLE));
+    EXPECT_CALL(mock, dlsym(_, StrEq("CreatePlugin"))).Times(1).WillRepeatedly(Return(reinterpret_cast<void*>(CreatePluginWrap)));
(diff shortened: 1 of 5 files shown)


tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
```diff
@@ -59,19 +59,20 @@ TEST_P(DispatcherPluginsTestParametrized, ShouldLoadFromGoodPluginsCfgCorrectly)
         // and the parsed plugin list is stored as a global variable in the
         // dispatcher lib. Therefore, the calls below will only be executed once.
         std::string path = std::string(MFX_PLUGINS_CONF_DIR) + "/plugins.cfg";
-        EXPECT_CALL(mock, fopen(StrEq(path.c_str()),_)).WillRepeatedly(Return(MOCK_FILE_DESCRIPTOR));
-        EXPECT_CALL(mock, fgets(_,_,MOCK_FILE_DESCRIPTOR)).WillRepeatedly(Invoke(&mock, &MockCallObj::FeedEmulatedPluginsCfgLines));
-        EXPECT_CALL(mock, fclose);
+        EXPECT_CALL(mock, fopen(StrEq(path.c_str()), _)).Times(AtLeast(1)).WillRepeatedly(Return(MOCK_FILE_DESCRIPTOR));
```
dvrogozh: I think we open plugins.cfg only once. Shouldn't we expect fopen/fclose `Times(1)`?

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
```diff
@@ -59,19 +59,20 @@ TEST_P(DispatcherPluginsTestParametrized, ShouldLoadFromGoodPluginsCfgCorrectly)
...
-    EXPECT_CALL(mock, GetPluginParam(&mock.m_mock_plugin,_)).WillRepeatedly(Return(MFX_ERR_NONE));
-    EXPECT_CALL(mock, MFXVideoUSER_Register(session,plugin.Type, &mock.m_mock_plugin)).WillRepeatedly(Return(MFX_ERR_NONE));
+    EXPECT_CALL(mock, dlopen(StrEq(plugin.Path.c_str()), _)).Times(AtLeast(1)).WillRepeatedly(Return(MOCK_DLOPEN_HANDLE));
+    EXPECT_CALL(mock, dlsym(_, StrEq("CreatePlugin"))).Times(AtLeast(1)).WillRepeatedly(Return(reinterpret_cast<void*>(CreatePluginWrap)));
+    EXPECT_CALL(mock, CreatePlugin(plugin.PluginUID, _)).Times(AtLeast(1)).WillRepeatedly(Invoke(&mock, &MockCallObj::ReturnMockPlugin));
+    EXPECT_CALL(mock, GetPluginParam(mock.m_mock_plugin.pthis, _)).Times(AtLeast(1)).WillRepeatedly(DoAll(SetArgPointee<1>(plugin), Return(MFX_ERR_NONE)));
+    EXPECT_CALL(mock, MFXVideoUSER_Register(MOCK_SESSION_HANDLE, plugin.Type, _)).Times(AtLeast(1)).WillRepeatedly(Return(MFX_ERR_NONE));
 
     mfxStatus sts = MFXVideoUSER_Load(session, &plugin.PluginUID, 0);
+    ASSERT_EQ(sts, MFX_ERR_NONE);
```
dvrogozh: Why we don't exercise UnLoad and dlclose() here?

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
```diff
@@ -59,19 +59,20 @@ TEST_P(DispatcherPluginsTestParametrized, ShouldLoadFromGoodPluginsCfgCorrectly)
...
+        EXPECT_CALL(mock, fclose).Times(AtLeast(1)).WillRepeatedly(Return(MFX_ERR_NONE));
         successfully_parsed_at_least_once = true;
     }
 
-    EXPECT_CALL(mock, dlopen(StrEq(plugin.Path.c_str()),_));
-    EXPECT_CALL(mock, dlsym(_,StrEq("CreatePlugin"))).WillRepeatedly(Return(reinterpret_cast<void*>(CreatePluginWrap)));
-    EXPECT_CALL(mock, CreatePlugin(plugin.PluginUID,_)).WillRepeatedly(Invoke(&mock, &MockCallObj::ReturnMockPlugin));
-    EXPECT_CALL(mock, GetPluginParam(&mock.m_mock_plugin,_)).WillRepeatedly(Return(MFX_ERR_NONE));
-    EXPECT_CALL(mock, MFXVideoUSER_Register(session,plugin.Type, &mock.m_mock_plugin)).WillRepeatedly(Return(MFX_ERR_NONE));
+    EXPECT_CALL(mock, dlopen(StrEq(plugin.Path.c_str()), _)).Times(AtLeast(1)).WillRepeatedly(Return(MOCK_DLOPEN_HANDLE));
```
dvrogozh: shouldn't we expect dlopen() `Times(1)` as well? same question for other expectations below.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
semelianova: You are right. It will be replaced.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
semelianova: You are right. It will be replaced.

tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp
semelianova: I think it is good idea to check unload plugin, but this case should check only load function as I understood. Maybe create new case for it or add here. What do you think?

dvrogozh: I would adopt the following rule. If test is designed to check for the successful loading/allocation/etc. then we should have symmetric checks meaning to check both dlopen/dlclose or malloc/free or Load/Unload.

So, I would add unload to _this_ test.

semelianova: Ok, it will be done.

PR #901, opened 2018-11-14
vc1d: Add check after Init

Losev-Ivan:
In case of complex profile Init() will return UMC_ERR_UNSUPPORTED, but we nowhere process this situation. As a result - we get sefault.


_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder.cpp
```diff
@@ -18,6 +18,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+
```
dvrogozh: too many empty lines.... remove.

PR #902, opened 2018-11-14
AVC fei: wait for processing to be finished in QueryStatus

### LINKED ISSUE #603: MFX_ERR_UNKNOWN is returned when HAVC PREENC calls MFXVideoCORE_SyncOperation() in 1toN process
running 1toN pipeline as following:
Input 1080i -->decoder-->h264 fei encoder1-->output1
                                      -->h264 fei encoder2-->output2
                                      -->h264 fei encoder2-->output3

issue in libmfxhw64.so:
When VideoENC_PREENC::QueryStatus() checks field0 and field1 hanle result by calling VAAPIFEIPREENCEncoder::QueryStatus(field0) and VAAPIFEIPREENCEncoder::QueryStatus(field1).
Field0 surfacc is ready(VASurfaceReady), calling m_statFeedbackCache.erase() to erase vasurface. Field1 is rendering(VASurfaceRendering), returns MFX_WRN_DEVICE_BUSY.
Call VideoENC_PREENC::QueryStatus() again to check field0 and field1 hanle result.
Field0 surfacc is ready(VASurfaceReady), calling m_statFeedbackCache.erase() to erase vasurface.
When check field1, m_statFeedbackCache is NULL, return MFX_ERR_UNKNOWN.
"MFX_CHECK(indxSurf != m_statFeedbackCache.size(), MFX_ERR_UNKNOWN);"

mgonchar:
FEI behavior aligned with MSDK

Fixes #603

### DIFF
--- a/_studio/mfx_lib/fei/h264_enc/mfx_h264_enc.cpp
+++ b/_studio/mfx_lib/fei/h264_enc/mfx_h264_enc.cpp
@@ -155,8 +155,8 @@ mfxStatus VideoENC_ENC::QueryStatus(DdiTask& task)
 
     for (mfxU32 f = f_start; f <= fieldCount; f++)
     {
-        sts = m_ddi->QueryStatus(task, task.m_fid[f]);
-        MFX_CHECK(sts != MFX_WRN_DEVICE_BUSY, MFX_TASK_BUSY);
+        while ((sts = m_ddi->QueryStatus(task, task.m_fid[f])) == MFX_WRN_DEVICE_BUSY)
+            vm_time_sleep(1);
         MFX_CHECK(sts == MFX_ERR_NONE, Error(sts));
     }
 
--- a/_studio/mfx_lib/fei/h264_pak/mfx_h264_fei_pak.cpp
+++ b/_studio/mfx_lib/fei/h264_pak/mfx_h264_fei_pak.cpp
@@ -217,8 +217,8 @@ mfxStatus VideoPAK_PAK::QueryStatus(DdiTask& task)
 
     for (mfxU32 f = f_start; f <= fieldCount; f++)
     {
-        sts = m_ddi->QueryStatus(task, task.m_fid[f]);
-        MFX_CHECK(sts != MFX_WRN_DEVICE_BUSY, MFX_TASK_BUSY);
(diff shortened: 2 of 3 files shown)


_studio/mfx_lib/fei/h264_enc/mfx_h264_enc.cpp
```diff
@@ -155,8 +155,8 @@ mfxStatus VideoENC_ENC::QueryStatus(DdiTask& task)
 
     for (mfxU32 f = f_start; f <= fieldCount; f++)
     {
-        sts = m_ddi->QueryStatus(task, task.m_fid[f]);
-        MFX_CHECK(sts != MFX_WRN_DEVICE_BUSY, MFX_TASK_BUSY);
+        while ((sts = m_ddi->QueryStatus(task, task.m_fid[f])) == MFX_WRN_DEVICE_BUSY)
+            vm_time_sleep(1);
```
dvrogozh: We are supposed to have blocking synchronization on Linux which the only one having FEI support. Why we can return MFX_WRN_DEVICE_BUSY at all?

mgonchar: From MSDK side this status means that
```
        case VASurfaceRendering:
        case VASurfaceDisplaying:
            sts = MFX_WRN_DEVICE_BUSY;
```
where switch is performed on the returned value from
```
vaSts = vaQuerySurfaceStatus(m_vaDisplay, waitSurface, &surfSts);
```

According to https://github.com/Intel-Media-SDK/MediaSDK/issues/603 it can happen. I don't have any idea on beneath implementation or what is causing it

dvrogozh: You are not supposed to use vaQuerySurfaceStatus() at the first place. You **must** use vaSyncSurface(). Please, redesign the code and avoid artificial sleeps altogether. If there are bug in the driver or somewhere else preventing them - please, provide a list.

mgonchar: @dvrogozh ok, I will redesign

mgonchar:
I will implement different approach

PR #903, opened 2018-11-14
core cleanup

dvrogozh:


_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
```diff
@@ -417,25 +413,34 @@ mfxStatus MFXVideoDECODE_DecodeHeader(mfxSession session, mfxBitstream *bs, mfxV
...
+
+    try
+    {
         // check existence of component
-        if (!session->m_pDECODE.get())
+        if (!session->m_pDECODE)
         {
             // create a new instance
             session->m_pDECODE.reset(session->Create<VideoDECODE>(*par));
             MFX_CHECK(session->m_pDECODE.get(), MFX_ERR_INVALID_VIDEO_PARAM);
```
mgonchar: .get() is unnecessary in such checks

_studio/shared/src/libmfx_core.cpp
```diff
@@ -272,7 +254,7 @@ mfxStatus CommonCORE::LockFrame(mfxHDL mid, mfxFrameData *ptr)
             return MFX_ERR_INVALID_HANDLE;
         return (*pAlloc->Lock)(pAlloc->pthis, mid, ptr);
     }
-    catch(MFX_CORE_CATCH_TYPE)
```
mgonchar: MFX_CORE_CATCH_TYPE is defined as ```...```, so maybe we can get rid of the define as well?

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
dvrogozh: It is unnesessary in the honest if (smart_ptr), but our MFX_CHECK transforms to if (!!!(smart_ptr)) and compiler goes crazy.

_studio/shared/src/libmfx_core.cpp
dvrogozh: That's the last change in the series.

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
mgonchar: got it

PR #911, opened 2018-11-16
scheduler: implement GetThreadNum in components

dvrogozh:
This PR is a final step after #885 to teach components report required number of threads to the scheduler.

### DIFF
--- a/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
@@ -644,6 +644,13 @@ mfxTaskThreadingPolicy VideoDECODEH264::GetThreadingPolicy(void)
     return MFX_TASK_THREADING_SHARED;
 }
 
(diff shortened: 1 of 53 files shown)


_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
```diff
@@ -644,6 +644,13 @@ mfxTaskThreadingPolicy VideoDECODEH264::GetThreadingPolicy(void)
     return MFX_TASK_THREADING_SHARED;
 }
 
+mfxStatus VideoDECODEH264::GetThreadNum(mfxU32& threadNum)
+{
+    MFX_CHECK(m_isInit, MFX_ERR_NOT_INITIALIZED);
```
mgonchar: hm, that's why it should return status

dvrogozh: Yes. Unfortunately we can't provide exact number of threads for our legacy components at all times.

PR #920, opened 2018-11-21
smt: fixed static analysis issues

mgonchar:
- correct printf option
- fix possible memory leak

### DIFF
--- a/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
+++ b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
@@ -183,10 +183,12 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
     // create sessions, allocators
     for (i = 0; i < m_InputParamsArray.size(); i++)
     {
-        GeneralAllocator* pAllocator = new GeneralAllocator;
+        std::unique_ptr<GeneralAllocator> pAllocator(new GeneralAllocator);
         sts = pAllocator->Init(m_pAllocParam.get());
         MSDK_CHECK_STATUS(sts, "pAllocator->Init failed");
-        m_pAllocArray.push_back(pAllocator);
+
+        m_pAllocArray.push_back(pAllocator.get());
+        pAllocator.release();
 
         std::unique_ptr<ThreadTranscodeContext> pThreadPipeline(new ThreadTranscodeContext);
         // extend BS processing init
--- a/samples/sample_multi_transcode/include/pipeline_transcode.h
+++ b/samples/sample_multi_transcode/include/pipeline_transcode.h
@@ -416,7 +416,7 @@ namespace TranscodingSample
             inline void PrintStatistics(mfxU32 numPipelineid, mfxF64 target_framerate = -1 /*default stands for infinite*/)
             {
                 // print timings in ms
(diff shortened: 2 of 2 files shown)


samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -183,10 +183,12 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
...
     for (i = 0; i < m_InputParamsArray.size(); i++)
     {
-        GeneralAllocator* pAllocator = new GeneralAllocator;
+        std::unique_ptr<GeneralAllocator> pAllocator(new GeneralAllocator);
         sts = pAllocator->Init(m_pAllocParam.get());
         MSDK_CHECK_STATUS(sts, "pAllocator->Init failed");
-        m_pAllocArray.push_back(pAllocator);
+
+        m_pAllocArray.push_back(pAllocator.get());
+        pAllocator.release();
```
vilichev: This can be combined if just : m_pAllocArray.push_back(pAllocator.release());

mgonchar: no, because push_back may throw exception if new node allocation fails in that case nothing protects memory from leakage. It would be possible if memory would be preallocated (i.e. reserve() was called) and copy/move constructors are marked as nothrow, in that case it would be safe to pass raw pointer to push_back

fzhar: Have you tried to run it ?
If memory was allocated by new - it's good idea to check where it was freed...
I don't see any changes in cleanup part of Close() function - pretty sure that it will crash there. Please make appropriate changes

mgonchar: @fedor why should it crash? I didn't change the behavior at all

fzhar: OK, sorry, I thought you're pushing smart pointer. What's the idea of this change then ? Just avoid memory leak in low-probable situation when pointer cannot be pushed into array ? Maybe it's better to handle exception and break program execution - it won't work correctly after that anyway?

mgonchar: not only exceptions, there is an exit point at line 188, where status is checked...
Well in situations with dynamic memory allocations you generally have 2 options:
- manual memory management = cover all exit points with deletes + writing try-catch block
- pass this responsibility to smart pointers in RAII paradigm

I believe second one is much better. Actually with smart pointers you should never call delete by yourself. So this particular change is fixing complain of static analysis tool. Better to use vector of smart pointers and don't do manual memory management at all

mgonchar: also, execution flow of program will break in case of error with current code as well. So I didn't get your concern

mgonchar: I don't think that this change is extra because it does exactly what I mention in comment - it fixes memory leak in two cases: in case of error on line 188 and in case of exception.

This is a bit ugly fix, but it looks so because of original design issue. I agree that better to have proper redesign which makes this code both workable and memory leak safe. But since it requires more effort, let's have this solution before redesign happens.

PR #927, opened 2018-11-23
AVBR: enable support for AVC in core library and sample_encode

ChenXiaomin36:
./sample_encode h264 -i xxx.yuv -o xxx.h264 .....-avbr -b TargetBitrate
-convergence Convergence -accuracy Accuracy
Convergence is in the unit of frame
Accuracy should be in [1, 100]

Change-Id: Ic57cf2c213a9345328d2b91d9bf455bec3283360

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -2413,6 +2413,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
        par.mfx.RateControlMethod != MFX_RATECONTROL_VBR &&
        par.mfx.RateControlMethod != MFX_RATECONTROL_QVBR &&
        par.mfx.RateControlMethod != MFX_RATECONTROL_CQP &&
+       par.mfx.RateControlMethod != MFX_RATECONTROL_AVBR &&
        par.mfx.RateControlMethod != MFX_RATECONTROL_ICQ &&
        !bRateControlLA(par.mfx.RateControlMethod))
     {
@@ -3843,6 +3844,13 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
 
     if (par.mfx.RateControlMethod == MFX_RATECONTROL_QVBR &&
         hwCaps.QVBRBRCSupport == 0)
+    {
+        par.mfx.RateControlMethod = 0;
+        unsupported = true;
+    }
+
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_AVBR &&
+        hwCaps.AVBRBRCSupport == 0)
     {
         par.mfx.RateControlMethod = 0;
         unsupported = true;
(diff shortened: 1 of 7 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -5389,7 +5397,7 @@ void MfxHwH264Encode::SetDefaults(
     if (par.mfx.RateControlMethod == MFX_RATECONTROL_AVBR)
     {
         if (par.mfx.Accuracy == 0)
-            par.mfx.Accuracy = AVBR_ACCURACY_MAX;
+            par.mfx.Accuracy = 100;
```
lakulako: Does AVBR have sense with 100% accuracy?

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -202,6 +202,15 @@ mfxStatus SetRateControl(
     rate_param->bits_per_second = GetMaxBitrateValue(par.calcParam.maxKbps) << (6 + SCALE_FROM_DRIVER);
     rate_param->window_size     = par.mfx.Convergence * 100;
 
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_AVBR)
+    {
+        rate_param->window_size = par.mfx.Convergence;
+    }
+    if (par.mfx.RateControlMethod == MFX_RATECONTROL_AVBR)
+    {
+        rate_param->bits_per_second = ((1000 * par.calcParam.targetKbps) >> (6 + SCALE_FROM_DRIVER)) << (6 + SCALE_FROM_DRIVER);
```
lakulako: Rounding down is OK?

samples/sample_encode/src/sample_encode.cpp
```diff
@@ -1119,6 +1144,14 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
         return MFX_ERR_UNSUPPORTED;
     }
 
+    if (pParams->nRateControlMethod == MFX_RATECONTROL_AVBR)
+    {
+        if (pParams->Accuracy > 100)
+        {
+            pParams->Accuracy = 100;
```
lakulako: Shouldn't it be warned?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
ChenXiaomin36: Yes. This accuracy corresponds to "target_percentage" in VAEncMiscParameterRateControl. Its valid range is [1, 100]. As to the real control result, it depends on AVBR algorithm and implementation in lower level. But its valid range is [1, 100].

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
ChenXiaomin36: Yes, it is OK. There are similar codes in above line:
rate_param->bits_per_second = GetMaxBitrateValue(par.calcParam.maxKbps) << (6 + SCALE_FROM_DRIVER);
And the implementation of GetMaxBitrateValue is as follows:
mfxU32 MfxHwH264Encode::GetMaxBitrateValue(mfxU32 kbps, mfxU32 scale)
{
    return (1000 * kbps) >> (6 + scale);
}
And the default value of scale is SCALE_FROM_DRIVER

lizhong1008:
This is not only add AVBR in sample_encode, but also add support in MSDK core libary. So, please update the PR name.

samples/sample_encode/src/sample_encode.cpp
ChenXiaomin36: I think the reply "Added prompt here" has lead in confusion here. Sorry for that. I mean I have added prompt info in source codes. Thanks!

PR #930, opened 2018-11-23
Android build: replace path to MFX API by header library

dmitrybrazhkin:


_studio/mfx_lib/Android.mk
```diff
@@ -180,6 +180,8 @@ LOCAL_CFLAGS := \
 LOCAL_CFLAGS_32 := $(MFX_CFLAGS_INTERNAL_32)
 LOCAL_CFLAGS_64 := $(MFX_CFLAGS_INTERNAL_64)
 
+LOCAL_HEADER_LIBRARIES := libmfx_headers
```
dvrogozh: I don't see where libmfx_headers is defined. Was it already defined somewehere but we did not use it?

dmitrybrazhkin: Yes, it is already defined in /android/Android.mk

PR #934, opened 2018-11-23
Fix mfx_dispatch build on Android

dmitrybrazhkin:

### DIFF
--- a/api/mfx_dispatch/linux/Android.mk
+++ b/api/mfx_dispatch/linux/Android.mk
@@ -5,13 +5,15 @@ LOCAL_PATH:= $(call my-dir)
 include $(CLEAR_VARS)
 include $(MFX_HOME)/android/mfx_defs.mk
 
-LOCAL_SRC_FILES := $(addprefix src/, \
+LOCAL_SRC_FILES := \
     mfxloader.cpp \
-    mfxparser.cpp)
+    mfxparser.cpp
 
 LOCAL_C_INCLUDES := $(MFX_INCLUDES)
 
-LOCAL_CFLAGS := $(MFX_CFLAGS_INTERNAL)
+LOCAL_CFLAGS := \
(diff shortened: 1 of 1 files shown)


api/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -442,7 +442,11 @@ mfxStatus MFXVideoUSER_Load(mfxSession session, const mfxPluginUID *uid, mfxU32
       if (MFX::g_GlobalCtx.m_plugins.empty()) {
         // Parsing plugin configuration file and loading information of
         // _all_ plugins registered on the system.
+#ifdef ANDROID
```
dvrogozh: I would try to avoid #ifdefs in the code as much as we can. We already have a lot. Thus, questions:
1. Why android defines some other name for plugins.cfg file? If you want to separate mediasdk config file from anything else, why you can't create mediasdk dedicated folder? For example: /vendor/etc/mediasdk/plugins.cfg
2. Why you can't pass /vendor/etc/mediasdk via MFX_PLUGINS_CONF_DIR?

api/mfx_dispatch/linux/Android.mk
```diff
@@ -5,9 +5,9 @@ LOCAL_PATH:= $(call my-dir)
 include $(CLEAR_VARS)
 include $(MFX_HOME)/android/mfx_defs.mk
 
-LOCAL_SRC_FILES := $(addprefix src/, \
+LOCAL_SRC_FILES := \
```
dvrogozh: This one is obvious. We can merge this in almost immediately if this will be separated out from the below plugins change...

api/mfx_dispatch/linux/mfxloader.cpp
dmitrybrazhkin: Agree, moved to Android.mk and replaced by plugins.cfg

PR #940, opened 2018-11-27
MPEG2e should return unsupported on BXT

tkuzne1x:
MDP-49092

### DIFF
--- a/_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
@@ -948,6 +948,8 @@ namespace MPEG2EncoderHW
         CHECK_CODEC_ID(par->mfx.CodecId, MFX_CODEC_MPEG2);
         MFX_CHECK (CheckExtendedBuffers(par) == MFX_ERR_NONE, MFX_ERR_INVALID_VIDEO_PARAM);
 
+        mfxStatus sts = core->IsGuidSupported(DXVA2_Intel_Encode_MPEG2, par);
+        MFX_CHECK_STS(sts);
 
         mfxExtCodingOption* ext = GetExtCodingOptions(par->ExtParam, par->NumExtParam);
         mfxExtCodingOptionSPSPPS* pSPSPPS = GetExtCodingOptionsSPSPPS (par->ExtParam, par->NumExtParam);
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_full_hw.cpp
```diff
@@ -155,6 +155,11 @@ mfxStatus FullEncode::Init(mfxVideoParam *par)
 
     if (is_initialized())
         return MFX_ERR_UNDEFINED_BEHAVIOR;
+
+    if (MFX_HW_APL == m_pCore->GetHWType())
```
otsirule: Why does it return MSDK, but not driver?

aulitin:
@tkuzne1x, IsGuidSupported checks Entrypoint and Profile and Resolution. If platform doesn't support something than MFX_ERR_UNSUPPORTED returned.  You need to use this function in QueryIOSurf (_studio\mfx_lib\encode_hw\mpeg2\src\mfx_mpeg2_encode_utils_hw.cpp). At the same time could you delete duplicate IsGuidSupported code?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -948,6 +948,8 @@ namespace MPEG2EncoderHW
         CHECK_CODEC_ID(par->mfx.CodecId, MFX_CODEC_MPEG2);
         MFX_CHECK (CheckExtendedBuffers(par) == MFX_ERR_NONE, MFX_ERR_INVALID_VIDEO_PARAM);
 
+        mfxStatus sts = core->IsGuidSupported(DXVA2_Intel_Encode_MPEG2, par);
```
alariono: Will this function's call work for Linux as well?

aulitin: Yes. Mpeg2 support FullHD and IsGuidSupported will check this resolution. Check resolution will not be needed on 8K in 990 and 984 strings. As well as IsGuidSupported checks encode support on platform.

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -976,19 +978,9 @@ namespace MPEG2EncoderHW
...
 
-        mfxStatus sts = CheckHwCaps(core, par, ext, &EncCaps);
+        sts = CheckHwCaps(core, par, ext, &EncCaps);
         MFX_CHECK_STS(sts);
 
-        if (par->mfx.FrameInfo.Width > 0x1fff || (par->mfx.FrameInfo.Width & 0x0f) != 0)
-        {
-            return MFX_ERR_INVALID_VIDEO_PARAM;
-        }
-        mfxU32 mask = (par->mfx.FrameInfo.PicStruct & MFX_PICSTRUCT_PROGRESSIVE)? 0x0f:0x1f;
```
aulitin: This code also checks alignment for width and height. Please, don't delete this check, because IsGuidSupported doesn't check alignment.

PR #942, opened 2018-11-27
[VP9e][Linux] Fixed returned status for MxN tiles

eshiryae:
Fixed returned status for cases with MxN tiles what are currently unsupported
by driver on Linux.

### DIFF
--- a/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
+++ b/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
@@ -1479,6 +1479,15 @@ mfxStatus CheckParameters(VP9MfxVideoParam &par, ENCODE_CAPS_VP9 const &caps)
         unsupported = true;
     }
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
```diff
@@ -1479,6 +1479,15 @@ mfxStatus CheckParameters(VP9MfxVideoParam &par, ENCODE_CAPS_VP9 const &caps)
         unsupported = true;
     }
 
+#if defined(MFX_VA_LINUX)
+    // TODO: For now driver on Linux doesn't support MxN tiles
+    if (rows > 1 && cols > 1)
```
lakulako: Maybe ||? Otherwise 5x1 is supported?

dvrogozh: Do we have driver query caps for that? If not did you request to add them?

PR #950, opened 2018-11-28
Remove using from headers

mgonchar:
Take under control namespace diffusion from headers


mgonchar:
@dmitry-gurulev agree, it introduced huge delta

onabiull:
Please rebase

PR #956, opened 2018-11-29
h265e: Code alignment

mgonchar:


_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h
```diff
@@ -274,8 +270,8 @@ class MfxFrameAllocResponse : public mfxFrameAllocResponse
 
 struct DpbFrame
 {
-    mfxI32              m_poc         = 0;
-    mfxU32              m_fo          = 0;          // FrameOrder
+    mfxI32              m_poc         = -1;
```
lakulako: -1 can be legal value. RADL for example

mgonchar: Yes, it is mostly alignment with previously existed approach with total memset 0xff of this struct. This field never used to check the whole struct for valid state.

PR #960, opened 2018-12-03
replace VP9_LOG

vsidorow:


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -746,15 +746,15 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     m_caps.EncodeFunc = 1;
     m_caps.HybridPakFunc = 1;
 
-    VP9_LOG(" \n CreateAuxilliaryDevice return ERR_NONE \n");
+    MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_HOTSPOTS, "  CreateAuxilliaryDevice return ERR_NONE ");
```
lakulako: In the previous logs there were useless hanging spaces before \n. What for they are often kept in new?

PR #963, opened 2018-12-03
replace VP9_LOG

vsidorow:


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw.cpp
```diff
@@ -802,15 +794,14 @@ mfxStatus MFXVideoENCODEVP9_HW::Execute(mfxThreadTask task, mfxU32 /*uid_p*/, mf
     {
         Task& frameToGet = m_submitted.front();
 
-        VP9_LOG("\n (VP9_LOG) Frame %d MFXVideoENCODEVP9_HW::QueryFrame +", frameToGet.m_frameOrder);
+        MFX_LTRACE_1(MFX_TRACE_LEVEL_DEFAULT, "MFXVideoENCODEVP9_HW::QueryFrame", "Frame %d", frameToGet.m_frameOrder);
```
lakulako: Is a space needed in the end of the message like at 747 and 517 or it isn't like here?

vsidorow: Yes, thanks for pointing that out

PR #966, opened 2018-12-04
[VP9e] Correct setting framerates

eshiryae:
Corrected setting framerates for VP9 encoder according to the libva


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -548,10 +542,12 @@ mfxStatus SetFrameRate(
 
     frameRateBufIds.resize(numTL);
 
-    mfxU32 base_framerate;
-    PackMfxFrameRate(par.mfx.FrameInfo.FrameRateExtN, par.mfx.FrameInfo.FrameRateExtD, base_framerate);
+    mfxU32 nom = par.mfx.FrameInfo.FrameRateExtN;
```
vshampor: Actually "N" in "FrameRateExtN" stands for "Numerator".

PR #968, opened 2018-12-04
[VP9e] Correct setting framerates

eshiryae:
Corrected setting framerates for VP9 encoder according to the libva

### DIFF
--- a/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
@@ -530,13 +530,7 @@ mfxStatus SetFrameRate(
     VAEncMiscParameterFrameRate *frameRate_param;
 
     mfxExtVP9TemporalLayers& extTL = GetExtBufferRef(par);
-
-    mfxU8 numTL = par.m_numLayers;
-    bool TL_attached = false;
-    if (numTL == 0) 
-        numTL = 1;
-    else
-        TL_attached = true;
+    mfxU16 numTL = std::max<mfxU16>(par.m_numLayers, 1);
 
     for (VABufferID id : frameRateBufIds)
     {
@@ -548,10 +542,10 @@ mfxStatus SetFrameRate(
 
     frameRateBufIds.resize(numTL);
 
-    mfxU32 base_framerate;
-    PackMfxFrameRate(par.mfx.FrameInfo.FrameRateExtN, par.mfx.FrameInfo.FrameRateExtD, base_framerate);
+    mfxU32 nom = par.mfx.FrameInfo.FrameRateExtN;
+    mfxU32 denom = par.mfx.FrameInfo.FrameRateExtD;
 
-    for (mfxU8 tl = 0; tl < frameRateBufIds.size(); tl++)
+    for (mfxU16 tl = 0; tl < numTL; tl++)
     {
         vaSts = vaCreateBuffer(m_vaDisplay,
                     m_vaContextEncode,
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -548,10 +544,12 @@ mfxStatus SetFrameRate(
 
     frameRateBufIds.resize(numTL);
 
-    mfxU32 base_framerate;
-    PackMfxFrameRate(par.mfx.FrameInfo.FrameRateExtN, par.mfx.FrameInfo.FrameRateExtD, base_framerate);
+    mfxU32 num = par.mfx.FrameInfo.FrameRateExtN;
```
lakulako: Note. N stands for nOminator, so "nom" would be proper name for the variable.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -570,7 +568,21 @@ mfxStatus SetFrameRate(
...
+        {
+            mfxU16 l2lRatio = extTL.Layer[tl + 1].FrameRateScale / extTL.Layer[tl].FrameRateScale;
+            if (num % l2lRatio == 0)
+            {
+                num /= l2lRatio;
+            }
+            else
+            {
+                denom *= l2lRatio;
+            }
```
lakulako: PackMfxFrameRate(...) involves fraction reducing, so no need in optimization here.
denom *= extTL.Layer[tl + 1].FrameRateScale; nom *= extTL.Layer[tl].FrameRateScale;
will work correctly regardless of FrameRateScale ratio.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
eshiryae: Doesn't N stand for numerator?

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
eshiryae: Fixed.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
lakulako: Nominator - denominator.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -570,7 +568,14 @@ mfxStatus SetFrameRate(
         misc_param->type = VAEncMiscParameterTypeFrameRate;
         frameRate_param = (VAEncMiscParameterFrameRate *)misc_param->data;
 
-        frameRate_param->framerate = TL_attached ? base_framerate / extTL.Layer[par.m_numLayers - 1 - tl].FrameRateScale : base_framerate;
+        if (tl < numTL - 1)
+        {
+            num *= extTL.Layer[tl].FrameRateScale;
+            denom *= extTL.Layer[tl + 1].FrameRateScale;
+        }
+        PackMfxFrameRate(num, denom, tl_framerate);
```
lakulako: Also move tl_framerate declaration here. Or don't use it at all.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -570,7 +566,13 @@ mfxStatus SetFrameRate(
         misc_param->type = VAEncMiscParameterTypeFrameRate;
         frameRate_param = (VAEncMiscParameterFrameRate *)misc_param->data;
 
-        frameRate_param->framerate = TL_attached ? base_framerate / extTL.Layer[par.m_numLayers - 1 - tl].FrameRateScale : base_framerate;
```
DmitryMalishev: Are you 100% sure that par.m_numLayers always > 1? In case of par.m_numLayers equals to 0, or equals to 1 with tl>1 you'll get invalid index for the array.

lakulako: Should add
if (tl == numTL-1)
     PackMfxFrameRate(nom, denom, frameRate_param->framerate);
else
    Pack... as here
It will work when either isn't extTL or it is.

PR #972, opened 2018-12-05
h264d: fix async model

dmitry-gurulev:
- Decoder triggered 'global task' not only when
   it needs to force async. operation completion
   but every time it needs more work surfaces.
   When blocking sync. is used this leads to
   [SyncOperation] call may take more time than it
   really needed. In current patch we use the same
   approach as it is in h265d component.

Issue: MDP-41203
Test: full decoding & transcoding validation

### DIFF
--- a/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
@@ -89,9 +89,13 @@ mfxExtBuffer* GetExtBuffer(mfxExtBuffer** ebuffers, mfxU32 nbuffers, mfxU32 Buff
 
 struct ThreadTaskInfo
 {
-    mfxFrameSurface1 *surface_work;
-    mfxFrameSurface1 *surface_out;
-    mfxU32            taskID; // for task ordering
+    mfxFrameSurface1* surface_work = nullptr;
+    mfxFrameSurface1* surface_out  = nullptr;
+
+    ThreadTaskInfo(mfxFrameSurface1* work, mfxFrameSurface1* out)
+        : surface_work(work)
+        , surface_out(out)
+    {}
 };
 
 enum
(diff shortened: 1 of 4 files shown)


_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
```diff
@@ -1114,39 +1125,26 @@ mfxStatus VideoDECODEH264::DecodeFrameCheck(mfxBitstream *bs,
 
     mfxStatus mfxSts = DecodeFrameCheck(bs, surface_work, surface_out);
 
-    if (MFX_ERR_NONE == mfxSts) // It can be useful to run threads right after first frame receive
+    if (MFX_ERR_NONE == mfxSts || (mfxStatus)MFX_ERR_MORE_DATA_SUBMIT_TASK == mfxSts) // It can be useful to run threads right after first frame receive
```
mgonchar: why do you need ```mfxStatus``` cast here?

_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
```diff
@@ -1269,10 +1267,10 @@ mfxStatus VideoDECODEH264::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1 *
                 umcAddSourceRes = umcFrameRes = umcRes = UMC::UMC_OK;
             }
 
-            if (umcRes == UMC::UMC_ERR_NOT_ENOUGH_BUFFER || umcRes == UMC::UMC_WRN_INFO_NOT_READY)
+            if (umcRes == UMC::UMC_ERR_NOT_ENOUGH_BUFFER || umcRes == UMC::UMC_WRN_INFO_NOT_READY || umcRes == UMC::UMC_ERR_NEED_FORCE_OUTPUT)
             {
-                force = umcRes == UMC::UMC_ERR_NOT_ENOUGH_BUFFER;
-                sts = MFX_WRN_DEVICE_BUSY;
+                force = umcRes == UMC::UMC_ERR_NEED_FORCE_OUTPUT;
+                sts = umcRes == UMC::UMC_ERR_NOT_ENOUGH_BUFFER ? (mfxStatus)MFX_ERR_MORE_DATA_SUBMIT_TASK: MFX_WRN_DEVICE_BUSY;
```
mgonchar: same cast

_studio/shared/umc/codec/h264_dec/src/umc_h264_va_supplier.cpp
```diff
@@ -327,40 +327,8 @@ Status VATaskSupplier::AddSource(MediaData * pSource)
...
-        pDPB->IsDisposableExist())
-        return MFXTaskSupplier::AddSource(pSource);
-
-    H264DecoderFrame* completed = 0;
-    Status umcRes = CompleteDecodedFrames(&completed);
-    if (umcRes != UMC_OK)
-        return pSource || !completed ? umcRes : UMC_OK;
-
-    return UMC_WRN_INFO_NOT_READY;
+    notifier0<LazyCopier> copy_slice_data(&m_lazyCopier, &LazyCopier::CopyAll);
```
mgonchar: so this is:
```
if (pSource)
    notifier0<LazyCopier> copy_slice_data(&m_lazyCopier, &LazyCopier::CopyAll);

return MFXTaskSupplier::AddSource(pSource);
```

dmitry-gurulev: No it isn't. In your case [LazyCopier] will always (when pSource != NULL) copy slice data while we needn't do this when full frame was submitted.

PR #973, opened 2018-12-05
Added branch if for va_fourcc == VA_FOURCC_AYUV then fromat is VA_RT_FORMAT_YUV444

dmenshov:


DenWolf:
Why so long name for pull request especially for this simple change? You have a pretty good commit name.

Regarding the failed build - apparently it was a temporary problem on CI, you change don't need to break something, so rebase/rerun CI build need to solve the problem

PR #976, opened 2018-12-05
vc1d: Added check for RTM field in case of Simple/Main profile

ivan-losev:
RTM (release-to-manufacturer) flag (also called Reserved6). Means old WMV3 encoding with
different bitstream format for P/B frames.

According to Annex J Table 263, this field should be set to one and other values shall be forbidden.


varistar:
Looks correct, but did you check VC-1 (WMV3) conformance including extended test set?

PR #984, opened 2018-12-05
test_monitor fixes for issues from static code analysis

andreyor:


samples/metrics_monitor/test/cttmetrics_gtest.cpp
```diff
@@ -133,10 +135,6 @@ int setGpuFrequency(int min, int max, int boost)
             res_max = write(fd_max, buf_max, sizeof(buf_max));
```
mgonchar: and now code is incorrect, because you will get sizeof pointer
Fix:
```
res_max = write(fd_max, buf_max.c_str(), buf_max.size());
```

samples/metrics_monitor/test/cttmetrics_gtest.cpp
```diff
@@ -106,14 +105,17 @@ int getGpuFrequency(char* path)
...
     int fd_min, fd_max, fd_boost, res = -1, res_min = -1, res_max = -1, res_boost = -1;
-    char buf_max[6] = {0}, buf_min[6] = {0}, buf_boost[6] = {0};
 
     if (!(min <= max && min <= boost && boost <= max))
         return res;
 
-    sprintf(buf_min, "%d", min);
-    sprintf(buf_max, "%d", max);
-    sprintf(buf_boost, "%d", boost);
+    std::ostringstream ss_min, ss_max, ss_boost;
```
mgonchar: this is an over complicated logic.
Use "std::to_string" instead:
```
std::string buf_min(std::to_string(min));
...
```

samples/metrics_monitor/test/cttmetrics_gtest.cpp
```diff
@@ -123,20 +121,16 @@ int setGpuFrequency(int min, int max, int boost)
     {
         if (min < getGpuFrequency(GPU_MIN_FREQ_FILE_PATH))
```
mgonchar: what is the purpose of this "if" if you do the same things in both branches? the order of writing matters?

andreyor: Yes, the order depends on the current min frequency.

mgonchar: But what is the effect of different order of writing from HW side?

andreyor: For instance if we have: gt_min_freq_mhz=700 / gt_max_freq_mhz=700 / gt_boost_freq_mhz=700 and we want to set frequency to 1000, we need write to files in this order: gt_boost_freq_mhz, gt_max_freq_mhz, gt_min_freq_mhz.
Overwise we wil get: 
# echo 1000 > /sys/class/drm/card0/gt_min_freq_mhz
-bash: echo: write error: Invalid argument
and -1 must be returned from write(...)

mgonchar: OK, now I got the idea

PR #988, opened 2018-12-06
[core] Propagated new pointer checks procedure to VAAPI core

vshampor:


_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -893,7 +896,7 @@ VAAPIVideoCORE::DoFastCopyWrapper(
     {
         if (srcMemType & MFX_MEMTYPE_SYSTEM_MEMORY)
         {
-            if (LumaIsNull(pSrc))
+            if (NULL == srcPtr)
```
mgonchar: please use ```nullptr``` at least in new code

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -893,7 +896,7 @@ VAAPIVideoCORE::DoFastCopyWrapper(
     {
         if (srcMemType & MFX_MEMTYPE_SYSTEM_MEMORY)
         {
-            if (LumaIsNull(pSrc))
```
alariono: Have you deleted LumaIsNull macro/function?

_studio/shared/src/libmfx_core_vaapi.cpp
alariono: Why do not use modern "nullptr"?

_studio/shared/src/libmfx_core_vaapi.cpp
vshampor: No, it is still used elsewhere.

_studio/shared/src/libmfx_core_vaapi.cpp
vshampor: Done

vshampor: Done

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -1041,12 +1044,19 @@ VAAPIVideoCORE::DoFastCopyExtended(
...
+    MFX_CHECK(MFX_SUCCEEDED(sts), MFX_ERR_UNDEFINED_BEHAVIOR);
+    sts = GetFramePointerChecked(pDst->Info, pDst->Data, &dstPtr);
+    MFX_CHECK(MFX_SUCCEEDED(sts), MFX_ERR_UNDEFINED_BEHAVIOR);
 
     // check that only memId or pointer are passed
     // otherwise don't know which type of memory copying is requested
     if (
-        (!LumaIsNull(pDst) && NULL != pDst->Data.MemId) ||
-        (!LumaIsNull(pSrc) && NULL != pSrc->Data.MemId)
+        (nullptr != dstPtr && nullptr != pDst->Data.MemId) ||
```
lakulako: "nullptr != pDst->Data.MemId" Looks strange. Does everyone has to now that MemId is defined as HANDLE which is defined as void*? Better compare to 0 to avoid confusing readers, or introduce INVALID_MEMID.

mgonchar: or maybe better to drop the actual value and use implicit conversion to ```bool```?

PR #990, opened 2018-12-06
[bs_parser_hevc] Use C++ primitives (memcpy => std::copy)

ndomnina:


tools/bs_parser_hevc/src/hevc2_headers.cpp
```diff
@@ -376,7 +376,7 @@ Bs32u Parser::parseEXT(Bs8u*& ExtData)
 
     ExtData = alloc<Bs8u>(0, (Bs32u)data.size());
 
-    memcpy(ExtData, data.data(), data.size());
+    std::copy(data.data(), data.data() + data.size(), ExtData);
```
mgonchar: use ```data.begin() / end()```

tools/bs_parser_hevc/src/hevc2_headers.cpp
```diff
@@ -466,10 +466,9 @@ void Parser::parseSLD(QM& qm)
             case 0:
                 if (refMatrixId < matrixId)
                 {
-                    memcpy(
-                        qm.ScalingFactor0[matrixId],
-                        qm.ScalingFactor0[refMatrixId],
-                        sizeof(qm.ScalingFactor0[0]));
+                    std::copy(qm.ScalingFactor0[refMatrixId],
```
mgonchar: why not direct assignment?

PR #992, opened 2018-12-06
[mpeg2d] New design of HW MPEG2 decoder

dmitryermilov:


_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
```diff
@@ -0,0 +1,282 @@
...
+        bool                              isField;
+        bool                              isBottomField;
+        FrameType                         frameType;
+        bool                              isFilled;
+        uint32_t                          sliceCount;
+        std::vector<MPEG2Slice*>          slices;
+
+        struct RefFramePair
+        {
+            MPEG2DecoderFrame*  forward;
```
mgonchar: Maybe better to add defaults here:
```
struct RefFramePair
        {
            MPEG2DecoderFrame*  forward = nullptr;
            MPEG2DecoderFrame*  backward = nullptr;
        };
```

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
```diff
@@ -0,0 +1,282 @@
...
+
+        UMC::FrameData* GetFrameData()
+        { return data.get(); }
+        UMC::FrameData const* GetFrameData() const
+        { return data.get(); }
+
+        // Wrapper for getting 'surface Index' FrameMID
+        UMC::FrameMemID GetMemID() const
+        {
+            VM_ASSERT(data);
```
mgonchar: need something more than just assert

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
```diff
@@ -0,0 +1,282 @@
...
+                    IncrementReference();
+
+                isRef = true;
+            }
+            else
+            {
+                bool wasRef = IsRef();
+                isRef = false;
+
+                if (wasRef && !IsRef())
```
mgonchar: so result may change between line 209 and 212?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+    {
+    }
+
+    UMC::Status MPEG2Decoder::Init(UMC::BaseCodecParams* vp)
+    {
+        if (!vp)
+            return UMC::UMC_ERR_NULL_PTR;
+
+        MPEG2DecoderParams* dp =
+            DynamicCast<MPEG2DecoderParams, UMC::BaseCodecParams>(vp);
```
mgonchar: maybe we can get rid of custom casts?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+        }
+
+        std::unique_ptr<MPEG2Slice> slice (new MPEG2Slice); // unique_ptr is to prevent a possible memory leak
+
+        size_t const size = in.end - in.begin - prefix_size;
+        slice->source.Alloc(size);
+        if (slice->source.GetBufferSize() < size)
+            throw mpeg2_exception(UMC::UMC_ERR_ALLOC);
+
+        memcpy(slice->source.GetDataPointer(), in.begin + prefix_size, size);
```
mgonchar: better to fall in std::copy for consistency

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+                    uint8_t* picExtBegin = RawHeaderIterator::FindStartCode(unit.begin + prefix_size + bitStream.BytesDecoded(), unit.end);
+                    if (!picExtBegin)
+                        return UMC::UMC_ERR_INVALID_STREAM;
+
+                    bitStream.Reset(picExtBegin + prefix_size, unit.end - picExtBegin);
+                    bitStream.SkipBits(8 + 4); // skip unit and extension types
+                    bitStream.GetPictureExtensionHeader(*picExtHdr.get());
+
+                    // Save headers
+                    m_currHeaders.picHdr = picHdr;
```
mgonchar: use move assignment for shared_ptr, it has quite expensive copy constructor because of atomic counters in control block

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+            return ex.GetStatus();
+        }
+        catch(...)
+        {
+            return UMC::UMC_ERR_INVALID_STREAM;
+        }
+
+        if (type == SEQUENCE_HEADER && m_firstMfxVideoParams.mfx.FrameInfo.Width)
+        {
+            const MPEG2SequenceHeader* currSeq = m_currHeaders.seqHdr.get();
```
mgonchar: as I see, you don't need ```currSeq``` at all. smart pointers have everything to act like a pointer in this code

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+            if (m_dpb.size() >= m_dpbSize)
+            {
+                return nullptr;
+            }
+
+            // Didn't find any. Let's create a new one
+            frame = new MPEG2DecoderFrame;
+
+            // Add to DPB
+            m_dpb.push_back(frame);
```
mgonchar: it throws, so you actually need to reserve one entry ahead or wrap to smart pointer (first is better I think:) )

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+                    m_payloads[i].isUsed = m_lastUsed;
+            }
+        }
+
+        m_lastUsed++;
+    }
+
+    const Payload_Storer::Message * Payload_Storer::GetPayloadMessage()
+    {
+        Payload_Storer::Message * msg = 0;
```
mgonchar: nullptr

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+            if (!m_payloads[i].isUsed)
+                continue;
+
+            Message & mmsg = m_payloads[i];
+
+            if ((m_offset + sz > mmsg.offset) &&
+                (m_offset < mmsg.offset + mmsg.msg_size))
+            {
+                m_payloads[i].isUsed = 0;
+                return 0;
```
mgonchar: nullptr

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+        }
+
+        return msg;
+    }
+
+    Payload_Storer::Message* Payload_Storer::AddMessage(const RawUnit & data, int32_t auIndex)
+    {
+        size_t sz = data.end - data.begin;
+
+        if (sz > (m_data.size() >> 2))
```
mgonchar: unusual place for optimization)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+            if ((m_offset + sz > mmsg.offset) &&
+                (m_offset < mmsg.offset + mmsg.msg_size))
+            {
+                m_payloads[i].isUsed = 0;
+                return 0;
+            }
+        }
+
+        size_t freeSlot = 0;
+        for (uint32_t i = 0; i < m_payloads.size(); i++)
```
mgonchar: so this is ```std::find_if```

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1147 @@
...
+                break;
+            }
+        }
+
+        if (m_payloads[freeSlot].isUsed)
+        {
+            if (m_payloads.size() >= MAX_ELEMENTS)
+                return 0;
+
+            m_payloads.push_back(Message());
```
mgonchar: same as ```push_back()```, but with one excessive constructor call

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+        slices.resize(0);
+        sliceCount = 0;
+        FreeReferenceFrames();
+
+    }
+
+    void MPEG2DecoderFrameInfo::AddSlice(MPEG2Slice* slice)
+    {
+        if (0 == sliceCount) // on the first slice
```
mgonchar: I believe we need pointer check

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1201 @@
...
-        return false;
-    }
-
-#if defined (MFX_VA_LINUX)
-    if (MFX_ERR_NONE != pCore->IsGuidSupported(sDXVA2_ModeMPEG2_VLD, par))
-    {
-        return false;
-    }
-#endif
+    mfxU32 asyncDepth = par->AsyncDepth ? par->AsyncDepth : MFX_AUTO_ASYNC_DEPTH_VALUE;
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1201 @@
...
 
-mfxU8 GetMfxCodecLevel(mfxI32 level)
+VideoDECODEMPEG2::VideoDECODEMPEG2(VideoCORE* core, mfxStatus* sts)
+    : m_core(core)
+    , m_platform(MFX_PLATFORM_SOFTWARE)
+    , m_opaque(false)
+    , m_first_run(true)
 {
-    switch (level)
+    memset(&m_stat, 0, sizeof(m_stat));
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1201 @@
...
-
-    picInfo.top_ff = ((p[7] & 0x80) != 0);
-    picInfo.repeat_ff = ((p[7] & 2) != 0);
-    picInfo.first_polarity = (picInfo.picture_structure == 3) ? (picInfo.top_ff ? 1 : 2) : picInfo.picture_structure;
-    // field and frame&repeat both have last==first
-    picInfo.last_polarity = (picInfo.picture_structure == 3 && !picInfo.repeat_ff) ? picInfo.first_polarity ^ 3 : picInfo.first_polarity;
-    return true;
-}
+    // allocate memory
+    mfxFrameAllocRequest request;
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
dmitryermilov: No. It will be excessive. see DecodeSliceHeader usage

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1201 @@
...
-        else
-        {
-            picStruct = MFX_PICSTRUCT_FIELD_BFF;
-        }
+        type = m_core->GetHWType();
     }
 
-    return picStruct;
-}
+    if (CheckVideoParamDecoders(par, m_core->IsExternalFrameAllocator(), type) < MFX_ERR_NONE)
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1201 @@
...
-    vPar.mfx.FrameInfo.Width = AlignValue(vPar.mfx.FrameInfo.CropW, MFX_MB_WIDTH);
-    vPar.mfx.FrameInfo.Height = AlignValue(vPar.mfx.FrameInfo.CropH, MFX_MB_WIDTH);
-    vPar.mfx.FrameInfo.ChromaFormat = Mpeg2GetMfxChromaFormatFromUmcMpeg2(sh.chroma_format);
-    GetMfxFrameRate((mfxU8)sh.frame_rate_code, &vPar.mfx.FrameInfo.FrameRateExtN, &vPar.mfx.FrameInfo.FrameRateExtD);
-
-    vPar.mfx.FrameInfo.PicStruct = GetMfxPicStruct(sh.progressive_sequence, ph.progressive_frame,
-                                                   ph.top_field_first, ph.repeat_first_field, ph.picture_structure,
-                                                   vPar.mfx.ExtendedPicStruct);
-}
+    bool internal = m_platform == !!(MFX_PLATFORM_SOFTWARE ?
```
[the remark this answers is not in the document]
dmitryermilov: Misprint. Thanks

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1201 @@
...
-        mfxExtOpaqueSurfaceAlloc *pOpaqAlloc = (mfxExtOpaqueSurfaceAlloc *)GetExtendedBuffer(par->ExtParam, par->NumExtParam, MFX_EXTBUFF_OPAQUE_SURFACE_ALLOCATION);
-        if (!pOpaqAlloc)
-            return MFX_ERR_INVALID_VIDEO_PARAM;
-
-        useInternal = m_isSWDecoder ? !(pOpaqAlloc->Out.Type & MFX_MEMTYPE_SYSTEM_MEMORY) : (pOpaqAlloc->Out.Type & MFX_MEMTYPE_SYSTEM_MEMORY);
-    }
-
-    if (!m_isSWDecoder || useInternal || (IOPattern & MFX_IOPATTERN_OUT_OPAQUE_MEMORY))
-    {
-        // positive condition means that decoder was configu
```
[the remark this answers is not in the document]
dmitryermilov: Then please fix it in mfx_common_int.h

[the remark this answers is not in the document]
dmitryermilov: ok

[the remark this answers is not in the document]
dmitryermilov: see CompleteProc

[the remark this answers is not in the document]
dmitryermilov: ok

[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -0,0 +1,264 @@
...
+
+        MPEG2DecoderParams()
+            : allocator(nullptr)
+            , async_depth(0)
+        {}
+
+    public:
+
+        UMC::FrameAllocator* allocator;
+        uint32_t             async_depth;
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
```diff
@@ -0,0 +1,297 @@
...
+        uint32_t  bit_rate_extension           = 0;
+        uint32_t  vbv_buffer_size_extension    = 0;
+        uint8_t   low_delay                    = 0;
+        uint8_t   frame_rate_extension_n       = 0;
+        uint8_t   frame_rate_extension_d       = 0;
+
+        void Reset()
+        {
+            MPEG2SequenceExtension b = {};
+            *this = b;
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
```diff
@@ -0,0 +1,297 @@
...
+        { Reset(); }
+
+        void Reset()
+        { _buffer.clear(); }
+
+        size_t GetSize() const
+        { return _buffer.size(); }
+
+        uint8_t * GetPointer()
+        { return _buffer.empty() ? nullptr : &_buffer[0]; }
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
dmitryermilov: ok

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -0,0 +1,264 @@
...
+        std::unique_ptr<Payload_Storer> m_messages;
+    };
+
+    class Payload_Storer
+    {
+    public:
+
+        struct Message
+        {
+            size_t             msg_size;
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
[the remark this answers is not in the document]
dmitryermilov: I don't really want to change this code

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -0,0 +1,264 @@
...
+            double             timestamp;
+            int32_t             isUsed;
+            int32_t             auID;
+            int32_t             inputID;
+            MPEG2DecoderFrame * frame;
+
+            Message()
+                : msg_size(0)
+                , offset(0)
+                , data(0)
```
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: I didn't get your idea

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: Ok. Thanks

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_va_packer.h
```diff
@@ -0,0 +1,102 @@
...
+        virtual void PackQmatrix(const MPEG2DecoderFrameInfo&) = 0;
+        virtual void PackSliceParams(const MPEG2Slice&, uint32_t) = 0;
+
+    protected:
+
+        UMC::VideoAccelerator *m_va;
+    };
+
+
+    class PackerVA
```
dmitry-gurulev: I'd rather put platform specific packer into separate file.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,259 @@
...
+        {
+            if (begin[0] == 0 && begin[1] == 0 && begin[2] == 1)
+            {
+                return begin;
+            }
+        }
+        return nullptr;
+    }
+
+    RawUnit RawHeaderIterator::FindRawUnit(mfxU8 * begin, mfxU8 * end)
```
dmitry-gurulev: Different types - _uint8_t_ in _FindStartCode_ but _mfxU8_ here. What is intention of that?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,259 @@
...
+    {
+        for (; begin < end - 4; ++begin)
```
dmitry-gurulev: Reuse _FindStartCode_ here?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+        m_it.LoadData(source); // Find new unit
+
+        // for loop is to handle cases like: seq header, seq header, seq extension
+        for (;;)
+        {
+            if (m_buff.size() && (m_it != RawHeaderIterator()))
+            {
+                size_t size = m_it->end - m_it->begin;
+                MPEG2_UNIT_TYPE_EXT type_ext = (MPEG2_UNIT_TYPE_EXT) ((m_it->type == EXTENSION && size >= 5) ? (m_it->begin[prefix_size + 1] >> 4) : 0 );
```
dmitry-gurulev: What is 5? Is it length of extension header? Need to either add self-descriptive constant or comment at least.
BTW, I'd rather make [size] to be method of _RawHeaderIterator_ to avoid mixing low level pointers logic and high level units handling.

The same concerns about _m_it->begin[prefix_size + 1] >> 4_

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,259 @@
...
+                }
+            }
+            else if (SEQUENCE_HEADER == m_it->type || PICTURE_HEADER == m_it->type)
+            {
+                m_typeExtToFind = (SEQUENCE_HEADER == m_it->type) ? SEQUENCE_EXTENSION : PICTURE_CODING_EXTENSION;
+                m_buff.insert(m_buff.end(), (uint8_t *)m_it->begin, (uint8_t *)m_it->end);
+
+                ++m_it; // Look for another unit
+
+                if (m_it != RawHeaderIterator())
```
dmitry-gurulev: This piece of code (under condition) looks to be the same as above. Is it possible to extract this as function (lambda)?

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -0,0 +1,245 @@
...
+        {
+            std::shared_ptr<MPEG2SequenceHeader>           seqHdr;
+            std::shared_ptr<MPEG2SequenceExtension>        seqExtHdr;
+            std::shared_ptr<MPEG2SequenceDisplayExtension> displayExtHdr;
+            std::shared_ptr<MPEG2PictureHeader>            picHdr;
+            std::shared_ptr<MPEG2PictureCodingExtension>   picExtHdr;
+            std::shared_ptr<MPEG2QuantMatrix>              qmatrix;
+        };
+
+        Headers                         m_currHeaders;
```
dmitry-gurulev: Where do we use _m_currHeaders_? Why do we need headers to be shared?

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_slice.h
```diff
@@ -0,0 +1,105 @@
...
+        // Obtain bitstream object
+        const MPEG2HeadersBitstream & GetBitStream()  const
+        { return m_bitStream; }
+
+    public:
+        UMC::MediaData       source;
+        MPEG2SliceHeader     sliceHeader;
+
+    protected:
+        std::shared_ptr<MPEG2SequenceHeader>         m_seqHdr;
```
dmitry-gurulev: Why do we need to share ownership for this (and below) headers?

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_va_packer.h
dmitryermilov: I believe it will be needed when Windows support appears

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
dmitryermilov: No special intention. If you want, I can replace with uint8_t

dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
dmitryermilov: >>What is 5
Because of m_it->begin[prefix_size + 1] >> 4. In the same line. Ok, I'll add a comment. 

>>BTW, I'd rather make [size] to be method of RawHeaderIterator to avoid mixing low level pointers logic and high level units handling.
It' will bring too many changes without actual benefits.  I'd like to avoid it in current PR.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
dmitryermilov: I'll think.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
dmitryermilov: Used for slice initialization. 
At the time of slice decoding, decoder may have headers different from the headers used for previous slice initialization.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_bitstream.h
```diff
@@ -0,0 +1,242 @@
...
+            m_pbs += (nbits / 8);
+
+            int32_t bitsOffset = nbits % 8;
+
+            if ((bitsOffset + m_bitOffset) >= 8)
+            {
+                ++m_pbs;
+                bitsOffset -= 8;
+            }
+            m_bitOffset += bitsOffset;
```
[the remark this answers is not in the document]
dmitryermilov: Ok.Thanks

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_bitstream.h
```diff
@@ -0,0 +1,242 @@
...
+        // Read N bits from bitstream array
+        inline uint32_t GetBits(uint32_t nbits)
+        {
+            uint32_t bits = 0;
+            for (; nbits > 0; --nbits)
+            {
+                bits <<= 1;
+                bits |= GetBit();
+            }
+
```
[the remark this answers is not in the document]
dmitryermilov: It can be of course. Suggest to do it later.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -0,0 +1,245 @@
...
+        mfxVideoParam                   m_firstMfxVideoParams;
+
+        DPBType                         m_dpb;     // store of decoded frames
+
+        MPEG2DecoderFrame*              m_currFrame;
+
+        struct Headers
+        {
+            std::shared_ptr<MPEG2SequenceHeader>           seqHdr;
+            std::shared_ptr<MPEG2SequenceExtension>        seqExtHdr;
```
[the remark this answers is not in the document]
dmitryermilov: Yes, for MPEG2 and no for MPEG1. Consider if SW fallback is enabled in the future.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
```diff
@@ -0,0 +1,289 @@
...
+
+namespace UMC_MPEG2_DECODER
+{
+    class MPEG2DecoderFrame;
+    typedef std::list<MPEG2DecoderFrame*> DPBType;
+
+    static uint8_t start_code_prefix[] = {0, 0, 1};
+    static size_t prefix_size = sizeof(start_code_prefix);
+
+    static uint16_t MFX_PROFILE_MPEG1 = 8;
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
```diff
@@ -0,0 +1,289 @@
...
+    public:
+
+        void Reset()
+        { m_buffer.clear(); }
+
+        size_t GetSize() const
+        { return m_buffer.size(); }
+
+        uint8_t * GetPointer()
+        { return m_buffer.empty() ? nullptr : &m_buffer[0]; }
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_slice.h
```diff
@@ -0,0 +1,105 @@
...
+
+        // Decode slice header and initialize slice structure with parsed values
+        bool Reset();
+
+    public:
+        // Decoder slice header
+        bool DecodeSliceHeader();
+
+        const MPEG2SequenceHeader & GetSeqHeader(void) const
+        { return *m_seqHdr.get(); }
```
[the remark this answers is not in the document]
dmitryermilov: It can't be in current implementation. Check how slice are initialized. "DecodeSliceHeader"

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_slice.h
dmitryermilov: Shared ownership is needed to manage memory for seq and pic headers. Do you suggest another way? Which one?

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
```diff
@@ -0,0 +1,280 @@
...
+        {
+            UpdateErrorWithRefFrameStatus();
+            decodingCompleted = true;
+        }
+
+        // Check reference frames for error status and flag current frame if error is found
+        void UpdateErrorWithRefFrameStatus();
+        bool CheckReferenceFrameError();
+
+        bool IsReadyToBeOutputted() const
```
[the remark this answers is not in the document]
dmitryermilov: I doubt. https://www.dictionary.com/browse/outputted

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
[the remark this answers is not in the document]
mgonchar: for empty container ```.data()``` may return non-null pointer

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -0,0 +1,245 @@
...
+
+        uint32_t                        m_dpbSize;
+
+        Splitter                        m_splitter;
+
+        double                          m_localDeltaFrameTime;
+        bool                            m_useExternalFramerate;
+        double                          m_localFrameTime;
+
+        std::unique_ptr<Payload_Storer> m_messages;
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
```diff
@@ -0,0 +1,289 @@
...
+        uint8_t   constrained_parameters_flag     = 0;
+        uint8_t   load_intra_quantiser_matrix     = 0;
+        uint8_t   intra_quantiser_matrix[64]      = {};
+        uint8_t   load_non_intra_quantiser_matrix = 0;
+        uint8_t   non_intra_quantiser_matrix[64]  = {};
+
+        void Reset()
+        {
+            *this = {};
+        }
```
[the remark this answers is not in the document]
dmitryermilov: I think it's not needed. Currently default values are handles in packer.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
```diff
@@ -0,0 +1,289 @@
...
+        MPEG2_P_PICTURE = 2,
+        MPEG2_B_PICTURE = 3
+    };
+
+    enum
+    {
+        FLD_STRUCTURE        = 1,
+        TOP_FLD_STRUCTURE    = 1,
+        BOTTOM_FLD_STRUCTURE = 2,
+        FRM_STRUCTURE        = 3,
```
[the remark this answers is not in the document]
dmitryermilov: Ok. Thanks

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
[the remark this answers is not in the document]
lakulako: They are compatible. No problem with mpeg1

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_defs.h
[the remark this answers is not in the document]
lakulako: Correct, but better use empty() or size() to decide whether the container can be accessed, rather that GetPointer result. imho.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_slice.h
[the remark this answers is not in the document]
lakulako: Implementation can be changed, the use of GetSeqHeader extended. Better to avoid code that can be improperly used.

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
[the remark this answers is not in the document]
lakulako: std::max )

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1145 @@
...
+    if (newPar->mfx.FrameInfo.Height > oldPar->mfx.FrameInfo.Height)
         return false;
 
-    bool progseq = ((m_implUmc->GetSequenceHeader()).progressive_sequence != 0);
+    if (newPar->mfx.FrameInfo.Width > oldPar->mfx.FrameInfo.Width)
+        return false;
 
-    if (!dec_frame_count)
-        m_fieldsInCurrFrame = 0;
+    if (newPar->mfx.FrameInfo.FourCC != oldPar->mfx.FrameInfo.FourCC)
```
lakulako: Already at 292

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1,3293 +1,1145 @@
...
-            case MFX_TIMESTAMPCALC_UNKNOWN:
-                break;
-
-            default:
-                return MFX_ERR_INVALID_VIDEO_PARAM;
-        }
-
-        m_vdPar.numThreads = (mfxU16)(par->AsyncDepth ? par->AsyncDepth : m_pCore->GetAutoAsyncDepth());
-
-        if (m_InitH ||
```
lakulako: For IDR it should be closed_gop (or may be broken link). Also, if second field is P, it refers to same polarity field in previous frame.

lakulako: "is should" sounds incorrect, maybe bShouldUpdate?

lakulako: better !H || !W. One of them has no sense.

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
dmitryermilov: Ok. Thanks

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
dmitryermilov: Sorry, I didn't get your point. MPEG2 doesn't have the term IDR . If you mean "MFX_FRAMETYPE_I | MFX_FRAMETYPE_REF | MFX_FRAMETYPE_IDR" ,  here I'm repeating logic of old implementation. Some tests rely on it.

dmitryermilov: Ok

dmitryermilov: ok

lakulako: Sure, mpeg2 doesn't have term IDR, but IDR has traditional sense: frames after it don't depend on frames before. There were two examples in first comment.
No need to care about it now.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_slice.h
dmitry-gurulev: I'd rather have ownership (and share) on frame (maybe field) level not at slices which have weak pointers (or even raw) instead.

dmitryermilov: Suggest to do it later. This PR has been more or less validated manually.

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -19,3275 +19,1133 @@
...
-
-    if (0 < allocResponse.NumFrameActual)
-    {
-        mfxRes = m_pCore->FreeFrames(&allocResponse);
-        memset(&allocResponse, 0, sizeof(mfxFrameAllocResponse));
-    }
-
-    if (0 < m_opaqueResponse.NumFrameActual && true == m_isOpaqueMemory)
-    {
-        mfxRes = m_pCore->FreeFra
```
akokoshn: what actually does flag "force"?, I haven't fount any check for it

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder_va.cpp
```diff
@@ -0,0 +1,156 @@
...
+        UMC::Status sts = UMC::UMC_OK;
+
+        sts = m_va->BeginFrame(frame.GetMemID());
+        if (sts != UMC::UMC_OK)
+            return sts;
+
+        m_packer->PackAU(frame, fieldIndex);
+
+        sts = m_va->EndFrame();
+        if (sts != UMC::UMC_OK)
```
mgonchar: here it is useless check

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder_va.cpp
```diff
@@ -0,0 +1,156 @@
...
+        : m_va(nullptr)
+    {}
+
+    UMC::Status MPEG2DecoderVA::SetParams(UMC::BaseCodecParams* info)
+    {
+        if (!info)
+            return UMC::UMC_ERR_NULL_PTR;
+
+        MPEG2DecoderParams* dp =
+            DynamicCast<MPEG2DecoderParams, UMC::BaseCodecParams>(info);
```
mgonchar: can we get rid of this self-written cast?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
```diff
@@ -0,0 +1,316 @@
...
+        {
+            const auto pic    = slice->GetPicHeader();
+            const auto picExt = slice->GetPicExtHeader();
```
mgonchar: auto doesn't deduce reference, so here most probably copy occurs. In almost all cases it is faster than indirection caused by usage of reference, so I won't treat it as problem

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder_va.cpp
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    if (InProgress(*f))
+                        decode_queue.push_back(f);
+                }
+            );
+        }
+
+        decode_queue.sort([](MPEG2DecoderFrame const* f1, MPEG2DecoderFrame const* f2) { return f1->decOrder < f2->decOrder; });
+
+        decode_queue.remove_if(
```
lakulako: May be don't add earlier frames to decode queue from the beginning?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder_va.cpp
```diff
@@ -0,0 +1,156 @@
...
+            [&frame](MPEG2DecoderFrame const * f)
+            {
+                return f->decOrder > frame.decOrder; // remove frames beyong the current frame (in dec order)
+            }
+        );
+
+        // iterate through frames submitted to the driver in decoded order
+        for (DPBType::iterator frm = decode_queue.begin(); frm != decode_queue.end(); frm++)
```
lakulako: shorter: for (DPBType::iterator frm : decode_queue)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
```diff
@@ -0,0 +1,316 @@
...
+                    references.forward = &frame;
+                    references.forward->IncrementReference();
+                    return;
+                }
+            }
+        }
+
+        // 7.6.1 Frame/Field prediction
+        DPBType refFrames = dpb;
+        refFrames.remove_if(
```
lakulako: Can copy_if be used instead of full copy and remove_if?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
```diff
@@ -0,0 +1,316 @@
...
+        // If no reference frame is found, use 'self' reference.
+        // This is to handle cases without leading I picture
+        if (!ref)
+        {
+            ref = &frame;
+        }
+
+        if (MPEG2_P_PICTURE == pic.picture_coding_type)
+        {
+            if (ref)
```
lakulako: Can't find how is implemented for second field in P frame, which utilizes same polarity field from previous frame and first field of the current frame. (7.6.2.1 in 13818-2 spec)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
```diff
@@ -0,0 +1,316 @@
...
+        }
+        else // MPEG2_B_PICTURE
+        {
+            if (ref)
+            {
+                references.backward = ref;
+                references.backward->IncrementReference();
+            }
+
+            refFrames.remove(ref);
```
lakulako: If !ref were possible:
if !ref references.backward is unassigned. And future computations can be skipped?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
lakulako: !ref seems to be impossible: see line 123.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_slice.cpp
```diff
@@ -18,38 +18,39 @@
...
-#define MPEG2_LEVEL_H14    6
-#define MPEG2_LEVEL_HIGH   4
-
-    class MPEG2VideoDecoderBase;
-    
+    bool MPEG2Slice::Reset()
+    {
+        m_bitStream.Reset((uint8_t *) source.GetDataPointer(), (uint32_t) source.GetDataSize());
+
+        if (source.GetDataSize() && false == DecodeSliceHeader())
```
lakulako: If source.GetDataSize() ==0 the method returns true. Is it OK?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+        // We have cached and got new piece of data
+        if (m_cache.size())
+        {
+            uint8_t* unitStart = FindStartCode(begin, end); // Check if new data have any start codes
+            if (unitStart) // Start code was found. Need to close the cached unit
+            {
+                readSize = unitStart - begin; // Up to found startcode
+                m_cache.insert(m_cache.end(), (uint8_t *)begin, (uint8_t *)begin + readSize); // load data to the cache
+
```
lakulako: Probably with m_cahce.insert(m_cache.end(), 3, 0); // append 3 zero bytes
can avoid checking end of data with every GetBits.
It would also require similar appending in FindRawUnit.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+        RawUnit naluTmp = FindRawUnit(begin, end);
+        if (-1 == naluTmp.type || (!naluTmp.begin && !naluTmp.end)) // nothing found
+        {
+            readSize = end - begin; // Skip all data
+        }
+        else if (naluTmp.begin && !naluTmp.end) // Found only where unit starts
+        {
+            // Consume all data from buffer except last 2 bytes.
+            // An exception from this is 'MFX_BITSTREAM_COMPLETE_FRAME' case.
+            // These two bytes are to prevent start code chopping between two separete FindRawUnit() calls.
```
lakulako: separAte

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+            uint32_t numBytesToLeave = !((m_source->GetFlags() & UMC::MediaData::FLAG_VIDEO_DATA_NOT_FULL_FRAME)) ? 0 : (prefix_size - 1);
```
lakulako: duplicated (( ))

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
```diff
@@ -0,0 +1,488 @@
...
+    // Check HW capabilities
+    bool IsNeedPartialAcceleration_MPEG2(mfxVideoParam* par)
+    {
+        return (par->mfx.CodecProfile == MFX_PROFILE_MPEG1) ? true : false;
+    }
+
+    // Returns implementation platform
+    eMFXPlatform GetPlatform_MPEG2(VideoCORE * core, mfxVideoParam * par)
+    {
+        (void)core;
```
lakulako: unnamed argument is preferable

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
```diff
@@ -0,0 +1,488 @@
...
+                out->IOPattern = MFX_IOPATTERN_OUT_VIDEO_MEMORY;
+
+            // DECODE's configurables
+            out->mfx.NumThread = 1;
+            out->AsyncDepth = MFX_AUTO_ASYNC_DEPTH_VALUE;
+        }
+        else
+        {
+            // checking mode
+            if (1 == in->mfx.DecodedOrder)
```
lakulako: Isn't DecodedOrder is supported?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
```diff
@@ -0,0 +1,488 @@
...
+            if (in->mfx.FrameInfo.FourCC != MFX_FOURCC_NV12)
+            {
+                out->mfx.FrameInfo.FourCC = 0;
+                return MFX_ERR_UNSUPPORTED;
+            }
+
+            if (in->NumExtParam != 0 && !in->ExtParam)
+                return MFX_ERR_UNSUPPORTED;
+
+            if (!in->NumExtParam && in->ExtParam)
```
lakulako: Merge 2 ifs: if( (in->NumExtParam == 0) != (in->ExtParam == nullptr) )

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
```diff
@@ -0,0 +1,488 @@
...
+                return MFX_LEVEL_MPEG2_HIGH;
+            default:
+                return MFX_LEVEL_UNKNOWN;
+        }
+    }
+
+    bool DARtoPAR(uint32_t width, uint32_t height, uint32_t dar_h, uint32_t dar_v,
+                uint16_t & par_h, uint16_t &par_v)
+    {
+        uint32_t simple_tab[] = {2,3,5,7,11,13,17,19,23,29,31,37,41,43,47,53,59};
```
lakulako: Euclidean algorithm for ratio simplification can be used.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
dmitryermilov: No. We even have test which check it,

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
dmitryermilov: This is copy paste from legacy implementation,

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+
+                rawHdr.Resize(it->end - it->begin);
+                std::copy(it->begin, it->end, rawHdr.GetPointer());
+            }
+            else if (EXTENSION == it->type)
+            {
+                uint8_t ext_type = it->begin[prefix_size + 1] >> 4;
+                if (SEQUENCE_EXTENSION == ext_type)
+                {
+                    // seq header already found
```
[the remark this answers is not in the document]
dmitryermilov: You forget about MPEG1

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+        if (picExt.picture_structure != FRM_PICTURE)
+        {
+            frame.bottom_field_flag[0] = (picExt.picture_structure == BOTTOM_FLD_PICTURE);
+            frame.bottom_field_flag[1] = !(frame.bottom_field_flag[0]);
+
+            frame.pictureStructure = FLD_PICTURE;
+        }
+        else
+        {
+            frame.bottom_field_flag[0] = 0;
```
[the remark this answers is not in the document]
dmitryermilov: I din't get the question. Please explain your point.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+            m_messages->SetAUID(m_currFrame->currFieldIndex);
+
+        // 6.1.1.11 - Reordering:
+        // - If a current frame is a B-frame, then output it immediately.
+        // - if a current frame is a I-frame or P-frame, then output previous (in decoded order) I-frame or P-frame.
+        if (m_currFrame->frameType == MPEG2_B_PICTURE)
+        {
+            m_currFrame->SetReadyToBeOutputted();
+
+            auto prev = FindPrevRefFrame(m_currFrame);
```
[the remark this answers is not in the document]
dmitryermilov: Previous in decoder order.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+                    uint32_t const id1 = f1->IsRef() ? f1->decOrder : (std::numeric_limits<uint32_t>::max)();
+                    uint32_t const id2 = f2->IsRef() ? f2->decOrder : (std::numeric_limits<uint32_t>::max)();
+
+                    return  id1 < id2;
+                }
+            );
+
+            auto oldest = (i == std::end(m_dpb)) ? nullptr : *i;
+            if (oldest)
+                oldest->SetIsRef(false); // Release the oldest reference frame ...
```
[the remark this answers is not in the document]
dmitryermilov: Ok.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+            frame = new MPEG2DecoderFrame;
+
+            // Add to DPB
+            m_dpb.push_back(frame);
+        }
+
+        return frame;
+    }
+
+    MPEG2DecoderFrame* MPEG2Decoder::FindPrevRefFrame(MPEG2DecoderFrame const* curr)
```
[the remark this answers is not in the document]
dmitryermilov: Then it will look like it works in terms of display order.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+        {
+            m_nalu = OnEOS();
+            return;
+        }
+
+        m_nalu = FindUnit(m_source);
+
+        if ( (-1 == m_nalu.type) &&  // didn't find complete unit
+                m_cache.size() &&       // but have beginning of the unit
+                !((m_source->GetFlags() & UMC::MediaData::FLAG_VIDEO_DATA_NOT_FULL_FRAME)) ) // and complete frame flag is set
```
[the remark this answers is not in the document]
dmitryermilov: There is no mistake here. Check FLAG_VIDEO_DATA_NOT_FULL_FRAME how is set.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
dmitryermilov: What ?

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
```diff
@@ -0,0 +1,280 @@
...
+
+        bool IsFilled () const
+        { return isFilled; }
+
+    private:
+        MPEG2DecoderFrame&                frame;
+        bool                              isField;
+        bool                              isBottomField;
+        FrameType                         frameType;
+        bool                              isFilled;
```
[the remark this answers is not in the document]
dmitryermilov: e.g. ?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+        SkipBits(1);  // extra_bit_slice, shall be 0
+
+        return UMC::UMC_OK;
+    }
+
+    // Parse full slice header
+    UMC::Status MPEG2HeadersBitstream::GetSliceHeaderFull(MPEG2Slice & slice, const MPEG2SequenceHeader & seq, const MPEG2SequenceExtension & seqExt)
+    {
+        UMC::Status sts = GetSliceHeader(slice.GetSliceHeader(), seq, seqExt);
```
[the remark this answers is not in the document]
dmitryermilov: This naming is used in AVC and HEVC.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+                macroblock_address_increment += 33;
+            }
+
+            SkipBits(bitsToRead);
+        }
+
+        if (cc >= 1024)
+        {
+            cc = GetBits(1);
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+                CompletePicture(*m_currFrame, m_currFrame->currFieldIndex);
+                m_currFrame->SetFullFrame(true);
+
+                m_currFrame = nullptr;
+
+                return UMC::UMC_OK; // exit from the loop on full frame
+            }
+        }
+
+        // EOS handling
```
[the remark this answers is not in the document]
dmitryermilov: EOS

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+                    auto seqHdr    = std::make_shared<MPEG2SequenceHeader>();
+                    auto seqExtHdr = std::make_shared<MPEG2SequenceExtension>();
+                    bitStream.GetSequenceHeader(*seqHdr.get());
+
+                    uint8_t* seqExtBegin = RawHeaderIterator::FindStartCode(unit.begin + prefix_size + bitStream.BytesDecoded(), unit.end);
+                    if (!seqExtBegin)
+                        return UMC::UMC_ERR_INVALID_STREAM;
+
+                    bitStream.Reset(seqExtBegin + prefix_size, unit.end - seqExtBegin);
+                    bitStream.SkipBits(8 + 4); // skip unit and extension types
```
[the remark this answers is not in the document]
dmitryermilov: It's guaranteed in splitter class. splitter appends seq and seq ext, pic and pic ext. If there is no pair, the unit is skipped.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+
+                    // Save headers
+                    m_currHeaders.picHdr = std::move(picHdr);
+                    m_currHeaders.picExtHdr = std::move(picExtHdr);
+                }
+                break;
+            case EXTENSION:
+            {
+                auto type_ext = (MPEG2_UNIT_TYPE_EXT)bitStream.GetBits(4);
+                switch(type_ext)
```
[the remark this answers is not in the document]
dmitryermilov: Then how to handle it at the upper level? We can either return an error to app or ignore. If ignore - then such handling isn't needed here.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
dmitryermilov: Agree, but let's leave this optimization later because it requires discussion. E.g. my proposal is to remove checks in GetBits and make one Check call after parsing the whole unit.

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
dmitryermilov: out of date

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
```diff
@@ -0,0 +1,383 @@
...
+        uint8_t bitsToRead;
+
+        for (;;)
+        {
+            bitsToRead = BitsLeft() <= 11 ? BitsLeft() : 11;
+
+            cc = GetBits(bitsToRead);
+            for (int32_t nbits = 11 - bitsToRead; nbits > 0; --nbits) // if left bits is less than 11
+                cc <<= 1;
+
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder_va.cpp
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
dmitryermilov: it's here. Just previous in dec order frame

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+                m_messages->AddMessage(unit, -1);
+                break;
+            default:
+                if (unit.type >= 0x1 && unit.type <= 0xAF) // Slice header start codes
+                {
+                    if (auto slice = DecodeSliceHeader(unit))
+                    {
+                        UMC::Status sts = AddSlice(slice);
+                        if (sts == UMC::UMC_ERR_NOT_ENOUGH_BUFFER || sts == UMC::UMC_OK)
+                            return sts;
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -0,0 +1,1145 @@
...
+
+    static
+    bool IsFieldOfOneFrame(const MPEG2PictureHeader& picHdr, const MPEG2PictureCodingExtension& picExtHdr,
+                           const MPEG2PictureHeader & newPictureHdr, const MPEG2PictureCodingExtension& newPicExtHdr)
+    {
+        if (picHdr.temporal_reference != newPictureHdr.temporal_reference)  // 6.3.9
+            return false;
+
+        if ((picExtHdr.picture_structure == FRM_PICTURE) != (newPicExtHdr.picture_structure == FRM_PICTURE)) // not the same frame structures
+            return false;
```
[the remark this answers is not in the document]
dmitryermilov: I'll remove this

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+        {
+            m_nalu = OnEOS();
+        }
+    }
+
+    RawHeaderIterator & RawHeaderIterator::operator++()
+    {
+        m_nalu = RawUnit(); // reset
```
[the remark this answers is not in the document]
dmitryermilov: Duplication is very small, I don't want to introduce a new function so far.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_slice.cpp
dmitryermilov: changed

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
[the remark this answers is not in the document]
lakulako: prepared, extracted, complete, ready.
frameFilled...

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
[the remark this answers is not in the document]
lakulako: Is mpeg1 supported?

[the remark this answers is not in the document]
dmitryermilov: Yes by DecodeHeader()

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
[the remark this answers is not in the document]
lakulako: LatestRef is latest in either order.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
lakulako: It is for "the first field is I"

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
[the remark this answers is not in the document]
lakulako: You can implement op++ as {LoadData(m_source); return *this;}

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp
lakulako: ok, improve in future

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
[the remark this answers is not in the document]
dmitryermilov: Why is it better?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_splitter.cpp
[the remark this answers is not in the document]
dmitryermilov: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
dmitryermilov: I don't understand what you're looking for. IP case is above. P/B frame are at 128 and below.

lakulako: I'm looking for 2nd P field case. Not necessary in IP, but in PP pair.
Second field refers to 2(!) frames - previous and current (first field of the current).

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_bitstream.h
```diff
@@ -0,0 +1,229 @@
...
+            return bits;
+        }
+
+        // Return bitstream position pointer N bits backward
+        inline void UnGetBits(uint32_t nbits)
+        {
+            m_pbs -= (nbits / 8);
```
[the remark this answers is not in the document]
dmitryermilov: Ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp
dmitryermilov: I see what you mean. For HW implementation we need to pass a reference surface(s).  So reference from the first field is available to the driver without an extra information from decoder

PR #998, opened 2018-12-08
[CMAKE] Refactoring lib and plugin information

aulitin:
Add new information to lib and plugin: MSDK api version, commit,
build info, MSDK version
Deleted information: MSDK file version, MSDK product version


_studio/mfx_lib/plugin/include/plugin_version_linux.h
```diff
@@ -22,22 +22,24 @@
 
 #pragma once
 
+#include "va/va.h"
```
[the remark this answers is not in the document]
aulitin: I think _studio/mfx_lib/plugin/include/plugin_version_linux.h and _studio/hevce_hw/h265/include/plugin_version_linux.h have a different meaning, I don't quite understand which of them should be removed.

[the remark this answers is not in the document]
onabiull: There is no need in hevc- or vp9-specific versions of this file. Let keep only generic one.

_studio/mfx_lib/shared/src/libmfxsw.cpp
```diff
@@ -39,27 +41,26 @@ void* g_hModule = NULL; // DLL handle received in DllMain
...
-#ifndef MFX_PRODUCT_VERSION
-#define MFX_PRODUCT_VERSION "0.0.000.0000"
-#endif
 
 #if defined(mfxhw64_EXPORTS) || defined(mfxhw32_EXPORTS) || defined(mfxsw64_EXPORTS) || defined(mfxsw32_EXPORTS)
-const char* g_MfxProductName = "mediasdk_product_name: Intel(R) Media SDK";
-const char* g_MfxCopyright = "mediasdk_copyright: Copyright(c) 2011-2018 Intel Corporation";
-const char* g_MfxFileVersion = "mediasdk_file_version: " MFX_FILE_VERSION;
-const char* g_MfxProductVersion = "mediasdk_product_version: " MFX_PRODUCT_VERSION;
+const char* g_MfxProductName     = "mediasdk_product_name: Intel(R) Media SDK";
```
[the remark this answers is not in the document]
dvrogozh: Not address. Please, drop all these defined(mfxhw64_EXPORTS) and use honest __linux__

_studio/mfx_lib/plugin/include/plugin_version_linux.h
```diff
@@ -22,22 +22,23 @@
 
 #pragma once
 
+
```
dvrogozh: "there can only be one"

_studio/hevce_hw/h265/include/plugin_version_linux.h
```diff
@@ -23,22 +23,22 @@
...
-#endif
 #ifndef MFX_FILE_DESCRIPTION
 #define MFX_FILE_DESCRIPTION "File info ought to be here"
 #endif
 
-const char* g_MfxProductName = "mediasdk_product_name: " MFX_PLUGIN_PRODUCT_NAME;
-const char* g_MfxCopyright = "mediasdk_copyright: Copyright(c) 2007-2018 Intel Corporation";
-const char* g_MfxProductVersion = "mediasdk_product_version: " MFX_PLUGIN_PRODUCT_VERSION;
-const char* g_MfxFileVersion = "mediasdk_file_version: " MFX_PLUGIN_FILE_VERSION;
+#define STRINGIZE(s) #s
```
DenWolf: Why do we need 2 defines when one of them call another just for only one purpose - converting marco to string ?
I mean why don't you squash 2 define to 1 :
#define CONVERT_TO_STRING(s) #s

_studio/mfx_lib/shared/src/libmfxsw.cpp
```diff
@@ -39,26 +38,19 @@ void* g_hModule = NULL; // DLL handle received in DllMain
...
-const char* g_MfxFileVersion = "mediasdk_file_version: " MFX_FILE_VERSION;
-const char* g_MfxProductVersion = "mediasdk_product_version: " MFX_PRODUCT_VERSION;
-#endif
+#define STRINGIZE(s) #s
+#define CONVERT_TO_STRING(s) STRINGIZE(s)
+
+const char* g_MfxProductName     = "mediasdk_product_name: Intel(R) Media SDK";
+const char* g_MfxCopyright       = "mediasdk_copyright: Copyright(c) 2011-2019 Intel Corporation";
+const char* g_MfxApiVersion      = "mediasdk_api_version: " MFX_API_VERSION;
+const char* g_MfxVersion         = "mediasdk_version: " CONVERT_TO_STRING(MFX_VERSION_MAJOR) "." CONVERT_TO_STRING(MFX_VERSION_MINOR);
```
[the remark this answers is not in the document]
onabiull: Having single file with version meta-information is pretty common practice. No objections from me :)

builder/FindGlobals.cmake
```diff
@@ -185,3 +183,52 @@ endfunction()
...
+    COMMAND lsb_release -i
+    OUTPUT_VARIABLE OUTPUT
+    ERROR_VARIABLE ERROR
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  if(NOT ERROR)
+    string(REPLACE  "Distributor ID:	" ""  MFX_LINUX_NAME "${OUTPUT}")
+    set(MFX_SYSTEM "${MFX_LINUX_NAME}")
+  endif()
+elseif(WIN32)
```
dvrogozh: WIN32 looks suspicious... Please, double check how to distinguish Linux.

builder/FindGlobals.cmake
```diff
@@ -185,3 +183,52 @@ endfunction()
 function(report_add_target var target)
   set(${ARGV0} ${${ARGV0}} ${ARGV1} CACHE INTERNAL "" FORCE)
 endfunction()
+
+# Defined OS name and version and build info and build commit
+if(APPLE)
+  set(MFX_SYSTEM "MAC")
+elseif(UNIX)
+  execute_process(
```
dvrogozh: What will happen if lsb_release is not available? I am specifically worry about such flavors as FreeBSD. Can you, please, account for that and at least make sure that we will print out "unknown"?

builder/FindGlobals.cmake
```diff
@@ -185,3 +183,52 @@ endfunction()
...
+  if(NOT ERROR)
+    string(REPLACE  "Distributor ID:	" ""  MFX_LINUX_NAME "${OUTPUT}")
+    set(MFX_SYSTEM "${MFX_LINUX_NAME}")
+  endif()
+elseif(WIN32)
+  set(MFX_SYSTEM "Windows")
+endif()
+
+execute_process(
+  COMMAND getconf GNU_LIBC_VERSION
```
dvrogozh: what if getconf is not available?

_studio/mfx_lib/shared/include/mediasdk_version.h
```diff
@@ -18,27 +18,23 @@
...
 
-const char* g_MfxProductName = "mediasdk_product_name: " MFX_PLUGIN_PRODUCT_NAME;
-const char* g_MfxCopyright = "mediasdk_copyright: Copyright(c) 2007-2018 Intel Corporation";
-const char* g_MfxProductVersion = "mediasdk_product_version: " MFX_PLUGIN_PRODUCT_VERSION;
-const char* g_MfxFileVersion = "mediasdk_file_version: " MFX_PLUGIN_FILE_VERSION;
-const char* g_MfxFileDescription = "mediasdk_file_description: " MFX_FILE_DESCRIPTION;
+const char* g_MfxProductName     = "mediasdk_product_name: Intel(R) Media SDK";
+const char* g_MfxCopyright       = "mediasdk_copyright: Copyright(c) 2007-2019 Intel Corporation";
+const char* g_MfxApiVersion      = "mediasdk_api_version: " MFX_API_VERSION;
+const char* g_MfxVersion         = "mediasdk_version: master";
```
dvrogozh: emmm. I have some feeling that we should have not only master here. From time to time at least:).

builder/FindGlobals.cmake
aulitin: Then lib will not have version glibc.

builder/FindGlobals.cmake
aulitin: If lsb_release is not available than output will be first string output command.
Search performed by parsing string with target words "Distributor ID:<Tab>" if these words are missing than the output will be first string output command. I will add check 
```
if(NOT "${OUTPUT}" STREQUAL "${MFX_LINUX_NAME}")
  set(MFX_SYSTEM "${MFX_LINUX_NAME}"
```

builder/FindGlobals.cmake
aulitin: This is correct check.
https://cmake.org/cmake/help/v3.0/variable/WIN32.html

aulitin:
@dvrogozh, If mediasdk is built outside git then output will
```
mediasdk_product_name: Intel(R) Media SDK
mediasdk_copyright: Copyright(c) 2011-2019 Intel Corporation
mediasdk_api_version: 1.28
mediasdk_version: 1.28
mediasdk_git_commit:
mediasdk_build_info: Ubuntu 4.19.5 | GNU 7.3.0 | glibc 2.27
```

_studio/mfx_lib/shared/include/mediasdk_version.h
onabiull: I believe it supposed to be corrected manually after branch out.

PR #1002, opened 2018-12-10
[bs_parser_hevc] Add required checks

ndomnina:


tools/bs_parser_hevc/src/hevc2_parser.cpp
```diff
@@ -759,7 +759,10 @@ Bs32u Parser::ParseSSDSubmit(SDThread*& pSDT, NALU* AU)
 
         sdpar.Emulation = true;
 
-        TargetRBSP = Bs32u(CurRBSP / m_sdt.size());
+        if (m_sdt.empty())
```
mgonchar: no need of this, I synced with Evgeniy, size of vector can't be zero here

tools/bs_parser_hevc/src/hevc2_parser.cpp
```diff
@@ -875,7 +878,11 @@ Bs32u Parser::ParseSSDSubmit(SDThread*& pSDT, NALU* AU)
         if (sdpar.ColPic)
         {
             lock(sdpar.ColPic);
-            sliceDep.push_back(spColPic);
+
+            if (spColPic)
```
mgonchar: @eistarov do we need this check?

tools/bs_parser_hevc/src/hevc2_parser.cpp
ndomnina: This check was removed

tools/bs_parser_hevc/src/hevc2_parser.cpp
ndomnina: This check was removed

tools/bs_parser_hevc/tools/hevc_fei_extractor/src/main.cpp
```diff
@@ -67,7 +67,7 @@ class FileHandler
             m_errorSts = true;// Error sts
         }
 
-        if ((m_pFile = fopen(fileName, mode)) == nullptr) {
+        if (fileName && (m_pFile = fopen(fileName, mode)) == nullptr) {
```
mgonchar: filename is checked at line 60 above. No need to check again

ndomnina: But we do not exit the function after check. If the file name is empty, we still try to open the file.

ndomnina: Fixed

PR #1004, opened 2018-12-11
[samples] fix drm drivers detection

vilichev:
Current mechanism to detect Intel drm drivers uses the following schema:
1) Walk through the list of PCI devices, get device properties
   (Vendor ID) and calculate number of PCI devices  that corresponds
   to ‘DISPLAY_CONTROLLER_CLASS’ (adapters_num)
2) Use adapters_num to get path to /dev/dri/cardN and /dev/dri/renderDN
   with checking for Vendor ID.
But this approach is not working in some cases. For example, if PCI device
is present but corresponding driver is not set up. So it can be attempt to
open wrong /dev/dri/renderN and /dev/dri/cardN  paths.

New approach is based on checking for /dev/dri/renderD<N> (N=0..15) with
drm request for driver name.

### DIFF
--- a/samples/sample_common/src/vaapi_utils_drm.cpp
+++ b/samples/sample_common/src/vaapi_utils_drm.cpp
@@ -22,216 +22,91 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "vaapi_utils_drm.h"
 #include "vaapi_allocator.h"
 #include <fcntl.h>
+#include <sys/ioctl.h>
 
-#include <dirent.h>
 #include <stdexcept>
 
+#include "vaapi_utils_drm.h"
 #include <drm_fourcc.h>
-#include <intel_bufmgr.h>
 
-#define MFX_PCI_DIR "/sys/bus/pci/devices"
-#define MFX_DRI_DIR "/dev/dri/"
-#define MFX_PCI_DISPLAY_CONTROLLER_CLASS 0x03
+constexpr mfxU32 MFX_DRI_MAX_NODES_NUM = 16;
+constexpr mfxU32 MFX_DRI_RENDER_START_INDEX = 128;
+constexpr mfxU32 MFX_DRI_CARD_START_INDEX = 0;
+constexpr  mfxU32 MFX_DRM_DRIVER_NAME_LEN = 4;
+const char* MFX_DRM_INTEL_DRIVER_NAME = "i915";
+const char* MFX_DRI_PATH = "/dev/dri/";
+const char* MFX_DRI_NODE_RENDER = "renderD";
+const char* MFX_DRI_NODE_CARD = "card";
 
-struct mfx_disp_adapters
+int get_drm_driver_name(int fd, char *name, int name_size)
 {
-    mfxU32 vendor_id;
-    mfxU32 device_id;
-};
-
-static int mfx_dir_filter(const struct dirent* dir_ent)
-{
-    if (!dir_ent) return 0;
-    if (!strcmp(dir_ent->d_name, ".")) return 0;
-    if (!strcmp(dir_ent->d_name, "..")) return 0;
-    return 1;
+    drm_version_t version = {};
+    version.name_len = name_size;
+    version.name = name;
+    return ioctl(fd, DRM_IOWR(0, drm_version), &version);
 }
 
-typedef int (*fsort)(const struct dirent**, const struct dirent**);
-
-static mfxU32 mfx_init_adapters(struct mfx_disp_adapters** p_adapters)
+int open_intel_adapter(int type)
 {
-    mfxU32 adapters_num = 0;
-    int i = 0;
-    struct mfx_disp_adapters* adapters = NULL;
-    struct dirent** dir_entries = NULL;
-    int entries_num = scandir(MFX_PCI_DIR, &dir_entries, mfx_dir_filter, (fsort)alphasort);
+    std::string adapterPath = MFX_DRI_PATH;
+    char driverName[MFX_DRM_DRIVER_NAME_LEN + 1] = {};
+    mfxU32 nodeIndex;
 
-    char file_name[300] = {};
-    char str[16] = {0};
-    FILE* file = NULL;
-
-    for (i = 0; i < entries_num; ++i)
-    {
-        long int class_id = 0, vendor_id = 0, device_id = 0;
-
-        if (!dir_entries[i])
-            continue;
-
-        // obtaining device class id
-        snprintf(file_name, sizeof(file_name)/sizeof(file_name[0]), "%s/%s/%s", MFX_PCI_DIR, dir_entries[i]->d_name, "class");
-        file = fopen(file_name, "r");
-        if (file)
-        {
-            if (fgets(str, sizeof(str), file))
-            {
-                class_id = strtol(str, NULL, 16);
-            }
-            fclose(file);
-
-            if (MFX_PCI_DISPLAY_CONTROLLER_CLASS == (class_id >> 16))
-            {
-                // obtaining device vendor id
-                snprintf(file_name, sizeof(file_name)/sizeof(file_name[0]), "%s/%s/%s", MFX_PCI_DIR, dir_entries[i]->d_name, "vendor");
-                file = fopen(file_name, "r");
-                if (file)
-                {
-                    if (fgets(str, sizeof(str), file))
-                    {
-                        vendor_id = strtol(str, NULL, 16);
-                    }
-                    fclose(file);
-                }
-                // obtaining device id
-                snprintf(file_name, sizeof(file_name)/sizeof(file_name[0]), "%s/%s/%s", MFX_PCI_DIR, dir_entries[i]->d_name, "device");
-                file = fopen(file_name, "r");
-                if (file)
-                {
-                    if (fgets(str, sizeof(str), file))
-                    {
-                        device_id = strtol(str, NULL, 16);
-                    }
-                    fclose(file);
-                }
-                // adding valid adapter to the list
-                if (vendor_id && device_id)
-                {
-                    struct mfx_disp_adapters* tmp_adapters = NULL;
-
-                    tmp_adapters = (mfx_disp_adapters*)realloc(adapters,
-                                                               (adapters_num+1)*sizeof(struct mfx_disp_adapters));
-
-                    if (tmp_adapters)
-                    {
-                        adapters = tmp_adapters;
-                        adapters[adapters_num].vendor_id = vendor_id;
-                        adapters[adapters_num].device_id = device_id;
-
-                        ++adapters_num;
-                    }
-                }
-            }
-        }
-        free(dir_entries[i]);
+    switch (type) {
+    case MFX_LIBVA_DRM:
+    case MFX_LIBVA_AUTO:
+        adapterPath += MFX_DRI_NODE_RENDER;
+        nodeIndex = MFX_DRI_RENDER_START_INDEX;
+        break;
+    case MFX_LIBVA_DRM_MODESET:
+        adapterPath += MFX_DRI_NODE_CARD;
+        nodeIndex = MFX_DRI_CARD_START_INDEX;
+        break;
+    default:
+        throw std::invalid_argument("Wrong libVA backend type");
     }
-    if (entries_num) free(dir_entries);
-    if (p_adapters) *p_adapters = adapters;
 
-    return adapters_num;
+    for (mfxU32 i = 0; i < MFX_DRI_MAX_NODES_NUM; ++i) {
+        std::string curAdapterPath = adapterPath + std::to_string(nodeIndex + i);
+
+        int fd = open(curAdapterPath.c_str(), O_RDWR);
+        if (fd < 0) continue;
+
+        if (!get_drm_driver_name(fd, driverName, MFX_DRM_DRIVER_NAME_LEN) &&
+            !strcmp(driverName, MFX_DRM_INTEL_DRIVER_NAME)) {
+            return fd;
+        }
+        close(fd);
+    }
+
+    return -1;
 }
 
 DRMLibVA::DRMLibVA(int type)
     : CLibVA(type)
     , m_fd(-1)
 {
-    const mfxU32 IntelVendorID = 0x8086;
-    //the first Intel adapter is only required now, the second - in the future
-    const mfxU32 numberOfRequiredIntelAdapter = 1;
-    const char nodesNames[][8] = {"renderD", "card"};
-
-    VAStatus va_res = VA_STATUS_SUCCESS;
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,108 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-    if (!strcmp(dir_ent->d_name, ".")) return 0;
-    if (!strcmp(dir_ent->d_name, "..")) return 0;
-    return 1;
-}
+    drm_version_t version = {};
+    version.name_len = name_size;
+    version.name = name;
 
-typedef int (*fsort)(const struct dirent**, const struct dirent**);
+    if (ioctl(fd, DRM_IOWR(0, drm_version), &version)) {
```
mgonchar: ok, so this is
```
int res = ioctl(fd, DRM_IOWR(0, drm_version), &version);

return res == 0;
```

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,108 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-
-#define MFX_PCI_DIR "/sys/bus/pci/devices"
-#define MFX_DRI_DIR "/dev/dri/"
-#define MFX_PCI_DISPLAY_CONTROLLER_CLASS 0x03
 
-struct mfx_disp_adapters
-{
-    mfxU32 vendor_id;
-    mfxU32 device_id;
+const mfxU32 MFX_DRI_MAX_NODES_NUM = 16;
```
mgonchar: you can make them constexpr

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,108 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+const mfxU32 MFX_DRI_RENDER_START_INDEX = 128;
+const mfxU32 MFX_DRM_DRIVER_NAME_LEN = 4;
+const std::string MFX_DRM_INTEL_DRIVER_NAME = "i915";
```
mgonchar: why strings? I think char* is enough

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,108 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-    {
-        long int class_id = 0, vendor_id = 0, device_id = 0;
+    for (mfxU32 i = 0; i < MFX_DRI_MAX_NODES_NUM; ++i) {
+        std::stringstream node;
+        node << MFX_DRI_NODE_RENDER_PATH.c_str() << MFX_DRI_RENDER_START_INDEX + i;
 
-        if (!dir_entries[i])
-            continue;
+        int fd = open(node.str().c_str(), O_RDONLY);
+        if (fd < 0) continue;
```
mgonchar: Are you trying to check existence of file? What if file exists, but ```open``` fails?

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,108 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-          adapterPaths[i] = NULL;
           continue;
         }
-        adapterPaths[i] = new char[sizeof(MFX_DRI_DIR) + sizeof(nodesNames[i]) + 3];
-        sprintf(adapterPaths[i], "%s%s%d", MFX_DRI_DIR, nodesNames[i], nodesNumbers[i]);
+
+        switch (i) {
+        case MFX_DRI_NODE_RENDER:
+            adapterPaths[i] << MFX_DRI_NODE_RENDER_PATH;
+            adapterPaths[i] << MFX_DRI_RENDER_START_INDEX + adapters.first();
```
mgonchar: ```.front()``` ?

mgonchar: Do not use ```+``` it will create unnecessary tmp strings, use ```<<``` instead

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: OK

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: yes, I will fix it

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: We need only those render nodes that can be opened

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: Agree

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,105 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+    
+    if (adapters.size() != numberOfRequiredIntelAdapter)
+        throw std::range_error("Number of required Intel adapters was not found");
 
-    // Search for the required display adapter
-    int i = 0, nFoundAdapters = 0;
-    int nodesNumbers[] = {0,0};
-    while ((i < adapters_num) && (nFoundAdapters != numberOfRequiredIntelAdapter))
+    // Initialize device paths using first found Intel adapter
+    std::stringstream adapterPaths[MFX_DRI_NODE_NUM];
```
mgonchar: I think it is an overkill, array of ```std::strings``` is just fine, only use ```+=``` for them and then you won't suffer from many temporary copies created

samples/sample_common/src/vaapi_utils_drm.cpp
mgonchar: ok, so you won't inform user if the required file exists but, for example, user don't have enough permissions to open it?

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: OK, I will do it.  But in this case it needs explicitly convert int to string

mgonchar: ```std::to_string``` will do the trick

vilichev: I meant the same )

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,104 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-          continue;
+        switch (i) {
+        case MFX_DRI_NODE_RENDER:
+            if (type == MFX_LIBVA_DRM_MODESET) continue;
+            adapterPaths[i] += MFX_DRI_NODE_RENDER_PATH +
+                               std::to_string(MFX_DRI_RENDER_START_INDEX + adapters.front());
+            break;
+        case MFX_DRI_NODE_CARD:
+            adapterPaths[i] += MFX_DRI_NODE_CARD_PATH +
+                               std::to_string(adapters.front());
```
dvrogozh: I don't see that you handle MFX_DRI_NODE_CARD in get_intel_adapters at all. You will have nothing in the vector.

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,104 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
 
-    char file_name[300] = {};
-    char str[16] = {0};
-    FILE* file = NULL;
-
-    for (i = 0; i < entries_num; ++i)
-    {
-        long int class_id = 0, vendor_id = 0, device_id = 0;
+    for (mfxU32 i = 0; i < MFX_DRI_MAX_NODES_NUM; ++i) {
+        std::string node = MFX_DRI_NODE_RENDER_PATH +
```
dvrogozh: So, we support only render nodes? Why we have cardN path then? or if you plan to support it where is a corresponding code?

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,104 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
     mfxStatus sts = MFX_ERR_NONE;
     int major_version = 0, minor_version = 0;
 
-    mfx_disp_adapters* adapters = NULL;
-    int adapters_num = mfx_init_adapters(&adapters);
+    std::vector<mfxU32> adapters;
+    get_intel_adapters(adapters);
+    
+    if (adapters.size() != numberOfRequiredIntelAdapter)
+        throw std::range_error("Number of required Intel adapters was not found");
```
dvrogozh: Message is ugly. Please, change. "Intel GPU was not found" would be better, I think.

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: OK

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,93 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+    int res = ioctl(fd, DRM_IOWR(0, drm_version), &version);
 
-static mfxU32 mfx_init_adapters(struct mfx_disp_adapters** p_adapters)
-{
-    mfxU32 adapters_num = 0;
-    int i = 0;
-    struct mfx_disp_adapters* adapters = NULL;
-    struct dirent** dir_entries = NULL;
-    int entries_num = scandir(MFX_PCI_DIR, &dir_entries, mfx_dir_filter, (fsort)alphasort);
+    return res;
```
mgonchar: well, now you don't actually need ```res```

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,169 +22,93 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+    if (type != MFX_LIBVA_DRM_MODESET) {
+        adapterPaths[MFX_DRI_NODE_RENDER] = MFX_DRI_NODE_RENDER_PATH;
+        adapterPaths[MFX_DRI_NODE_RENDER] += std::to_string(MFX_DRI_RENDER_START_INDEX + adapters.front());
     }
+    adapterPaths[MFX_DRI_NODE_CARD] = MFX_DRI_NODE_CARD_PATH;
+    adapterPaths[MFX_DRI_NODE_CARD] += std::to_string(adapters.front());
 
     // Loading display. At first trying to open render nodes, then card.
-    for (int i=0; i<2; i++)
+    for (int i = MFX_DRI_NODE_RENDER; i < MFX_DRI_NODE_NUM; i++)
```
mgonchar: this cycle is again example of tricky logic without a need in it. But that wasn't introduced by you, so I'm ok to leave it as is, but this is next candidate for refactoring

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: The fix was reworked

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,216 +22,91 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+        int major_version = 0, minor_version = 0;
+        va_res = m_libva.vaInitialize(m_va_dpy, &major_version, &minor_version);
+        sts = va_to_mfx_status(va_res);
     }
-
-    for (int i=0; i<2; i++)
-    {
-        delete [] adapterPaths[i];
+    else {
+        m_libva.vaTerminate(m_va_dpy);
```
mgonchar: I think terminate should be in first branch, where libva was initialized

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,216 +22,91 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-        adapterPaths[i] = new char[sizeof(MFX_DRI_DIR) + sizeof(nodesNames[i]) + 3];
-        sprintf(adapterPaths[i], "%s%s%d", MFX_DRI_DIR, nodesNames[i], nodesNumbers[i]);
-    }
+    m_fd = open_intel_adapter(type);
+    if (m_fd < 0) throw std::range_error("Intel GPU was not found");
 
-    // Loading display. At first trying to open render nodes, then card.
-    for (int i=0; i<2; i++)
+    m_va_dpy = m_vadrmlib.vaGetDisplayDRM(m_fd);
+    if (m_va_dpy)
```
mgonchar: I think this redesign is simpler:
```
m_va_dpy = m_vadrmlib.vaGetDisplayDRM(m_fd);

if (!m_va_dpy)
{
    fclose(m_fd);
    throw std::runtime_error("Loading of VA display was failed");
}

int major_version = 0, minor_version = 0;
VAStatus va_res = m_libva.vaInitialize(m_va_dpy, &major_version, &minor_version);

if (va_to_mfx_status(va_res) != MFX_ERR_NONE)
{
    m_libva.vaTerminate(m_va_dpy);
    close(m_fd);
    throw std::runtime_error("Loading of VA display was failed");
}
```

samples/sample_common/src/vaapi_utils_drm.cpp
```diff
@@ -22,216 +22,91 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
-                    }
-                    fclose(file);
-                }
-                // adding valid adapter to the list
-                if (vendor_id && device_id)
-                {
-                    struct mfx_disp_adapters* tmp_adapters = NULL;
+    std::string adapterPath = MFX_DRI_PATH;
+    char driverName[MFX_DRM_DRIVER_NAME_LEN + 1] = {};
+    mfxU32 nodeIndex = MFX_DRI_RENDER_START_INDEX;
```
dvrogozh: Can you leave nodeIndex either not initialized or initialized to 0? Move MFX_DRI_RENDER_START_INDEX to render start section.

samples/sample_common/src/vaapi_utils_drm.cpp
dvrogozh: +1

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: fixed

samples/sample_common/src/vaapi_utils_drm.cpp
vilichev: @dvrogozh, please review last changes

PR #1008, opened 2018-12-12
h265e: returned utility macro for parameters copy

mgonchar:


onabiull:
Maybe it's time to abandon macros? I believe templates here would give safer solution.

mgonchar:
how you suggest to implement such thing without macro? Currently c++ doesn't have reflection. The original request came from @ogolikov, she wants to see all parameters listed in a way I did in current commit

_studio/shared/include/mfx_utils.h
```diff
@@ -131,4 +131,7 @@ template<class T> inline T AlignValue(T value, mfxU32 alignment)
 //#undef  FAILED
 //#define FAILED(hr)      (MFX_STS_TRACE(hr) < 0)
 
+#define mfxCopyField(Field)      buf_dst.Field = buf_src.Field
```
dmitry-gurulev: Macros should not mimic to ordinary function. Please re-name it.

mgonchar: @dmitry-gurulev your suggestion on name changing? caps lock?

PR #1017, opened 2018-12-14
Check surface frame info

maksenox:


_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
```diff
@@ -406,6 +368,46 @@ namespace MFX_VPX_Utility
         return true;
     }
 
+    bool CheckInfoMFX(const mfxInfoMFX &mfx, mfxU32 codecId, eMFXPlatform platform,  eMFXHWType hwtype)
+    {
+        if (codecId != mfx.CodecId)
```
mgonchar: you can do this with ```MFX_CHECK``` that will allow to see what check was failed in logs with enabled tracing

_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
```diff
@@ -406,6 +368,46 @@ namespace MFX_VPX_Utility
...
+            if (mfx.CodecLevel != MFX_LEVEL_UNKNOWN)
+                return false;
+        }
+
+        return CheckFrameInfo(mfx.FrameInfo, codecId, platform, hwtype);
+    }
+
+    bool CheckVideoParam(mfxVideoParam const*p_in, mfxU32 codecId, eMFXPlatform platform, eMFXHWType hwtype)
+    {
+        if (!p_in)
```
mgonchar: same

PR #1018, opened 2018-12-14
[smt] cpu usage grow fix

andreyor:
Issue: MDP-52379

### DIFF
--- a/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
+++ b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
@@ -387,7 +387,10 @@ void Launcher::DoTranscoding()
             if (!m_pThreadContextArray[i]->handle.valid())
                 continue;
 
-            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(1));
+            //Payslip interval to check the state of working threads:
+            //such interval is usually a realtime, i.e. for 30 fps this would be 33ms,
+            //66ms typically mean either 1/fps or 2/fps payslip checks.
+            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(66));
             if (waitSts == std::future_status::ready)
             {
                 // Invoke get() of the handle just to reset the valid state.


samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -387,7 +387,7 @@ void Launcher::DoTranscoding()
             if (!m_pThreadContextArray[i]->handle.valid())
                 continue;
 
-            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(1));
+            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(1000));
```
mgonchar: did you try another values?

fzhar: It's better reduce timeout value - 1 second is too large, it might cause slowdowns and issues in case of many multiple sessions (checks will be done too rare and if some session is closed abnormally, it might be detected too late)

andreyor: Yes, I tried. Starting from 10 ms I see stable low cpu usage. What do you think the optimal value ?

dvrogozh: Pretty often the goal is to have real time transcoding. For this reason typical implementation is to have some payslip interval at which something should happen, for example, scheduling of the next portion of the data. Such interval is usually a realtime, i.e. for 30 fps this would be 33ms. Thus, it would be nice to emulate this behavior and select timeout value accordingly. I would probably go with 66ms which would typically mean either 1/fps or 2/fps payslip checks.

fzhar: @dvrogozh , this timeout is NOT inside session, so it would not affect realtime operation at all (data processing is done in different thread). This check is for session completion only - normal or abnormal. So it will affect sample closing routine only. 1 second is too large because full check interval will be 1 sec * number of sessions. Some scenarios got up to 64 sessions, so to avoid delay more than 1 seconds while closing, I'd suggest 1s/64 = 15 ms or less.

andreyor: Ok, will set 15 ms.

dvrogozh: @fzhar: I know that timeout is not inside the session. And still the whole my explanation fully holds. Payslip interval by its idea don't have any relation with sessions actually. 15ms or less - is a no go. It is a number from the sky. @andreyor : please, program 66ms.

andreyor: done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -387,7 +387,7 @@ void Launcher::DoTranscoding()
             if (!m_pThreadContextArray[i]->handle.valid())
                 continue;
 
-            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(1));
+            auto waitSts = m_pThreadContextArray[i]->handle.wait_for(std::chrono::milliseconds(66));
```
dvrogozh: Add a comment above. Basically repeat my explanation about payslip interval to check the state of working threads.

andreyor: done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
fzhar: @dvrogozh : "For this reason typical implementation is to have some payslip interval at which something should happen, for example, scheduling of the next portion of the data.". Scheduling next portion of data happens in separate thread, so your calculations are actually sky-fallen. Anyway, I know that @dvrogozh will keep his opinion to the end, and for me it's no big difference between 15 and 66 - slowdown will be seen only in case of many session and only at the end of processing - so I'm fine with it.

dvrogozh: That's why I used word "emulation".

Issue #1014, opened 2018-12-14
Possible cmake missprint


adydychk:
I tried to build latest MediaSDK on CentOS 7.4 **without** OpenCL and was surprised with output:

```
...
No package 'libmfx' found
Intel OpenCL SDK was not found (optional). The following will not be built: rotate_opencl plugin. <- !!!
...
Enable:
ENABLE_OPENCL : ON <- !!!
ENABLE_X11_DRI3 : OFF
ENABLE_WAYLAND : OFF
ENABLE_ITT : OFF
ENABLE_TEXTLOG : OFF
ENABLE_STAT : OFF
...
Build:
BUILD_RUNTIME : ON
BUILD_DISPATCHER : ON
BUILD_SAMPLES : ON
BUILD_TESTS : OFF
BUILD_TOOLS : OFF
BUILD_KERNELS : OFF
...
```
I surprised that cmake wrote that "OpenCL not found" (which is fully correct) but in section "Enable" I see "ENABLE_OPENCL : ON". That is very strange.
It seems that cmake has a misprint...

@onabiull Can you take a look on it? :)

onabiull:
This option enabled by default. 
https://github.com/Intel-Media-SDK/MediaSDK/blob/f7145a6b7bbea73dc37bc0b4e7a825a1b6ccbe8d/CMakeLists.txt#L46
And thus generate error, as dependency missed.
You should explicitly turn if OFF. 
`cmake3 ../MediaSDK -DENABLE_OPENCL=OFF`

adydychk:
Oh, thanks, understood the logic but it is unclear if the user see only only output of cmake...   
Maybe it is better to change "enable" to "Enabled project configuration (by default or parameters)"

dvrogozh:
The plan is to use the rule for all the enable options: if option is explicitly enabled and required prerequisites are missing - error out. This behavior was not simply yet implemented for OCL.

PR #1022, opened 2018-12-17
[cmake] Fix OpenCL enable print

adydychk:
Fixes https://github.com/Intel-Media-SDK/MediaSDK/issues/1014


adydychk:
It is unclear from "enable" that these parameters not env specific but enabled by default or by parameter (so it is not final result).  
I tried to formulate it more clear if you have another suggestions - pls welcome :)

CMakeLists.txt
```diff
@@ -173,7 +173,7 @@ endif()
 message("  CMAKE_INSTALL_FULL_INCLUDEDIR           : ${CMAKE_INSTALL_FULL_INCLUDEDIR}")
 message("  CMAKE_INSTALL_FULL_LIBDIR               : ${CMAKE_INSTALL_FULL_LIBDIR}")
 message("  CMAKE_INSTALL_FULL_DATADIR              : ${CMAKE_INSTALL_FULL_DATADIR}")
-message("Enable:")
+message("Enabled project configuration (by default or parameters):")
```
dvrogozh: I personally don't understand what you mean under this phrase. To me it is misleading. The previous "Enable" is honest and straightforward: here is list of variables and here is their values. Just read this literally. This is not indication whether feature is really enabled or not, that's list of variables. The fact that OCL could be OFF while variable is ON - that's a bug. We need to fix it, not workaround it.

CMakeLists.txt
```diff
@@ -174,7 +174,9 @@ message("  CMAKE_INSTALL_FULL_INCLUDEDIR           : ${CMAKE_INSTALL_FULL_INCLUD
 message("  CMAKE_INSTALL_FULL_LIBDIR               : ${CMAKE_INSTALL_FULL_LIBDIR}")
 message("  CMAKE_INSTALL_FULL_DATADIR              : ${CMAKE_INSTALL_FULL_DATADIR}")
 message("Enable:")
-message("  ENABLE_OPENCL                           : ${ENABLE_OPENCL}")
+if(DEFINED OPENCL_FOUND)
+  message("  ENABLE_OPENCL                           : ${ENABLE_OPENCL}")
+endif()
```
adydychk: @dvrogozh 
I hope that it is better...

I tried to "play" with something as:
```
option( ENABLE_OPENCL "Build targets dependent on OpenCL?" FALSE )
                                                                                                           ^
```
But it is all time defined as TRUE... I don\`t know why possible I did it wrong...

adydychk: Or I can implement:
```
if(DEFINED OPENCL_FOUND)
  message("  ENABLE_OPENCL                           : ${ENABLE_OPENCL}")
else()
  message("  ENABLE_OPENCL                           : OFF")
endif()
```

Output:
```
Enable:
  ENABLE_OPENCL                           : OFF
  ENABLE_X11_DRI3                         : OFF
  ENABLE_WAYLAND                          : OFF
  ENABLE_ITT                              : OFF
  ENABLE_TEXTLOG                          : OFF
  ENABLE_STAT                             : OFF
Build:
...
```

adydychk: @onabiull Your opinion on this https://github.com/Intel-Media-SDK/MediaSDK/pull/1022#discussion_r245482645? Is it ready to merge? Or should I to re-implement as in https://github.com/Intel-Media-SDK/MediaSDK/pull/1022#discussion_r245483252 ?

PR #1032, opened 2018-12-18
[vp9d] Clear mfx.Shift for packed format

maksenox:
DecodeHeader and UpdateVideoParam functions
set Shift to one for planar formats
but didn't clear it for packed formats (like Y410).

Issue: vsmgwl-10697
Test: vp9d_*_format_change./*

### DIFF
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -574,9 +574,6 @@ mfxStatus VideoDECODEVP9_HW::DecodeHeader(VideoCORE* core, mfxBitstream* bs, mfx
     mfxStatus sts = MFX_VP9_Utility::DecodeHeader(core, bs, par);
     MFX_CHECK_STS(sts);
 
-    if (par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010)
-        par->mfx.FrameInfo.Shift = 1;
-
     return sts;
 }
 
@@ -747,16 +744,6 @@ mfxStatus VideoDECODEVP9_HW::GetVideoParam(mfxVideoParam *par)
     par->mfx.FrameInfo.AspectRatioW = m_vInitPar.mfx.FrameInfo.AspectRatioW;
 
     return MFX_ERR_NONE;
-}
-
-void VideoDECODEVP9_HW::UpdateVideoParam(mfxVideoParam *par, VP9DecoderFrame const & frameInfo)
-{
-    VM_ASSERT(par);
-
-    MFX_VP9_Utility::FillVideoParam(m_core, frameInfo, par);
-
-    if (par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010)
-        par->mfx.FrameInfo.Shift = 1;
 }
 
 mfxStatus VideoDECODEVP9_HW::GetDecodeStat(mfxDecodeStat *pStat)
@@ -1075,7 +1062,7 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     sts = DecodeFrameHeader(bs, frameInfo);
     MFX_CHECK_STS(sts);
 
-    UpdateVideoParam(&m_vPar, frameInfo);
+    MFX_VP9_Utility::FillVideoParam(m_core->GetPlatformType(), frameInfo, m_vPar);
 
     // check resize
     if (m_vPar.mfx.FrameInfo.Width > surface_work->Info.Width || m_vPar.mfx.FrameInfo.Height > surface_work->Info.Height)
--- a/_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode.h
+++ b/_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode.h
@@ -36,7 +36,7 @@
 namespace MFX_VP9_Utility
 {
     mfxStatus DecodeHeader(VideoCORE * /*core*/, mfxBitstream *bs, mfxVideoParam *params);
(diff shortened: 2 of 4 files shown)


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -522,7 +522,13 @@ mfxStatus VideoDECODEVP9_HW::DecodeHeader(VideoCORE* core, mfxBitstream* bs, mfx
 
     if (   par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010
         )
+    {
```
mgonchar: it could be simply:
```
par->mfx.FrameInfo.Shift = mfxU16(par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010);
```

DenWolf: In my opinion concrete values "0" and "1"are better for understanding, especially for people who see source code the first time, so I think we can keep this code in this style or we can change to 1-line style, for example:

par->mfx.FrameInfo.Shift = (par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010) ? 1 : 0;

mgonchar: I think oneliner is better in this situation. I'm OK with your suggestion to use operator ?:

dmitry-gurulev: What about y210?

maksenox: VP9 HW decoder doesn't support y210, only nv12, ayuv, p010, p016.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -573,10 +569,6 @@ void VideoDECODEVP9_HW::UpdateVideoParam(mfxVideoParam *par, VP9DecoderFrame con
     VM_ASSERT(par);
 
     MFX_VP9_Utility::FillVideoParam(m_core, frameInfo, par);
```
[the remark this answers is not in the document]
maksenox: Done. Added FillVideoParam(m_core, frameInfo, par) status check.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -738,11 +734,8 @@ void VideoDECODEVP9_HW::UpdateVideoParam(mfxVideoParam *par, VP9DecoderFrame con
 {
     VM_ASSERT(par);
 
-    MFX_VP9_Utility::FillVideoParam(m_core, frameInfo, par);
-
-    if (   par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010
-        )
-        par->mfx.FrameInfo.Shift = 1;
+    mfxStatus sts = MFX_VP9_Utility::FillVideoParam(m_core, frameInfo, par);
+    MFX_SUCCEEDED(sts);
```
mgonchar: it doesn't do anything if check fails. Correct way is:
```
MFX_CHECK_STS(sts);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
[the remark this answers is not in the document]
dmitry-gurulev: I'd rather remove status from `MFX_VP9_Utility::FillVideoParam` it looks quite useless. Function semantic is format MFX parameters from domain specific headers. The latter ones must be valid at time when `FillVideoParam` is invoked, and actually they are not checked. The only thing that is checked by the `FillVideoParam` is its arguments - `MFX_CHECK_NULL_PTR2(core, params)`. This may be resolved by
1. Using of reference instead of pointer, or since `FillVideoParam` is internal we may use assertion for `mfxVideoParam*`.
2.  Pointer to core is needed only for checking if platform is HW and set `Shift` properly. We may pass platform type as argument.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -749,14 +746,12 @@ mfxStatus VideoDECODEVP9_HW::GetVideoParam(mfxVideoParam *par)
     return MFX_ERR_NONE;
 }
 
-void VideoDECODEVP9_HW::UpdateVideoParam(mfxVideoParam *par, VP9DecoderFrame const & frameInfo)
+mfxStatus VideoDECODEVP9_HW::UpdateVideoParam(mfxVideoParam *par, VP9DecoderFrame const & frameInfo)
```
mgonchar: why can't you pass ```mfxVideoParam``` by reference? This will allow you to get rid of pointer check and error status return.

Or if you still prefer to return error status, you need to add status check after ```UpdateVideoParam``` call

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -749,14 +746,12 @@ mfxStatus VideoDECODEVP9_HW::GetVideoParam(mfxVideoParam *par)
...
 {
-    VM_ASSERT(par);
-
-    MFX_VP9_Utility::FillVideoParam(m_core, frameInfo, par);
-
-    if (par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010)
-        par->mfx.FrameInfo.Shift = 1;
+    MFX_CHECK_NULL_PTR1(par);
+    eMFXPlatform platform = m_core->GetPlatformType();
```
mgonchar: no need in this tmp variable

PR #1037, opened 2018-12-19
[VPP] colorfill disable option

IlyaSergeyev:
#605 was splitted into #891 and this one

### DIFF
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -2050,6 +2050,12 @@ mfxStatus VideoVPPHW::GetVideoParams(mfxVideoParam *par) const
             MFX_CHECK_NULL_PTR1(bufFRC);
             bufFRC->Algorithm = m_executeParams.frcModeOrig;
         }
+        else if (MFX_EXTBUFF_VPP_COLORFILL == bufferId)
+        {
+            mfxExtVPPColorFill *bufColorfill = reinterpret_cast<mfxExtVPPColorFill *>(par->ExtParam[i]);
+            MFX_CHECK_NULL_PTR1(bufColorfill);
+            bufColorfill->Enable = m_executeParams.iBackgroundColor?MFX_CODINGOPTION_ON:MFX_CODINGOPTION_OFF;
+        }
     }
 
     return MFX_ERR_NONE;
@@ -5130,6 +5136,12 @@ mfxU64 get_background_color(const mfxVideoParam & videoParam)
 {
     for (mfxU32 i = 0; i < videoParam.NumExtParam; i++)
     {
+        if (videoParam.ExtParam[i]->BufferId == MFX_EXTBUFF_VPP_COLORFILL)
+        {
+            mfxExtVPPColorFill *extCF = reinterpret_cast<mfxExtVPPColorFill *>(videoParam.ExtParam[i]);
+            if (!IsOn(extCF->Enable))
+                return 0;
+        }
         if (videoParam.ExtParam[i]->BufferId == MFX_EXTBUFF_VPP_COMPOSITE)
         {
             mfxExtVPPComposite *extComp = (mfxExtVPPComposite *) videoParam.ExtParam[i];
--- a/samples/sample_vpp/include/sample_vpp_config.h
+++ b/samples/sample_vpp/include/sample_vpp_config.h
@@ -125,6 +125,17 @@ typedef struct
 
 } sDenoiseParam;
 
+struct sColorFillParam: public mfxExtVPPColorFill
+{
+    FilterConfig mode= VPP_FILTER_DISABLED;
+    sColorFillParam():
+        mfxExtVPPColorFill()
+    {
+        Header.BufferId = MFX_EXTBUFF_VPP_COLORFILL;
+        Header.BufferSz = sizeof(mfxExtVPPColorFill);
+    }
+} ;
+
 #ifdef ENABLE_MCTF
 typedef struct
 {
--- a/samples/sample_vpp/src/sample_vpp_config.cpp
+++ b/samples/sample_vpp/src/sample_vpp_config.cpp
@@ -131,6 +131,13 @@ mfxStatus ConfigVideoEnhancementFilters( sInputParams* pParams, sAppResources* p
         pVppParam->ExtParam[pVppParam->NumExtParam++] = (mfxExtBuffer*)&(pResources->mirroringConfig);
     }
 
+    if( VPP_FILTER_ENABLED_CONFIGURED == pParams->colorfillParam[paramID].mode )
+    {
+        pResources->colorfillConfig = pParams->colorfillParam[paramID];
+        pVppParam->ExtParam[pVppParam->NumExtParam++] = &(pResources->colorfillConfig.Header);
+
+    }
+
     if( VPP_FILTER_ENABLED_CONFIGURED == pParams->procampParam[paramID].mode )
     {
         pResources->procampConfig.Header.BufferId = MFX_EXTBUFF_VPP_PROCAMP;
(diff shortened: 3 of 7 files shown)


samples/sample_vpp/include/sample_vpp_config.h
```diff
@@ -125,6 +125,15 @@ typedef struct
 
 } sDenoiseParam;
 
+struct sColorFillParam: public mfxExtVPPColorFill {
+    FilterConfig mode;
+    sColorFillParam():mfxExtVPPColorFill(), mode(VPP_FILTER_DISABLED)
+    {
+        Header.BufferId = MFX_EXTBUFF_VPP_COLORFILL;
+        Header.BufferSz = sizeof(mfxExtVPPColorFill);
+    };
```
lakulako: No need in semicolon after function body

samples/sample_vpp/src/sample_vpp_config.cpp
```diff
@@ -131,6 +131,13 @@ mfxStatus ConfigVideoEnhancementFilters( sInputParams* pParams, sAppResources* p
         pVppParam->ExtParam[pVppParam->NumExtParam++] = (mfxExtBuffer*)&(pResources->mirroringConfig);
     }
 
+    if( VPP_FILTER_ENABLED_CONFIGURED == pParams->colorfillParam[paramID].mode )
+    {
+        pResources->colorfillConfig = pParams->colorfillParam[paramID];
+        pVppParam->ExtParam[pVppParam->NumExtParam++] = (mfxExtBuffer*)&(pResources->colorfillConfig);
```
lakulako: Instead of (mfxExtBuffer*)&(pResources->colorfillConfig) use &(pResources->colorfillConfig.Header)

samples/sample_vpp/src/sample_vpp_parser.cpp
```diff
@@ -155,6 +155,8 @@ msdk_printf(MSDK_STRING("                                  GlobalAlpha=128\n"));
 msdk_printf(MSDK_STRING("                                  LumaKeyEnable=1\n"));
 msdk_printf(MSDK_STRING("                                  LumaKeyMin=250\n"));
 msdk_printf(MSDK_STRING("                                  LumaKeyMax=255\n"));
+msdk_printf(MSDK_STRING("   [-cf_disable]                - disable colorfill\n"));
```
lakulako: remove '-' before "disable colorfill"

samples/sample_vpp/include/sample_vpp_config.h
IlyaSergeyev: It's not my idea, it's a style of whole file.

samples/sample_vpp/src/sample_vpp_config.cpp
IlyaSergeyev: same

fzhar: AFAIR, ExtParam should keep pointer to the external buffer structure, so it's more correct to use (mfxExtBuffer*)&(pResources->colorfillConfig) However, there's no difference as long as Header is the first element in buffer (and in memory)

samples/sample_vpp/include/sample_vpp_config.h
lakulako: No, but do it in the code you modify.

samples/sample_vpp/src/sample_vpp_config.cpp
lakulako: Header type is exactly mfxExtBuffer, it is similar to base class. It was designed to be used this way. Casting can hide possible errors.
Again, fix it only in modified code.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2050,6 +2050,11 @@ mfxStatus VideoVPPHW::GetVideoParams(mfxVideoParam *par) const
             MFX_CHECK_NULL_PTR1(bufFRC);
             bufFRC->Algorithm = m_executeParams.frcModeOrig;
         }
+        else if (MFX_EXTBUFF_VPP_COLORFILL == bufferId)
+        {
+            mfxExtVPPColorFill *bufColorfill = reinterpret_cast<mfxExtVPPColorFill *>(par->ExtParam[i]);
+            MFX_CHECK_NULL_PTR1(bufColorfill);
```
dvrogozh: I think you are supposed to assign currently configured values for the color fill. Idea is that user calls GetVideoParams and sees how VPP is configured. Please, add assignment.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5130,6 +5135,13 @@ mfxU64 get_background_color(const mfxVideoParam & videoParam)
 {
     for (mfxU32 i = 0; i < videoParam.NumExtParam; i++)
     {
+        if (videoParam.ExtParam[i]->BufferId == MFX_EXTBUFF_VPP_COLORFILL)
+        {
+            mfxExtVPPColorFill *extCF = reinterpret_cast<mfxExtVPPColorFill *>(videoParam.ExtParam[i]);
+            if (extCF->Enable != MFX_CODINGOPTION_ON) {
```
mgonchar: there id function ```IsOn```

samples/sample_vpp/include/sample_vpp_config.h
```diff
@@ -125,6 +125,15 @@ typedef struct
 
 } sDenoiseParam;
 
+struct sColorFillParam: public mfxExtVPPColorFill {
```
mgonchar: as for me, it is hard to read such code, several newlines required. My suggestion:
```
struct sColorFillParam: public mfxExtVPPColorFill
{
     FilterConfig mode = VPP_FILTER_DISABLED;

     sColorFillParam()
     : mfxExtVPPColorFill()
     {
         Header.BufferId = MFX_EXTBUFF_VPP_COLORFILL;
         Header.BufferSz = sizeof(mfxExtVPPColorFill);
     }
};
```

samples/sample_vpp/src/sample_vpp_config.cpp
mgonchar: agree with @lakulako. Try to avoid excessive casts

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
mgonchar: +1

samples/sample_vpp/include/sample_vpp_config.h
IlyaSergeyev: done

samples/sample_vpp/src/sample_vpp_config.cpp
IlyaSergeyev: done

samples/sample_vpp/src/sample_vpp_parser.cpp
IlyaSergeyev: done

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
IlyaSergeyev: We have 3 copies of IsOn() function. I added another one for vpp. I thing this function shuld be moved into api/include/mfxstructures.h file (where MFX_CODINGOPTION_* definitions are defined)

samples/sample_vpp/include/sample_vpp_config.h
IlyaSergeyev: It's ok, but it will look differently from other similar classes.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
IlyaSergeyev: done

PR #1040, opened 2018-12-19
Remove vm mutex 

akokoshn:


_studio/mfx_lib/shared/include/mfx_frames.h
```diff
@@ -77,31 +77,28 @@
     class clExtTasks1
     {
     public:
-        clExtTasks1()
+        clExtTasks1():
+            m_mGuard()
```
dvrogozh: emmm. why? this is useless and should be done automatically by compiler.

_studio/mfx_lib/shared/include/mfx_frames.h
```diff
@@ -167,32 +160,29 @@
     class clExtTasks2
     {
     public:
-        clExtTasks2()
+        clExtTasks2():
+            m_mGuard()
```
dvrogozh: same

_studio/mfx_lib/shared/include/mfx_frames.h
mgonchar: agree

mgonchar:
@dvrogozh could you look again?

PR #1050, opened 2018-12-20
rebuild kernels with fresh compiler

artem-shaporenko:
fix ICL kernels, decrease size of other kernels


dmitryermilov:
@artem-shaporenko , did you run any sanity tests to check that LA works fine with updated kernels?

artem-shaporenko:
yes, verified

PR #1051, opened 2018-12-20
Rebuild kernels with new compiler

artem-shaporenko:
Fix ICL kernels, improve previous platforms


dmitryermilov:
@artem-shaporenko , did you run any sanity tests to check that LA works fine with updated kernels?

artem-shaporenko:
yes, verified

PR #1054, opened 2018-12-20
Remove umc array

akokoshn:


_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
```diff
@@ -1212,11 +1212,13 @@ void MVC_Extension::ChooseLevelIdc(const H264SeqParamSetMVCExtension * extension
 
     m_level_idc = 0;
 
-    for (uint32_t i = 0; i <= extension->num_level_values_signalled_minus1; i++)
+    assert(extension->num_level_values_signalled_minus1 + 1 == extension->levelInfo.size());
+    for (uint32_t i = 0; i < extension->levelInfo.size(); i++)
```
mgonchar: ok, I think all this scary construction is a good example of how stl algorithms can simplify things:
```
    for (const auto& levelInfo : extension->levelInfo)
    {
        assert((levelInfo.num_applicable_ops_minus1 + 1) == levelInfo.opsInfo.size());
        
        for (const auto& operationPoint : levelInfo.opsInfo)
        {            
            if (std::find_first_of(m_viewIDsList.begin(), m_viewIDsList.end(),
                                   operationPoint.applicable_op_target_view_id.begin(), operationPoint.applicable_op_target_view_id.end())
                        == m_viewIDsList.end())
                break;

            if (!m_level_idc || m_level_idc > levelInfo.level_idc)
            {
                m_level_idc = levelInfo.level_idc; // <- looks weird, we operate with levelInfo on upper level, we have assignment of loop invariant here?
            }    
        }
    }
```

Could you also find an answer on my question?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -507,7 +504,7 @@ void mfxSchedulerCore::RegisterTaskDependencies(MFX_SCHEDULER_TASK  *pTask)
         if (pTask->param.task.pDst[i])
         {
             // find empty table entry
-            while (m_pDependencyTable[tableIdx].p)
+            while ((MFX_MAX_NUMBER_TASK * 2 - 1) > tableIdx && m_pDependencyTable[tableIdx].p)
```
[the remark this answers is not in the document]
akokoshn: MFX_MAX_NUMBER_TASK * 2 - size of m_pDependencyTable

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -431,9 +427,10 @@ void mfxSchedulerCore::RegisterTaskDependencies(MFX_SCHEDULER_TASK  *pTask)
 
     // check if the table have empty position(s),
     // If so decrement the index of the last table entry.
-    i = m_numDependencies;
+    assert(MFX_MAX_NUMBER_TASK * 2 > m_numDependencies);
+    i = (MFX_MAX_NUMBER_TASK * 2 > m_numDependencies) ? m_numDependencies : 0;
     while ((0 < i) &&
-           (NULL == m_pDependencyTable[i - 1].p))
+           (nullptr == m_pDependencyTable[i - 1].p))
```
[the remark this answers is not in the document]
mgonchar: Sorry, I made a mistake in comment and provided incorrect bounds for reverse iterators. Thanks to @lakulako for pointing me on it

So correct code would be:
```
i = 0;
if (MFX_MAX_NUMBER_TASK * 2 /* .size() I believe */ > m_numDependencies)
{
    auto it = std::find_if(std::rend(m_pDependencyTable) - m_numDependencies, std::rend(m_pDependencyTable),
    [](const MFX_DEPENDENCY_ITEM & item){ return item.p != nullptr; });

    i = std::rend(m_pDependencyTable) - it;
}
```

_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
```diff
@@ -395,12 +394,12 @@ class mfxSchedulerCore : public MFXIScheduler2
     // Dependency table.
     // There are an index: it holds the number of values valid in the array.
     // The array may have larger size.
-    UMC::Array<MFX_DEPENDENCY_ITEM> m_pDependencyTable;
+    std::vector<MFX_DEPENDENCY_ITEM> m_pDependencyTable;
     volatile
     mfxU32 m_numDependencies;
```
lakulako: Why not to use m_pDependencyTable.size() instead of  "volatile mfxU32 m_numDependencies"

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -68,27 +68,15 @@ mfxStatus mfxSchedulerCore::Initialize2(const MFX_SCHEDULER_PARAM2 *pParam)
...
-    }
+    m_ppTaskLookUpTable.resize(MFX_MAX_NUMBER_TASK, nullptr);
 
     // allocate the dependency table
-    m_pDependencyTable.Alloc(MFX_MAX_NUMBER_TASK * 2);
-    if (UMC::UMC_OK != umcRes)
-    {
-        return MFX_ERR_MEMORY_ALLOC;
-    }
+    m_pDependencyTable.resize(MFX_MAX_NUMBER_TASK * 2, MFX_DEPENDENCY_ITEM());
```
lakulako: As far as I see it is the only place where m_pDependencyTable can be changed. If so then no need to check in many places for (smth < MFX_MAX_NUMBER_TASK * 2). If not so, then check only where m_pDependencyTable.size() grows.

_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
akokoshn: find_first_of returns !end() one  element from m_viewIDsList contains in applicable_op_target_view_id,
but old code check that **all** elements of m_viewIDsList  contain in applicable_op_target_view_id

_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
akokoshn: work with m_pDependencyTable designed as for const size array.
I guess we shouldn't rework current behavior in this PR

_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
mgonchar: yep, you are right... does the order of elements matters? can we sort them and just use ```std::includes``` ? The other variant is to construct two ```std::set``` and check that one is a subset of another (but I don't like this approach, because of overhead on sets construction)

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -431,13 +427,13 @@ void mfxSchedulerCore::RegisterTaskDependencies(MFX_SCHEDULER_TASK  *pTask)
...
-    i = m_numDependencies;
-    while ((0 < i) &&
-           (NULL == m_pDependencyTable[i - 1].p))
-    {
-        i -= 1;
-    }
-    m_numDependencies = i;
+    assert(m_pDependencyTable.size() > m_numDependencies);
+
+    i = (m_pDependencyTable.size() > m_numDependencies) ? m_numDependencies : 0;
```
mgonchar: you writing an assert and then use the same condition as the correct one in following code... looks suspicious

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -431,13 +427,13 @@ void mfxSchedulerCore::RegisterTaskDependencies(MFX_SCHEDULER_TASK  *pTask)
...
+    auto it = std::find_if(m_pDependencyTable.rend() - i, m_pDependencyTable.rend(),
```
mgonchar: I suggest to put this code under if

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -328,7 +318,8 @@ mfxStatus mfxSchedulerCore::WaitForDependencyResolved(const void *pDependency)
 
         for (curIdx = 0; curIdx < m_numDependencies; curIdx += 1)
         {
-            if (m_pDependencyTable[curIdx].p == pDependency)
+            assert(m_pDependencyTable.size() > curIdx);
```
mgonchar: I think you introduced too many excessive checks. You may use ```std::vector::at``` if you really fear of out-of-bounds access

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
akokoshn: assert - only for debug.
what is wrong?
assert(true)// OK, so i = m_numDependencies
I have missed some thing?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -150,13 +149,10 @@ void mfxSchedulerCore::Close(void)
     }
 
     // delete task objects
-    for (n = 0; n < m_ppTaskLookUpTable.Size(); n += 1)
+    for (auto &it : m_ppTaskLookUpTable)
```
mgonchar: no need to use refs here, it just brings unnecessary level of indirection, just ```auto it```

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -328,11 +317,11 @@ mfxStatus mfxSchedulerCore::WaitForDependencyResolved(const void *pDependency)
 
         for (curIdx = 0; curIdx < m_numDependencies; curIdx += 1)
         {
-            if (m_pDependencyTable[curIdx].p == pDependency)
+            if (m_pDependencyTable.at(curIdx).p == pDependency)
             {
                 // get the handle before leaving protected section
-                waitHandle.taskID = m_pDependencyTable[curIdx].pTask->taskID;
-                waitHandle.jobID = m_pDependencyTable[curIdx].pTask->jobID;
+                waitHandle.taskID = m_pDependencyTable.at(curIdx).pTask->taskID;
```
mgonchar: You've already checked that this item exists at line 320

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_mfx.cpp
```diff
@@ -47,25 +47,25 @@ Status FillVideoParamExtension(const UMC_H264_DECODER::H264SeqParamSetMVCExtensi
     size_t numberOfPoints = 0;
     size_t numberOfTargets = 0;
 
-    for (size_t i = 0; i < seqEx->levelInfo.Size(); i++)
+    for (size_t i = 0; i < seqEx->levelInfo.size(); i++)
```
mgonchar: use range-based fo here

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_mfx.cpp
```diff
@@ -47,25 +47,25 @@ Status FillVideoParamExtension(const UMC_H264_DECODER::H264SeqParamSetMVCExtensi
...
         const UMC_H264_DECODER::H264LevelValueSignaled & levelSignal = seqEx->levelInfo[i];
 
-        if ((uint32_t)levelSignal.num_applicable_ops_minus1 + 1 != levelSignal.opsInfo.Size())
+        if ((uint32_t)levelSignal.num_applicable_ops_minus1 + 1 != levelSignal.opsInfo.size())
         {
             return UMC_ERR_FAILED;
         }
 
-        for (size_t j = 0; j < levelSignal.opsInfo.Size(); j++)
+        for (size_t j = 0; j < levelSignal.opsInfo.size(); j++)
```
mgonchar: and here

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_mfx.cpp
```diff
@@ -78,7 +78,7 @@ Status FillVideoParamExtension(const UMC_H264_DECODER::H264SeqParamSetMVCExtensi
         return UMC_ERR_NOT_ENOUGH_BUFFER;
     }
 
-    for (size_t i = 0; i < seqEx->viewInfo.Size(); i++)
+    for (size_t i = 0; i < seqEx->viewInfo.size(); i++)
```
mgonchar: here and below as well

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
mgonchar: I mean that your code handles this condition (when assert is false) without any problems... So if the number of dependencies in array is incorrect, isn't it a fatal error?

_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
lakulako: Why volatile, if it is size of const size array? I'd rather see
constexpr size_t m_numDependencies = MFX_MAX_NUMBER_TASK * 2;

akokoshn: m_numDependencies - is actual number of dependency (<= MFX_MAX_NUMBER_TASK * 2)

lakulako: Ok, you are right, it is possible to leave it for future.
Anyway, while array size is m_pDependencyTable.capacity() equal to MFX_MAX_NUMBER_TASK * 2, using m_pDependencyTable.size() instead of m_numDependencies would be more clear.
And it is related to the change. Better do it now, I think.

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_mfx.cpp
```diff
@@ -101,33 +98,31 @@ Status FillVideoParamExtension(const UMC_H264_DECODER::H264SeqParamSetMVCExtensi
...
-
             operationPoints->TemporalId = ops.applicable_op_temporal_id;
             operationPoints->LevelIdc = levelSignal.level_idc;
             operationPoints->NumViews = ops.applicable_op_num_views_minus1 + 1;
             operationPoints->NumTargetViews = ops.applicable_op_num_target_views_minus1 + 1;
             operationPoints->TargetViewId = targetViews;
 
-            for (size_t k = 0; k < ops.applicable_op_target_view_id.Size(); k++)
+            size_t k = 0;
+            for (auto targetView : ops.applicable_op_target_view_id)
```
mgonchar: use ```std::copy``` instead of this cycle

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_slice_decoder_decode_pic.cpp
```diff
@@ -591,25 +590,24 @@ void H264Slice::ReOrderRefPicList(H264DecoderFrame **pRefPicList,
...
+            const auto currView = std::find_if(m_pSeqParamSetMvcEx->viewInfo.begin(), m_pSeqParamSetMvcEx->viewInfo.end(),
+                                         [this](const UMC_H264_DECODER::H264ViewRefInfo &item){ return item.view_id == m_SliceHeader.nal_ext.mvc.view_id; });
+            if (currView == m_pSeqParamSetMvcEx->viewInfo.end())
             {
-                if (m_SliceHeader.nal_ext.mvc.view_id == m_pSeqParamSetMvcEx->viewInfo[viewIdx].view_id)
-                {
-                    currVOIdx = viewIdx;
-                    break;
-                }
+                assert(false);
```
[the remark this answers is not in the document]
mgonchar: +1

_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
```diff
@@ -1186,13 +1183,13 @@ void MVC_Extension::ChooseLevelIdc(const H264SeqParamSetMVCExtension * extension
...
         return;
 
-    VM_ASSERT(extension->viewInfo.Size() == extension->num_views_minus1 + 1);
+    assert(extension->viewInfo.size() == extension->num_views_minus1 + 1);
 
-    if (!m_viewIDsList.size())
+    if (m_viewIDsList.empty())
     {
-        for (size_t i = 0; i <= extension->num_views_minus1; i++)
+        for (const auto & viewInfo : extension->viewInfo)
```
mgonchar: before the for loop add small optimization:
```
m_viewIDsList.reserve(extension->viewInfo.size());
```

akokoshn: std::list<T> hasn't member reserve

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -431,13 +427,13 @@ void mfxSchedulerCore::RegisterTaskDependencies(MFX_SCHEDULER_TASK  *pTask)
 
     // check if the table have empty position(s),
     // If so decrement the index of the last table entry.
```
mgonchar: second comment is outdated

akokoshn:
WIP (need functional validation before merge)

mgonchar:
@akokoshn any progress with testing this PR?

akokoshn:
@mgonchar , still wait for validation

_studio/mfx_lib/scheduler/src/mfx_scheduler_core.cpp
```diff
@@ -150,13 +149,10 @@ void mfxSchedulerCore::Close(void)
...
-    for (n = 0; n < m_ppTaskLookUpTable.Size(); n += 1)
+    for (auto it : m_ppTaskLookUpTable)
     {
-        if (m_ppTaskLookUpTable[n])
-        {
-            delete m_ppTaskLookUpTable[n];
-            m_ppTaskLookUpTable[n] = NULL;
-        }
+        delete it;
+        it = nullptr;
```
onabiull: Shouldn't it be reference to make this line feasible?

dvrogozh: This will not affect the array, you need `auto& it` instead.

mgonchar: I suggest to shrink vector to 0 after deletions instead of references usage.
Anyway, you need to do something with this code

mgonchar: Following code is better from performance perspective:
for (auto ptr : m_ppTaskLookUpTable)
{
    delete ptr;
}
m_ppTaskLookUpTable.clear();

mgonchar:
@dvrogozh could you please look at this PR again?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -598,6 +585,10 @@ mfxStatus mfxSchedulerCore::AddTask(const MFX_TASK &task, mfxSyncPoint *pSyncPoi
         {
             return mfxRes;
         }
+        if (m_occupancyTable.size() <= occupancyIdx)
```
mgonchar: Why do use this approach instead of ```.at()``` as you did previously?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_task_management.cpp
```diff
@@ -423,18 +422,19 @@ void mfxSchedulerCore::MarkTaskCompleted(const MFX_CALL_INFO *pCallInfo,
     (void)pCallInfo;
     (void)threadNum;
 
-    bool taskReleased = false;
-    mfxU32 nTraceTaskId = 0;
-
-    MFX_SCHEDULER_TASK *pTask = m_ppTaskLookUpTable[pCallInfo->taskHandle.taskID];
-    mfxU32 curTime;
+    MFX_SCHEDULER_TASK *pTask = nullptr;
```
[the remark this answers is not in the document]
mgonchar: make this a oneliner

PR #1056, opened 2018-12-21
[samples] vaapi allocator surfaces attribs fixed

IlyaSergeyev:
As it was done previously in lib version of vaapi allocator.

### DIFF
--- a/samples/sample_common/src/vaapi_allocator.cpp
+++ b/samples/sample_common/src/vaapi_allocator.cpp
@@ -154,7 +154,6 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
     VAStatus  va_res  = VA_STATUS_SUCCESS;
     unsigned int va_fourcc = 0;
     VASurfaceID* surfaces = NULL;
-    VASurfaceAttrib attrib;
     vaapiMemId *vaapi_mids = NULL, *vaapi_mid = NULL;
     mfxMemId* mids = NULL;
     mfxU32 fourcc = request->Info.FourCC;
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -206,23 +205,27 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
         if( VA_FOURCC_P208 != va_fourcc )
         {
             unsigned int format;
+            VASurfaceAttrib attrib[2];
```
mgonchar: better to make it ```std::vector```

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -206,23 +205,27 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
...
+            attrib[attrCnt].value.type    = VAGenericValueTypeInteger;
+            attrib[attrCnt++].value.value.i = va_fourcc;
             format               = va_fourcc;
 
             if (fourcc == MFX_FOURCC_VP8_NV12)
             {
                 // special configuration for NV12 surf allocation for VP8 hybrid encoder is required
-                attrib.type          = (VASurfaceAttribType)VASurfaceAttribUsageHint;
-                attrib.value.value.i = VA_SURFACE_ATTRIB_USAGE_HINT_ENCODER;
+                attrib[attrCnt].type          = (VASurfaceAttribType)VASurfaceAttribUsageHint;
```
mgonchar: I believe explicit cast to ```VASurfaceAttribType``` is excessive

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -247,7 +250,7 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
                                     request->Info.Width, request->Info.Height,
                                     surfaces,
                                     surfaces_num,
-                                    &attrib, 1);
+                                    &attrib[0], attrCnt);
```
mgonchar: here please use ```vector::data()``` and ```vector::size()```

samples/sample_common/src/vaapi_allocator.cpp
IlyaSergeyev: why? std::vector is too complicated for this purpose ... VASurfaceAttrib is a small-size struct

samples/sample_common/src/vaapi_allocator.cpp
alariono: Agree with @IlyaSergeyev ; lets keep "attrib" as array

samples/sample_common/src/vaapi_allocator.cpp
mgonchar: because you are using explicit indexing and it has potential issue of out of bounds indexing if this code will be changed (for example one more attrib will be added without changing array size).  Using vector + reserve + push_back much more error prone and overhead of heap allocation is negligible

PR #1057, opened 2018-12-21
JPEGe/JPEGGd code cleanup

Alexandr-Konovalov:
1) Explicitly delete unused ctors/assignments.
2) Remove unused CJPEGEncoder::SelectScanScripts(). Current implementation
of CJPEGEncoder::SelectScanScripts() is broken, and the function is not
called currently.
3) Check for NULL before dereference of pointers.

Issue: MDP-52543


Alexandr-Konovalov:
I believe move ctors/move assignments are implicitly defined only when there are no user-declared destructors.

mgonchar:
Not only in that cases, if you declare assignment/copy ctor they are also not declared (see https://stackoverflow.com/questions/44430905/rule-of-3-default-member-deprecation-in-c11). I'm only saying that better to follow rule of five and define all five or none of them

Alexandr-Konovalov:
@mgonchar ok, let's do it this way.

PR #1058, opened 2018-12-21
Add checking for arguments of IPP functions against NULL

Alexandr-Konovalov:
Issue: MDP-52543


Alexandr-Konovalov:
Thanks, fixed.

PR #1062, opened 2018-12-25
hevce: enable control for max slice size

dmenshov:
enable control for max slice size

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -445,6 +445,40 @@ mfxStatus SetFrameRate(
     return MFX_ERR_NONE;
 }
 
+mfxStatus SetMaxSliceSize(
+    MfxVideoParam const & par,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & maxSliceSizeBuf_id)
+{
+    MFX_DESTROY_VABUFFER(maxSliceSizeBuf_id, vaDisplay);
+
+    VAStatus vaSts = vaCreateBuffer(vaDisplay,
+                   vaContextEncode,
+                   VAEncMiscParameterBufferType,
+                   sizeof(VAEncMiscParameterBuffer) + sizeof(VAEncMiscParameterMaxSliceSize),
+                   1,
+                   nullptr,
+                   &maxSliceSizeBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    VAEncMiscParameterBuffer *misc_param;
+    vaSts = vaMapBuffer(vaDisplay,
+                 maxSliceSizeBuf_id,
+                (void **)&misc_param);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    misc_param->type = VAEncMiscParameterTypeMaxSliceSize;
+    auto maxSliceSize_param = reinterpret_cast<VAEncMiscParameterMaxSliceSize *>(misc_param->data);
+
+    maxSliceSize_param->max_slice_size = par.m_ext.CO2.MaxSliceSize;
+
+    vaSts = vaUnmapBuffer(vaDisplay, maxSliceSizeBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    return MFX_ERR_NONE;
+}
+
 
 mfxStatus SetQualityLevelParams(
     MfxVideoParam const & par,
@@ -896,6 +930,7 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
         m_caps.MaxPicHeight = 1088;
 
     m_caps.SliceStructure = 4;
+    m_caps.SliceByteSizeCtrl = 1; //It means that GPU may further split the slice region that slice control data specifies into finer slice segments based on slice size upper limit (MaxSliceSize).
 
     if (attrs[ idx_map[VAConfigAttribEncMaxRefFrames] ].value != VA_ATTRIB_NOT_SUPPORTED)
     {
@@ -1053,6 +1088,11 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_FrameRate, 1)), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevelParams(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_QualityLevel, 1)), MFX_ERR_DEVICE_FAILED);
+    if(par.m_ext.CO2.MaxSliceSize != 0)
+    {
+        mfxStatus sts = SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1));
+        MFX_CHECK_WITH_ASSERT(sts == MFX_ERR_NONE, MFX_ERR_DEVICE_FAILED);
+    }
     FillConstPartOfPps(par, m_pps);
     FillSliceBuffer(par, m_sps, m_pps, m_slice);
 
@@ -1077,7 +1117,11 @@ mfxStatus VAAPIEncoder::Reset(MfxVideoParam const & par, bool bResetBRC)
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_FrameRate, 1)), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevelParams(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_QualityLevel, 1)), MFX_ERR_DEVICE_FAILED);
-
+    if(par.m_ext.CO2.MaxSliceSize != 0)
+    {
+        mfxStatus sts = SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1));
+        MFX_CHECK_WITH_ASSERT(sts == MFX_ERR_NONE, MFX_ERR_DEVICE_FAILED);
+    }
 
     FillConstPartOfPps(par, m_pps);
     FillSliceBuffer(par, m_sps, m_pps, m_slice);
--- a/_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
+++ b/_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
@@ -83,6 +83,7 @@ public:
 
         , VABID_ROI
         , VABID_RIR
+        , VABID_MaxSliceSize
 
         , VABID_END_OF_LIST // Remain this item last in the list
     };
@@ -144,6 +145,12 @@ mfxStatus SetFrameRate(
     VADisplay    vaDisplay,
     VAContextID  vaContextEncode,
     VABufferID & frameRateBuf_id);
+
+mfxStatus SetMaxSliceSize(
+    MfxHwH265Encode::MfxVideoParam const & par,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & maxSliceSizeBuf_id);
 
 mfxStatus SetSkipFrame(
     VADisplay    vaDisplay,


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -445,6 +445,43 @@ mfxStatus SetFrameRate(
...
+    MfxVideoParam const & par,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & maxSliceSizeBuf_id)
+{
+    VAStatus vaSts;
+    VAEncMiscParameterBuffer *misc_param;
+    VAEncMiscParameterMaxSliceSize *maxSliceSize_param;
+
+    MFX_DESTROY_VABUFFER(maxSliceSizeBuf_id, vaDisplay);
```
mgonchar: do not use this macro, it doesn't check returned status from ```vaDestroyBuffer```
You can use following function for example:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h#L53

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -445,6 +445,43 @@ mfxStatus SetFrameRate(
...
+                   &maxSliceSizeBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    vaSts = vaMapBuffer(vaDisplay,
+                 maxSliceSizeBuf_id,
+                (void **)&misc_param);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    misc_param->type = VAEncMiscParameterTypeMaxSliceSize;
+    maxSliceSize_param = (VAEncMiscParameterMaxSliceSize *)misc_param->data;
```
mgonchar: declare ```maxSliceSize_param``` right here:
```
auto maxSliceSize_param = reinterpret_cast<VAEncMiscParameterMaxSliceSize *>(misc_param->data);
```

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -445,6 +445,43 @@ mfxStatus SetFrameRate(
...
+    vaSts = vaCreateBuffer(vaDisplay,
+                   vaContextEncode,
+                   VAEncMiscParameterBufferType,
+                   sizeof(VAEncMiscParameterBuffer) + sizeof(VAEncMiscParameterMaxSliceSize),
+                   1,
+                   NULL,
+                   &maxSliceSizeBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    vaSts = vaMapBuffer(vaDisplay,
```
mgonchar: declare ```misc_param``` right before usage

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -445,6 +445,43 @@ mfxStatus SetFrameRate(
...
+    VAEncMiscParameterMaxSliceSize *maxSliceSize_param;
+
+    MFX_DESTROY_VABUFFER(maxSliceSizeBuf_id, vaDisplay);
+
+    vaSts = vaCreateBuffer(vaDisplay,
+                   vaContextEncode,
+                   VAEncMiscParameterBufferType,
+                   sizeof(VAEncMiscParameterBuffer) + sizeof(VAEncMiscParameterMaxSliceSize),
+                   1,
+                   NULL,
```
mgonchar: nullptr

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1053,6 +1091,10 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_FrameRate, 1)), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevelParams(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_QualityLevel, 1)), MFX_ERR_DEVICE_FAILED);
+    if(par.m_ext.CO2.MaxSliceSize != 0)
+    {
+        MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1)), MFX_ERR_DEVICE_FAILED);
```
mgonchar: I suggest to pull this function out of macro. In debug build expansion of this macro leads to double call of function. It would be nice if you will fix this issue for other similar call which I see in near by code

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1077,7 +1119,10 @@ mfxStatus VAAPIEncoder::Reset(MfxVideoParam const & par, bool bResetBRC)
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_FrameRate, 1)), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevelParams(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_QualityLevel, 1)), MFX_ERR_DEVICE_FAILED);
-
+    if(par.m_ext.CO2.MaxSliceSize != 0)
+    {
+        MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1)), MFX_ERR_DEVICE_FAILED);
```
mgonchar: same

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -445,6 +458,40 @@ mfxStatus SetFrameRate(
     return MFX_ERR_NONE;
 }
 
+mfxStatus SetMaxSliceSize(
+    MfxVideoParam const & par,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & maxSliceSizeBuf_id)
+{
+    mfxStatus mfxRes = CheckAndDestroyVAbuffer(vaDisplay, maxSliceSizeBuf_id);
```
mgonchar: you have to check returned statuses

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1053,6 +1101,10 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_FrameRate, 1)), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevelParams(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_QualityLevel, 1)), MFX_ERR_DEVICE_FAILED);
+    if(par.m_ext.CO2.MaxSliceSize != 0)
+    {
+        SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1));
```
mgonchar: no, that is not what I meant in previous review. You need to check the status, but don't put the function call inside the macro:
```
mfxStatus sts = SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1));
MFX_CHECK_WITH_ASSERT(sts == MFX_ERR_NONE);
```

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -35,6 +35,19 @@
 #define MFX_CHECK_WITH_ASSERT(EXPR, ERR) { assert(EXPR); MFX_CHECK(EXPR, ERR); }
 //#define PARALLEL_BRC_support
 
+inline mfxStatus CheckAndDestroyVAbuffer(VADisplay display, VABufferID & buffer_id)
+{
```
lakulako: What the reason to introduce one more kind of MFX_DESTROY_VABUFFER?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -445,6 +458,41 @@ mfxStatus SetFrameRate(
...
 }
 
+mfxStatus SetMaxSliceSize(
+    MfxVideoParam const & par,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & maxSliceSizeBuf_id)
+{
+    mfxStatus sts = CheckAndDestroyVAbuffer(vaDisplay, maxSliceSizeBuf_id);
+    MFX_CHECK_WITH_ASSERT(sts == MFX_ERR_NONE, MFX_ERR_DEVICE_FAILED);
```
lakulako: Direct call to vaDestroyBuffer would be more organic here.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1053,6 +1102,11 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
 
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_FrameRate, 1)), MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevelParams(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_QualityLevel, 1)), MFX_ERR_DEVICE_FAILED);
+    if(par.m_ext.CO2.MaxSliceSize != 0)
+    {
+        mfxStatus sts = SetMaxSliceSize(par, m_vaDisplay, m_vaContextEncode, VABufferNew(VABID_MaxSliceSize, 1));
```
lakulako: Does it make sense to move VABufferNew(VABID_MaxSliceSize, 1) into SetMaxSliceSize()?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
dmenshov: Because macro MFX_DESTROY_VABUFFER doesn't check returned result.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
mgonchar: I would say yes. It makes sense

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
DenWolf: Agree with @lakulako . I don't see any reason of using function CheckAndDestroyVAbuffer, especially in this place.

Function CheckAndDestroyVAbuffer only call vaDestroyBuffer and check VA return status - if it not success -> return immediately MFX_ERR_DEVICE_FAILED with assert, that's all.
In current place right after calling CheckAndDestroyVAbuffer you check again returning status and return immediately MFX_ERR_DEVICE_FAILED with assert. Why ? I don't see any benefits of this.

As result, we have a pretty good function vaDestroyBuffer and I don't understand why we need to add new level of calling stack with wrapper function which do nothing and double check of the same status. Also, I don't see any other places where this function is used. Or I missed something ?

mgonchar: This function is used and was first introduced in AVC code.

Ok, to get rid of this function and use ```vaDestroy``` we need to make sure that this function is resistant across invalid input, since ```CheckAndDestroyVAbuffer``` check buffer id as well.
Looking here 
https://github.com/intel/libva/blob/e4cd982197a4f0131a6a5a43dd83f69f0a724db7/va/va.c#L1393 
and here
 https://github.com/intel/media-driver/blob/8e41388b5d17c80b126cc7f36505b090f3b1942d/media_driver/linux/common/ddi/media_libva.cpp#L3035
we can say that it should be.
```CheckAndDestroyVAbuffer``` is also set invalid id to buffer id, so we need to pull this assignment out of the wrapper and do on this level (because if subsequent call to ```vaCreateBuffer``` fails, we will have valid ID for deleted buffer - unwelcome situation).
Saying this I see reasons to use ```CheckAndDestroyVAbuffer``` wrapper

**Comment by dmitryermilov** [comment]
@mgonchar, while I agree with you that the approach with VA_DESTROY_BUFFER is not good, it's currently used in the HEVCe packer for all other functional like SetRoi, SetIntraRefresh etc. So introduction of another way to handle vaBuffers will look weird in current code.  So I asked Dmitry to revert to original version of PR.

**Comment by dmitryermilov** [comment]
>>Create an issue and fix on component-by-component basis?
Yes, it would be a good option

File: _studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp [hunk 12]
```diff
@@ -896,6 +930,7 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
         m_caps.MaxPicHeight = 1088;
 
     m_caps.SliceStructure = 4;
+    m_caps.SliceByteSizeCtrl = 1;
```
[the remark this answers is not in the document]
dmenshov: It means that GPU may further split the slice region that slice control data specifies into finer slice segments based on slice size upper limit (MaxSliceSize).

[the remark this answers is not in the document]
dmitryermilov: I meant to leave a comment in code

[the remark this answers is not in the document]
dmenshov: Done

PR #1064, opened 2018-12-26
Add support for AVC max frame size in legacy mode

horacelan:
Change-Id: I64db4f45d11a4386b4ddbb652e9b87c369a2fbeb

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
@@ -2305,6 +2305,15 @@ void MfxHwH264Encode::ConfigureTask(
         task.m_maxPBFrameSize = 0;
     }
 
+    if (video.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
+    {
+        task.m_repackMaxIFrameSize = video.mfx.MaxIFrameSize;
+        task.m_repackMaxPFrameSize = video.mfx.MaxPFrameSize;
+        task.m_repackMaxBFrameSize = video.mfx.MaxBFrameSize;
+        task.m_repackNumPasses = video.mfx.NumPasses;
+        memcpy_s(task.m_repackDeltaQP, task.m_repackNumPasses, video.mfx.DeltaQP, task.m_repackNumPasses);
+    }
+
     task.m_numMbPerSlice = extOpt2.NumMbPerSlice;
     task.m_numSlice[ffid] = (task.m_type[ffid] & MFX_FRAMETYPE_I) ? extOpt3.NumSliceI :
         (task.m_type[ffid] & MFX_FRAMETYPE_P) ? extOpt3.NumSliceP : extOpt3.NumSliceB;
--- a/_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
+++ b/_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
@@ -251,6 +251,7 @@ namespace MfxHwH264Encode
         VABufferID m_frameRateId;               // VAEncMiscParameterFrameRate
         VABufferID m_qualityLevelId;            // VAEncMiscParameterBufferQualityLevel
         VABufferID m_maxFrameSizeId;            // VAEncMiscParameterFrameRate
+        VABufferID m_multiPassFrameSizeId;            // VAEncMiscParameterBufferMultiPassFrameSize
         VABufferID m_quantizationId;            // VAEncMiscParameterQuantization
         VABufferID m_rirId;                     // VAEncMiscParameterRIR
         VABufferID m_qualityParamsId;           // VAEncMiscParameterEncQuality
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -309,6 +309,55 @@ static mfxStatus SetMaxFrameSize(
     return MFX_ERR_NONE;
 }
 
+static mfxStatus SetMultiPassFrameSize(
+    DdiTask const & task,
+    mfxU32       fieldId,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & frameSizeBuf_id)
+{
+    VAEncMiscParameterBuffer             *misc_param;
+    VAEncMiscParameterBufferMultiPassFrameSize *p_multiPassFrameSize;
+
+    mfxStatus mfxSts = CheckAndDestroyVAbuffer(vaDisplay, frameSizeBuf_id);
+    MFX_CHECK_STS(mfxSts);
+
+    VAStatus vaSts = vaCreateBuffer(vaDisplay,
+                   vaContextEncode,
+                   VAEncMiscParameterBufferType,
+                   sizeof(VAEncMiscParameterBuffer) + sizeof(VAEncMiscParameterBufferMultiPassFrameSize),
+                   1,
+                   NULL,
+                   &frameSizeBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    vaSts = vaMapBuffer(vaDisplay, frameSizeBuf_id, (void **)&misc_param);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    misc_param->type = VAEncMiscParameterTypeMultiPassFrameSize;
+    p_multiPassFrameSize = (VAEncMiscParameterBufferMultiPassFrameSize *)misc_param->data;
+
+    if (task.m_type[fieldId] & MFX_FRAMETYPE_I)
+        p_multiPassFrameSize->max_frame_size = task.m_repackMaxIFrameSize;
+    else if (task.m_type[fieldId] & MFX_FRAMETYPE_P)
+        p_multiPassFrameSize->max_frame_size = task.m_repackMaxPFrameSize;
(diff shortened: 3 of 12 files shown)


_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
```diff
@@ -2305,6 +2305,15 @@ void MfxHwH264Encode::ConfigureTask(
         task.m_maxPBFrameSize = 0;
     }
 
+    if (video.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
+    {
+        task.m_repackMaxIFrameSize = video.mfx.MaxIFrameSize;
+        task.m_repackMaxPFrameSize = video.mfx.MaxPFrameSize;
+        task.m_repackMaxBFrameSize = video.mfx.MaxBFrameSize;
+        task.m_repackNumPasses = video.mfx.NumPasses;
+        memcpy_s(task.m_repackDeltaQP, task.m_repackNumPasses, video.mfx.DeltaQP, task.m_repackNumPasses);
```
mgonchar: Please, do not use _s functions. We also cleaned MSDK code from memcpy calls, switching to std::copy
Here you will have:
```
std::copy(std::begin(video.mfx.DeltaQP), std::begin(video.mfx.DeltaQP) +task.m_repackNumPasses, std::begin(task.m_repackDeltaQP));
```

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -2753,6 +2809,10 @@ mfxStatus VAAPIEncoder::Execute(
                                                           m_vaContextEncode, m_maxFrameSizeId), MFX_ERR_DEVICE_FAILED);
     configBuffers[buffersCount++] = m_maxFrameSizeId;
 
+    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetMultiPassFrameSize(task, fieldId, m_vaDisplay,
+                                                          m_vaContextEncode, m_multiPassFrameSizeId), MFX_ERR_DEVICE_FAILED);
+    configBuffers[buffersCount++] = m_multiPassFrameSizeId;
```
mgonchar: this code traditionally uses dangerous way to store buffers in vector. Could please check that initial resize of vector takes into account current buffer? I mean that your change won't cause out of bounds writing. See line https://github.com/Intel-Media-SDK/MediaSDK/blob/373f2a0250590e51ebe28b04bfdb4b290c1f1671/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp#L2088

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -309,6 +313,55 @@ static mfxStatus SetMaxFrameSize(
...
+    else if (task.m_type[fieldId] & MFX_FRAMETYPE_B)
+        p_multiPassFrameSize->max_frame_size = task.m_repackMaxBFrameSize;
+    else // bad coding type
+        return MFX_ERR_DEVICE_FAILED;
+
+    p_multiPassFrameSize->num_passes = task.m_repackNumPasses;
+    p_multiPassFrameSize->delta_qp = (unsigned char *)task.m_repackDeltaQP;
+
+    {
+        MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaUnmapBuffer");
```
mgonchar: if you trace unmap, then add a trace to map, or remove both

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -2753,6 +2809,10 @@ mfxStatus VAAPIEncoder::Execute(
                                                           m_vaContextEncode, m_maxFrameSizeId), MFX_ERR_DEVICE_FAILED);
     configBuffers[buffersCount++] = m_maxFrameSizeId;
 
+    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetMultiPassFrameSize(task, fieldId, m_vaDisplay,
```
mgonchar: Also, please, put this function call out of the macro. It's expansion in debug build will lead to double function call. I suggest:
```
sts = SetMultiPassFrameSize(task, fieldId, m_vaDisplay,m_vaContextEncode, m_multiPassFrameSizeId);
MFX_CHECK_WITH_ASSERT(sts == MFX_ERR_NONE, MFX_ERR_DEVICE_FAILED);
```

api/include/mfxstructures.h
```diff
@@ -272,7 +272,12 @@ enum {
 
 /* Transcoding Info */
 typedef struct {
-    mfxU32  reserved[7];
+    mfxU32  reserved;
+    mfxU32  MaxIFrameSize;
+    mfxU32  MaxPFrameSize;
+    mfxU32  MaxBFrameSize;
+    mfxU32  NumPasses;
+    mfxU8   DeltaQP[8];
```
lakulako: I'm against extending of topmost mfx structure mfxInfoMFX with very custom BRC related parameters. It would be better to create additional ExtBuffer. Custom feature should be isolated as much as possible.

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -414,6 +414,29 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+            MSDK_FOPEN(m_pRepackCtrl_in, pInParams->repackctrlFile, MSDK_CHAR("rb"));
+            if (m_pRepackCtrl_in == NULL)
+            {
+                msdk_printf(MSDK_STRING("ERROR: Can't open file %s\n"), pInParams->repackctrlFile);
+                return MFX_ERR_UNSUPPORTED;
+            }
+            msdk_printf(MSDK_STRING("Using Frame Size control input file: %s\n"), pInParams->repackctrlFile);
+
+            SAFE_FREAD(&(m_mfxEncParams.mfx.MaxIFrameSize), sizeof(mfxU32), 1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_mfxEncParams.mfx.MaxPFrameSize), sizeof(mfxU32), 1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
```
lakulako: Much better using object name in sizeof than its implicit type.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
```diff
@@ -2305,6 +2305,15 @@ void MfxHwH264Encode::ConfigureTask(
         task.m_maxPBFrameSize = 0;
     }
 
+    if (video.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
```
lakulako: may be if MFX_RATECONTROL_QVBR?
No sense in max frame size for CQP.

_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
```diff
@@ -251,6 +251,7 @@ namespace MfxHwH264Encode
         VABufferID m_frameRateId;               // VAEncMiscParameterFrameRate
         VABufferID m_qualityLevelId;            // VAEncMiscParameterBufferQualityLevel
         VABufferID m_maxFrameSizeId;            // VAEncMiscParameterFrameRate
+        VABufferID m_multiPassFrameSizeId;            // VAEncMiscParameterBufferMultiPassFrameSize
```
lakulako: Please follow the formatting.

cwang64: Will modify to follow the formatting.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
cwang64: The max frame size feature is only enabled in CQP with multipack case.

samples/sample_encode/src/pipeline_encode.cpp
cwang64: Will replace the type with object name.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
cwang64: Will move the function call out of the macro.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
cwang64: Will remove the trace here.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
cwang64: Will add one for MAX_CONFIG_BUFFERS_COUNT macro.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
cwang64: Will replace the memcpy_s with std::copy.

api/include/mfxstructures.h
artem-shaporenko: Not that problem actually, change is done right way in general, but we already have most of these options in mfxExtCodingOption3 (MaxFrameSize for I and P frames and place for different B frame as well as delta QP)

lakulako: Agree, mfxExtCodingOption3 seems to be the best choice.

cwang64:
I have created a new pull request for this feature, so @horacelan please close this pull request, thanks!

horacelan:
Chengwei submmited a new pull request for this feature(https://github.com/Intel-Media-SDK/MediaSDK/pull/1075)  So close this one.

PR #1066, opened 2018-12-27
[hevce] SAO luma/chroma unsupported for CQP

vsidorow:
On Gen10+ VDEnc SAO for only Luma/Chroma in CQP mode
isn't supported by driver


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -2423,7 +2423,7 @@ mfxStatus CheckVideoParam(MfxVideoParam& par, ENCODE_CAPS_HEVC const & caps, boo
             (par.m_ext.HEVCParam.SampleAdaptiveOffset == (mfxU16)MFX_SAO_ENABLE_LUMA || par.m_ext.HEVCParam.SampleAdaptiveOffset == (mfxU16)MFX_SAO_ENABLE_CHROMA))
         {
             par.m_ext.HEVCParam.SampleAdaptiveOffset = (mfxU16)MFX_SAO_DISABLE;
-            changed++;
+            invalid++;
```
lakulako: Why? invalid goes for MFX_ERR_INVALID_VIDEO_PARAM, what means it is unacceptable for any config or makes no sense and can't be corrected. changed goes for MFX_WRN_INCOMPATIBLE_VIDEO_PARAM and can be fixed, what is done in the above line.

lakulako: OK, it then means UNSUPPORTED. Let it be so.

PR #1075, opened 2019-01-08
 Add support for AVC max frame size in legacy mode

cwang64:


samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -414,6 +414,34 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+
+            SAFE_FREAD(&(m_CodingOption3.RepackMaxIFrameSize), sizeof(m_CodingOption3.RepackMaxIFrameSize), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_CodingOption3.RepackMaxPFrameSize), sizeof(m_CodingOption3.RepackMaxPFrameSize), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_CodingOption3.RepackMaxBFrameSize), sizeof(m_CodingOption3.RepackMaxBFrameSize), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_CodingOption3.RepackNumPasses), sizeof(m_CodingOption3.RepackNumPasses), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(m_CodingOption3.RepackDeltaQP, sizeof(m_CodingOption3.RepackDeltaQP[0]) * 8,
```
lakulako: Better sizeof(m_CodingOption3.RepackDeltaQP)

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -624,6 +652,10 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
 
         m_EncExtParams.push_back((mfxExtBuffer *)&m_CodingOption3);
     }
+    else if (m_CodingOption3.RepackMaxIFrameSize || m_CodingOption3.RepackMaxPFrameSize || m_CodingOption3.RepackMaxBFrameSize)
```
lakulako: looks like this block must be merged with previous. Otherwise the buffer pushed here will contain many unset fields.

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -414,6 +414,34 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_CodingOption3.RepackMaxPFrameSize), sizeof(m_CodingOption3.RepackMaxPFrameSize), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_CodingOption3.RepackMaxBFrameSize), sizeof(m_CodingOption3.RepackMaxBFrameSize), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(&(m_CodingOption3.RepackNumPasses), sizeof(m_CodingOption3.RepackNumPasses), 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            SAFE_FREAD(m_CodingOption3.RepackDeltaQP, sizeof(m_CodingOption3.RepackDeltaQP[0]) * 8, 
+                       1, m_pRepackCtrl_in, MFX_ERR_MORE_DATA);
+            if (m_CodingOption3.RepackNumPasses > 4)
```
lakulako: Why not 8, i.e. buffer size?

samples/sample_encode/src/sample_encode.cpp
```diff
@@ -112,6 +112,7 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage, ...)
 #if (MFX_VERSION >= 1027)
     msdk_printf(MSDK_STRING("   [-round_offset_in file]  - use this file to set per frame inter/intra rounding offset(for AVC only)\n"));
 #endif
+    msdk_printf(MSDK_STRING("   [-repackctrl file]       - file to input max encoded frame size,number of pass and delta qp for each frame(ENCODE only)\n"));
```
lakulako: ...size, number of passes and delta qp for each pass ...

samples/sample_encode/src/pipeline_encode.cpp
cwang64: Will modify it, thanks!

samples/sample_encode/src/pipeline_encode.cpp
cwang64: AVC encoder HW only support up to 4 passes.

samples/sample_encode/src/sample_encode.cpp
cwang64: Will modify it, thanks!

samples/sample_encode/src/pipeline_encode.cpp
cwang64: Repack max frame size parameters are independent with other parameters in m_CodingOption3, so I don't think we should merge this with previous.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
```diff
@@ -2312,6 +2312,15 @@ void MfxHwH264Encode::ConfigureTask(
         task.m_maxPBFrameSize = 0;
     }
 
+    if (video.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
+    {
+        task.m_repackMaxIFrameSize = video.mfx.MaxIFrameSize;
+        task.m_repackMaxPFrameSize = video.mfx.MaxPFrameSize;
+        task.m_repackMaxBFrameSize = video.mfx.MaxBFrameSize;
+        task.m_repackNumPasses = video.mfx.NumPasses;
+        memcpy_s(task.m_repackDeltaQP, task.m_repackNumPasses, video.mfx.DeltaQP, task.m_repackNumPasses);
```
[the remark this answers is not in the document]
cwang64: Merged two commits into one commit. Thanks!

samples/sample_encode/src/pipeline_encode.cpp
lakulako: Probably this number will differ for different versions, codecs, platforms. May be better to check it in specific codec::Query but not in common pipeline.

_studio/shared/include/mfx_config.h
```diff
@@ -88,6 +88,9 @@
     #if MFX_VERSION >= 1027
         #define MFX_ENABLE_H264_ROUNDING_OFFSET
     #endif
+    #if MFX_VERSION >= 1028
```
onabiull: Common way to introduce new functionality, which requires API change, is to put it under MFX_VERSION >= MFX_VERSION_NEXT. Thus anyone who build master w/o specifying -DAPI:STRING=latest would get only released API features. When feature considered ready,  MFX_VERSION_NEXT replaced by next API number.  For this one it would be at least 1.29, as API 1.28 finalized three months ago and published as Q3 release.

_studio/shared/src/mfx_static_assert_structs.cpp
```diff
@@ -822,6 +822,13 @@
 #if (MFX_VERSION >= 1026)
         MSDK_STATIC_ASSERT_STRUCT_OFFSET(mfxExtCodingOption3                ,ExtBrcAdaptiveLTR             ,184  )
 #endif
+#if (MFX_VERSION >= 1028)
```
dvrogozh: shouldn't we see MFX_ENABLE_H264_MAX_FRAME_SIZE here? If not, use MFX_VERSION_NEXT.

api/include/mfxstructures.h
```diff
@@ -796,7 +796,18 @@ typedef struct {
 #else
     mfxU16      reserved9;
 #endif
-    mfxU16      reserved[163];
+#if (MFX_VERSION >= 1028)
```
dvrogozh: MFX_VERSION_NEXT

api/mediasdk_structures/ts_struct_decl.h
```diff
@@ -890,6 +890,11 @@ STRUCT(mfxExtCodingOption3,
     FIELD_T(mfxU16, EncodedUnitsInfo)
     FIELD_T(mfxU16, EnableNalUnitType)
     FIELD_T(mfxU16, ExtBrcAdaptiveLTR)     /* tri-state option for ExtBrcAdaptiveLTR */
+    FIELD_T(mfxU32, RepackMaxIFrameSize)
```
dvrogozh: version check missed entirely

samples/sample_common/src/parameters_dumper.cpp
```diff
@@ -366,6 +366,15 @@ void CParametersDumper::SerializeExtensionBuffer(msdk_ostream& sstr,msdk_string
             SERIALIZE_INFO_ARRAY(NumRefActiveBL0);
             SERIALIZE_INFO_ARRAY(NumRefActiveBL1);
             SERIALIZE_INFO(BRCPanicMode);
+#if (MFX_VERSION >= 1028)
```
dvrogozh: MFX_VERSION_NEXT

samples/sample_encode/include/pipeline_encode.h
```diff
@@ -180,6 +180,9 @@ struct sInputParams
 
 #if (MFX_VERSION >= 1027)
     msdk_char *RoundingOffsetFile;
+#endif
+#if (MFX_VERSION >= 1028)
```
dvrogozh: MFX_VERSION_NEXT here and below

cwang64: Hi Dmitry, when you upgrade MFX_VERSION to 1029, will you replace all current MFX_VERSION_NEXT with 1028. I think this feature should be started from version 1028, so I use 1028 here.

api/mediasdk_structures/ts_struct_decl.h
cwang64: The version check is define in line 838.

_studio/shared/include/mfx_config.h
cwang64: Thanks for the explanation, I will modify them.

api/mediasdk_structures/ts_struct_decl.h
dvrogozh: You define new fields in the primary mediasdk uAPI. You need another version check with MFX_VERSION_NEXT.

cwang64: Yes, it is. Please find it in line 838: #if (MFX_VERSION >= MFX_VERSION_NEXT)

samples/sample_encode/src/pipeline_encode.cpp
cwang64: OK, I will move this check into mfx_h264_encode_hw_utils_new.cpp

api/mediasdk_structures/ts_struct_decl.h
dvrogozh: Hm. Yes, indeed. But this looks to be a bug. @onabiull : this definition in ts_struct_decl.h don't correspond to uAPI definition...

api/mediasdk_structures/ts_struct_decl.h
```diff
@@ -890,6 +890,11 @@ STRUCT(mfxExtCodingOption3,
...
+    FIELD_T(mfxU32, RepackMaxPFrameSize)
+    FIELD_T(mfxU32, RepackMaxBFrameSize)
```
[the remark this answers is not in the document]
cwang64: We cannot use MaxFrameSizeI/P/B since they are used by max frame size for BRC not multi PAK.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp
```diff
@@ -2312,6 +2312,19 @@ void MfxHwH264Encode::ConfigureTask(
...
     }
 
+#ifdef MFX_ENABLE_H264_MAX_FRAME_SIZE
+    if (video.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
+    {
+        task.m_repackMaxIFrameSize = extOpt3.RepackMaxFrameSizeI;
+        task.m_repackMaxPFrameSize = extOpt3.RepackMaxFrameSizeP;
+        task.m_repackMaxBFrameSize = extOpt3.RepackMaxFrameSizeB;
+        task.m_repackNumPasses = extOpt3.RepackNumPasses;
+        assert(task.m_repackNumPasses <= 4); // for AVC encoder, NumPasses should be less than or equal to 4
```
mgonchar: this check won't be present in release build. I believe we need it. To make it correctly you need to check ```extOpt3.RepackNumPasses``` somewhere in ```QueryVideoParamsQueryLike```

cwang64: Moved this check to CheckVideoParamQueryLike function. Thanks!

PR #1079, opened 2019-01-10
[VP9D] aspect ratio check on init fixed

IlyaSergeyev:
with some kinds of aspect ratio MFX_VPX_Utility::CheckVideoParam fails
MDP-51420

PR #1081, opened 2019-01-10
Fix aspect ratio checks

fzhar:
If the aspect ratio is not set by the user (contains zero values)
then treat it as (1,1).
If the user set the aspect ratio
then check it is valid (AspectRatioW != 0 and AspectRatioH != 0)

Change-Id: I5007707e986d21c366a1f4e40db45fb2a1fdc077

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
@@ -212,23 +212,15 @@ namespace MFX_VPX_Utility
                 p_out->mfx.FrameInfo.FrameRateExtD = p_in->mfx.FrameInfo.FrameRateExtD;
             }
 
-            if ((p_in->mfx.FrameInfo.AspectRatioW || p_in->mfx.FrameInfo.AspectRatioH) &&
-                (!p_in->mfx.FrameInfo.AspectRatioW || !p_in->mfx.FrameInfo.AspectRatioH))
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
```diff
@@ -212,23 +212,15 @@ namespace MFX_VPX_Utility
                 p_out->mfx.FrameInfo.FrameRateExtD = p_in->mfx.FrameInfo.FrameRateExtD;
             }
 
-            if ((p_in->mfx.FrameInfo.AspectRatioW || p_in->mfx.FrameInfo.AspectRatioH) &&
-                (!p_in->mfx.FrameInfo.AspectRatioW || !p_in->mfx.FrameInfo.AspectRatioH))
+            if ((p_in->mfx.FrameInfo.AspectRatioW == 0 && p_in->mfx.FrameInfo.AspectRatioH == 0) ||
+                (p_in->mfx.FrameInfo.AspectRatioW != 0 && p_in->mfx.FrameInfo.AspectRatioH != 0))
```
lakulako: Simplify if() with (p_in->mfx.FrameInfo.AspectRatioW == 0) == (p_in->mfx.FrameInfo.AspectRatioH == 0),
and similar below. Same as in #1079.

fzhar: This will make code smaller but less readable. I'd better leave it as is.

lakulako: Less? Disagree. The same "both equal to 0 or not".

PR #1084, opened 2019-01-11
hevce: replacing macro MFX_DESTROY_VABUFFER to function CheckAndDestroyVAbuffer

dmenshov:
It is need to align code according AVC encoder.
CheckAndDestroyVAbuffer is also set invalid id to buffer id, so we need to pull this assignment out of the wrapper and do on this level (because if subsequent call to vaCreateBuffer fails, we will have valid ID for deleted buffer - unwelcome situation).

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
+++ b/_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
@@ -37,6 +37,8 @@
 #include <map>
 #include <algorithm>
 
+#define MFX_CHECK_WITH_ASSERT(EXPR, ERR) { assert(EXPR); MFX_CHECK(EXPR, ERR); }
+
 #define MFX_DESTROY_VABUFFER(vaBufferId, vaDisplay)    \
 do {                                               \
     if ( vaBufferId != VA_INVALID_ID)              \
@@ -45,6 +47,19 @@ do {                                               \
         vaBufferId = VA_INVALID_ID;                \
     }                                              \
 } while (0)
+
+inline mfxStatus CheckAndDestroyVAbuffer(VADisplay display, VABufferID & buffer_id)
+{
+    if (buffer_id != VA_INVALID_ID)
+    {
+        VAStatus vaSts = vaDestroyBuffer(display, buffer_id);
+        MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+        buffer_id = VA_INVALID_ID;
+    }
+
+    return MFX_ERR_NONE;
+}
 
 namespace MfxHwH265Encode
 {
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -32,7 +32,6 @@
 #include "mfx_h265_encode_hw_utils.h"
 #include <unordered_map>
 
-#define MFX_CHECK_WITH_ASSERT(EXPR, ERR) { assert(EXPR); MFX_CHECK(EXPR, ERR); }
 //#define PARALLEL_BRC_support
 
 namespace MfxHwH265Encode
@@ -91,7 +90,8 @@ static mfxStatus SetROI(
     VAEncMiscParameterBufferROI *roi_Param;
     unsigned int roi_buffer_size = sizeof(VAEncMiscParameterBufferROI);
 
-    MFX_DESTROY_VABUFFER(roiParam_id, vaDisplay);
+    mfxStatus mfxSts = CheckAndDestroyVAbuffer(vaDisplay, roiParam_id);
+    MFX_CHECK_STS(mfxSts);
 
     vaSts = vaCreateBuffer(vaDisplay,
             vaContextEncode,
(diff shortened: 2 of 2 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -225,8 +226,12 @@ VABufferID& VABuffersHandler::VABuffersNew(mfxU32 id, mfxU32 pool, mfxU32 num)
 
     begin = it; idBegin = idIt;
 
+    mfxStatus mfxSts;
+
     for (;it != end && *idIt == id; it++, idIt++)
-        MFX_DESTROY_VABUFFER(*it, m_vaDisplay);
+    {
+        mfxSts = CheckAndDestroyVAbuffer(m_vaDisplay, *it);
```
DenWolf: Why don't you check mfxSts here? I suppose the main idea of replacing MFX_DESTROY_VABUFFER to CheckAndDestroyVAbuffer is checking returning status every time

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -260,8 +265,12 @@ VABufferID& VABuffersHandler::VABuffersNew(mfxU32 id, mfxU32 pool, mfxU32 num)
 
 void VABuffersHandler::VABuffersDestroy()
 {
+    mfxStatus mfxSts;
+
     for (size_t i = 0; i < m_buf.size(); i++)
-        MFX_DESTROY_VABUFFER(m_buf[i], m_vaDisplay);
+    {
+        mfxSts = CheckAndDestroyVAbuffer(m_vaDisplay, m_buf[i]);
```
DenWolf: The same as above about checking mfxSts

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -278,8 +287,11 @@ void VABuffersHandler::VABuffersDestroyPool(mfxU32 pool)
 
     std::vector<mfxU32>::iterator idBegin = m_id.begin() + std::distance(m_buf.begin(), begin);
 
+    mfxStatus mfxSts;
     for (it = begin; it != end; it++)
-        MFX_DESTROY_VABUFFER(*it, m_vaDisplay);
+    {
+        mfxSts = CheckAndDestroyVAbuffer(m_vaDisplay, *it);
```
DenWolf: The same as above about checking mfxSts

_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
```diff
@@ -46,6 +48,19 @@ do {                                               \
     }                                              \
 } while (0)
```
DenWolf: If we replace MFX_DESTROY_VABUFFER to new function and we will not use this define anymore - I suggest to delete MFX_DESTROY_VABUFFER

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
lakulako: Have to continue destroying another buffers and somehow report errors.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
lakulako: No need in status in destructor. Just log error maybe.

_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
```diff
@@ -37,6 +37,8 @@
 #include <map>
 #include <algorithm>
 
+#define MFX_CHECK_WITH_ASSERT(EXPR, ERR) { assert(EXPR); MFX_CHECK(EXPR, ERR); }
+
```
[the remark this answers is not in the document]
DenWolf: define MFX_CHECK_WITH_ASSERT just moved from source file (mfx_h265_encode_vaapi,h) - please see the next file in this PR - so that's OK

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
DenWolf: Ok, I see that function VABuffersDestroy return "void" - in this case I suggest to remove "mfxStatus mfxSts;" if it's not used

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
mgonchar: yes, we need some kind of logging. at least we should see this status in debug

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
mgonchar: as I stated above - better to have status at least in debug. If no good idea how to use it, I suggest something like
```
#if DEBUG
mfxStatus sts = 
#endif 
CheckAndDestroyVAbuffer(m_vaDisplay, m_buf[i]);
```

mgonchar:
No need in this change anymore. It was fully covered in #1194, so I close this PR.

Issue #864, opened 2018-11-05
FFmpeg-qsv encoding failed when NumMbPerSlice < 50


lizhong1008:
Apply patch: https://patchwork.ffmpeg.org/patch/10778/ 
Then run:
./ffmpeg -hwaccel qsv -c:v h264_qsv -i bbb_sunflower_1080p_30fps_normal.mp4 -vframes 100 -c:v h264_qsv -flags qscale -global_quality 25 -num_mb_per_slice 49 -y test.mp4 -loglevel verbose

Encoding failed with error message:
[h264_qsv @ 0x2e9b480] Error during encoding: device failed (-17)
Video encoding failed

if set num_mb_per_slice >=50, everything is ok. 

The test input clip can be downloaded from: http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_1080p_30fps_normal.mp4

dvrogozh:
@lizhong1008 : https://patchwork.ffmpeg.org/patch/10778/ does not apply by itself, it has some prerequisites. Could you, please, list them all here?

lizhong1008:
@dvrogozh , I don't know why ffmpeg patchwork just received 5 patches (7 patches in this series).
However, you can get all of them from https://github.com/intel-media-ci/ffmpeg/pull/4

andreasunterhuber:
Any update?

dvrogozh:
After https://github.com/Intel-Media-SDK/MediaSDK/commit/58835b09202e472830774fc556d3227b34dfccbe I see error gone for num_mb_per_slice=49, but now it (this can be other error inside mediasdk, but it has the same signature on ffmpeg level) happens on the threshold of 31/32 (at least in case of my clip):

```
ffmpeg \
  -hwaccel qsv -c:v h264_qsv -i 1080p_30fpsl.mp4 -vframes 100 \
  -c:v h264_qsv -flags qscale -global_quality 25 -num_mb_per_slice 31 -mfmode off -y test.mp4 -loglevel verbose
[h264_qsv @ 0x1fe6180] Using the constant quantization parameter (CQP) ratecontrol method
[h264_qsv @ 0x1fe6180] MFMode:1
[AVHWDeviceContext @ 0x21cb440] VAAPI driver: Intel iHD driver - 1.0.0.
[AVHWDeviceContext @ 0x21cb440] Driver not found in known nonstandard list, using standard behaviour.
[...]

PR #1085, opened 2019-01-11
[AVCe] Fix for slice header packing

ogolikov:
Fixes: #864

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -8963,22 +8963,22 @@ std::vector<ENCODE_PACKEDHEADER_DATA> const & HeaderPacker::PackSlices(
     DdiTask const & task,
     mfxU32          fieldId)
 {
+    const mfxU32 maxSliceHeaderSize = 50;  //maximum coded slice header size in bytes
+    size_t numSlices = task.m_SliceInfo.size() ?
+        task.m_SliceInfo.size() :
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -8963,22 +8963,22 @@ std::vector<ENCODE_PACKEDHEADER_DATA> const & HeaderPacker::PackSlices(
     DdiTask const & task,
     mfxU32          fieldId)
 {
+    const mfxU32 maxSliceHeaderSize = 50;
```
lakulako: Where did 50 come from? sizeof? at least explain in code.
Better use size_t for all types here and thus avoid all castings. Old style castings.

dvrogozh:
@ogolikov : can you, please, update the comment and let's close this PR?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -8963,22 +8963,22 @@ std::vector<ENCODE_PACKEDHEADER_DATA> const & HeaderPacker::PackSlices(
     DdiTask const & task,
     mfxU32          fieldId)
 {
+    const mfxU32 maxSliceHeaderSize = 50;  //maximum coded slice header size in bytes
+    size_t numSlices = task.m_SliceInfo.size() ?
```
mgonchar: better to use ```.empty()``` to check for emptiness

PR #1092, opened 2019-01-14
vp8d: Fixed playback

Losev-Ivan:
On Android O we use old version of driver and should use
special code for 1st partition size computation (for count == 8)

### DIFF
--- a/_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw.cpp
@@ -1284,11 +1284,20 @@ mfxStatus VideoDECODEVP8_HW::DecodeFrameHeader(mfxBitstream *in)
         while (++i < 2);
     }
 
+// On Android O we use old version of driver and should use special code for 1st partition size computation (for count == 8)
+#if (MFX_ANDROID_VERSION >= MFX_P)
     // Header info consumed bits
     m_frame_info.entropyDecSize = m_boolDecoder[VP8_FIRST_PARTITION].pos() * 8 - 3*8 - m_boolDecoder[VP8_FIRST_PARTITION].bitcount();
 
     // Subtract completely consumed bytes + current byte. Current is completely consumed if bitcount is 8.
     m_frame_info.firstPartitionSize = first_partition_size - ((m_frame_info.entropyDecSize + 7) >> 3);
+#else
+    // Header info consumed bits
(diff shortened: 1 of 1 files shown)


Losev-Ivan:
Related to #493

_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw.cpp
```diff
@@ -1284,11 +1284,20 @@ mfxStatus VideoDECODEVP8_HW::DecodeFrameHeader(mfxBitstream *in)
         while (++i < 2);
     }
 
+// On Android O we use old version of driver and should use special code for 1st partition size computation (for count == 8)
```
dvrogozh: Why you don't use old version of mediasdk then? What is this driver on Android O: open or closed source?

Losev-Ivan: On Android O  we have frozen version of Open-Source driver (~Spring-Summer 2018), and there is no plans to change source code of driver.
We have a plan to provide new features and bug fixing from latest Media SDK source code, and now we blocked only by broken vp8d, current PR fixed the problem (by recovering old source code).

_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw.cpp
```diff
@@ -1284,11 +1284,20 @@ mfxStatus VideoDECODEVP8_HW::DecodeFrameHeader(mfxBitstream *in)
...
+#if (MFX_ANDROID_VERSION >= MFX_P)
     // Header info consumed bits
     m_frame_info.entropyDecSize = m_boolDecoder[VP8_FIRST_PARTITION].pos() * 8 - 3*8 - m_boolDecoder[VP8_FIRST_PARTITION].bitcount();
 
     // Subtract completely consumed bytes + current byte. Current is completely consumed if bitcount is 8.
     m_frame_info.firstPartitionSize = first_partition_size - ((m_frame_info.entropyDecSize + 7) >> 3);
+#else
+    // Header info consumed bits
+    m_frame_info.entropyDecSize = m_boolDecoder[VP8_FIRST_PARTITION].pos() * 8 - 16 - m_boolDecoder[VP8_FIRST_PARTITION].bitcount();
```
akokoshn: why not 2*8 like above (line 1290)?
just looks better, not required:)

DenWolf: Agree, that it would be better, but current patch recovers original code for calculating macroblock for VP8D without any changes, and original code  - "16"

PR #1093, opened 2019-01-14
Fix BDW Lookahead kernel

artem-shaporenko:
Fixed issue with lookahead BDW kernel loading, earlier kernel was not rebuilt with updated interfaces


artem-shaporenko:
> I don't see a change in the kernel source file.

There were no changes in kernel code, earlier kernel wasn't rebuilt properly.

PR #1095, opened 2019-01-14
[vp9d] VideoDECODEVP9_HW::Reset with Y410 on ICL causes error

IlyaSergeyev:
[vp9d] VideoDECODEVP9_HW::Reset with Y410 on ICL causes INVALID_VIDEO_PARAM

MDP-50999

hwtype does not passed properly


_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
```diff
@@ -304,8 +304,6 @@ namespace MFX_VPX_Utility
 
     bool CheckVideoParam(mfxVideoParam const*p_in, mfxU32 codecId, eMFXPlatform platform, eMFXHWType hwtype)
     {
-        (void)hwtype;
-
         if (!p_in)
```
dvrogozh: Why?
1. I still don't see this parameter referenced. On clang/android this probably will give a warning
2. This looks to be not related to the fix you are doing. Why change the code then?

IlyaSergeyev: There are condition (row 361) where  hwtype is checked

dvrogozh: ok, I see it now. However, this check is macro conditional for api 1.27. I believe we have Android builds for api 1.26 (@dmitrybrazhkin : please, comment).

IlyaSergeyev:
Other commit with same changes merged

PR #1101, opened 2019-01-14
vp9d: add flushing

akokoshn:
-Return locked surfcae and sync point if input bs = 0:
 actuall in case of latest frame is not  "show frame"

### DIFF
--- a/_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode_hw.h
+++ b/_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode_hw.h
@@ -37,6 +37,8 @@
 
 namespace UMC_VP9_DECODER { class Packer; }
 
+class FrameStorage;
+
 class VideoDECODEVP9_HW : public VideoDECODE, public MfxCriticalErrorHandler
 {
 public:
@@ -66,7 +68,8 @@ protected:
     void CalculateTimeSteps(mfxFrameSurface1 *);
     static mfxStatus QueryIOSurfInternal(eMFXPlatform, mfxVideoParam *, mfxFrameAllocRequest *);
 
-    mfxStatus UpdateRefFrames(const mfxU8 refreshFrameFlags, UMC_VP9_DECODER::VP9DecoderFrame & info);
+    mfxStatus UpdateRefFrames();
+    mfxStatus CleanRefList();
 
     mfxStatus DecodeSuperFrame(mfxBitstream *in, UMC_VP9_DECODER::VP9DecoderFrame & info);
     mfxStatus DecodeFrameHeader(mfxBitstream *in, UMC_VP9_DECODER::VP9DecoderFrame & info);
@@ -98,6 +101,7 @@ private:
     std::unique_ptr<mfx_UMC_FrameAllocator> m_FrameAllocator;
 
     std::unique_ptr<UMC_VP9_DECODER::Packer>  m_Packer;
+    std::unique_ptr<FrameStorage> m_framesStorage;
 
     mfxFrameAllocRequest     m_request;
     mfxFrameAllocResponse    m_response;
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -86,6 +86,173 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
     return true;
 }
 
+mfxStatus VideoDECODEVP9_HW::UpdateRefFrames()
+{
+    if (m_frameInfo.show_existing_frame)
+        return MFX_ERR_NONE;
+
+    for (mfxI32 ref_index = 0; ref_index < NUM_REF_FRAMES; ++ref_index)
+    {
+        mfxU8 mask = m_frameInfo.refreshFrameFlags & (1 << ref_index);
+        if (mask != 0 // we should update the reference according to the bitstream data
+
+            // The next condition is for decoder robustness.
+            // After the first keyframe is decoded this frame occupies all ref slots
+            // its mask is 011111111 and the condition "minus one" never hits.
+            // But if the first frame is Inter-frame encoded in the intra-only mode
+            // then we can receive "minus one" ref_frame_map item.
+            // The decoder never should touch these references.
+            // But if the stream is broken then decode may touch not initialized reference
+            // So to prevent crash let's put something as the reference frame.
+            || (m_frameInfo.ref_frame_map[ref_index] == (UMC::FrameMemID)-1))
+        {
+            MFX_CHECK((m_FrameAllocator->IncreaseReference(m_frameInfo.currFrame) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+
+            if (m_frameInfo.ref_frame_map[ref_index] >= 0)
+                MFX_CHECK((m_FrameAllocator->DecreaseReference(m_frameInfo.ref_frame_map[ref_index]) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+            m_frameInfo.ref_frame_map[ref_index] = m_frameInfo.currFrame;
+
+            m_frameInfo.sizesOfRefFrame[ref_index].width = m_frameInfo.width;
+            m_frameInfo.sizesOfRefFrame[ref_index].height = m_frameInfo.height;
+        }
+    }
+    return MFX_ERR_NONE;
+}
+mfxStatus VideoDECODEVP9_HW::CleanRefList()
+{
+    for (mfxI32 ref_index = 0; ref_index < NUM_REF_FRAMES; ++ref_index)
+    {
+        if (m_frameInfo.ref_frame_map[ref_index] >= 0)
+            MFX_CHECK((m_FrameAllocator->DecreaseReference(m_frameInfo.ref_frame_map[ref_index]) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+
+        m_frameInfo.ref_frame_map[ref_index] = -1;
+    }
+    return MFX_ERR_NONE;
+}
+
+
+class FrameStorage
+{
+private:
+    std::vector<UMC_VP9_DECODER::VP9DecoderFrame> m_submittedFrames;
+    mfx_UMC_FrameAllocator *m_frameAllocator;
+
+    void LockResources(const UMC_VP9_DECODER::VP9DecoderFrame & frame) const
+    {
+        // lock current frame for decode
+        VP9_CHECK_AND_THROW((m_frameAllocator->IncreaseReference(frame.currFrame) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+        // lock frame for copy
+        if (frame.show_existing_frame)
+        {
+            VP9_CHECK_AND_THROW((m_frameAllocator->IncreaseReference(frame.ref_frame_map[frame.frame_to_show]) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+        }
+        // lock all references
+        else
+        {
+            for (mfxI32 ref_index = 0; ref_index < NUM_REF_FRAMES; ++ref_index)
+            {
+                const UMC::FrameMemID mid = frame.ref_frame_map[ref_index];
+                if (mid >= 0)
+                {
+                     VP9_CHECK_AND_THROW((m_frameAllocator->IncreaseReference(mid) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+                }
+            }
+        }
+    }
+
+    void UnLockResources(const UMC_VP9_DECODER::VP9DecoderFrame & frame) const
+    {
+        // decoded frame should be unlocked in async routine
+
+        // unlock frame for copy
+        if (frame.show_existing_frame)
+        {
+            VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(frame.ref_frame_map[frame.frame_to_show]) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+        }
+        // unlock all references
+        else
+        {
+            for (mfxI32 ref_index = 0; ref_index < NUM_REF_FRAMES; ++ref_index)
+            {
+                const UMC::FrameMemID mid = frame.ref_frame_map[ref_index];
+                if (mid >= 0)
+                {
+                    VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(mid) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+                }
+            }
+        }
+    }
+public:
+    FrameStorage(mfx_UMC_FrameAllocator *frameAllocator):
+        m_submittedFrames(),
+        m_frameAllocator(frameAllocator)
+    {
+    }
+
+    ~FrameStorage()
+    {
+        // unlock all locked resources
+        try
+        {
+            for (auto & it : m_submittedFrames)
+                UnLockResources(it);
+        }
+        catch (vp9_exception const& e)
+        {
+            m_submittedFrames.shrink_to_fit();
+            m_submittedFrames.clear();
+            VM_ASSERT(0);
+        }
+        m_submittedFrames.shrink_to_fit();
+        m_submittedFrames.clear();
+    }
+
+    void Add(UMC_VP9_DECODER::VP9DecoderFrame & frame)
+    {
+        VP9_CHECK_AND_THROW((frame.currFrame >= 0), MFX_ERR_UNKNOWN);
+
+        // lock refereces for current frame
+        LockResources(frame);
+
+        // avoid double submit
+        auto find_it = std::find_if(m_submittedFrames.begin(), m_submittedFrames.end(),
+            [frame](const UMC_VP9_DECODER::VP9DecoderFrame & item){ return item.currFrame == frame.currFrame; });
+        VP9_CHECK_AND_THROW((find_it == m_submittedFrames.end()), MFX_ERR_UNKNOWN);
+
+        m_submittedFrames.push_back(frame);
+    }
+
+    void DecodeFrame(UMC::FrameMemID frameId)
+    {
+        auto find_it = std::find_if(m_submittedFrames.begin(), m_submittedFrames.end(),
+            [frameId](const UMC_VP9_DECODER::VP9DecoderFrame & item){ return item.currFrame == frameId; });
+
+        if (find_it != m_submittedFrames.end())
+        {
+            find_it->isDecoded = true;
+        }
+    }
+
+    void CompleteFrames()
+    {
+        for( auto it = m_submittedFrames.begin(); it != m_submittedFrames.end(); )
+        {
+            if (it->isDecoded)
+            {
+                UnLockResources(*it);
+                it = m_submittedFrames.erase(it);
+            }
+            else
+                it = std::next(it);
+        }
+    }
+
+    bool IsAllFramesCompleted() const
+    {
+        return m_submittedFrames.empty();
+    }
+};
+
 VideoDECODEVP9_HW::VideoDECODEVP9_HW(VideoCORE *p_core, mfxStatus *sts)
     : m_isInit(false),
       m_is_opaque_memory(false),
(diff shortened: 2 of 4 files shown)


akokoshn:
WIP, need an verification

dmitry-gurulev:
I think u actually mean 'flush' not 'flash'.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -751,7 +757,43 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     if (NeedToReturnCriticalStatus(bs))
         return ReturningCriticalStatus();
 
-    MFX_CHECK_NULL_PTR2(surface_work, surface_out);
+    if (bs && !bs->DataLength)
+        return MFX_ERR_MORE_DATA;
+
+    MFX_CHECK_NULL_PTR1(surface_out);
+
+    *surface_out = 0;
```
mgonchar: nullptr

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -730,6 +734,8 @@ mfxStatus MFX_CDECL VP9DECODERoutine(void *p_state, void * /* pp_param */, mfxU3
     if (data.currFrameId != -1)
        decoder.m_FrameAllocator->DecreaseReference(data.currFrameId);
 
+    decoder.CheckFreeBufferedSurface(data.currFrameId);
```
DenWolf: The same as upper - I think it's better to move this to "if (data.currFrameId != -1)" condition on line 734

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -751,7 +760,43 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
...
+            if (surface != nullptr && surface->Data.Locked)
+            {
+                isAllRefFramesUnlocked = false;
+                break;
+            }
+        }
+        if (isAllRefFramesUnlocked)
+            return MFX_ERR_MORE_DATA;
+        else
+            return MFX_WRN_DEVICE_BUSY;
```
dmitry-gurulev: I'd rather change explicit loop w/ STL algo, like this:
```
// check buffered surfaces
auto locked = std::any_of(std::begin(bufferedSurfaces), std::end(bufferedSurfaces),
    [this](UMC::FrameMemID id)
    {
        auto surface = m_FrameAllocator->GetSurfaceByIndex(id);
        return surface && surface->Data.Locked;
    }
);

return locked ? MFX_WRN_DEVICE_BUSY : MFX_ERR_MORE_DATA;

```

_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode_hw.h
```diff
@@ -124,6 +124,10 @@ class VideoDECODEVP9_HW : public VideoDECODE, public MfxCriticalErrorHandler
     typedef std::list<mfxFrameSurface1 *> StatuReportList;
     StatuReportList m_completedList;
 
+    std::vector<UMC::FrameMemID> m_bufferedSurfaces;
```
dmitry-gurulev: We have [m_frameInfo.ref_frame_map], why do we need yet another array of IDs?
Anyway, even If we really need something like that, u have to sync w/ [PR #1114](https://github.com/Intel-Media-SDK/MediaSDK/pull/1114) which introduces _ReferenceFrameStorage_ that in turn has _m_lockedRefFrames_ map & _m_decodedFramesIDs_ array.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,78 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+        }
+
+    }
+
+public:
+    ReferenceFrameStorage(mfx_UMC_FrameAllocator *frameAllocator):m_frameAllocator(frameAllocator)
+    {
+    }
+
+    void Add(UMC_VP9_DECODER::VP9DecoderFrame currFrame)
```
dmitry-gurulev: what is benefit to pass [currFrame] by value, not by const reference?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,78 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    {
+        m_submittedFrames.push_back(currFrame);
+    }
+
+    void FrameDecoded(UMC::FrameMemID frameId)
+    {
+        auto find_it = std::find_if(m_submittedFrames.begin(), m_submittedFrames.end(), [frameId](UMC_VP9_DECODER::VP9DecoderFrame frame){ return (frame.currFrame == frameId); });
```
dmitry-gurulev: It would be better to adjust formatting, like it is done at [FreeUnusedReferenceFrames] method.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,78 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+        if (find_it != m_submittedFrames.end())
+        {
+            find_it->isDecoded = true;
+        }
+        return;
```
dmitry-gurulev: `return` is redundant

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,78 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    mfx_UMC_FrameAllocator *m_frameAllocator;
+
+    void FreeReferenceFrames(UMC::FrameMemID frameId, bool force)
+    {
+        auto find_it = std::find_if(m_submittedFrames.begin(), m_submittedFrames.end(), [frameId](UMC_VP9_DECODER::VP9DecoderFrame frame){ return (frame.currFrame == frameId); });
+        if (find_it != m_submittedFrames.end())
+        {
+            // if force=TRUE need to unlock all reference frames
+            // otherwiseref useless references wil be unlocked in VideoDECODEVP9_HW::UpdateRefFrames()
+            if (force)
```
dmitry-gurulev: `force` looks like a dirty hack. Consider using of `FreeReferenceFrames` from `UpdateRefFrames`

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,87 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+
+public:
+    ReferenceFrameStorage(mfx_UMC_FrameAllocator *frameAllocator):m_frameAllocator(frameAllocator)
+    {
+    }
+
+    void Add(UMC_VP9_DECODER::VP9DecoderFrame frame)
+    {
+        auto find_it = std::find_if(m_submittedFrames.begin(), m_submittedFrames.end(),
+            [frame](UMC_VP9_DECODER::VP9DecoderFrame _frame)
```
dmitry-gurulev: const reference?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,87 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
     return true;
 }
 
+class ReferenceFrameStorage
+{
+private:
+    std::vector<UMC_VP9_DECODER::VP9DecoderFrame> m_submittedFrames;
+    mfx_UMC_FrameAllocator *m_frameAllocator;
+
+    void FreeReferenceFrames(UMC::FrameMemID frameId)
```
dmitry-gurulev: `VP9DecoderFrame` instead of `FrameMemID` to avoid redundant search

_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode_hw.h
```diff
@@ -37,6 +37,8 @@
 
 namespace UMC_VP9_DECODER { class Packer; }
 
+class ReferenceFrameStorage;
```
maksenox: The class name is confused. It is not Reference frames it is frames submitted to the driver or frames in decoding. My concern is VP9DecoderFrame structure contains ref_frame_map  which is the list of reference frames.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,87 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    {
+        auto find_it = std::find_if(m_submittedFrames.begin(), m_submittedFrames.end(), [frameId](UMC_VP9_DECODER::VP9DecoderFrame frame){ return (frame.currFrame == frameId); });
+        if (find_it != m_submittedFrames.end())
+        {
+            for (mfxI32 ref_index = 0; ref_index < NUM_REF_FRAMES; ++ref_index)
+            {
+                const UMC::FrameMemID mid = find_it->ref_frame_map[ref_index];
+                if (mid >= 0 && mid != find_it->currFrame)
+                {
+                    m_frameAllocator->DecreaseReference(mid);
```
maksenox: IncreaseReference outside Storage (Decoder UpdateReference) DecreaseReference inside Sotrage class.
I think it would be better to put reference counting logic in one place this Storage class, add new frame increases all required references, mark frame as decoded, free unused frames decreases all reference back and it should be maid from synchronous code DecodeFrameAsync, move all DecreaseReference functions from async DecodeRoutine to the decoder function.

_studio/shared/umc/codec/vp9_dec/include/umc_vp9_frame.h
```diff
@@ -107,6 +107,8 @@ class VP9DecoderFrame
 
     uint32_t frameCountInBS;
     uint32_t currFrameInBS;
+
+    bool isDecoded;
```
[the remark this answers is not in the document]
akokoshn: isDecoded - is state of frame, related to one frame. We can't use condition "frame in storage, so it was submitted, but not decoded", otherwise we have to make an delete from storage in async part.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +87,133 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    mfxStatus Add(UMC_VP9_DECODER::VP9DecoderFrame & frame)
+    {
+        MFX_CHECK((frame.currFrame >= 0), MFX_ERR_UNKNOWN);
+
+        mfxStatus sts = UpdateRefFrames(frame);
+        MFX_CHECK_STS(sts);
+
+        if (!frame.show_existing_frame)
+        {
+            sts = FreeUnusedReferenceFrames();
```
dmitry-gurulev: Why is it only for `!show_existing_frame`?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,87 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
     return true;
 }
 
+class ReferenceFrameStorage
+{
+private:
+    std::vector<UMC_VP9_DECODER::VP9DecoderFrame> m_submittedFrames;
```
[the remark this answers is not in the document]
dmitry-gurulev: As for  me, I don't see any problems w/ `std::find` since number of elements is small enough. At the same time, having same mem ID both as a map's key and as member of `VP9DecoderFrame` is redundant and confusing. I'd rather use `vector` or `set` at least.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -369,14 +508,9 @@ mfxStatus VideoDECODEVP9_HW::Close()
...
 {
-    for (mfxU8 i = 0; i < sizeof(m_frameInfo.ref_frame_map)/sizeof(m_frameInfo.ref_frame_map[0]); i++)
-    {
-        const UMC::FrameMemID oldMid = m_frameInfo.ref_frame_map[i];
-        if (oldMid >= 0)
-        {
-            m_FrameAllocator->DecreaseReference(oldMid);
-        }
-    }
+    if (m_framesStorage)
```
maksenox: Useless if. You should add the destructor to the class. And you missed the cleanup in the destructor. CompleteFrames removes decoded frame from the map and frees the resources. IsAllFramesCompleted should check that the map is empty. The destructor should clean up the map.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -751,7 +883,43 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
...
+            if (m_frameInfo.ref_frame_map[ref_index] >= 0)
+            {
+                MFX_CHECK((m_FrameAllocator->DecreaseReference(m_frameInfo.ref_frame_map[ref_index]) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+                m_frameInfo.ref_frame_map[ref_index] = -1;
+            }
+        }
+        bool isDecodingFinished = false;
+        try
+        {
+            m_framesStorage->CompleteFrames();
```
maksenox: Method CompleteFrames should be the first statement in DecodeCheckFrame because it releases decoded frames and its reference.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +87,143 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+        {
+            const UMC::FrameMemID mid = frame.ref_frame_map[ref_index];
+            if (mid >= 0)
+            {
+                VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(mid) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+            }
+        }
+    }
+
+    void UpdateRefFrames(UMC_VP9_DECODER::VP9DecoderFrame & frame) const
```
maksenox: UpdateRefFrames should not be the method of FrameStorage class. FrameStorage tracks the list of frames in decoding state and locks all resources required for the frames. UpdateRefFrames tracks the list of current reference frames and updates the current list for the next frame.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +87,143 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    }
+
+    void Add(UMC_VP9_DECODER::VP9DecoderFrame & frame)
+    {
+        VP9_CHECK_AND_THROW((frame.currFrame >= 0), MFX_ERR_UNKNOWN);
+
+        UpdateRefFrames(frame);
+        LockResources(frame);
+        CompleteFrames();
+
```
maksenox: Wrong function calls order. First LockResources for the current frame. After the resources are locked we can call UpdateRefFrames which will update the reference frame list for the next frame and this function should be called from the decoder, not from the FrameStorage (this change should break async decoding, did you test it). I believe CompleteFrames should not be called from here. I see the only place where it should be called is the first statement in DecodeFrameCheck, because of guard(mutex) no one will be able to modify FrameStorage.

akokoshn: Sure I  tested change on conformance suites

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,164 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    {
+    }
+
+    ~FrameStorage()
+    {
+        // unlock all locked resources
+        for (auto & it : m_submittedFrames)
+            UnLockResources(it);
+        m_submittedFrames.shrink_to_fit();
+        m_submittedFrames.clear();
```
dmitry-gurulev: This (free of unused capacity, clearing) will be done by `vector`'s dtor, u don't need to do this explicitly.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,164 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+                {
+                     VP9_CHECK_AND_THROW((m_frameAllocator->IncreaseReference(mid) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+                }
+            }
+        }
+    }
+
+    void UnLockResources(const UMC_VP9_DECODER::VP9DecoderFrame & frame) const
+    {
+        // decoded frame should be unlocked in async routine
```
dmitry-gurulev: I believe we have to unlock decoded frame here instead of doing that at async routine. What is a reason of not doing that?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,164 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+        {
+            for (mfxI32 ref_index = 0; ref_index < NUM_REF_FRAMES; ++ref_index)
+            {
+                const UMC::FrameMemID mid = frame.ref_frame_map[ref_index];
+                if (mid >= 0)
+                {
+                    VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(mid) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+                }
+            }
+        }
```
maksenox: I suggest to unlock all frames and if some DecreaseReference has failed then the last statement throws an exception.

akokoshn: if you means remove line 177 ("if (mid >= 0)"), I'm not agree: exception here leads critical error, in case of 1-th frame frame.ref_frame_map[ref_index] = -1 for all ref_index.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +85,164 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+        m_submittedFrames(),
+        m_frameAllocator(frameAllocator)
+    {
+    }
+
+    ~FrameStorage()
+    {
+        // unlock all locked resources
+        for (auto & it : m_submittedFrames)
+            UnLockResources(it);
```
maksenox: Unlock resources may throw an exception. The destructors cannot throw the exceptions. Add try/catch.

PR #1103, opened 2019-01-15
[VP9D] Reopen Lock reference frames before UpdateRefList

IlyaSergeyev:
Make reference frames are locked (increase reference counter)
before UpdateRefList.

UpdateRefList prepares the references for the next frame and
removes frames from the current list, when it removes it calls
DecreaseRefCount function.
If frame refCount is zero Free function is called
this function clears Locked member and the reference frame
could be used as new target frame
while still in use by HW decoder.

Issue: MDP-52110
Test: Linux/ICL


IlyaSergeyev:
Copy of #1114

PR #1107, opened 2019-01-16
hevce: add setting weights/offsets for driver

otsirule:
This commit is prerequisite for enabling weighted prediction,
but support for all platforms is still disabled, because
driver doesn't have capability reporting for this functionality.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -554,7 +554,7 @@ void FillConstPartOfPps(
     pps.pic_fields.bits.transform_skip_enabled_flag                = par.m_pps.transform_skip_enabled_flag;
     pps.pic_fields.bits.cu_qp_delta_enabled_flag                   = par.m_pps.cu_qp_delta_enabled_flag;
     pps.pic_fields.bits.weighted_pred_flag                         = par.m_pps.weighted_pred_flag;
-    pps.pic_fields.bits.weighted_bipred_flag                       = par.m_pps.weighted_pred_flag;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -679,18 +679,37 @@ void UpdateSlice(
...
-        slice.luma_offset_l1
-        slice.delta_chroma_weight_l1
-        slice.chroma_offset_l1
-*/
+
+#if defined(MFX_ENABLE_HEVCE_WEIGHTED_PREDICTION)
+    if ((slice.slice_type == SLICE_P && pps.pic_fields.bits.weighted_pred_flag) ||
+        (slice.slice_type == SLICE_B && pps.pic_fields.bits.weighted_bipred_flag))
+    {
+        const mfxU16 Y = 0, Cb = 1, Cr = 2, W = 0, O = 1;
```
lakulako: O looks too much like 0. Possible to rename?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -679,18 +679,37 @@ void UpdateSlice(
...
+
+        slice.luma_log2_weight_denom           = task.m_sh.luma_log2_weight_denom;
+        slice.delta_chroma_log2_weight_denom   = task.m_sh.chroma_log2_weight_denom - slice.luma_log2_weight_denom;
```
lakulako: Needs [static_]cast to avoid warnings. Similar in the loop below.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
otsirule: Done

PR #1114, opened 2019-01-17
Lock ref frames

maksenox:


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -671,7 +799,7 @@ mfxStatus MFX_CDECL VP9DECODERoutine(void *p_state, void * /* pp_param */, mfxU3
 
     if (!data.surface_work)
         return MFX_ERR_UNDEFINED_BEHAVIOR;
-
+    printf("%d", data.currFrameId);
```
mgonchar: You shouldn'd upload debug prints to upstream code

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -85,6 +87,130 @@ bool CheckHardwareSupport(VideoCORE *p_core, mfxVideoParam *p_video_param)
...
+    {
+        if(m_lockedRefFrames.find(currFrame) != m_lockedRefFrames.end())
+        {
+            // currFrame should not be in m_lockedRefFrames map
+            // but if it is there free resources (decrease the references).
+            FreeReferenceFrames(currFrame);
+            (void)MFX_STS_TRACE(MFX_ERR_UNKNOWN);
+        }
+
+        for (mfxI32 i = 0; i < NUM_REF_FRAMES; ++i)
```
mgonchar: This line is inconsistent with declaration at line 163. I suggest to use same constant for declaration as well

PR #1118, opened 2019-01-17
vp8d: Fixed incorrect computation of 1st partition size

Losev-Ivan:
Fixes: 88f8b98cf0a280efb35d54da4843d713438ecf84 ("vp8: Fixed vp8 playback")


_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2019 Intel Corporation
```
DenWolf: Something wrong with Copyright years - need to be 2018-2019

PR #1128, opened 2019-01-22
mpeg2d: enable flushing with zero workSurface

akokoshn:
-Enable flushing with zero workSurface for unlock buffered surface
 without using extra (useless) surface


_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -772,7 +777,8 @@ mfxStatus VideoDECODEMPEG2::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surf
 
     std::unique_ptr<TaskInfo> info(new TaskInfo); // to prevent memory leak
```
mgonchar: this comment is useless

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -651,31 +651,36 @@ mfxStatus VideoDECODEMPEG2::DecodeFrameCheck(mfxBitstream* bs, mfxFrameSurface1*
...
-        sts = CheckFrameInfoCodecs(&surface_work->Info, MFX_CODEC_MPEG2);
-        MFX_CHECK(MFX_ERR_NONE == sts, MFX_ERR_UNSUPPORTED);
+        if (m_opaque)
+        {
+            sts = CheckFrameInfoCodecs(&surface_work->Info, MFX_CODEC_MPEG2);
+            MFX_CHECK(MFX_ERR_NONE == sts, MFX_ERR_UNSUPPORTED);
 
-        if (surface_work->Data.MemId || surface_work->Data.Y || surface_work->Data.R || surface_work->Data.A || surface_work->Data.UV) // opaq surface
-            return MFX_ERR_UNDEFINED_BEHAVIOR;
+            if (surface_work->Data.MemId || surface_work->Data.Y || surface_work->Data.R || surface_work->Data.A || surface_work->Data.UV) // opaq surface
```
mgonchar: I would suggest to use ```MFX_CHECK``` macro, it will allow to trace triggering of this condition

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -651,31 +651,36 @@ mfxStatus VideoDECODEMPEG2::DecodeFrameCheck(mfxBitstream* bs, mfxFrameSurface1*
     MFX_CHECK(m_core, MFX_ERR_UNDEFINED_BEHAVIOR);
     MFX_CHECK(m_decoder, MFX_ERR_NOT_INITIALIZED);
 
-    MFX_CHECK_NULL_PTR2(surface_work, surface_out);
+    MFX_CHECK_NULL_PTR1(surface_out);
 
     mfxStatus sts = bs ? CheckBitstream(bs) : MFX_ERR_NONE;
```
mgonchar: I don't like this construction
It should be
```
mfxStatus sts = MFX_ERR_NONE;
if (bs)
{
    sts = CheckBitstream(bs);
    MFX_CHECK_STS(sts);
}
```

But since it wasn't introduced by you I'm not forcing you to change it

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1030,9 +1036,16 @@ void VideoDECODEMPEG2::FillOutputSurface(mfxFrameSurface1* surface_work, mfxFram
 {
     const auto fd = frame->GetFrameData();
```
mgonchar: I think you should check input pointers which you are using

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -772,7 +777,8 @@ mfxStatus VideoDECODEMPEG2::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surf
 
     std::unique_ptr<TaskInfo> info(new TaskInfo); // to prevent memory leak
 
-    info->surface_work = GetOriginalSurface(surface_work);
+    if (surface_work)
+        info->surface_work = GetOriginalSurface(surface_work);
```
dmitryermilov: info->surface_work = surface_work ? GetOriginalSurface(surface_work) : nullptr;

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -651,31 +651,36 @@ mfxStatus VideoDECODEMPEG2::DecodeFrameCheck(mfxBitstream* bs, mfxFrameSurface1*
...
     MFX_CHECK(m_decoder, MFX_ERR_NOT_INITIALIZED);
 
-    MFX_CHECK_NULL_PTR2(surface_work, surface_out);
+    MFX_CHECK_NULL_PTR1(surface_out);
 
     mfxStatus sts = bs ? CheckBitstream(bs) : MFX_ERR_NONE;
     MFX_CHECK_STS(sts);
 
-    if (m_opaque)
+    if (surface_work != nullptr)
```
dmitryermilov: Please add a comment in code what you want to achieve. For future readers

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1030,9 +1043,10 @@ void VideoDECODEMPEG2::FillOutputSurface(mfxFrameSurface1* surface_work, mfxFram
 {
     const auto fd = frame->GetFrameData();
```
mgonchar: at his point you used ```frame``` which can be null (because it is input you can't rely on outside code). As alternative you can use reference here

_studio/mfx_lib/decode/mpeg2/src/mfx_mpeg2_decode.cpp
```diff
@@ -1030,9 +1043,10 @@ void VideoDECODEMPEG2::FillOutputSurface(mfxFrameSurface1* surface_work, mfxFram
...
 
-    mfxVideoParam vp;
-    *surf_out = m_allocator->GetSurface(fd->GetFrameMID(), surface_work, &vp);
-     if (m_opaque)
+    *surf_out = (surface_work != nullptr) ? m_allocator->GetSurface(fd->GetFrameMID(), surface_work, &m_video_par) :
```
mgonchar: at least ```GetSurfaceByIndex``` can return null, so you need to check ```*surf_out``` pointer before dereferencing

Issue #970, opened 2018-12-05
ffmpeg mjpeg encode: set FrameType to MFX_FRAMETYPE_I 


fulinjie:
In mjpeg_qsv: 
pict_type was decided by bs->FrameType and it equals AV_PICTURE_TYPE_NONE.

In native mjpeg encode:
pict_type was set to AV_PICTURE_TYPE_I in avcodec_encode_video2() after calling ff_mpv_encode_picture();

So, I think it may be better to set bs Frame Type to MFX_FRAMETYPE_I in MSDK than setting pict_type to AV_PICTURE_TYPE_NONE in ff_qsv_encode()?

Details in https://trac.ffmpeg.org/ticket/7593

dvrogozh:
@Alexandr-Konovalov: I thought, maybe you will take a look since you work with jpeg now?

PR #1129, opened 2019-01-22
mjpege: change frame type to MFX_FRAMETYPE_I from MFX_FRAMETYPE_UNKNOWN

Alexandr-Konovalov:
Fixes: #970

### DIFF
--- a/_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
+++ b/_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
@@ -659,10 +659,10 @@ mfxStatus MFXVideoENCODEMJPEG::MJPEGENCODECompleteProc(void *pState, void *pPara
     }
 
     pTask->bs->DataLength += (mfxU32)(pDataOut.GetDataSize());
-    
+
     obj.m_frameCount++;
 
-    if(pTask->bs->DataLength - pTask->m_initialDataLength) 
+    if(pTask->bs->DataLength - pTask->m_initialDataLength)
     {
         obj.m_encodedFrames++;
         obj.m_totalBits += (pTask->bs->DataLength - pTask->m_initialDataLength) * 8;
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
```diff
@@ -659,10 +659,11 @@ mfxStatus MFXVideoENCODEMJPEG::MJPEGENCODECompleteProc(void *pState, void *pPara
     }
 
     pTask->bs->DataLength += (mfxU32)(pDataOut.GetDataSize());
-    
+    pTask->bs->FrameType = MFX_FRAMETYPE_I;
```
dvrogozh: I always thought that such frame encoding characteristics as gop structure should be assigned within EncodeFrameAsync and only data is filled once SyncOperation is completed. From this perspective this assignment looks incorrect - the one which you did in EncodeFrameCheck should be enough.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_hw.cpp
```diff
@@ -972,6 +972,7 @@ mfxStatus MFXVideoENCODEMJPEG_HW::EncodeFrameCheck(
 
     bs->TimeStamp = surface->Data.TimeStamp;
     bs->DecodeTimeStamp = surface->Data.TimeStamp;
+    bs->FrameType = MFX_FRAMETYPE_I;
```
dvrogozh: What about SW path? And are there a shared SW/HW code where we can do that?

Alexandr-Konovalov: SW path is covered by change in mfx_mjpeg_encode.cpp. I found no right place to put `MFX_FRAMETYPE_I` assignment that is called from `EncodeFrameAsync` and shared by HW and SW pathes.

_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
Alexandr-Konovalov: Sure, moved to `MFXVideoENCODEMJPEG::EncodeFrameCheck()`.

PR #1130, opened 2019-01-22
Fix ROI task settings to delta QP ROI in CQP case

vshampor:


akokoshn:
looks OK, but not sure MFX_ROI_MODE_QP_DELTA is right value for RC = MFX_RATECONTROL_CQP or not

vshampor:
> 
> 
> looks OK, but not sure MFX_ROI_MODE_QP_DELTA is right value for RC = MFX_RATECONTROL_CQP or not

The other option is the "priority" ROI mode, which is not supported by the driver. Furthermore, I have found no documentation on how it's actually supposed work, but probably the priority mode is only relevant for BRC types other than CQP.

PR #1132, opened 2019-01-23
Static analysis issues fixes

dmitrybrazhkin:

### DIFF
--- a/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
@@ -786,7 +786,11 @@ mfxStatus VideoDECODEH265::GetDecodeStat(mfxDecodeStat *stat)
     m_stat.NumSkippedFrame = m_pH265VideoDecoder->GetSkipInfo().numberOfSkippedFrames;
     m_stat.NumCachedFrame = 0;
 
-    H265DecoderFrame *pFrame = m_pH265VideoDecoder->GetDPBList()->head();
+    H265DBPList *dpb = m_pH265VideoDecoder->GetDPBList();
+    if (!dpb)
+        return MFX_ERR_UNDEFINED_BEHAVIOR;
+
+    H265DecoderFrame *pFrame = dpb->head();
     for (; pFrame; pFrame = pFrame->future())
     {
         if (!pFrame->wasOutputted() && !isAlmostDisposable(pFrame))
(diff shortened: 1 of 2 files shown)


_studio/shared/umc/codec/h264_dec/src/umc_h264_mfx_supplier.cpp
```diff
@@ -388,7 +388,7 @@ Status MFXTaskSupplier::DecodeSEI(NalUnit *nalUnit)
         bitStream.Reset((uint8_t*)swappedMem.GetPointer(), (uint32_t)swappedMem.GetDataSize());
 
         NAL_Unit_Type nal_unit_type;
-        uint32_t nal_ref_idc;
+        uint32_t nal_ref_idc = 0;
```
dvrogozh: Why we don't initialize nal_unit_type then? And what we actually get in the bitStream.GetNALUnitType(nal_unit_type, nal_ref_idc); - both values or only one?

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -786,7 +786,11 @@ mfxStatus VideoDECODEH265::GetDecodeStat(mfxDecodeStat *stat)
     m_stat.NumSkippedFrame = m_pH265VideoDecoder->GetSkipInfo().numberOfSkippedFrames;
     m_stat.NumCachedFrame = 0;
 
-    H265DecoderFrame *pFrame = m_pH265VideoDecoder->GetDPBList()->head();
+    H265DBPList *dpb = m_pH265VideoDecoder->GetDPBList();
+    if (!dpb)
+        return MFX_ERR_NULL_PTR;
```
dmitry-gurulev: `MFX_ERR_NULL_PTR` is described as 
> NULL pointer in the input or output arguments
Having null pointer for DPB here looks to be unexpected, so I think `MFX_ERR_UNDEFINED_BEHAVIOR` status would be better.

dmitrybrazhkin: Fixed

_studio/shared/umc/codec/h264_dec/src/umc_h264_mfx_supplier.cpp
dvrogozh: I would either initialize both or have error check for GetNALUnitType.

dmitrybrazhkin: Done

PR #1140, opened 2019-01-25
[VP9D] decoding DRC with external system memory allocator

IlyaSergeyev:
MFX_MEMTYPE_SYSTEM_MEMORY does not support adaptive mode
Decoder uses internal frames for decoding in that case, that are
need to be re-allocated simultaneously with external frames

MDP-50893

Original #1124 was splitted into #1141 and this one

Test cases

./mfx_player -i Stress_VP9_FC2p1ss444_1920x1080_250_extra_stress_2.2.0.vp9 -o:i444 -hw -sys -async 5 -crc -async_depth 5 -vp9_drc -gpu_copy 1 -adaptive_playback

PR #1124, opened 2019-01-18
[VP9D]  show_existing_frame with DRC

IlyaSergeyev:
1) With MFX_MEMTYPE_EXTERNAL_FRAME and MFX_MEMTYPE_DXVA2_DECODER_TARGET
copying MFX_MEMTYPE_EXTERNAL_FRAME frames fails without locking frame
2) We should re-allocate internal D3D decoding surface in same way as external output surface in case of REALLOC_SURFACE 

MDP-50343

Test:
./mfx_player -i Stress_VP9_FC2p1ss444_1920x1080_250_extra_stress_2.2.0.vp9 -o:i444 -hw -d3d -async 5 -crc -async_depth 5 -vp9_drc

./mfx_player -i Stress_VP9_FC2p1ss444_1920x1080_250_extra_stress_2.2.0.vp9 -o:i444 -hw -sys -async 5 -crc -async_depth 5 -vp9_drc -gpu_copy 1 -adaptive_playback

PR #1141, opened 2019-01-25
[VP9D] show existing frame with external vaapi frame failed

IlyaSergeyev:
show_existing_flag with external vaapi allocator case
copying frame with DoFastCopyWrapper without locking temporary source fails

(Original #1124 was splitted into #1140 and this one)

MDP-51343

Testcase:

./mfx_player -i Stress_VP9_FC2p1ss444_1920x1080_250_extra_stress_2.2.0.vp9 -o:i444 -hw -d3d -async 5 -crc -async_depth 5 -vp9_drc

### DIFF
--- a/_studio/shared/src/libmfx_core_vaapi.cpp
+++ b/_studio/shared/src/libmfx_core_vaapi.cpp
@@ -916,8 +916,19 @@ VAAPIVideoCORE::DoFastCopyWrapper(
         {
             sts = GetExternalFrameHDL(srcMemId, &srcHandle);
             MFX_CHECK_STS(sts);
-
-            srcTempSurface.Data.MemId = srcHandle;
+            // VP9D show_existing_frame with external D3D allocator case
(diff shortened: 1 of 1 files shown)


IlyaSergeyev:
> We don't have d3d windows support here as of now. What is this change about?
Windows? In Linux without locking MFX_MEMTYPE_DXVA2_DECODER source copying (that occurs when VP9D meet show_existing_frame flag) fails. I did not dare to remove the code completely, so I left ifdefs.

dvrogozh:
This patch looks incorrect.  The code which you remove looks reasonable. Please, describe failing case. Pay attention what memory types you are dealing with and where they were allocated.

IlyaSergeyev:
> This patch looks incorrect. The code which you remove looks reasonable. Please, describe failing case. Pay attention what memory types you are dealing with and where they were allocated.


Yes, you are right. Using ifdefs here is not good idea. Commit  rewritten with checking memory types.

PR #1142, opened 2019-01-28
Clamp utility function

### LINKED ISSUE #926: Eliminate multipile self-written std::clamp
Please, substitute multiple self-defined variants of std::clamp (https://en.cppreference.com/w/cpp/algorithm/clamp). Since it comes from c++17, implement single mfx::clamp version and comment appropriately. Examples:
`_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_utils.h:78:template<class T> inline T Clip3(T min, T max, T x) { return Min(Max(min, x), max); }
_studio/mfx_lib/decode/vp8/include/mfx_vp8_dec_decode_vp8_defs.h:155:#define clip(x) MFX_MIN(256, MFX_MAX(x, -256))
_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h:44:#define CLIPVAL(MINVAL, MAXVAL, VAL) (MFX_MAX(MINVAL, MFX_MIN(MAXVAL, VAL)))
_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_brc.cpp:30:#define CLIPVAL(MINVAL, MAXVAL, VAL) (MFX_MAX(MINVAL, MFX_MIN(MAXVAL, VAL)))
_studio/mfx_lib/mctf_package/mctf/include/mctf_common.h:130:#define CLIPVAL(VAL, MINVAL, MAXVAL) MAX(MINVAL, MIN(MAXVAL, VAL))
`
Mind the list can be partial. I would also address this before merging fix for #858.

mgonchar:
- Introduce mfx::clamp
- Clean up code from multiple self-written clamps

Fixes #926

### DIFF
--- a/_studio/mfx_lib/vpp/src/mfx_detail_enhancement_vpp.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_detail_enhancement_vpp.cpp
@@ -29,8 +29,6 @@
 #define ALIGN16(SZ) (((SZ + 15) >> 4) << 4) // round up to a multiple of 16
 #endif
 
-#define VPP_RANGE_CLIP(val, min_val, max_val)  MFX_MAX( MFX_MIN(max_val, val), min_val )
-
 #define VPP_DETAIL_GAIN_MIN      (0)
 #define VPP_DETAIL_GAIN_MAX      (63)
 #define VPP_DETAIL_GAIN_MAX_REAL (63)
@@ -57,7 +55,7 @@ mfxStatus MFXVideoVPPDetailEnhancement::Query( mfxExtBuffer* pHint )
 
     if( pParam->DetailFactor > VPP_DETAIL_GAIN_MAX_USER_LEVEL )
     {
-        VPP_RANGE_CLIP(pParam->DetailFactor, VPP_DETAIL_GAIN_MIN, VPP_DETAIL_GAIN_MAX_USER_LEVEL);
+        pParam->DetailFactor = VPP_DETAIL_GAIN_MAX_USER_LEVEL;
 
         sts = MFX_WRN_INCOMPATIBLE_VIDEO_PARAM;
     }
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -5095,9 +5095,9 @@ mfxU64 make_back_color_yuv(mfxU16 bit_depth, mfxU16 Y, mfxU16 U, mfxU16 V)
 
     return
         ((mfxU64) max_val << 48) |
-        ((mfxU64) VPP_RANGE_CLIP(Y, (16 << shift), (235 << shift))     << 32) |
-        ((mfxU64) VPP_RANGE_CLIP(U, (16 << shift), (240 << shift))     << 16) |
-        ((mfxU64) VPP_RANGE_CLIP(V, (16 << shift), (240 << shift))     <<  0);
+        ((mfxU64) mfx::clamp<mfxI32>(Y, (16 << shift), (235 << shift)) << 32) |
+        ((mfxU64) mfx::clamp<mfxI32>(U, (16 << shift), (240 << shift)) << 16) |
+        ((mfxU64) mfx::clamp<mfxI32>(V, (16 << shift), (240 << shift)) <<  0);
 }
 
 inline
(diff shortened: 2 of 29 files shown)


_studio/shared/include/mfx_utils.h
```diff
@@ -125,6 +127,18 @@ template<class T> inline T AlignValue(T value, mfxU32 alignment)
     return static_cast<T>((value + alignment - 1) & ~(alignment - 1));
 }
 
+namespace mfx
+{
+// TODO: switch to std::clamp when C++17 support will be enabled
+
+// Clip value v to range [lo, hi]
+template<class T>
+const T clamp(const T v, const T lo, const T hi)
```
dmitry-gurulev: Why not `const&` ?

lakulako: Suppose it is not good for rvalues and constants: clamp(x+1, 0, 2)

lakulako: What for const for passes by value parameters?

mgonchar: Rvalue is not a problem, it will move-construct element which would be used as reference. the main reason is that this code always used for built-in types, so no need to introduce additional indirection with references. If there were some nontrivial classes for which we need ```clamp```, the right solution still will be to use SFINAE to redirect calls for built-in types to implementation with current signature

_studio/mfx_lib/vpp/src/mfx_detail_enhancement_vpp.cpp
```diff
@@ -57,7 +55,7 @@ mfxStatus MFXVideoVPPDetailEnhancement::Query( mfxExtBuffer* pHint )
 
     if( pParam->DetailFactor > VPP_DETAIL_GAIN_MAX_USER_LEVEL )
     {
-        VPP_RANGE_CLIP(pParam->DetailFactor, VPP_DETAIL_GAIN_MIN, VPP_DETAIL_GAIN_MAX_USER_LEVEL);
+        pParam->DetailFactor = VPP_DETAIL_GAIN_MAX_USER_LEVEL;
```
lakulako: differs too

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5109,9 +5109,9 @@ mfxU64 make_back_color_argb(mfxU16 bit_depth, mfxU16 R, mfxU16 G, mfxU16 B)
     mfxU64 const max_val = (1 << bit_depth) - 1;
 
     return ((mfxU64) max_val << 48) |
-           ((mfxU64) VPP_RANGE_CLIP(R, 0, max_val) << 32) |
-           ((mfxU64) VPP_RANGE_CLIP(G, 0, max_val) << 16) |
-           ((mfxU64) VPP_RANGE_CLIP(B, 0, max_val) << 0);
+           ((mfxU64) mfx::clamp<mfxU64>(R, 0, max_val) << 32) |
+           ((mfxU64) mfx::clamp<mfxU64>(G, 0, max_val) << 16) |
+           ((mfxU64) mfx::clamp<mfxU64>(B, 0, max_val) << 0);
 }
```
lakulako: No need to cast to mfxU64 in return computations.

_studio/mfx_lib/vpp/src/mfx_detail_enhancement_vpp.cpp
mgonchar: see the condition of above ```if```

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
mgonchar: right

_studio/shared/include/mfx_utils.h
mgonchar: actually, ```const``` can be dropped, yes

dmitry-gurulev: I'd rather use standard signature to avoid surprises when "C++17 support will be enabled"

lakulako: @dmitry-gurulev , agree. Lets follow C++17 prototype.

dvrogozh:
CI don't pass. Please, fix.

_studio/shared/include/mfx_utils.h
eistarov: c++17 version is much better for heavy classes with custom comparing operator.
If we are introducing mfx-global utility function to emulate it, lets be consistent and replicate its behavior properly.
And please add this version, as well:
template<class T, class Compare>
constexpr const T& clamp( const T& v, const T& lo, const T& hi, Compare comp );

mgonchar: ok, let's do it this way

PR #1157, opened 2019-02-01
h264d: Implement output decoded frames regarding num_reorder_frames

DenWolf:
Flag SPS.VUI.num_reorder_frames may reduce latency
on outputting decoded frames, this flag is presented
when SPS.VUI.bitstream_restriction_flag == 1

This feature is needed on Android O MR1


DenWolf:
> @DenWolf, are u sure that this feature need to be turned off for any Android O, not for certain customer? There are many streams where this syntax element equal to 0 but reordering actually exists.

@dmitry-gurulev , good point - I remake the patch - now this feature will be enabled by "make" cmd for need Android versions and projects

DenWolf:
> I think it should be always enabled, without ifdefs. It would affect output order only if a stream contains incorrect NumReorderedFrames. Decoders have to follow it.

@lakulako , according to comment from @dmitry-gurulev 
> There are many streams where this syntax element equal to 0 but reordering actually exists.

and my comment in the code : https://github.com/Intel-Media-SDK/MediaSDK/pull/1157/files#diff-205bc80d363db7407fbe2ddd4136cb43R319

we have a risks to break decoding functionality for some old non-specification-compatible videos. So, for now we decided to hide this feature under define and enable it for special project on demand

lakulako:
OK with define. Then I would propose to reduce use of the define in the code. I think 
"#ifdef OUTPUT_FRAMES_REGARDING_SPS_NUM_REORDER_FRAMES
                || countNumDisplayable > view.maxNumReorderFrames
#endif"
should be the single place in the code where define is used.
maxNumReorderFrames declaration should keep current comments and note that it is used with the define.

DenWolf:
@lakulako , when I created this change I thought to make it right as you suggested, but I decided to hide under define every meaning of new field "maxNumReorderFrames" and I explain why:

If I hide under define only checking (agree, it makes sense) -> when we will not use this define/feature we will get new member "maxNumReorderFrames" in the class which doesn't do anything, just exists. Also, in my opinion, hiding by define every using of "maxNumReorderFrames" is more representative for another programmers when they will look at the code - filed "maxNumReorderFrames" is needed only when enabled define/feature.
As result, I suggest to keep this change -> hide under define every using "maxNumReorderFrames".

@dmitry-gurulev , @onabiull , @lakulako - what do you think ?

lakulako:
maxNumReorderFrames is related to h264 specification, not bad to have it ready.
ifdefs are heavier than the code enclosed.
Code inside ifdefs is spoiling during the time.

dmitry-gurulev:
+1 for @lakulako comment
> maxNumReorderFrames is related to h264 specification, not bad to have it ready.

It need to be mentioned that accord. to spec. _maxNumReorderFrames_ have to be 0 when _constraint_set3_flag_ == 1 and profile is one of 
- H264_PROFILE_CAVLC444_INTRA
- H264_PROFILE_SCALABLE_HIGH
- H264_PROFILE_HIGH
- H264_PROFILE_HIGH10
- H264_PROFILE_HIGH422
- H264_PROFILE_HIGH444_PRED
(actually, _Intra_ profiles)

IMO, prev. version _ENABLE_MAX_NUM_REORDER_FRAMES_OUTPUT_ was better - closer to spec. and clean enough for understanding.

DenWolf:
@dmitry-gurulev , @lakulako - thank you for review! I remade the patch regarding your comments and our discussion, could you please make a review again ?

PR #1161, opened 2019-02-04
[test_monitor] intel_gt(dev_id) function fixed

andreyor:
Earlier scheme of decoding device id to get GT size
doesn't work on some platforms anymore,
so they are were determined in platform tables.

Issue: MDP-51279


samples/metrics_monitor/test/igt_load.cpp
```diff
@@ -174,8 +174,18 @@ unsigned intel_gen(uint16_t devid)
 
 unsigned intel_gt(uint16_t devid)
 {
+    const struct intel_device_info *devinfo = intel_get_device_info(devid);
     unsigned mask = intel_gen(devid);
```
mgonchar: move declaration to line 188 to be closer to usage

andreyor: Done

PR #1162, opened 2019-02-05
AVC FEI DSO: fixed high CPU usage of std::copy

mgonchar:
Moved to structure-wise copy instead of byte-wise

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -826,8 +826,8 @@ Status PackerVA::QueryStreamOut(H264DecoderFrame* pFrame)
     VM_ASSERT(dynamic_cast<FEIVideoAccelerator*>(m_va) &&
               "VA should be [FEIVideoAccelerator] if [streamout] is enabled");
 
-    FEIVideoAccelerator* fei_va =
-        static_cast<FEIVideoAccelerator*>(m_va);
+    if (!pFrame)
+        return UMC_ERR_FAILED;
 
     FrameData const* fd = pFrame->GetFrameData();
     VM_ASSERT(fd);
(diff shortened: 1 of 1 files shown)


dmitry-gurulev:
@mgonchar, quite strange that struct wise copy gives u improvement. In case of byte-wise copying we should fall into builtin `memmove` wile for struct-wise we'll fall into explicit loop w/ per-struct member-wise copying through compiler's generated assignment operator of `mfxFeiDecStreamOutMBCtrl`. The latter is because `std::is_trivially_copyable` gives `false` for that type.

dmitry-gurulev:
@mgonchar 
>`std::is_trivially_copyable` gives `false` for that type

double check this - it returns `true` for `mfxFeiDecStreamOutMBCtrl`, so `std::copy` for it should also fall into `memcpy`. So, there should not be differences w/ byte-wise copying.

mgonchar:
@dmitry-gurulev well, from my understanding, ```std::copy``` should be not less efficient than ```memcpy```. Also from my understanding the more you know about types you are copying the more efficient this copy should be. Since ```memcpy``` doesn't actually know size of blocks and it's actual alignment (since ```memcpy``` accepts ```void*```) and ```std::copy``` knows everything about type it could apply more aggressive optimizations. But let's focus on background of this fix. Initial problem found in validation described in ticket MDP-53044. CPU usage increased from 26% to 31% after ```memcpy``` to ```std::copy``` change. After current fix it returned to previous 26% (this is confirmed by our validation). I don't really know the source of such behavior, my assumption that it is because ```char*``` pointers is a special case for compiler (but using non-standardized ```restrict``` key-word which is supported by gcc didn't help). I examined the assembly produced by gcc 6.3.1 and in case of byte-wise ```std::copy``` it produces tons of instructions and for struct-wise only a few. Neither of variants with ```std::copy``` had fallen to ```memmove``` or ```memcpy```.

mgonchar:
@dmitry-gurulev of course I was talking about trivially copyable types. POD structs without user defined constructors should be trivially copyable.

dmitry-gurulev:
@mgonchar  
> Neither of variants with ```std::copy``` had fallen to ```memmove``` or ```memcpy```.

Looks like we see inlined `memcpy` version. Just a version - in case of inlined `memcpy` compiler knows certain type anyway and it is able to generate more optimal code for struct-wise.

mgonchar:
For those who still interested why we had high CPU usage with initial variant.
After investigation I found an answer.
Initial type of ```src``` is ```const mfxU8*``` which is ```const unsigned char*``` but ```dst``` type is ```char*```. Now looking at stl realization by gcc we see following https://github.com/gcc-mirror/gcc/blob/866cd688d1b72b0700a7e001428bdf2fe73fbf64/libstdc%2B%2B-v3/include/bits/stl_algobase.h#L397 
and
https://github.com/gcc-mirror/gcc/blob/866cd688d1b72b0700a7e001428bdf2fe73fbf64/libstdc%2B%2B-v3/include/bits/stl_algobase.h#L329
Now its clear why compiler fallback to loop assignment, while simple change of one pointer type (```char``` => ```uchar```) forces to fallback to memmove. Performance of byte-wise (with same pointers type) and structure-wise is on par.

PR #1163, opened 2019-02-05
[VPP] UYVY enable fix

IlyaSergeyev:
This commit fixes previous (UYVY was disabled for any platforms except ICL by mistake)


IlyaSergeyev:
> Have you uploaded changes to file mfx_vp9_dec_decode_hw.cpp by mistake?
yes, it already updated

wakuken:
Is UYVY format enabled for only VPP's input on any platforms?
Does the UYVY output does not work?

IlyaSergeyev:
> Can it be derived from caps or whatever instead of hardcoding?

Yes, all calls of CheckFrameInfo function are from Init, Reset and QueryIOSurf. Checking caps will not affect performance.

lakulako:
There are no changesfor YUV2 as stated in commit message.

onabiull:
Please fix commit message

IlyaSergeyev:
Commit message fixed

Issue #1165, opened 2019-02-06
Use installed googletests/gmcok as default, do not force installing


chivakker:
GoogleTests integration in MediaSDK is broken.

Do not modify filesystem with googletest installation.  Use pkgconfig gtest and pkgconfig gmock to find if system already supports.  If not then force user to enable it.  Install it in a particular prefix as you don't know if user wants to install and suggest installing it in a temp folder.

dvrogozh:
Ok, I reproduced it with the following on CentOS 7:
```
# scl enable devtoolset-7 -- cmake -DCMAKE_INSTALL_PREFIX=`pwd`/install -DBUILD_TESTS=ON ..
# make
# make install
# tree `pwd`/install
install
├── include
│   ├── gmock
│   │   ├── gmock-actions.h
│   │   ├── gmock-cardinalities.h
│   │   ├── gmock-generated-actions.h
│   │   ├── gmock-generated-actions.h.pump
│   │   ├── gmock-generated-function-mockers.h
│   │   ├── gmock-generated-function-mockers.h.pump
│   │   ├── gmock-generated-matchers.h
│   │   ├── gmock-generated-matchers.h.pump
│   │   ├── gmock-generated-nice-strict.h
│   │   ├── gmock-generated-nice-strict.h.pump
│   │   ├── gmock.h
│   │   ├── gmock-matchers.h
│   │   ├── gmock-more-actions.h
│   │   ├── gmock-more-matchers.h
│   │   ├── gmock-spec-builders.h
│   │   └── internal
│   │       ├── custom
│   │       │   ├── gmock-generated-actions.h
│   │       │   ├── gmock-generated-actions.h.pump
│   │       │   ├── gmock-matchers.h
│   │       │   ├── gmock-port.h
│   │       │   └── README.md
│   │       ├── gmock-generated-internal-utils.h
│   │       ├── gmock-generated-internal-utils.h.pump
│   │       ├── gmock-internal-utils.h
│   │       └── gmock-port.h
│   ├── gtest
│   │   ├── gtest-death-test.h
│   │   ├── gtest.h
│   │   ├── gtest-message.h
│   │   ├── gtest-param-test.h
│   │   ├── gtest-param-test.h.pump
│   │   ├── gtest_pred_impl.h
│   │   ├── gtest-printers.h
│   │   ├── gtest_prod.h
│   │   ├── gtest-spi.h
│   │   ├── gtest-test-part.h
[...]

PR #1166, opened 2019-02-06
build: do not install gtest

dvrogozh:
Fixes: #1165

cmake option() rewrites set() without FORCE flag which we tried
to use to instuct gtest to avoid installation path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,8 +111,12 @@ endif()
 if (BUILD_TESTS)
   enable_testing()
 
+  # Our intent is to use gtest for building, we do not want to install it.
+  # To achieve that we need to force set INSTALL_GTEST=OFF otherwise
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -111,8 +111,12 @@ endif()
 if (BUILD_TESTS)
   enable_testing()
 
+  # Our intent is to use gtest for building, we do not want to install it.
+  # To achieve that we need to force set INSTALL_GTEST=OFF otherwrite
```
vshampor: otherwise

CMakeLists.txt
```diff
@@ -111,8 +111,12 @@ endif()
...
+  # according to cmake rules it will be overwriten to option() default
```
vshampor: overwritten by

dvrogozh:
@vshampor : updated

PR #925, opened 2018-11-22
Remove minmax

mgonchar:
This pull request clean up MSDK code from MIN / MAX macro and replaces them with std::min / max

Fixes #858

PR #1171, opened 2019-02-07
h265d: Use int32_t instead of int for style alignment

mgonchar:


_studio/shared/umc/codec/h265_dec/src/umc_h265_bitstream_headers.cpp
```diff
@@ -396,10 +396,10 @@ void H265HeadersBitstream::xDecodeScalingList(H265ScalingList *scalingList, unsi
 {
     VM_ASSERT(scalingList);
 
-    int i,coefNum = MFX_MIN(MAX_MATRIX_COEF_NUM,(int)g_scalingListSize[sizeId]);
-    int nextCoef = SCALING_LIST_START_VALUE;
+    int32_t i,coefNum = MFX_MIN(MAX_MATRIX_COEF_NUM,(int32_t)g_scalingListSize[sizeId]);
```
dvrogozh: When do you plan to return to #925. You eventually multiply conflicts...

mgonchar: I'm almost finished reworking #925 as background activity and now it is in a good shape. I need to clarify one piece of code and then I will upload new version. I know about introduced conflict by this patch, not a big deal to rebase)

PR #1177, opened 2019-02-08
Fix truncation

mgonchar:
No sense of taking min with 0 for unsigned types (difference of two ```size_t``` is still ```size_t```)


mgonchar:
> 
> 
> It makes sense to also re-name _sz_ to something like _sizeToAdd_ (or _szToAdd_)

Done

PR #1179, opened 2019-02-08
gtest,gmock: force building static libraries

vshampor:
GTest and GMock will be built as static even if the variable
-DBUILD_SHARED_LIBS:BOOL is specified as ON via CMake CLI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,15 @@ if (BUILD_TESTS)
   # which might be ON.
   set(INSTALL_GTEST OFF CACHE BOOL "Do _not_ install gtest" FORCE)
 
+  # Make sure that gtest/gmock is built as static libs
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -117,8 +117,16 @@ if (BUILD_TESTS)
   # which might be ON.
   set(INSTALL_GTEST OFF CACHE BOOL "Do _not_ install gtest" FORCE)
 
+  # Make sure that gtest/gmock is built as static libs
+  set(BUILD_SHARED_LIBS OFF CACHE BOOL "")
```
dvrogozh: Looks ugly:). Can we consider/try use these instead:
```
-DGTEST_CREATE_SHARED_LIBRARY=0 
-DGTEST_LINKED_AS_SHARED_LIBRARY=0 
```
I would like to avoid rewriting/changing widely used cmake-defined variables as much as possible.

PR #1181, opened 2019-02-11
samples: avoid loading default plugin for HEVCe on Android

akokoshn:
-On Android implementaion of HEVC encoder is placed in libmfx
 That's why we don't need default plugins for this codec

### DIFF
--- a/samples/sample_common/src/plugin_utils.cpp
+++ b/samples/sample_common/src/plugin_utils.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************\
-Copyright (c) 2005-2018, Intel Corporation
+Copyright (c) 2005-2019, Intel Corporation
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/plugin_utils.cpp
```diff
@@ -142,8 +142,12 @@ const mfxPluginUID & msdkGetPluginUID(mfxIMPL impl, msdkComponentType type, mfxU
         case MSDK_VENCODE:
             switch(uCodecid)
             {
+// On Android implementaion of HEVC encoder is placed in libmfx
+// That's why we don't need default plugins for this codec
+#if !defined(ANDROID)
```
dvrogozh: Isn't that the case now for non-Android as well? I thought this plugin is just a stub. In this case, why differentiate?

akokoshn: @dvrogozh , the reason of difference - different build configuration: for android build no target of default plugin, but for another system it still exist.
But generally you are right, if run HW HEVC encode on linux without plugin - should works OK.

akokoshn: by the way same actual for line 129

dvrogozh: That's a question what we want to demonstrate in our samples. If we consider start switching users to work with hw codecs right from the library (which makes sense since it avoids separate path to load plugins), we better provide common code for all platforms.

PR #1182, opened 2019-02-11
vp9d: fix frame reference counting

maksenox:
After the new frame was decoded mark it as fully decoded frame.
And DecreaseReference function call will be made from decoder
synchronous part (DecodeFrameAsync).

### DIFF
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -181,6 +181,7 @@ private:
                 }
             }
         }
+        VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(frame.currFrame) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
     }
 public:
     FrameStorage(mfx_UMC_FrameAllocator *frameAllocator):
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -181,6 +181,7 @@ class FrameStorage
                 }
             }
         }
+        VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(frame.currFrame) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
```
mgonchar: please do not put function call to macro. Currently this code is ok, but what will happen, if somebody will change the macro? It could lead to double call of ```DecreaseReference``` and the issue would be hard to find. I suggest you to rewrite current code as following:
```
auto umc_sts = m_frameAllocator->DecreaseReference(frame.currFrame);
VP9_CHECK_AND_THROW(umc_sts == UMC::UMC_OK, MFX_ERR_UNKNOWN);
```

akokoshn: @maksenox , could you please check  decoding with system memory, I have seen an problem on this use case when tried to move decrease references of decoded frame to synchronous part

maksenox: Yes, there is a problem, moving the reference counting to synchronous part cause different free frame order (the order in which player takes free surfaces and puts them into decoder). As the result, vaPutImage function receives different resolution of the source and destination frames. vaPutImage has a bug copying such frames and DRC streams on system memory on Linux may fail. That is just a coincidence that with this fix there are a little bit more failed tests.

maksenox:
New pull request was created #1189

PR #1183, opened 2019-02-11
[VP9D] Check if FourCC of frame different from video params

IlyaSergeyev:
MDP-50900


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -1014,6 +1014,12 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     sts = CheckBitstream(bs);
     MFX_CHECK_STS(sts);
 
+    if (m_vPar.mfx.FrameInfo.FourCC != surface_work->Info.FourCC)
```
dmitry-gurulev: Maybe it would be  better to check this after _UpdateVideoParam_?

IlyaSergeyev: yes

PR #1184, opened 2019-02-11
jpeg: width and height were mixed up during vpp rotation

Alexandr-Konovalov:
Issue: MDP-52482

### DIFF
--- a/_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
+++ b/_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
@@ -217,8 +217,23 @@ mfxStatus mfx_UMC_FrameAllocator_D3D_Converter::StartPreparingToOutput(mfxFrameS
                 isD3DToSys = !((pOpaqAlloc->Out.Type & MFX_MEMTYPE_SYSTEM_MEMORY) == 0);
             }
 
-            pSrc.Info.CropW = surface_work->Info.CropW;
-            pSrc.Info.CropH = surface_work->Info.CropH;
+            /* JPEG standard does not support crops as it is done in AVC, so:
+               - CropX and CropY are always 0,
+               - CropW and CropH represents picture size for current frame (in case of rotation,
+                 surface_work has rotated CropW and CropH),
+               - Width and Height represents surface allocation size (they are initialized
+                 in decoder Init and are correct).
+             */
+            if (par->mfx.Rotation == MFX_ROTATION_0 || par->mfx.Rotation == MFX_ROTATION_180)
+            {
+                pSrc.Info.CropW = surface_work->Info.CropW;
+                pSrc.Info.CropH = surface_work->Info.CropH;
+            }
+            else
+            {
+                pSrc.Info.CropW = surface_work->Info.CropH;
+                pSrc.Info.CropH = surface_work->Info.CropW;
+            }
             sts = DoHwJpegCc(pCc, m_pCore, surface_work, &pSrc, par, isD3DToSys, taskId);
             if (sts < MFX_ERR_NONE)
                 return sts;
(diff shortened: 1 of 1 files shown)


_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -267,10 +275,20 @@ mfxStatus mfx_UMC_FrameAllocator_D3D_Converter::StartPreparingToOutput(mfxFrameS
...
+            if (par->mfx.Rotation == MFX_ROTATION_0 || par->mfx.Rotation == MFX_ROTATION_180)
+            {
+                pSrcTop.Info.CropW = surface_work->Info.CropW;
+                pSrcTop.Info.CropH = surface_work->Info.CropH / 2;
+                pSrcBottom.Info.CropW = surface_work->Info.CropW;
+                pSrcBottom.Info.CropH = surface_work->Info.CropH / 2;
+            }
+            else
+            {
+                pSrcTop.Info.CropW = surface_work->Info.CropH;
```
Alexandr-Konovalov: I unable to test this functionality: for some reason vpp first runs field weaving to deinterlace, and does rotation on a separate step.

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -217,8 +217,16 @@ mfxStatus mfx_UMC_FrameAllocator_D3D_Converter::StartPreparingToOutput(mfxFrameS
...
-            pSrc.Info.CropW = surface_work->Info.CropW;
-            pSrc.Info.CropH = surface_work->Info.CropH;
+            if (par->mfx.Rotation == MFX_ROTATION_0 || par->mfx.Rotation == MFX_ROTATION_180)
+            {
+                pSrc.Info.CropW = surface_work->Info.CropW;
+                pSrc.Info.CropH = surface_work->Info.CropH;
+            }
+            else
+            {
+                pSrc.Info.CropW = surface_work->Info.CropH;
```
dvrogozh: You assign CropW = CropH. Why you don't do the same for Width = Height?

dmitrybrazhkin: Let me answer. For some cases including rotation case, JPEG decoder works in 2 steps: decoding step and VPP step. This is not common MSDK VPP but internal JPEG decoder VPP and it is not seen for the application. So we need a set of some internal surfaces to share them between decoder and VPP.  pSrc is one of such surfaces. Width and height of pSrc are initialized in decoder Init and are correct. But in some cases the real size of the frame is lower than allocated surface and so we need to set crop values without decoder re-initialization, that is why we copy crop values here (but not width and height).
In case of rotation, surface_work has rotated CropW and CropH, but pSrc should have not rotated crops. That is why they are swapped here. Width and height should not be changed here

dvrogozh: ok. This explains Width and Height. How about CropX and CropY :)?

dmitrybrazhkin: JPEG standard does not support crops as it is done in AVC, so
- CropX and CropY are always 0,
- CropW and CropH represents picture size for current frame,
- Width and Height represents surface allocation size

dvrogozh: This would make very good comment in the code and avoid similar questions in the future. @Alexandr-Konovalov : can you, please, add it to the code? After that I will be ok with the change.

Alexandr-Konovalov: @dvrogozh Sure.

PR #1185, opened 2019-02-11
vp9d: move release references to async part

akokoshn:
-mfxSurface should be unlocked after SyncOperation(), so
 we can't wait next call DecodeFrameAsync() for free references

### DIFF
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -224,26 +224,16 @@ public:
 
     void DecodeFrame(UMC::FrameMemID frameId)
     {
+        if (frameId >= 0)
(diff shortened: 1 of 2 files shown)


akokoshn:
WIP

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -224,26 +224,16 @@ class FrameStorage
 
     void DecodeFrame(UMC::FrameMemID frameId)
     {
+        if (frameId >= 0)
+            VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(frameId) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
```
mgonchar: please, don't put function calls inside macro

IlyaSergeyev:
Why closed?

akokoshn:
@IlyaSergeyev , it was particular revert of my previously merged change for fix issue from validation, but investigation show that this change useless.
Generally  lock/unlock should be in synchronous part.

PR #1187, opened 2019-02-13
[mpeg2d] Fix for missing slice

dmitryermilov:
In case of absent of a free frame
we need to save current slice for later processing.

### DIFF
--- a/_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
+++ b/_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
@@ -238,6 +238,8 @@ namespace UMC_MPEG2_DECODER
         double                          m_localFrameTime;
 
         std::unique_ptr<Payload_Storage> m_messages;
+
+        std::unique_ptr<MPEG2Slice>     m_lastSlice; // slice which could't be processed
     };
 
     class Payload_Storage
--- a/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
+++ b/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
@@ -132,6 +132,8 @@ namespace UMC_MPEG2_DECODER
 
         Skipping_MPEG2::Reset();
 
+        m_lastSlice.reset();
+
         return UMC::UMC_OK;
     }
 
@@ -327,7 +329,14 @@ namespace UMC_MPEG2_DECODER
     // Find units in new bitstream buffer and process them
     UMC::Status MPEG2Decoder::AddOneFrame(UMC::MediaData * source)
     {
-        // The main parsing loop
+        if (m_lastSlice)
+        {
+            UMC::Status sts = AddSlice(m_lastSlice.release());
+            if (sts == UMC::UMC_ERR_NOT_ENOUGH_BUFFER || sts == UMC::UMC_OK)
+                return sts;
+        }
+
+        // The main processing loop
         for (RawUnit unit = m_splitter.GetUnits(source); unit.begin && unit.end; unit = m_splitter.GetUnits(source))
         {
             switch (unit.type)
@@ -486,7 +495,7 @@ namespace UMC_MPEG2_DECODER
             auto umcRes = StartFrame(slice);
             if (umcRes != UMC::UMC_OK || !m_currFrame)
             {
-                delete slice;
+                m_lastSlice.reset(slice);
                 return umcRes;
             }
         }


_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -238,6 +238,8 @@ namespace UMC_MPEG2_DECODER
         double                          m_localFrameTime;
 
         std::unique_ptr<Payload_Storage> m_messages;
+
+        MPEG2Slice*                     m_lastSlice; // slice which could't be processed
```
mgonchar: I suggest to use ```std::unique_ptr``` instead

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -449,6 +460,8 @@ namespace UMC_MPEG2_DECODER
     // Add a new slice to picture
     UMC::Status MPEG2Decoder::AddSlice(MPEG2Slice * slice)
     {
+        m_lastSlice = 0;
```
mgonchar: ```nullptr```

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -486,7 +495,7 @@ namespace UMC_MPEG2_DECODER
             auto umcRes = StartFrame(slice);
             if (umcRes != UMC::UMC_OK || !m_currFrame)
             {
-                delete slice;
+                m_lastSlice.reset(slice);
```
mgonchar: it is not what it was previously... maybe you wanted to
```
m_lastSlice.reset();
```

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -327,7 +329,14 @@ namespace UMC_MPEG2_DECODER
     // Find units in new bitstream buffer and process them
     UMC::Status MPEG2Decoder::AddOneFrame(UMC::MediaData * source)
     {
-        // The main parsing loop
+        if (m_lastSlice)
+        {
+            UMC::Status sts = AddSlice(m_lastSlice.release());
```
mgonchar: are you sure, didn't you mean ```.get()```? You are loosing ownership of pointer and you need a guarantee that somebody will delete the owned memory inside.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -327,7 +329,14 @@ namespace UMC_MPEG2_DECODER
...
+            if (sts == UMC::UMC_ERR_NOT_ENOUGH_BUFFER || sts == UMC::UMC_OK)
+                return sts;
```
mgonchar: If status is OK we should exit and if error - proceed to execute current function?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: yes, it is not as before. in case of "m_lastSlice.reset();" the slice will be skipped (as without the change)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: I'm 99 % sure:)

mgonchar: ok, I just mention things that looks suspicious :)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
mgonchar: ok

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: Currently AddSlice can return UMC::UMC_OK or UMC::UMC_ERR_NOT_ENOUGH_BUFFER or UMC::UMC_ERR_NOT_ENOUGH_DATA. The 2st two means we should exit from function. The 3rd one means we should continue parsing.

mgonchar: okay

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -486,7 +495,7 @@ namespace UMC_MPEG2_DECODER
             auto umcRes = StartFrame(slice);
             if (umcRes != UMC::UMC_OK || !m_currFrame)
```
dmitry-gurulev: @dmitryermilov ,
If _StartFrame_ may return an error only when we have no free frame (and this means _m_currFrame_ is null) why do we need to check both - error result and _m_currFrame_? I'd rather make _StartFrame_ to return _MPEG2DecoderFrame_ pointer and use it as `m_currFrame = StartFrame(slice)`.

dmitryermilov: @dmitry-gurulev , thanks ! I'll address this in another PR

PR #1188, opened 2019-02-13
HEVC decoder: consume current bs even if decoder has a frame to display

dmitrybrazhkin:


dmitrybrazhkin:
PR need to be validated before merge

PR #1189, opened 2019-02-13
vp9d: fix frame reference counting

maksenox:
Move DecreaseReference to the decoder synchronous part.

### DIFF
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -130,21 +130,45 @@ mfxStatus VideoDECODEVP9_HW::CleanRefList()
     return MFX_ERR_NONE;
 }
 
+static void CheckUmcStatus(const UMC::Status status)
+{
+    VP9_CHECK_AND_THROW((status == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+}
 
 class FrameStorage
 {
 private:
+    class TraceLastUmcError
+    {
+    private:
+        UMC::Status m_status = UMC::UMC_OK;
+    public:
+        TraceLastUmcError& operator=(const UMC::Status status)
+        {
+            if(status != UMC::UMC_OK)
+            {
+                m_status = status;
+            }
+            return *this;
+        }
+
+        operator UMC::Status()
+        {
+            return m_status;
+        }
+    };
+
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -134,17 +134,36 @@ mfxStatus VideoDECODEVP9_HW::CleanRefList()
 class FrameStorage
 {
 private:
+    void CheckUmcStatus(UMC::Status status) {
```
dmitry-gurulev: The method doesn't use any _FrameStorage_ state, why it isn't free function?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -177,10 +195,12 @@ class FrameStorage
                 const UMC::FrameMemID mid = frame.ref_frame_map[ref_index];
                 if (mid >= 0)
                 {
-                    VP9_CHECK_AND_THROW((m_frameAllocator->DecreaseReference(mid) == UMC::UMC_OK), MFX_ERR_UNKNOWN);
+                    status = m_frameAllocator->DecreaseReference(mid);
                 }
             }
         }
+        status = m_frameAllocator->DecreaseReference(frame.currFrame);
+        CheckStatus(status);
```
dmitry-gurulev: if `DecreaseReference` returns an error, what does it mean? It looks like a logical error, we can't do any graceful handling here, just throw an exception immediately, not at the end of method.

maksenox: We had successfully locked tons of resources in LockResources function. If some DecreaseRefernce fails we throw an exception handling the exception we have two major possibilities the first one is to turn off MediaSDK engine at all, the second one tries to proceed work form some point using existing resources (after Reset, Close/Init, etc.), in that case, we can't guarantee the locked surfaces will be propper uninitialized or freed. So my point is clean up all resources before throw exception.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
maksenox: Done

PR #1190, opened 2019-02-13
[RFC] Skip status report

akokoshn:


akokoshn:
DO NOT MERGE! (WIP)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder_va.cpp
```diff
@@ -101,29 +101,32 @@ namespace UMC_MPEG2_DECODER
         // iterate through frames submitted to the driver in decoded order
         for (MPEG2DecoderFrame* frm : decode_queue)
         {
-            uint16_t surfCorruption = 0;
-            auto sts = m_packer->SyncTask(frm, &surfCorruption);
-
             frm->CompleteDecoding(); // complete frame even if there are errors
 
-            if (sts < UMC::UMC_OK)
+            if (m_va->IsUseStatusReport())
```
dmitryermilov: I believe just "if (!m_va->IsUseStatusReport()) return;" will look better, less {} depth.

dmitry-gurulev: @dmitryermilov +1, but `continue` instead of `return`

dmitry-gurulev: Moreover, I'd rather move frame completion (incl. status reporting) into separate private method, something like
```
void MPEG2DecoderVA::CompleteFrame(MPEG2DecoderFrame* frm)
{
    assert(frm);

    frm->CompleteDecoding();

    if (!m_va->IsUseStatusReport())
        return;
    //...
}

bool MPEG2DecoderVA::QueryFrames(MPEG2DecoderFrame& frame)
{
    //...
    std::for_each(std::begin(decode_queue), std::end(decode_queue),
        [this](MPEG2DecoderFrame* frm) { CompleteFrame(frm); }
    );
    //...
}
```

_studio/shared/umc/codec/h265_dec/src/umc_h265_segment_decoder_dxva.cpp
```diff
@@ -135,6 +135,24 @@ void TaskBrokerSingleThreadDXVA::Start()
 
     TaskBroker_H265::Start();
     m_completedQueue.clear();
+
+    H265_DXVA_SegmentDecoder * dxva_sd = static_cast<H265_DXVA_SegmentDecoder *>(m_pTaskSupplier->m_pSegmentDecoder[0]);
+
+    if (dxva_sd == nullptr || dxva_sd->m_va == nullptr || dxva_sd->m_va->IsUseStatusReport())
```
dmitry-gurulev: 1. Having `dxva_sd == nullptr || dxva_sd->m_va == nullptr` is logic error, I prefer to throw exception here.
2. Is see that u try to minimize code changes, but for me, it would be better not to spread completion logic over several functions and do all things at one place - `TaskBrokerSingleThreadDXVA::GetNextTaskInternal`

PR #1194, opened 2019-02-14
Destroy va buffers correctly

### LINKED ISSUE #1065: HEVCe: replacing macro MFX_DESTROY_VABUFFER to function CheckAndDestroyVAbuffer

mgonchar:
This PR replaces macro which ignored returned status with function that handles it.

All components patched. Fixes #1065

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_fei_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_fei_vaapi.cpp
@@ -59,25 +59,23 @@ mfxStatus VAAPIFEIPREENCEncoder::Destroy()
 {
     MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_HOTSPOTS, "FEI::PreENC::Destroy");
 
-    mfxStatus sts = MFX_ERR_NONE;
-
-    MFX_DESTROY_VABUFFER(m_statParamsId, m_vaDisplay);
+    mfxStatus sts = CheckAndDestroyVAbuffer(m_vaDisplay, m_statParamsId);
+    std::ignore = MFX_STS_TRACE(sts);
 
     for( mfxU32 i = 0; i < m_statMVId.size(); i++ )
     {
-        MFX_DESTROY_VABUFFER(m_statMVId[i], m_vaDisplay);
+        sts = CheckAndDestroyVAbuffer(m_vaDisplay, m_statMVId[i]);
+        std::ignore = MFX_STS_TRACE(sts);
     }
 
     for( mfxU32 i = 0; i < m_statOutId.size(); i++ )
     {
-        MFX_DESTROY_VABUFFER(m_statOutId[i], m_vaDisplay);
+        sts = CheckAndDestroyVAbuffer(m_vaDisplay, m_statOutId[i]);
+        std::ignore = MFX_STS_TRACE(sts);
     }
 
-    sts = VAAPIEncoder::Destroy();
-
-    return sts;
-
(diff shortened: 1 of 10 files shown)


_studio/mfx_lib/shared/src/mfx_h264_fei_vaapi.cpp
```diff
@@ -1775,17 +1774,29 @@ mfxStatus VAAPIFEIPAKEncoder::Destroy()
...
-    mfxStatus sts = MFX_ERR_NONE;
-
-    MFX_DESTROY_VABUFFER(m_statParamsId,     m_vaDisplay);
-    MFX_DESTROY_VABUFFER(m_statMVId,         m_vaDisplay);
-    MFX_DESTROY_VABUFFER(m_statOutId,        m_vaDisplay);
-    MFX_DESTROY_VABUFFER(m_codedBufferId[0], m_vaDisplay);
-    MFX_DESTROY_VABUFFER(m_codedBufferId[1], m_vaDisplay);
-
-    sts = VAAPIEncoder::Destroy();
+#if defined(_DEBUG)
```
dvrogozh: emmm. looks ugly:). https://en.cppreference.com/w/cpp/utility/tuple/ignore - can this help somehow?

mgonchar: No, my idea was a bit different. Since there is no good way to react on error status in destructor, I just want to make those statuses visible for debug. ```std::ignore``` doesn't allow to look for assigned value

mgonchar:
@dvrogozh any other concerns about this PR?

_studio/mfx_lib/shared/src/mfx_h264_fei_vaapi.cpp
```diff
@@ -1775,17 +1774,29 @@ mfxStatus VAAPIFEIPAKEncoder::Destroy()
...
+    mfxStatus sts;
+    sts =
+#endif
```
dvrogozh: I don't like this section. If someone needs to see variable in debug - let him patch according and rebuild. I vote for dropping these if-defs.

dmitry-gurulev: @dvrogozh, +1
Anyway, stepping into `CheckAndDestroyVAbuffer` is also an option.

mgonchar: @dmitry-gurulev @dvrogozh 
what about following solution:
```
mfxStatus sts = CheckAndDestroyVAbuffer(...);
MFX_STS_TRACE(sts);
```

onabiull: Please don't introduce __DEBUG stuff. MFX_STS_TRACE(sts); a bit better still ideally we should actually handle errors, not just log them.

mgonchar: @onabiull this is ```Close``` function, i.e. = destructor. What is your suggestion here?

onabiull: As a short-term logging is enough. But considering possible fail sources - corrupted heap, corrupted VA context, probably we should just fail here.

mgonchar: @dvrogozh @onabiull @dmitry-gurulev please see new version

mgonchar:
@dvrogozh any concerns about latest version?

PR #1198, opened 2019-02-15
[VP9D] params checking function was made codec-specific

IlyaSergeyev:
checkInfoCodecs function was moved into decoder and error codes were corrected

MDP-50900

### DIFF
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -558,8 +558,7 @@ mfxStatus VideoDECODEVP9_HW::DecodeHeader(VideoCORE* core, mfxBitstream* bs, mfx
     mfxStatus sts = MFX_VP9_Utility::DecodeHeader(core, bs, par);
     MFX_CHECK_STS(sts);
 
-    if (   par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010
-        )
+    if (par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010)
         par->mfx.FrameInfo.Shift = 1;
 
     return sts;
@@ -740,8 +739,7 @@ void VideoDECODEVP9_HW::UpdateVideoParam(mfxVideoParam *par, VP9DecoderFrame con
 
     MFX_VP9_Utility::FillVideoParam(m_core, frameInfo, par);
 
-    if (   par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010
-        )
+    if (par->mfx.FrameInfo.FourCC == MFX_FOURCC_P010)
         par->mfx.FrameInfo.Shift = 1;
 }
 
@@ -945,6 +943,44 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus CheckFrameInfo(mfxFrameInfo const &currInfo, mfxFrameInfo &info)
+{
+    MFX_SAFE_CALL(CheckFrameInfoCommon(&info, MFX_CODEC_VP9));
+
+    switch (info.FourCC)
+    {
+        case MFX_FOURCC_NV12:
+        case MFX_FOURCC_AYUV:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y410:
+#endif
+            break;
+        case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y210:
+#endif
+            MFX_CHECK(info.Shift == 1, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
+            break;
+        default:
+            MFX_CHECK_STS(MFX_ERR_INVALID_VIDEO_PARAM);
+    }
+
+    switch(info.ChromaFormat)
+    {
+        case MFX_CHROMAFORMAT_YUV420:
+        case MFX_CHROMAFORMAT_YUV444:
+            break;
+        default:
+            MFX_CHECK_STS(MFX_ERR_INVALID_VIDEO_PARAM);
+    }
+
+    MFX_CHECK(currInfo.FourCC == info.FourCC, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
+
+    return MFX_ERR_NONE;
+}
+
+
 mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1 *surface_work, mfxFrameSurface1 **surface_out, MFX_ENTRY_POINT * p_entry_point)
 {
     UMC::AutomaticUMCMutex guard(m_mGuard);
@@ -998,7 +1034,7 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
             return MFX_ERR_UNDEFINED_BEHAVIOR;
     }
 
-    sts = CheckFrameInfoCodecs(&surface_work->Info, MFX_CODEC_VP9, true);
+    sts = CheckFrameInfo(m_vPar.mfx.FrameInfo, surface_work->Info);
     MFX_CHECK_STS(sts);
 
     sts = CheckFrameData(surface_work);
@@ -1007,8 +1043,7 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     eMFXHWType type = m_core->GetHWType();
     if (!MFX_VPX_Utility::CheckFrameInfo(surface_work->Info, MFX_CODEC_VP9, m_platform, type))
     {
-        sts = MFX_ERR_INVALID_VIDEO_PARAM;
-        MFX_CHECK_STS(sts);
+        MFX_CHECK_STS(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
     }
 
     sts = CheckBitstream(bs);


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,183 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus checkFrameInfoCommon(mfxFrameInfo  *info, mfxU32 /* codecId */)
```
mgonchar: If it is now codec-specific, why we need second parameter?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,183 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+            return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
+        break;
+
+    default:
+        if (info->FourCC != MFX_FOURCC_NV12)
+            return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
+        break;
+    }
+
+    switch (codecId)
```
mgonchar: why we have switch among different codecs in file named ```mfx_vp9_dec_decode_hw.cpp``` ?

IlyaSergeyev:
excess codecs code reomoved

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,124 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus checkFrameInfoCommon(mfxFrameInfo  *info)
```
dmitry-gurulev: U actually don't need to bring _checkFrameInfoCommon_ here, it's actually very common checks that are not dependent on codec.

IlyaSergeyev: Yes, removed and refactored

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,48 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus CheckFrameInfo(mfxFrameInfo &currInfo, mfxFrameInfo &info)
+{
+    mfxStatus sts = CheckFrameInfoCommon(&info, MFX_CODEC_VP9);
+
+    if (sts != MFX_ERR_NONE)
```
mgonchar: I think you can use macro like ```MFX_CHECK``` here. It is preferable, since they support error tracing. Same for all checks below

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -1007,7 +1049,7 @@ mfxStatus VideoDECODEVP9_HW::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     eMFXHWType type = m_core->GetHWType();
     if (!MFX_VPX_Utility::CheckFrameInfo(surface_work->Info, MFX_CODEC_VP9, m_platform, type))
     {
-        sts = MFX_ERR_INVALID_VIDEO_PARAM;
+        sts = MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
```
mgonchar: make it a oneliner:
```
MFX_CHECK_STS(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: Tests failed with this macro, so it has little different behaviour

mgonchar: What do you mean different? If you write:
```
MFX_CHECK_STS(sts);
```
or
```
MFX_CHECK(sts == MFX_ERR_NONE, sts);
```
it will have exactly the same behavior but will allow you to trace the error origin

IlyaSergeyev: Yes, it works, updated

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,50 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus CheckFrameInfo(mfxFrameInfo &currInfo, mfxFrameInfo &info)
+{
+    mfxStatus sts = CheckFrameInfoCommon(&info, MFX_CODEC_VP9);
+    MFX_CHECK(sts == MFX_ERR_NONE, sts);
```
mgonchar:
```
MFX_CHECK_STS(sts);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,50 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y410:
+#endif
+            break;
+        case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y210:
+#endif
+           if (info.Shift != 1)
+              return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
```
mgonchar:
```
MFX_CHECK(info.Shift == 1, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,50 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+           break;
+        default:
+            return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
```
mgonchar:
```
MFX_CHECK_STS(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,50 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+
+    switch(info.ChromaFormat)
+    {
+        case MFX_CHROMAFORMAT_YUV420:
+        case MFX_CHROMAFORMAT_YUV400:
+        case MFX_CHROMAFORMAT_YUV422:
+        case MFX_CHROMAFORMAT_YUV444:
+            break;
+        default:
+            return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
```
mgonchar:
```
MFX_CHECK_STS(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,50 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+    }
+
+    //check if FourCC changed
+    if (currInfo.FourCC != info.FourCC)
+        return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
```
mgonchar:
```
MFX_CHECK(currInfo.FourCC == info.FourCC, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,50 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+        case MFX_CHROMAFORMAT_YUV420:
+        case MFX_CHROMAFORMAT_YUV400:
+        case MFX_CHROMAFORMAT_YUV422:
+        case MFX_CHROMAFORMAT_YUV444:
+            break;
+        default:
+            return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
+    }
+
+    //check if FourCC changed
```
mgonchar: not much sense in this comment it is quite obvious what you are doing

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: MFX_CHECK_STS has 2 arguments, second is return code. I think, we need return actual return code from CheckFrameInfoCommon here, not other constant.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: Ok, accepted

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
dmitry-gurulev: Do u really support 400/422?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: same

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: ok

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: Why? sts == MFX_ERR_NONE here ... We don't need to check it.

mgonchar: @IlyaSergeyev didn't get your idea

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
mgonchar: what about this one?

IlyaSergeyev: Ok, i'll do it. But I still think, that return ERR is more understandable, than MFX_CHECK_STS(err)!

mgonchar: I agree with you, that this code looks a bit artificial. But, it allows you to trace the error to origin. Reason is that you will see which exact check was triggering and sent an error to your application. See how this mechanism works in https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/include/mfx_utils.h#L38  you need to enable ```MFX_DEBUG_TRACE``` to see the full log

IlyaSergeyev: Maybe, we need to have MFX_RETURN macro for such cases. It will be much more readable.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,45 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+#endif
+            break;
+        case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y210:
+#endif
+            MFX_CHECK(info.Shift == 1, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
+            break;
+        default:
+            MFX_CHECK_STS(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```
dmitry-gurulev: Invalid?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +945,45 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus CheckFrameInfo(mfxFrameInfo &currInfo, mfxFrameInfo &info)
```
dmitry-gurulev: const?

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
IlyaSergeyev: yes. testcase 33!

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +943,44 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
     return MFX_ERR_NONE;
 }
 
+
+static mfxStatus CheckFrameInfo(mfxFrameInfo const &currInfo, mfxFrameInfo &info)
+{
+    MFX_CHECK_STS(CheckFrameInfoCommon(&info, MFX_CODEC_VP9));
```
mgonchar: please, use ```MFX_SAFE_CALL``` to avoid double call of function

IlyaSergeyev:
>Every codec (should) have own function. Dependent on SW or use HwCaps for HW.
I think so too.

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -945,6 +943,44 @@ mfxStatus VideoDECODEVP9_HW::PrepareInternalSurface(UMC::FrameMemID &mid, mfxFra
...
+        case MFX_FOURCC_AYUV:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y410:
+#endif
+            break;
+        case MFX_FOURCC_P010:
+#if (MFX_VERSION >= 1027)
+        case MFX_FOURCC_Y210:
+#endif
+            MFX_CHECK(info.Shift == 1, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
```
dmitry-gurulev: AFAIR, [Shift == 0] is allowed for system memory. Did u check decoding it?

IlyaSergeyev: No. But the purpose of this commit is only moving checking function into VP9 decoder code. There are no such check in mfx_common_int, so Shift == 0 with system memory wasn't work.

PR #1199, opened 2019-02-15
h265d: fix static analysis warning

akokoshn:
-Add initialization HeaderSet::m_Header

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/include/umc_h265_headers.h
+++ b/_studio/shared/umc/codec/h265_dec/include/umc_h265_headers.h
@@ -38,7 +38,8 @@ class HeaderSet
 public:
 
     HeaderSet(Heap_Objects  *pObjHeap)
-        : m_pObjHeap(pObjHeap)
+        : m_Header()
+        , m_pObjHeap(pObjHeap)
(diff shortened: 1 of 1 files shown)


_studio/shared/umc/codec/h265_dec/include/umc_h265_headers.h
```diff
@@ -38,7 +38,8 @@ class HeaderSet
 public:
 
     HeaderSet(Heap_Objects  *pObjHeap)
-        : m_pObjHeap(pObjHeap)
+        : m_Header()
```
mgonchar: ```std::vector``` has default constructor that will be called without this change as well as with it. What was the warning from static analysis?

akokoshn: @mgonchar , "'this->m_Header' is not initialized in this constructor"

mgonchar: Ok. Generally I don't mind on merging this. But please change wording from "issue" to "warning", because initial code hadn't got any real issue with initialization.

PR #1203, opened 2019-02-18
[samples] Force AYUV output format support for sample_decode

IlyaSergeyev:


IlyaSergeyev:
@fzhar No. Default output format will be NV12.

IlyaSergeyev:
@fzhar Yes, it is possible using AYUV input stream. Forced AYUV output is a IOTG request.

PR #1207, opened 2019-02-19
vp8d: don't convert a signed integer to an unsigned integer

xhaihao:
A signed integer is required when updating loop filter level for
segmentation however picParams->loop_filter_level[i] is an unsigned
integer

### DIFF
--- a/_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw_vaapi.cpp
+++ b/_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw_vaapi.cpp
@@ -147,9 +147,8 @@ mfxStatus VideoDECODEVP8_HW::PackHeaders(mfxBitstream *p_bistream)
                  picParams->loop_filter_level[i] = m_frame_info.segmentFeatureData[VP8_ALT_LOOP_FILTER][i];
              else
              {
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_hw_vaapi.cpp
```diff
@@ -147,9 +147,8 @@ mfxStatus VideoDECODEVP8_HW::PackHeaders(mfxBitstream *p_bistream)
                  picParams->loop_filter_level[i] = m_frame_info.segmentFeatureData[VP8_ALT_LOOP_FILTER][i];
              else
              {
-                 picParams->loop_filter_level[i] = m_frame_info.loopFilterLevel + m_frame_info.segmentFeatureData[VP8_ALT_LOOP_FILTER][i];
-                 picParams->loop_filter_level[i] = (picParams->loop_filter_level[i] >= 0) ?
-                     ((picParams->loop_filter_level[i] <= 63) ? picParams->loop_filter_level[i] : 63) : 0;
+                 int filter_level = m_frame_info.loopFilterLevel + m_frame_info.segmentFeatureData[VP8_ALT_LOOP_FILTER][i];
```
mgonchar: please, use ```int_32t``` type

xhaihao: Fixed, thanks

PR #1208, opened 2019-02-19
vpp: UYVY can be used as output

xhaihao:
This fixes the regression caused by b49b03b3.


xhaihao:
any comment ? @dmitryermilov @onabiull @alariono @snowpiel . You may run the command below to test UYVY output:

gst-launch-1.0 videotestsrc ! video/x-raw,format=YUY2 ! msdkvpp ! video/x-raw,format=UYVY ! fakesink

You need the latest gst-plugin-bad from https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad and media-driver from https://github.com/intel/media-driver

PR #654, opened 2018-09-20
h264e: enable ICQ rate control

dvrogozh:
Signed-off-by: Dmitry Rogozhkin <dmitry.v.rogozhkin@intel.com>

PR #1209, opened 2019-02-19
MFXVideoENCODE_Init(): reset return value.

djie1:
If the return value is "MFX_WRN_PARTIAL_ACCELERATION",
reset it to "MFX_ERR_INVALID_VIDEO_PARAM" as no sw callback.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
@@ -836,7 +836,8 @@ mfxStatus ImplementationAvc::Init(mfxVideoParam * par)
 
     mfxStatus checkStatus = CheckVideoParam(m_video, m_caps, m_core->IsExternalFrameAllocator(), m_currentPlatform, m_currentVaType, *pMFXGTConfig, true);
     if (checkStatus == MFX_WRN_PARTIAL_ACCELERATION)
-        return MFX_WRN_PARTIAL_ACCELERATION;
+        // there is no sw fallback for AVC encoder, so reset the return value
+        return MFX_ERR_INVALID_VIDEO_PARAM;
     else if (checkStatus < MFX_ERR_NONE)
         return checkStatus;
     else if (checkStatus == MFX_ERR_NONE)


_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -705,6 +705,11 @@ mfxStatus MFXVideoENCODE_Init(mfxSession session, mfxVideoParam *par)
         }
 
         mfxRes = session->m_pENCODE->Init(par);
+        if (MFX_WRN_PARTIAL_ACCELERATION == mfxRes)
+        {
+            // there is no sw callback
```
dvrogozh: There is mjpeg sw fallback. See line 227.

Please, provide a command line example where you see the issue. If some HW component returned this status - then this is bug in this component which we should fix rather than provide a workaround here.

djie1: It's our internal test "beh_fei_encode_init" failure, tracked by MDP-47377.
It's AVCe/AVC FEI Encode returns this status, so let me reset the return value inside ImplementationAVC::Init().

djie1:
> Same change is required for Query, QueryIOSurf. I think it's better to change all API functions within single commit.

Seems it's not required for MFXVideoENCODE_Query() and MFXVideoENCODE_QueryIOSurf(), see line#564-569, and line#654-659 in file libmfxsw_encode.cpp.

stasefremov:
Sorry for long silence. Then my concern is implementation of same logic (i.e. "return error if no SW fallback") on different abstraction layers for different API functions. 

@onabiull, could you please clarify why logic at lines #564-569 and #654-659 in libmfxsw_encode.cpp exists for Query/QueryIOSurf. But there is no such logic for Init?

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -836,7 +836,8 @@ mfxStatus ImplementationAvc::Init(mfxVideoParam * par)
 
     mfxStatus checkStatus = CheckVideoParam(m_video, m_caps, m_core->IsExternalFrameAllocator(), m_currentPlatform, m_currentVaType, *pMFXGTConfig, true);
     if (checkStatus == MFX_WRN_PARTIAL_ACCELERATION)
-        return MFX_WRN_PARTIAL_ACCELERATION;
+        // there is no sw fallback for AVC encoder, so reset the return value
+        return MFX_ERR_INVALID_VIDEO_PARAM;
```
mgonchar: it is definitely wrong status to return. We should report UNSUPPORTED if no appropriate SW fallback

mgonchar:
I think correct fix is #1349

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
dmitry-gurulev: AFAIR, it's `Query` function that is allowed to return UNSUPPORTED, `Init` have to return INVALID_VIDEO_PARAM

mgonchar:
#1349 is merged. I'm closing this one

Issue #333, opened 2018-06-21
ffmpeg vp8/vc1 decoding failure due to MFX_FRAMETYPE is uninitialized


lizhong1008:
Reproduce: 
Using the latest ffmpeg( commit_id: deb9a04d54045c8021816cfe019bf2bba0e9595d) 

./ffmpeg -hwaccel qsv -init_hw_device qsv -c:v vp8_qsv -i /root/lizhong/Videos/fate-suite/vp8-test-vectors-r1/vp80-00-comprehensiv
e-001.ivf -vf "hwdownload,format=nv12" -pix_fmt yuv420p -f md5 -
ffmpeg version N-91336-gdeb9a04 Copyright (c) 2000-2018 the FFmpeg developers
  built with gcc 5.4.0 (Ubuntu 5.4.0-6ubuntu1~16.04.9) 20160609
  configuration: --enable-libmfx --enable-libx264 --enable-gpl --disable-optimizations --enable-debug=3
  libavutil      56. 18.102 / 56. 18.102
  libavcodec     58. 20.103 / 58. 20.103
  libavformat    58. 17.101 / 58. 17.101
  libavdevice    58.  4.101 / 58.  4.101
  libavfilter     7. 25.100 /  7. 25.100
  libswscale      5.  2.100 /  5.  2.100
  libswresample   3.  2.100 /  3.  2.100
  libpostproc    55.  2.100 / 55.  2.100
Input #0, ivf, from '/root/lizhong/Videos/fate-suite/vp8-test-vectors-r1/vp80-00-comprehensive-001.ivf':
  Duration: 00:00:00.97, start: 0.000000, bitrate: 131 kb/s
    Stream #0:0: Video: vp8 (VP80 / 0x30385056), yuv420p(progressive), 176x144, 30 tbr, 30 tbn, 30 tbc
Stream mapping:
  Stream #0:0 -> #0:0 (vp8 (vp8_qsv) -> rawvideo (native))
Press [q] to stop, [?] for help
Assertion 0 failed at libavcodec/qsv.c:219
Aborted (core dumped)

It is failed in ff_qsv_map_pictype() checking frame_type and found it is uninitialized(return a value of MFX_FRAMETYPE_UNKNOWN).  Other codes (h264/vc1/hevc) looks good except vp8.

lizhong1008:
[...]

PR #1212, opened 2019-02-20
mjpegd: add reporting FrameType

akokoshn:
-Add reporting type of decoded frame via extended buffer
 MFX_EXTBUFF_DECODED_FRAME_INFO

Fixes: #333


mgonchar:
Extension buffers, not extended :)

akokoshn:
@mgonchar, right. Thanks)

PR #1214, opened 2019-02-20
[H264][Dec] Fix for h264 decode

aulitin:
To prevent an infinite loop

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/include/umc_h264_bitstream_headers.h
+++ b/_studio/shared/umc/codec/h264_dec/include/umc_h264_bitstream_headers.h
@@ -19,6 +19,8 @@
 // SOFTWARE.
 
 #include "umc_defs.h"
+#include <mfx_utils.h>
+
 #if defined (MFX_ENABLE_H264_VIDEO_DECODE)
 
 #ifndef __UMC_H264_BITSTREAM_HEADERS_H_
--- a/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
+++ b/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
@@ -775,6 +775,8 @@ inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t heig
         case H264VideoDecoderParams::H264_LEVEL_52:
             level_idc = H264VideoDecoderParams::H264_LEVEL_MAX;
             break;
+        default:
+            throw h264_exception(UMC_ERR_FAILED);
         }
     }
 
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_bitstream_headers.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_bitstream_headers.cpp
@@ -335,6 +335,39 @@ H264HeadersBitstream::H264HeadersBitstream(uint8_t * const pb, const uint32_t ma
 {
 }
 
+inline bool CheckLevel(uint8_t level_idc)
+{
+    switch(level_idc)
+    {
+    case H264VideoDecoderParams::H264_LEVEL_1:
+    case H264VideoDecoderParams::H264_LEVEL_11:
+    case H264VideoDecoderParams::H264_LEVEL_12:
+    case H264VideoDecoderParams::H264_LEVEL_13:
+
+    case H264VideoDecoderParams::H264_LEVEL_2:
+    case H264VideoDecoderParams::H264_LEVEL_21:
+    case H264VideoDecoderParams::H264_LEVEL_22:
+
+    case H264VideoDecoderParams::H264_LEVEL_3:
+    case H264VideoDecoderParams::H264_LEVEL_31:
+    case H264VideoDecoderParams::H264_LEVEL_32:
+
+    case H264VideoDecoderParams::H264_LEVEL_4:
+    case H264VideoDecoderParams::H264_LEVEL_41:
+    case H264VideoDecoderParams::H264_LEVEL_42:
+
+    case H264VideoDecoderParams::H264_LEVEL_5:
+    case H264VideoDecoderParams::H264_LEVEL_51:
+    case H264VideoDecoderParams::H264_LEVEL_52:
+
+    case H264VideoDecoderParams::H264_LEVEL_9:
+        return true;
+
+    default:
+        return false;
+    }
+}
+
 // ---------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
```diff
@@ -775,6 +775,8 @@ inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t heig
         case H264VideoDecoderParams::H264_LEVEL_52:
             level_idc = H264VideoDecoderParams::H264_LEVEL_MAX;
             break;
+        default:
+            throw h264_exception(UMC_ERR_INVALID_PARAMS);
```
dvrogozh: @dmitry-gurulev : if this function CalculateDPBSize will be called with the invalid level value _unknown_ to this switch, we could enter infinite loop. I guess that's what @aulitin was trying to say. If I understand author's commit message correctly, level_idc was initially passed as =240. Question @aulitin : where we managed to take this value from?

So, the above switch is designed to fix/adjust _correct_ value of level_ids only. If it will be wrong we will blow up.

dmitry-gurulev: @dvrogozh 
>if this function CalculateDPBSize will be called with the invalid level

My point is that it shouldn't be called in that way and we need to fix initial problem - appearing wrong level value. The second thing is it's a logic error and exception should be `UMC_ERR_FAILED`, not `UMC_ERR_INVALID_PARAMS`

dvrogozh: @dmitry-gurulev : my point is the same. I completely agree. This patch attempt to fix only the second part. The first part with wrong level value is not fixed and probably is not fully understood.

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_bitstream_headers.cpp
```diff
@@ -1111,6 +1111,10 @@ Status H264HeadersBitstream::GetSequenceParamSetMvcExt(H264SeqParamSetMVCExtensi
 
         // decode the level's profile idc
         levelInfo.level_idc = (uint8_t) GetBits(8);
+        if (H264VideoDecoderParams::H264_LEVEL_MAX <= levelInfo.level_idc)
```
dvrogozh: According to our code:
```
_studio/shared/umc/codec/h264_dec/include/umc_h264_dec.h:87:        H264_LEVEL_MAX  = 52,
_studio/shared/umc/codec/h264_dec/include/umc_h264_dec.h:86:        H264_LEVEL_52   = 52,
```
Do I understand correctly that after this change all H264 level 5.2 will stop working and will return an error?

@dmitry-gurulev : we support levels up to 5.2, but not including 5.2?

dmitry-gurulev: @dvrogozh, +1 we support 5.2, so proposed condition is incorrect.
@aulitin, I'd rather extract level check from _H264HeadersBitstream::GetSequenceParamSet_ into separate inline function, let's say `inline bool CheckLevel(uint8_t profile_idc, uint8_t level_idc)` and use it both at `GetSequenceParamSet` and `GetSequenceParamSetMvcExt`

_studio/shared/umc/codec/h264_dec/include/umc_h264_bitstream_headers.h
```diff
@@ -228,6 +230,8 @@ class H264HeadersBitstream : public H264BaseBitstream
 
     Status GetPredWeightTable(UMC_H264_DECODER::H264SliceHeader *hdr, const UMC_H264_DECODER::H264SeqParamSet *sps,
         UMC_H264_DECODER::PredWeightTable *pPredWeight_L0, UMC_H264_DECODER::PredWeightTable *pPredWeight_L1);
+
+    inline bool CheckLevel(uint8_t profile_idc, uint8_t level_idc);
```
mgonchar: no sense of ```inline``` key word. All class member functions are inline by default

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_bitstream_headers.cpp
```diff
@@ -335,6 +335,44 @@ H264HeadersBitstream::H264HeadersBitstream(uint8_t * const pb, const uint32_t ma
...
+
+    case H264VideoDecoderParams::H264_LEVEL_5:
+    case H264VideoDecoderParams::H264_LEVEL_51:
+    case H264VideoDecoderParams::H264_LEVEL_52:
+        return true;
+    case H264VideoDecoderParams::H264_LEVEL_9:
+        if (profile_idc != H264VideoDecoderParams::H264_PROFILE_BASELINE &&
+            profile_idc != H264VideoDecoderParams::H264_PROFILE_MAIN &&
+            profile_idc != H264VideoDecoderParams::H264_PROFILE_EXTENDED) {
+                level_idc = H264VideoDecoderParams::H264_LEVEL_1b;
```
mgonchar: no sense in this assignment, since ```level_idc``` is not a reference. Anyway, should this function correct value of ```level_idc``` somehow? If so, the name is missleading

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_bitstream_headers.cpp
```diff
@@ -335,6 +335,44 @@ H264HeadersBitstream::H264HeadersBitstream(uint8_t * const pb, const uint32_t ma
 {
 }
 
+//return ture if level_idc is correct
```
mgonchar: typo + I don't see any sense in this comment, I suggest to remove it

_studio/shared/umc/codec/h264_dec/include/umc_h264_bitstream_headers.h
dmitry-gurulev: `CheckLevel` doesn't use any non-public state of `H264HeadersBitstream`, it 's not necessary to make it a class method, I'd rather make it free function.

_studio/shared/umc/codec/h264_dec/src/umc_h264_dec_bitstream_headers.cpp
dmitry-gurulev: @aulitin, please leave `sps->level_idc` adjustment at its prev. place - at `GetSequenceParamSet`

PR #1216, opened 2019-02-21
h264e: restore check adaptive max frame size

akokoshn:
-AdaptiveMaxFrameSize can't be ON if MaxFrameSizeP = 0 or LowPower is ON

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -784,8 +784,16 @@ namespace
 
         if (hwCaps.UserMaxFrameSizeSupport == 0 && !IsEnabledSwBrc && IsOn(extOpt3.AdaptiveMaxFrameSize))
         {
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -787,6 +787,14 @@ namespace
             extOpt3.AdaptiveMaxFrameSize = 0;
             unsupported = true;
         }
+
+        if (hwCaps.UserMaxFrameSizeSupport == 1 && !IsEnabledSwBrc &&
+            (extOpt3.MaxFrameSizeP == 0 || IsOn(par.mfx.LowPower)) &&
+            IsOn(extOpt3.AdaptiveMaxFrameSize))
+        {
+            extOpt3.AdaptiveMaxFrameSize = 0;
```
mgonchar: Please, use proper enums, not constants.
0 is ```MFX_CODINGOPTION_UNKNOWN```

PR #1220, opened 2019-02-26
vp9d: fix InitDequantizer() for DRC

akokoshn:
-Remove VideoDECODEVP9_HW::m_*_delta_q as duplicate,
 current values already stored in m_frameInfo

### DIFF
--- a/_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode_hw.h
+++ b/_studio/mfx_lib/decode/vp9/include/mfx_vp9_dec_decode_hw.h
@@ -144,9 +144,6 @@ private:
     mfxBitstream m_bs;
 
     mfxI32 m_baseQIndex;
-    mfxI32 m_y_dc_delta_q;
-    mfxI32 m_uv_dc_delta_q;
-    mfxI32 m_uv_ac_delta_q;
 };
 
 #endif // _MFX_VP9_DECODE_HW_H_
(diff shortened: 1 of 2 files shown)


mgonchar:
According to your description you are not actually fixing something, but removing excessive code. Please, correct commit message

akokoshn:
@dmitry-gurulev , agree  m_baseQIndex should be removed, but I wouldn't like mix current change with removing useless code.
I guess better do it by separate PR

mgonchar:
@akokoshn it could be same PR but another commit. Anyway your PR should be named something like "Cleaning up vp9d excessive variables"

akokoshn:
@mgonchar, removing m_baseQIndex not relate to issue which should be fixed by this PR.
Let's don't mix

mgonchar:
@akokoshn do you really fixing anything? From your description I treated this commit as enhancement and refactoring. Could you describe what is the issue?

akokoshn:
@mgonchar, this change fix corruption issue on DRC: in case of changing resolution  and quantilization coefficients decoder doesn't update  *_delta_q.
The reason updating happens if coefficients were changed ("old != new"), due to double read bit stream in case of DRC (1th read for detect resolution change). On second read  "old" values already updated (stored in m_*_delta_q), so we miss changing quantilization coefficients

mgonchar:
ok, now the purpose of change is clear. so please put this description to your first comment, so it will appear in git history after merge of your commit

PR #1225, opened 2019-02-27
sample_common: fix mismatched new [] / delete

mgonchar:
m_fDestMVC is allocated as array


mgonchar:
Memory is allocated at https://github.com/Intel-Media-SDK/MediaSDK/blob/51f5d5e1bc1f3a468fd5b26e378f3891870dd92f/samples/sample_common/src/sample_utils.cpp#L783

PR #1226, opened 2019-02-27
sample_common: replace safe_array class with std::unique_ptr[]

mgonchar:


dmitry-gurulev:
I actually meant `std::vector`.

mgonchar:
this code works with memory allocated on heap, see https://github.com/Intel-Media-SDK/MediaSDK/blob/47608e1e2dbbe48aa0523aa72c42f66b2d0089a7/samples/sample_common/src/sysmem_allocator.cpp#L347 so we can't change to data type which stores memory on stack

mgonchar:
@fzhar the only reason why you need smart pointer here is to guarantee exception safety. Currently I just leave same logic as were here before, but removing excessive self-written class

mgonchar:
@dmitry-gurulev talking about ```std::vector``` the problem is the same - we can't use it without changing ```mfxFrameAllocResponse``` struct layout and making it non-POD, which is actually not acceptable here in my understanding.

dmitry-gurulev:
@mgonchar, got it, tnx.

PR #1229, opened 2019-02-28
vc1d: fix frame type reporting

akokoshn:
-Add handling P/B frame types

### DIFF
--- a/_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
+++ b/_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
@@ -49,7 +49,13 @@ static void SetFrameType(const UMC::VideoData &mediaData, mfxFrameSurface1 &surf
         case UMC::I_PICTURE:
             extFrameInfo->FrameType = MFX_FRAMETYPE_I;
             break;
-        // NONE_PICTURE, D_PICTURE (skipped)
+        case UMC::P_PICTURE:
+            extFrameInfo->FrameType = MFX_FRAMETYPE_P;
+            break;
+        case UMC::B_PICTURE:
+            extFrameInfo->FrameType = MFX_FRAMETYPE_B;
+            break;
+        // NONE_PICTURE
         default:
             extFrameInfo->FrameType = MFX_FRAMETYPE_UNKNOWN;
     }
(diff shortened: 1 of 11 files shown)


akokoshn:
WIP

_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
```diff
@@ -49,6 +49,12 @@ static void SetFrameType(const UMC::VideoData &mediaData, mfxFrameSurface1 &surf
         case UMC::I_PICTURE:
             extFrameInfo->FrameType = MFX_FRAMETYPE_I;
             break;
+        case UMC::P_PICTURE:
+            extFrameInfo->FrameType = MFX_FRAMETYPE_P;
+            break;
+        case UMC::B_PICTURE:
```
lizhong1008: VC1 has B and BI type.  Probably the issue you commented at https://github.com/Intel-Media-SDK/MediaSDK/issues/333#issuecomment-468352510 is caused BI type is not handled here.

akokoshn: BI-Frames handled as B:
`case  VC1_B_FRAME:
 case  VC1_BI_FRAME:
     out_data->SetFrameType(B_PICTURE);
     break;`
umc_vc1_video_decoder_hw.h: line 169

lizhong1008: Hm, I see. So what is the frame_tpye which is still reported as unknown?

_studio/shared/umc/codec/vc1_dec/include/umc_vc1_video_decoder_hw.h
```diff
@@ -157,10 +157,22 @@ namespace UMC
...
+                case  VC1_P_FRAME:
+                    out_data->SetFrameType(P_PICTURE);
+                    break;
+                case  VC1_B_FRAME:
+                case  VC1_BI_FRAME:
+                    out_data->SetFrameType(B_PICTURE);
+                    break;
+                case  VC1_SKIPPED_FRAME:
+                    out_data->SetFrameType(D_PICTURE); // means skipped
+                    break;
```
dvrogozh: default is missing. Can we potentially come into it here?

akokoshn: Default value (NONE) sets on initialization  out_data, we can reset NONE here, but I guess it's extra action

_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
akokoshn: VC1_SKIPPED_FRAME reported as unknown, should be fixed in the latest update this PR

_studio/shared/umc/codec/vc1_dec/include/umc_vc1_video_decoder_hw.h
```diff
@@ -157,10 +157,19 @@ namespace UMC
...
+                case  VC1_I_FRAME:
+                    out_data->SetFrameType(I_PICTURE);
+                    break;
+                case  VC1_P_FRAME:
+                    out_data->SetFrameType(P_PICTURE);
+                    break;
+                case  VC1_B_FRAME:
+                case  VC1_BI_FRAME:
+                    out_data->SetFrameType(B_PICTURE);
+                    break;
```
dvrogozh: You know, I am nervous when I see switch without the default case. This is origin of errors. Hence this is the wonderful place to add `assert()` even if you believe we won't be here.

akokoshn:
@dvrogozh, sure, I will provide test results

PR #1231, opened 2019-03-01
[VP9D] resolution change check removed if VA used

IlyaSergeyev:
Can been removed when VA surface reallocation patch implemented.

MDP-53745


dmitry-gurulev:
@IlyaSergeyev, isn't this the same as MDP-52037?

IlyaSergeyev:
> @IlyaSergeyev, isn't this the same as MDP-52037?

Yes

PR #1235, opened 2019-03-04
Add va status check

mgonchar:
This PR adds missed status check for VA functions


mgonchar:
@dvrogozh I don't think so, but when you ignore return statuses you can't say for sure :)

PR #1236, opened 2019-03-05
[jpegd] fix HW jped decoding support

Alexandr-Konovalov:
1) Image size checks are done via CAPs.
2) Enable HW/image combinations that are working.

Issue: MDP-52482

### DIFF
--- a/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
@@ -579,11 +579,6 @@ mfxStatus VideoDECODEMJPEG::QueryIOSurfInternal(VideoCORE *core, mfxVideoParam *
     }
     else
     {
-    // large pictures are not supported on Linux so far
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1025,6 +1020,7 @@ bool MFX_JPEG_Utility::IsNeedPartialAcceleration(VideoCORE * core, mfxVideoParam
 
     if (core->GetHWType() == MFX_HW_APL
```
onabiull: Shouldn't we re-phrase this one as type == APL || type >= KBL ? I.e. cover ICL and future platforms.

dvrogozh: +1

Alexandr-Konovalov: @onabiull @dvrogozh Done.

PR #1238, opened 2019-03-05
hevce: implement mbbrc for HEVCe

dmenshov:
implement mbbrc for HEVCe

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -417,6 +417,10 @@ mfxStatus SetRateControl(
 
     rate_param->initial_qp = par.m_pps.init_qp_minus26 + 26;
 
+    //  MBBRC control
+    // Control VA_RC_MB 0: default, 1: enable, 2: disable, other: reserved
+    rate_param->rc_flags.bits.mb_rate_control = IsOn(par.m_ext.CO2.MBBRC) ? 1 : IsOff(par.m_ext.CO2.MBBRC) ? 2 : 0;
+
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_vaapi.h
```diff
@@ -297,6 +297,8 @@ mfxStatus SetSkipFrame(
 
         UMC::Mutex m_guard;
         HeaderPacker m_headerPacker;
+
+        mfxU32 m_mbbrc;
```
eistarov: I see zero reasons to make it class-member. In both cases when it is used (CreateAccelerationService and Reset), its' value constructed/assigned right before usage and the stored one is ignored.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -417,6 +418,10 @@ mfxStatus SetRateControl(
 
     rate_param->initial_qp = par.m_pps.init_qp_minus26 + 26;
 
+    //  MBBRC control
+    // Control VA_RC_MB 0: default, 1: enable, 2: disable, other: reserved
+    rate_param->rc_flags.bits.mb_rate_control = mbbrc & 0xf;
```
mgonchar: Can't you directly use value from ```par``` here?
i.e. 
```
rate_param->rc_flags.bits.mb_rate_control = IsOn(par.m_ext.CO2.MBBRC) ? 1 : IsOff(par.m_ext.CO2.MBBRC) ? 2 : 0;
```

dmenshov: I have corrected.

PR #1247, opened 2019-03-06
Refactoring of internal VAAPI allocator

mgonchar:
Removed heap allocation
Simplified code
Increased consistency of operations with internal state

### DIFF
--- a/_studio/shared/src/libmfx_allocator_vaapi.cpp
+++ b/_studio/shared/src/libmfx_allocator_vaapi.cpp
@@ -625,7 +625,7 @@ mfxDefaultAllocatorVAAPI::LockFrameHW(
     mfxFrameData*  ptr)
 {
     MFX_CHECK(pthis, MFX_ERR_INVALID_HANDLE);
-    MFX_CHECK(mid, MFX_ERR_INVALID_HANDLE);
+    MFX_CHECK(mid,   MFX_ERR_INVALID_HANDLE);
     MFX_CHECK_NULL_PTR1(ptr);
 
     auto vaapi_mids = reinterpret_cast<vaapiMemIdInt*>(mid);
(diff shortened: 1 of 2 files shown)


mgonchar:
Only smoke dev testing done for now

_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -637,51 +637,59 @@ mfxDefaultAllocatorVAAPI::LockFrameHW(
...
-            va_res =  vaMapBuffer(pSelf->pVADisplay, *(vaapi_mids->m_surface), (void **)(&pBuffer));
+        {
+            mfxU8* pBuffer = nullptr;
+            {
+                MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaMapBuffer");
+                va_res = vaMapBuffer(pSelf->pVADisplay, *(vaapi_mids->m_surface), (void **)(&pBuffer));
+                MFX_CHECK(va_res == VA_STATUS_SUCCESS, VA_TO_MFX_STATUS(va_res));
+            }
+
+            ptr->Y = pBuffer;
```
dvrogozh: Do we allow double lock? Do we want to check it? (better in separate commit in the end of refactor to simplify bisect)

_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -697,39 +697,41 @@ mfxStatus mfxDefaultAllocatorVAAPI::UnlockFrameHW(
...
-    if (!vaapi_mids || !(vaapi_mids->m_surface)) return MFX_ERR_INVALID_HANDLE;
+    VAStatus va_res = VA_STATUS_SUCCESS;
 
     mfxU32 mfx_fourcc = ConvertVP8FourccToMfxFourcc(vaapi_mids->m_fourcc);
-
     if (MFX_FOURCC_P8 == mfx_fourcc)   // bitstream processing
     {
         MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaUnmapBuffer");
-        vaUnmapBuffer(pSelf->pVADisplay, *(vaapi_mids->m_surface));
+        va_res = vaUnmapBuffer(pSelf->pVADisplay, *(vaapi_mids->m_surface));
```
dvrogozh: what if it is already unlocked? again, maybe better in separate commit...

_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -279,196 +279,171 @@ mfxDefaultAllocatorVAAPI::AllocFramesHW(
...
+            codedbuf_size = request->Info.Width;
+            codedbuf_num  = request->Info.Height;
+            codedbuf_type = VAEncMacroblockMapBufferType;
         }
         else
         {
-            VAContextID context_id = request->AllocId;
-            mfxU32 codedbuf_size, codedbuf_num;
-            VABufferType codedbuf_type;
+            int width32  = 32 * ((request->Info.Width  + 31) >> 5);
```
dmitry-gurulev: `width32/height32` doesn't look to be good names. It would be better to use something like `aligned_width` or just add a comment that we need aligned value. BTW, we have a lot of alignment functions/macros like `UMC::align_value`, using of them would make code more readable than `32 * ((request->Info.Width  + 31) >> 5)`.

_studio/shared/src/libmfx_allocator_vaapi.cpp
mgonchar: I don't want to introduce the logic that wasn't initially in the design of this class. and previously this class didn't care about such things. We may think about it and add this functionality in separate PR.

Actually I don't know what will return ```vaMap```/```vaUnmap``` for already mapped/unmapped buffer

_studio/shared/src/libmfx_allocator_vaapi.cpp
mgonchar: This is for separate PR. I know several places in code where we have exactly same calculations. I suppose we need to change them all in separate PR

mgonchar:
@dmitry-gurulev agree to clean up styles. Which one will we pick?)

dmitry-gurulev:
@mgonchar, I believe it should be the one which is mostly used in that file. I didn't count, but it looks like it is snake notation (we may leave class members 'm_xxx' as-is).

_studio/shared/src/libmfx_allocator_vaapi.cpp
mgonchar: addressed in #1265

mgonchar:
smoke validation showed no regression

PR #1249, opened 2019-03-07
vc1d: fix reporting frame type for buffered frames

### LINKED ISSUE #1239: vc1d reports frame types incorrectly for some frames
Follow up issue after #333. Per @lizhong1008 observations:

> Tested SA00040.vc1 and SA10091.vc1 (you can get all of them from ffmpeg suite: http://ffmpeg.org/fate.html). Two issues found:
> * No I frame exposed for SA00040.vc1. All frames reported as B frames.
> * Some last frames of SA10091.vc1 are still reported as Unknown.

@akokoshn : can you, please, continue debug?

akokoshn:
-In case of reordering m_InternMediaDataOut.GetFrameType() may be not
 actual.
 So set correct value immediately after decode frame header

Fixes #1239

### DIFF
--- a/_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
+++ b/_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
@@ -38,26 +38,27 @@ using namespace UMC;
 class VC1TaskStore;
 
 
-static void SetFrameType(const UMC::VideoData &mediaData, mfxFrameSurface1 &surface)
+static void SetFrameType(const uint32_t type, mfxFrameSurface1 &surface)
 {
     auto extFrameInfo = reinterpret_cast<mfxExtDecodedFrameInfo *>(GetExtendedBuffer(surface.Data.ExtParam, surface.Data.NumExtParam, MFX_EXTBUFF_DECODED_FRAME_INFO));
     if (extFrameInfo == nullptr)
         return;
 
-    switch (mediaData.GetFrameType())
+    switch (type & VC1_BI_FRAME)
     {
-        case UMC::I_PICTURE:
+        case  VC1_I_FRAME:
             extFrameInfo->FrameType = MFX_FRAMETYPE_I;
             break;
-        case UMC::P_PICTURE:
+        case  VC1_P_FRAME:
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
```diff
@@ -1923,6 +1922,9 @@ mfxStatus   MFXVideoDECODEVC1::FillOutputSurface(mfxFrameSurface1 *surface)
                 surface->Info.PicStruct |= MFX_PICSTRUCT_FIELD_REPEATED;
         }
     }
+
+    SetFrameType(m_pVC1VideoDecoder->m_pStore->GetLastDS()->m_pContext->m_picLayerHeader->PTYPE, *surface);
```
mgonchar: Here is a chain of 5 pointers dereferencing... Is there a chance for some of them to be null?

_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
```diff
@@ -38,26 +38,27 @@ using namespace UMC;
...
-        case UMC::B_PICTURE:
+        case  VC1_B_FRAME:
+        case  VC1_BI_FRAME:
             extFrameInfo->FrameType = MFX_FRAMETYPE_B;
             break;
-        // NONE_PICTURE
-        default:
+        default:// unexpected type
             extFrameInfo->FrameType = MFX_FRAMETYPE_UNKNOWN;
+            assert(0);
```
mgonchar: add a ```break``` for consistency

_studio/mfx_lib/decode/vc1/src/mfx_vc1_decode.cpp
akokoshn: @mgonchar , same construction using above in this function (see line 1891).
But I agree, make sense add check.
Right now there pointers can't be null, but if call function from another place possible an issues

onabiull: Considering this function - looks like no. Still it doesn't look like good coding practice.

PR #1251, opened 2019-03-08
vpp: clean up some usage of __UINT64

mgonchar:


mgonchar:
This is clean up of real usage of these types. There are more typedefs in different headers, which should be cleaned up after we made some decision about type styles

PR #1253, opened 2019-03-09
[sample_multi_transcode] Enable tiles for VP9 encoder

jonrecker:
New test suite: smt_vp9e_tile


jonrecker:
Thanks. Sample is updated to show a warning that these parameters are ignored for codec != VP9.
Updated commit = 4566324

PR #1257, opened 2019-03-12
[H265][ENC] Deletes hardcode

aulitin:


_studio/mfx_lib/encode_hw/h265/include/mfx_h265_encode_hw_ddi.h
```diff
@@ -119,7 +119,7 @@ typedef enum tagENCODER_TYPE
 } ENCODER_TYPE;
 
 DriverEncoder* CreatePlatformH265Encoder(VideoCORE* core, ENCODER_TYPE type = ENCODER_DEFAULT);
-mfxStatus QueryHwCaps(VideoCORE* core, GUID guid, ENCODE_CAPS_HEVC & caps);
+mfxStatus QueryHwCaps(VideoCORE* core, GUID guid, ENCODE_CAPS_HEVC & caps, mfxU32 width, mfxU32 height);
```
mgonchar: I don't like expansion of interfaces. Looks like better solution is to pass ```const MfxVideoParam &``` object here + remove guid, since you can obtain it from ```MfxVideoParam```

dmitry-gurulev: @mgonchar, from that perspective passing of `const MfxVideoParam &` is more dramatically expansion. Proposed solution uses minimum required info when `const MfxVideoParam &` is like a `void*`

mgonchar: I like my variant more for more elegant interface. In future we may need some more parameters and just adding new inputs to function is a bad practice as for me.
Ok, since there are different opinions, I won't put -1 for this

PR #1262, opened 2019-03-13
[samples][docs] Converted xml dita for samples docs to markdown

MattClinebell:
Deleted samples readme pdfs. Converted xml dita for each sample to markdown. Updated samples command line options with up to date -? output from samples.

### DIFF
--- a/doc/samples/readme-decode_linux.md
+++ b/doc/samples/readme-decode_linux.md
@@ -1,0 +1,190 @@
+
+# Decoding Sample
+
+## Overview
+
+**Decoding Sample** works with **Intel® Media Server Studio 2018 - SDK for Linux\* Server** \(hereinafter referred to as "**SDK**"\).
+
+It demonstrates how to use the **SDK** API to create a sample console application that performs decoding of various video compression formats.
+
+The sample is able to work with **HEVC Decoder & Encoder** \(hereinafter referred to as "**HEVC**"\).
+
+**Note:** To run HEVC, please read the instructions in the “HEVC Plugin” section carefully.
+
+## Features
+
+**Decoding Sample** supports the following video formats:
+
+|Format type| |
+|---|---|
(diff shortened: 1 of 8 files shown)


doc/samples/readme-decode_linux.md
```diff
@@ -0,0 +1,190 @@
+
+# Decoding Sample
+
+## Overview
+
+**Decoding Sample** works with **Intel® Media Server Studio 2018 - SDK for Linux\* Server** \(hereinafter referred to as "**SDK**"\).
```
dvrogozh: I think we should not refer to MSS in open source documentation. We have a flow of people who are confused which version to use. Moreover, these people submit issues which turn out to be a wild mix of closed source and open source components, hence nothing work for them.

jonrecker: That sounds fine. What is the correct product name to use instead of MSS in the samples documentation?

I see we also refer to MSS with certain branches and tags in open source, e.g.
https://github.com/Intel-Media-SDK/MediaSDK/tree/mss2018_r2
https://github.com/Intel-Media-SDK/MediaSDK/releases/tag/MSS-2018-R2.2

dvrogozh: I think that's just "Intel Media SDK", but better to double check with @onabiull.

Yes, unfortunately we have some branches/tags with MSS names. I believe we agreed that there will be no more release branches other than open source release branches. I hope we will be able to stick to this. As for the tags, then optimally there should be no other tags except open source release tags. Correct thing to do seems to reuse approach from Linux distros which are doing something similar to:
* They create a repo where they store package generation information. For example: https://salsa.debian.org/debian/intel-mediasdk. This repo might have mediasdk sources inside or it might just have a .spec file which points to the particular open source release.
* They might have some custom patches on top of the release if needed. For example: https://salsa.debian.org/multimedia-team/intel-media-driver-non-free/tree/master/debian/patches

It will take a while to adopt such scheme though.

onabiull: Just "Media SDK" would be correct term. Please remove specific release information. I believe it would be much better to refer to supported Media SDK API version.

PR #1263, opened 2019-03-14
samples: use std syncronization primitives

mgonchar:
Replaced MSDKMutex with std::mutex and appropriate guards in most trivial cases

### DIFF
--- a/samples/sample_common/include/sample_utils.h
+++ b/samples/sample_common/include/sample_utils.h
@@ -1,5 +1,5 @@
 /******************************************************************************\
-Copyright (c) 2005-2018, Intel Corporation
+Copyright (c) 2005-2019, Intel Corporation
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
@@ -24,6 +24,9 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include <string>
 #include <sstream>
 #include <vector>
+#include <map>
+#include <stdexcept>
+#include <mutex>
 
 #include "mfxstructures.h"
 #include "mfxvideo.h"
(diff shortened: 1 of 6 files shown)


samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -4599,25 +4607,22 @@ mfxStatus SafetySurfaceBuffer::ReleaseSurface(mfxFrameSurface1* pSurf)
             if (0 == it->Locked)
             {
                 m_SList.erase(it);
-                m_mutex.Unlock();
+                lock.unlock();
 
                 // event operation should be out of synced context
```
dmitry-gurulev: Why is it so?

mgonchar: Hard to say. I believe it is related to event synchronization somehow. I plan to get rid of ```MSDKEvent``` in favor of ```std::condition_variable``` then I believe this code will be gone

fzhar: I believe that was done to avoid deadlock - because there's internal mutex in Event class

mgonchar:
@fzhar yes, that is the idea :) We will, but after refactoring of ```CBuffering``` class in mfx_buffering. It is not as straightforward as these changes, so I decided to split clean up in separate PRs. The goal to eliminate all custom synchronization primitives and classes not only including mutexes, but also custom semaphores, events and threads

samples/sample_common/include/sample_utils.h
```diff
@@ -47,7 +47,8 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "avc_nal_spl.h"
 
 #ifdef ENABLE_MCTF
-#include  <stdexcept>
+#include <stdexcept>
```
dvrogozh: how stdexcept and mutex include files depend on MCTF feature? I suggest to avoid overuse of ifdefs whenever possible, this makes code simpler. I don't see harm to always have these headers.

mgonchar: I think idea was to drop headers used only by MCTF (MCTF code is also under define in this file). Agree with you to drop extra preprocessor logic

mgonchar: Done

PR #1265, opened 2019-03-14
Use template utility function for values alignment

mgonchar:


_studio/shared/include/mfx_utils.h
```diff
@@ -145,6 +145,12 @@ constexpr const T& clamp( const T& v, const T& lo, const T& hi, Compare comp )
 {
     return comp(v, lo) ? lo : comp(hi, v) ? hi : v;
 }
+
+template<class T> inline
+T align_value(size_t value, size_t alignment = 16)
+{
```
dmitry-gurulev: This will only work for `[alignment = power of two]` and fail for, as e.g.:
`x = align_value(2, 3); //expects 3 but result is 4`.
IMO, we need to add assertion that `alignment % 2 == 0`.
Yet another option is
```
size_t align(size_t value, size_t alignment)
{
    return alignment ?
        alignment * ((value + alignment - 1) / alignment) : value;
};
```
but it looks overkill on our case (MSDK doesn't use such weird alignments)

The second thing is using of arg. type deduction instead of specifying it explicitly -
```
template <typename T>
inline T align_value(T value, size_t alignment = 16);
```
since common usage pattern is 
`value  = mfx::align_value<type_of_value>(value,  alignment);`

Using explicit specification is error-prone, consider this
`value  = mfx::align_value<int16>(value,  alignment);` when actual type of value is `int32`

mgonchar: @dmitry-gurulev good catch!
actually, ```alignment % 2 == 0``` doesn't mean that ```alignment``` is power of two :)
I think we can add in if statement to switch between realization to avoid division (actually need to benchmark and figure out if it has any sense to introduce complicated logic at all)

Agree with you about type deduction.

So currently I'm thinking about:
```
template<class T> inline
constexpr T align_value(size_t value, size_t alignment = 16)
{
   if (is_power_of_2(alignment))
      return static_cast<T> ((value + (alignment - 1)) & ~(alignment - 1));

   return alignment ?
        alignment * ((value + alignment - 1) / alignment) : value;
}
```

dmitry-gurulev: @mgonchar,
> actually, alignment % 2 == 0 doesn't mean that alignment is power of two :)

Yes, u're right =) Maybe this one - `assert((alignment & (~alignment + 1)) == alignment);` ?

>I think we can add in if statement to switch between realization to avoid division

I don't think the division is show stopper, compilers are smart enough to replace it w/ shift when `alignment` is compile time constant, for me branching is worse. Agree that benchmark is required rather than speculations though.

Anyway, I still think this is overkill for MSDK, we don't use such alignments and we aren't common purpose lib like STL.

mgonchar: @dmitry-gurulev If we decide to drop support of non-power of two alignments (which is OK as for me), I think we should change the name to reflect it somehow. Like ```align_value_to_power_of_2```.

Yes, your variant of power of two checking should work) But it will say true for zero, not sure if it is an actual problem

dmitry-gurulev: align2_value

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode.cpp
```diff
@@ -177,8 +177,8 @@ Status MJPEGVideoDecoderMFX::Close(void)
 
 void MJPEGVideoDecoderMFX::AdjustFrameSize(mfxSize & size)
 {
-    size.width  = mfx::align_value<uint32_t>(size.width, 16);
-    size.height = mfx::align_value<uint32_t>(size.height, m_interleaved ? 32 : 16);
+    size.width  = mfx::align2_value<uint32_t>(size.width, 16);
+    size.height = mfx::align2_value<uint32_t>(size.height, m_interleaved ? 32 : 16);
```
dmitry-gurulev: Do we need exp. spec. here?

_studio/shared/umc/core/umc/src/umc_video_data.cpp
```diff
@@ -293,7 +294,7 @@ Status VideoData::SetColorFormat(ColorFormat cFormat)
             align *= pFormat->m_PlaneFormatInfo[i].m_iAlignMult;
         }
         m_pPlaneData[i].m_nPitch =
-            align_value<size_t>(m_pPlaneData[i].m_ippSize.width * bpp, align);
+            mfx::align2_value<size_t>(m_pPlaneData[i].m_ippSize.width * bpp, align);
```
mgonchar: I'm worrying about this one. It looks that ```align``` here could be not power of two

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode.cpp
mgonchar: I think - not

_studio/shared/umc/core/umc/src/umc_video_data.cpp
dmitry-gurulev: Well, I think actual problem is
```
bpp = m_pPlaneData[i].m_iSampleSize * m_pPlaneData[i].m_iSamples;
align = MFX_MAX(m_iAlignment, bpp); 
```
because `m_iAlignMult` is either 2 or 1 for all supported `ColorFormatInfo`
`m_iAlignment` is not an issue because it is power of two (see `VideoData::SetAlignment`). A problem would be `bpp` which is `m_iSampleSize * m_iSamples`, for example for RGB24 `m_iSampleSize = (iBitDepth+7)>>3`  == 3. But `ColorFormatInfo` is defined w/ `m_iMinAlign` = 4 and we have `align = MFX_MAX(align, pFormat->m_iMinAlign)`

Anyway, we just improve (add assertion) behavior because old function works in the same way. I suggest to move on.

mgonchar: Sure, I know that behavior here is not changed

mgonchar:
@dmitry-gurulev do you have any other concerns?

_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -361,9 +361,9 @@ mfxDefaultAllocatorVAAPI::AllocFramesHW(
             }
             else
             {
-                int width32 = 32 * ((request->Info.Width + 31) >> 5);
-                int height32 = 32 * ((request->Info.Height + 31) >> 5);
-                codedbuf_size = static_cast<mfxU32>((width32 * height32) * 400LL / (16 * 16));
+                int aligned_width  = mfx::align_value<int>(request->Info.Width,  32);
+                int aligned_height = mfx::align_value<int>(request->Info.Height, 32);
+                codedbuf_size = static_cast<mfxU32>((aligned_width * aligned_height) * 400LL / (16 * 16));
```
dmitry-gurulev: `400LL` - do we really need `long long` here?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -1178,24 +1178,20 @@ namespace MPEG2EncoderHW
         {
             bInvalid = true;
         }
-        if ((pFrameInfo->Width  & 15) !=0)
+        if ((pFrameInfo->Width & 15) != 0)
```
dmitry-gurulev: IMO, it makes sense to add utility function like `bool aligned(value, alignment)` for places like this

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -1178,24 +1178,20 @@ namespace MPEG2EncoderHW
...
+            pFrameInfo->Width = mfx::align_value<mfxU16>(pFrameInfo->CropW, 16);
 
-        if (bProgressiveSequence)
-        {
-            if ((pFrameInfo->Height  & 15) !=0)
-                return MFX_ERR_INVALID_VIDEO_PARAM;
-            pFrameInfo->Height = pFrameInfo->CropH ? ((pFrameInfo->CropH + 15)/16)*16 : pFrameInfo->Height;
+        mfxU32 heightAlignment = bProgressiveSequence ? 16 : 32;
+
+        if ((pFrameInfo->Height & (heightAlignment - 1)) != 0)
```
dmitry-gurulev: The same as for line 1181

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1246,15 +1246,14 @@ void CUQPMap::Init (mfxU32 picWidthInLumaSamples, mfxU32 picHeightInLumaSamples)
 {
 
     //16x32 only: driver limitation
-    m_width        = (picWidthInLumaSamples   + 31) / 32;
-    m_height       = (picHeightInLumaSamples  + 31) / 32;
-    m_pitch  = (((picWidthInLumaSamples + 31) >> 5) + 63) &~ 63;
-    m_h_aligned = (((picHeightInLumaSamples + 31) >> 5) + 3) &~ 3;
+    m_width        = (picWidthInLumaSamples  + 31) / 32;
+    m_height       = (picHeightInLumaSamples + 31) / 32;
+    m_pitch        = mfx::align_value<mfxU32>(m_width, 64);
```
dmitry-gurulev: IMO, it makes sense to add utility function like bool aligned(value, alignment) for places like this

_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
```diff
@@ -426,7 +426,7 @@ void PackerVA::CreateSliceDataBuffer(H264DecoderFrameInfo * pSliceInfo)
         size += NalUnitSize;
     }
 
-    uint32_t const AlignedNalUnitSize = mfx::align_value<uint32_t>(size, 128);
+    uint32_t const AlignedNalUnitSize = mfx::align2_value<uint32_t>(size, 128);
```
dmitry-gurulev: Do we still need explicit `uint32_t` here?

_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer.cpp
```diff
@@ -378,7 +378,7 @@ void PackerVA::CreateSliceDataBuffer(H265DecoderFrameInfo const* sliceInfo)
         size += NalUnitSize + 3;
     }
 
-    AlignedNalUnitSize = mfx::align_value<int32_t>(size, 128);
+    AlignedNalUnitSize = mfx::align2_value<int32_t>(size, 128);
```
dmitry-gurulev: Do we still need explicit `int32_t` here?

BTW, `size/AlignedNalUnitSize` are naturally unsigned, need to fix this at separate PR.

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode.cpp
dmitry-gurulev: Maybe I missed something but I don't see a change.

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_dec_frame_descr.cpp
```diff
@@ -50,9 +50,9 @@ bool VC1FrameDescriptor::Init(uint32_t         DescriporID,
     if (!m_pContext)
     {
         uint8_t* ptr = NULL;
-        ptr += mfx::align_value<uint32_t>(sizeof(VC1Context));
-        ptr += mfx::align_value<uint32_t>(sizeof(VC1PictureLayerHeader)*VC1_MAX_SLICE_NUM);
-        ptr += mfx::align_value<uint32_t>((HeightMB*seqLayerHeader->MaxWidthMB*VC1_MAX_BITPANE_CHUNCKS));
+        ptr += mfx::align2_value<uint32_t>(sizeof(VC1Context));
+        ptr += mfx::align2_value<uint32_t>(sizeof(VC1PictureLayerHeader)*VC1_MAX_SLICE_NUM);
+        ptr += mfx::align2_value<uint32_t>((HeightMB*seqLayerHeader->MaxWidthMB*VC1_MAX_BITPANE_CHUNCKS));
```
dmitry-gurulev: Do we still need explicit `uint32_t` here?

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_dec_frame_descr.cpp
```diff
@@ -66,11 +66,11 @@ bool VC1FrameDescriptor::Init(uint32_t         DescriporID,
         m_pContext->bp_round_count = -1;
         ptr = (uint8_t*)m_pContext;
 
-        ptr += mfx::align_value<uint32_t>(sizeof(VC1Context));
+        ptr += mfx::align2_value<uint32_t>(sizeof(VC1Context));
         m_pContext->m_picLayerHeader = (VC1PictureLayerHeader*)ptr;
         m_pContext->m_InitPicLayer = m_pContext->m_picLayerHeader;
 
-        ptr += mfx::align_value<uint32_t>((sizeof(VC1PictureLayerHeader)*VC1_MAX_SLICE_NUM));
+        ptr += mfx::align2_value<uint32_t>((sizeof(VC1PictureLayerHeader)*VC1_MAX_SLICE_NUM));
```
dmitry-gurulev: Do we still need explicit `uint32_t` here?

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_dec_task_store.cpp
```diff
@@ -177,16 +177,16 @@ namespace UMC
...
         for (uint32_t counter = 0; counter < m_iNumFramesProcessing; counter++)
         {
                 if (pMainVC1Decoder->m_va)
                 {
-                    Size += mfx::align_value<uint32_t>(sizeof(VC1FrameDescriptorVA_Linux<VC1PackerLVA>));
+                    Size += mfx::align2_value<uint32_t>(sizeof(VC1FrameDescriptorVA_Linux<VC1PackerLVA>));
                 }
                 else
-                    Size += mfx::align_value<uint32_t>(sizeof(VC1FrameDescriptor));
+                    Size += mfx::align2_value<uint32_t>(sizeof(VC1FrameDescriptor));
```
dmitry-gurulev: Do we still need explicit `uint32_t` here and above?

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder.cpp
```diff
@@ -310,14 +310,14 @@ Status VC1VideoDecoder::ContextAllocation(uint32_t mbWidth,uint32_t mbHeight)
...
-        ptr += mfx::align_value<uint32_t>(sizeof(VC1TSHeap));
+        ptr += mfx::align2_value<uint32_t>(sizeof(VC1Context));
+        ptr += mfx::align2_value<uint32_t>(sizeof(VC1VLCTables));
+        ptr += mfx::align2_value<uint32_t>(sizeof(int16_t)*mbHeight*mbWidth*2*2);
+        ptr += mfx::align2_value<uint32_t>(mbHeight*mbWidth);
+        ptr += mfx::align2_value<uint32_t>(sizeof(VC1TSHeap));
         if(m_stCodes == NULL)
         {
-            ptr += mfx::align_value<uint32_t>(START_CODE_NUMBER*2*sizeof(uint32_t)+sizeof(MediaDataEx::_MediaDataEx));
+            ptr += mfx::align2_value<uint32_t>(START_CODE_NUMBER*2*sizeof(uint32_t)+sizeof(MediaDataEx::_MediaDataEx));
```
dmitry-gurulev: Do we still need explicit `uint32_t` here and above?

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder.cpp
```diff
@@ -467,7 +467,7 @@ Status VC1VideoDecoder::StartCodesProcessing(uint8_t*   pBStream,
                 (uint8_t*)m_frameData->GetDataPointer() + *pOffsets,
                 UnitSize);
             //use own buffer
-            SwapData(m_dataBuffer, mfx::align_value<uint32_t>(UnitSize));
+            SwapData(m_dataBuffer, mfx::align2_value<uint32_t>(UnitSize));
```
dmitry-gurulev: A lot of explicit `uint32_t` here and above, do we really need them?

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder_hw.cpp
```diff
@@ -127,13 +127,13 @@ uint32_t VC1VideoDecoderHW::CalculateHeapSize()
...
+    Size += mfx::align2_value<uint32_t>(sizeof(VC1TaskStore));
     if (!m_va)
-        Size += mfx::align_value<uint32_t>(sizeof(Frame)*(2*m_iMaxFramesInProcessing + 2*VC1NUMREFFRAMES));
+        Size += mfx::align2_value<uint32_t>(sizeof(Frame)*(2*m_iMaxFramesInProcessing + 2*VC1NUMREFFRAMES));
     else
-        Size += mfx::align_value<uint32_t>(sizeof(Frame)*(m_SurfaceNum));
+        Size += mfx::align2_value<uint32_t>(sizeof(Frame)*(m_SurfaceNum));
 
-    Size += mfx::align_value<uint32_t>(sizeof(MediaDataEx));
+    Size += mfx::align2_value<uint32_t>(sizeof(MediaDataEx));
```
dmitry-gurulev: Do we still need explicit `uint32_t` here and above?

_studio/shared/umc/core/umc/src/umc_video_data.cpp
dmitry-gurulev: Do we still need explicit `size_t` here and above?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
mgonchar: agree, but this would be the topic for different PR

_studio/shared/src/libmfx_allocator_vaapi.cpp
mgonchar: This question need more investigation. This code is some copypaste from driver code as I remember. Need just to check which result we expected here

_studio/shared/umc/codec/h264_dec/src/umc_h264_va_packer.cpp
mgonchar: no, we don't. I missed this one in cleanup commit. Will update

_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer.cpp
mgonchar: yes, I will update

_studio/shared/umc/codec/jpeg_dec/src/umc_mjpeg_mfx_decode.cpp
mgonchar: I missed this one in my update commit

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder_hw.cpp
mgonchar: no

_studio/shared/umc/core/umc/src/umc_video_data.cpp
mgonchar: no

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder.cpp
mgonchar: no

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_video_decoder.cpp
mgonchar: no

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_dec_task_store.cpp
mgonchar: no

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_dec_frame_descr.cpp
mgonchar: no

_studio/shared/umc/codec/vc1_dec/src/umc_vc1_dec_frame_descr.cpp
mgonchar: no

mgonchar:
@dvrogozh fixed

PR #1266, opened 2019-03-14
vp9e: disable copy constructor for MfxFrameAllocResponse class

eshiryae:


_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
```diff
@@ -435,6 +435,8 @@ template <typename T> mfxStatus RemoveExtBuffer(T & par, mfxU32 id)
     public:
         MfxFrameAllocResponse();
 
+        MfxFrameAllocResponse(MfxFrameAllocResponse const &) = delete;
```
mgonchar: It was disabled previously as well. You did it in more correct way according to modern c++ standard. But you didn't fix static analysis issue by this change. You need to do something with copy assignment as well (and with move ctor and move assignment)

_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
```diff
@@ -435,6 +435,9 @@ template <typename T> mfxStatus RemoveExtBuffer(T & par, mfxU32 id)
...
+        MfxFrameAllocResponse & operator =(MfxFrameAllocResponse const &) = delete;
```
mgonchar: Let's follow rule of five and define move ctor and move assignment explicitly as well (I believe they should also be deleted)

eshiryae: Okay, I will add them too.
and I guess the same should be done for other encoders?

mgonchar: Well, ideally - yes :) But not as part of this PR, because this one is dedicated to vp9e. Idea is that if you changing some code it is time to fix problems that you see in it. But it is up to you whether you decide to continue patching other classes

Issue #1027, opened 2018-12-18
[MFX] [HWVPP] HEVC 10 bit hwdownload VPP issue (unmatched result )


fulinjie:
CMD Line: 
$ ffmpeg -v debug -y -hwaccel qsv -c:v hevc_qsv -load_plugin hevc_hw -i $src_file -vf "hwdownload,format=p010" qsv.yuv
$(ffmpeg-3.4.1) ffmpeg -v verbose -y -c:v hevc_qsv -load_plugin hevc_hw -i $src_file -pix_fmt p010le qsv_341_no_hw.yuv
$ ffmpeg -hwaccel vaapi -v verbose -y -i $src_file -pix_fmt p010le  vaapi.yuv
$ ffmpeg -v verbose -y -i $src_file -pix_fmt p010le -vframes 1 ref.yuv

compare the output:
# md5sum *.yuv
653e42290412ebc17c3681d7fea1ab90  qsv_341_no_hw.yuv
b9c9436b729f2a1bb0e0ab0d5c25ccdc  qsv.yuv
653e42290412ebc17c3681d7fea1ab90  ref.yuv
653e42290412ebc17c3681d7fea1ab90  vaapi.yuv

Apply the patch in libva to enable va_TraceSurface support for P010, and dump the surface to check the decode result:
https://github.com/intel/libva/pull/260
export LIBVA_TRACE_SURFACE=qsvdec

Checked the libva output file from vaapi, qsv, qsv_without_HWVPP, they all got the same output.
I think it means decoding is all correct.

Tracing into MSDK and Media driver, and dump the data from Surface ID to check where this issue occurs:

va_TraceSurface// same Y data with ref.yuv

 VAAPIVideoCORE::DoFastCopyWrapper:
sts = GetFrameHDL(srcMemId, &srcHandle);// already wrong different Y data

Dump the [PostAll].p010 and [preALL].p010 from surface After VPP and pre-VPP:

$ od -A x -t x1z -v 'surfdump_loc[postALL]_lyr[0]_f[000]_w[3840]_h[2176]_p[7680].p010'  | head -n 2
000000 40 b2 80 b2 c0 b2 00 b3 80 b3 00 b4 80 b4 00 b5  >@...............<
000010 c0 b5 40 b6 c0 b6 40 b7 80 b7 c0 b7 00 b8 00 b8  >..@...@.........<

[...]

PR #1268, opened 2019-03-15
[vpp] Enable copy pass thru for VAAPI

Alexandr-Konovalov:
Issue: MDP-53850
Fixes: #1027

### DIFF
--- a/_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
+++ b/_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
@@ -908,6 +908,8 @@ namespace MfxHwVideoProcessing
             mfxFrameSurface1 *pInputSurface,
             mfxFrameSurface1 *pOutputSurface);
 
+        bool UseCopyPassThrough(const DdiTask *pTask) const;
+
         mfxStatus PreWorkOutSurface(ExtSurface & output);
         mfxStatus PreWorkInputSurface(std::vector<ExtSurface> & surfQueue);
 
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -2962,6 +2962,28 @@ mfxStatus VideoVPPHW::Close()
 } // mfxStatus VideoVPPHW::Close()
 
 
+bool VideoVPPHW::UseCopyPassThrough(const DdiTask *pTask) const
+{
+    if (!m_config.m_bCopyPassThroughEnable
+        || IsRoiDifferent(pTask->input.pSurf, pTask->output.pSurf))
+    {
+        return false;
+    }
+
+    if (m_pCore->GetVAType() != MFX_HW_VAAPI || m_ioMode != D3D_TO_D3D)
+    {
+        return true;
+    }
+
+    // Under VAAPI for D3D_TO_D3D either CM copy or VPP pipeline are preferred by
+    // performance reasons. So, before enabling CopyPassThrough for a task we check
+    // for CM availability to do fallback to VPP if there is no CM.
+    // This can't be done in ConfigureExecuteParams() because CmCopy status is set later.
+    const VAAPIVideoCORE *vaapiCore = dynamic_cast<VAAPIVideoCORE*>(m_pCore);
+    MFX_CHECK_WITH_ASSERT(vaapiCore, false);
+    return vaapiCore->CmCopy();
+}
+
 mfxStatus VideoVPPHW::VppFrameCheck(
                                     mfxFrameSurface1 *input,
                                     mfxFrameSurface1 *output,
@@ -3115,9 +3137,11 @@ mfxStatus VideoVPPHW::VppFrameCheck(
 #endif
     MFX_CHECK_STS(sts);
 
+    bool useCopyPassThrough = UseCopyPassThrough(pTask);
+
     if (VPP_SYNC_WORKLOAD == m_workloadMode)
     {
-        pTask->bRunTimeCopyPassThrough = (true == m_config.m_bCopyPassThroughEnable && false == IsRoiDifferent(pTask->input.pSurf, pTask->output.pSurf));
+        pTask->bRunTimeCopyPassThrough = useCopyPassThrough;
 
         // submit task
         SyncTaskSubmission(pTask);
@@ -3136,8 +3160,7 @@ mfxStatus VideoVPPHW::VppFrameCheck(
     }
     else
     {
-        if (false == m_config.m_bCopyPassThroughEnable ||
-            (true == m_config.m_bCopyPassThroughEnable && true == IsRoiDifferent(pTask->input.pSurf, pTask->output.pSurf)))
+        if (!useCopyPassThrough)
         {
             pTask->bRunTimeCopyPassThrough = false;
 
@@ -6104,7 +6127,6 @@ mfxStatus ConfigureExecuteParams(
         executeParams.bSceneDetectionEnable = false;
     }
 
-#if defined(WIN64) || defined (WIN32)
     if ( (0 == memcmp(&videoParam.vpp.In, &videoParam.vpp.Out, sizeof(mfxFrameInfo))) &&
          executeParams.IsDoNothing() )
     {
@@ -6115,7 +6137,6 @@ mfxStatus ConfigureExecuteParams(
         config.m_bCopyPassThroughEnable = false;// after Reset() parameters may be changed,
                                             // flag should be disabled
     }
-#endif//m_bCopyPassThroughEnable == false for another OS
 
     if (inDNRatio == outDNRatio && !executeParams.bVarianceEnable && !executeParams.bComposite &&
             !(config.m_extConfig.mode == IS_REFERENCES) )
--- a/_studio/shared/include/libmfx_core_vaapi.h
+++ b/_studio/shared/include/libmfx_core_vaapi.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -121,9 +121,10 @@ public:
         CMEnabledCoreAdapter(VAAPIVideoCORE *pVAAPICore): m_pVAAPICore(pVAAPICore)
         {
         };
-        virtual mfxStatus SetCmCopyStatus(bool enable)
+        virtual mfxStatus SetCmCopyStatus(bool enable) override
         {
-            return m_pVAAPICore->SetCmCopyStatus(enable);
+            m_pVAAPICore->SetCmCopy(enable);
+            return MFX_ERR_NONE;
         };
     protected:
         VAAPIVideoCORE *m_pVAAPICore;
@@ -175,7 +176,9 @@ public:
     mfxStatus              GetVAService(VADisplay *pVADisplay);
 
     // this function should not be virtual
-    mfxStatus SetCmCopyStatus(bool enable);
+    void SetCmCopy(bool enable);
+
+    bool CmCopy() const { return m_bCmCopy; }
 
 protected:
     VAAPIVideoCORE(const mfxU32 adapterNum, const mfxU32 numThreadsAvailable, const mfxSession session = NULL);
--- a/_studio/shared/src/libmfx_core_vaapi.cpp
+++ b/_studio/shared/src/libmfx_core_vaapi.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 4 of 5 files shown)


_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
```diff
@@ -908,6 +908,8 @@ namespace MfxHwVideoProcessing
             mfxFrameSurface1 *pInputSurface,
             mfxFrameSurface1 *pOutputSurface);
 
+        mfxStatus IsCopyPassThroughPreferred(bool &preferred) const;
```
mgonchar: This design looks ugly, this function should simply return ```bool```

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -3115,9 +3134,16 @@ mfxStatus VideoVPPHW::VppFrameCheck(
 #endif
     MFX_CHECK_STS(sts);
 
+    bool useCopyPassThrough = false;
+    if (true == m_config.m_bCopyPassThroughEnable && false == IsRoiDifferent(pTask->input.pSurf, pTask->output.pSurf))
```
mgonchar: I don't like ```bool``` to ```bool``` conversions.
As for me
```
if (m_config.m_bCopyPassThroughEnable && !IsRoiDifferent(pTask->input.pSurf, pTask->output.pSurf))
```
is much better.
But I won't insist of changing this if you like your approach

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2962,6 +2962,25 @@ mfxStatus VideoVPPHW::Close()
 } // mfxStatus VideoVPPHW::Close()
 
 
+mfxStatus VideoVPPHW::IsCopyPassThroughPreferred(bool &preferred) const
+{
+    if (m_pCore->GetVAType() != MFX_HW_VAAPI || m_ioMode != D3D_TO_D3D)
```
mgonchar: Doesn't this check mean that vaapi core is used?
I.e. cast on line 2979 wouldn't fail?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: This is legacy. But I'm for the change, sure.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: @mgonchar  Yes, the cast must not fail. I don't like dropping the check, however. What can I do here, as there is no asserts in release and you think returning mfxStatus is ugly?

mgonchar: well, currently I see two different ways:
1) Leave the current code, but return false instead of error if cast fails (which is supposed to be almost impossible situation). So we just silently hide the error (not actually so, because we can use trace macro to at least signal to stdout if traces are on)
2) Move to references casting, then some upper layers should deal with possible exception (which is supposed to be almost impossible situation). So we loose information about place of error occurrence. 

So both of these ways are definitely not much better than current approach. I will try to figure out better solution...

_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
dvrogozh: I don't like this function:). Both prototype (+1 to @mgonchar) and how it named/what it does. It returns not deterministic result: "I check copy pass thru, maybe I like it maybe not - decide later". I would like to see function named useSOMETHING() with the obvious inclination.

To be honest, I got lost in CopyPassThrough name. What it actually mean? Copy with memcpy or CM?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
dvrogozh: Asserts _are_ enabled in release builds of master branch. They are switched off in builds of release branches. So, use assert if you feel it right.

_studio/shared/include/libmfx_core_vaapi.h
```diff
@@ -177,6 +177,8 @@ class VAAPIVideoCORE : public CommonCORE
     // this function should not be virtual
     mfxStatus SetCmCopyStatus(bool enable);
 
+    bool GetCmCopyStatus() const { return m_bCmCopy; }
```
dvrogozh: Status in function name is misleading. Can we just have `SetCmCopy(bool)` and `bool CmCopy()`?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: Here I want to be protected from some fearless code refactoring. So checking on master' release is just fine.

_studio/mfx_lib/vpp/include/mfx_vpp_hw.h
Alexandr-Konovalov: I tried to say something like "performance of CopyPassThrough is good, you want to use it" vs "it's better to omit CopyPassThrough" and to hide low-level details what is used inside and what is faster.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
dmitry-gurulev: Having virtual `GetCmCopyStatus` in core is not an option to avoid this cast which is ugly on its own?

Alexandr-Konovalov: I believe adding new virtual functions to `CommonCORE` is prohibited. That's why the cast was applied here.

_studio/shared/include/libmfx_core_vaapi.h
Alexandr-Konovalov: Done.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: Done.

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -813,7 +813,7 @@ VAAPIVideoCORE::GetVAService(
 } // mfxStatus VAAPIVideoCORE::GetVAService(...)
 
 mfxStatus
-VAAPIVideoCORE::SetCmCopyStatus(bool enable)
+VAAPIVideoCORE::SetCmCopy(bool enable)
```
mgonchar: I believe no sense for this function to return status

Alexandr-Konovalov: @mgonchar  Is it possible to change `mfxStatus CMEnabledCoreInterface::SetCmCopy(bool enable)` to return void as well? Can it affect ABI of some plugins?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2962,6 +2962,22 @@ mfxStatus VideoVPPHW::Close()
 } // mfxStatus VideoVPPHW::Close()
 
 
+bool VideoVPPHW::UseCopyPassThrough() const
+{
+    if (m_pCore->GetVAType() != MFX_HW_VAAPI || m_ioMode != D3D_TO_D3D)
+    {
+        return true; // preferred unconditionally
```
dvrogozh: after function rename this comment is misleading

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -3115,9 +3131,13 @@ mfxStatus VideoVPPHW::VppFrameCheck(
 #endif
     MFX_CHECK_STS(sts);
 
+    bool useCopyPassThrough =
+        (m_config.m_bCopyPassThroughEnable && !IsRoiDifferent(pTask->input.pSurf, pTask->output.pSurf))?
```
dvrogozh: Why you don't want to move this condition to UseCopyPassThrough() as well?

_studio/shared/src/libmfx_core_vaapi.cpp
dvrogozh: @onabiull : please, comment.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: Done.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: Done.

_studio/shared/src/libmfx_core_vaapi.cpp
onabiull: If we wan't old plugins to remain functional, we need change interfaces by adding new ones. This would affect windows version in the first hand.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2962,6 +2962,28 @@ mfxStatus VideoVPPHW::Close()
...
+        return true;
+    }
+
+    // Under VAAPI for D3D_TO_D3D either CM copy or VPP pipeline are preferred by
+    // performance reasons. So, before enabling CopyPassThrough for a task we check
+    // for CM availability to do fallback to VPP if there is no CM.
+    // This can't be done in ConfigureExecuteParams() because CmCopy status is set later.
+    const VAAPIVideoCORE *vaapiCore = dynamic_cast<VAAPIVideoCORE*>(m_pCore);
+    assert(vaapiCore);
+    return vaapiCore->CmCopy();
```
dvrogozh: you know, I am not sure I like this code. I would probably go with just:
```
const VAAPIVideoCORE *vaapiCore = dynamic_cast<VAAPIVideoCORE*>(m_pCore);
return (vaapiCore)? vaapiCore->CmCopy(): false;
```

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2962,6 +2962,28 @@ mfxStatus VideoVPPHW::Close()
 } // mfxStatus VideoVPPHW::Close()
 
 
+bool VideoVPPHW::UseCopyPassThrough(const DdiTask *pTask) const
```
dvrogozh: the name of the function should probably be: Use**Cm**CopyPassThrough(), right?

_studio/shared/src/libmfx_core_vaapi.cpp
dvrogozh: @onabiull : Does this apply to VAAPIVideoCORE or VideoCORE or both?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: There is a huge "Can't happen!" and we are ignoring it silently, aren't we? Why not add 
`if(!vaapiCore) MFX_STS_TRACE();` at least?

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
Alexandr-Konovalov: The plan was to hide mere presence of Cm. Why may a user of CopyPassThrought be interested in what is used inside? One just want fastest version, I think.

dvrogozh: Because it seems that this function implies that CM is going to be used. At least that's what I think looking into last line in its code.

dvrogozh: Ok, I got it after offline discussion. Under copy pass through we understand any mem copy operation without VP algorithms. This can be map+memcpy or cmcopy. In different modes we prefer different copy type.

_studio/shared/src/libmfx_core_vaapi.cpp
Alexandr-Konovalov: If I understand Oleg correctly, `CMEnabledCoreInterface` is part of plugin API and so change of return value's type is prohibited.

So, rename `virtual mfxStatus CMEnabledCoreInterface::SetCmCopyStatus(bool enable)` to `virtual mfxStatus CMEnabledCoreInterface::SetCmCopy(bool enable)` is also incorrect and must be reverted.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2962,6 +2962,28 @@ mfxStatus VideoVPPHW::Close()
...
+    {
+        return true;
+    }
+
+    // Under VAAPI for D3D_TO_D3D either CM copy or VPP pipeline are preferred by
+    // performance reasons. So, before enabling CopyPassThrough for a task we check
+    // for CM availability to do fallback to VPP if there is no CM.
+    // This can't be done in ConfigureExecuteParams() because CmCopy status is set later.
+    const VAAPIVideoCORE *vaapiCore = dynamic_cast<VAAPIVideoCORE*>(m_pCore);
+    assert(vaapiCore);
```
mgonchar: I think here you can use following macro to simplify code + use assert + use traces:
```
MFX_CHECK_WITH_ASSERT(vaapiCore, false);

return vaapiCore->CmCopy();
```

Alexandr-Konovalov: Yep.

PR #1271, opened 2019-03-15
[sample_decode] Fix typo in logging condition

Alexandr-Konovalov:


samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1922,7 +1922,7 @@ mfxStatus CDecodingPipeline::RunDecoding()
 
     if (m_nFrames == m_output_count)
     {
-        if (!sts)
+        if (sts)
```
mgonchar: ```sts != MFX_ERR_NONE``` is more clear as for me

Alexandr-Konovalov: Done.

PR #1273, opened 2019-03-19
[jpegd] fix rotation when output in system memory

Alexandr-Konovalov:
Issue: MDP-53713, MDP-52247

### DIFF
--- a/_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
+++ b/_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -115,11 +115,10 @@ public:
 
 
 protected:
-    VideoVppJpegD3D9 *m_pCc;
-    // True if we need special VPP color conversion after decoding
-    bool   m_needVpp;
     // Decoder's array
     std::unique_ptr<UMC::MJPEGVideoDecoderMFX_HW> m_pMJPEGVideoDecoder;
+    // True if we need special VPP color conversion after decoding
+    bool   m_needVpp;
 
     // Number of pictures collected
     mfxU32 m_numPic;
--- a/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
@@ -1495,7 +1495,6 @@ VideoDECODEMJPEGBase_HW::VideoDECODEMJPEGBase_HW()
     m_va = 0;
     m_dst = 0;
     m_numPic = 0;
-    m_pCc    = NULL;
     m_needVpp = false;
 }
 
@@ -1533,8 +1532,6 @@ mfxStatus VideoDECODEMJPEGBase_HW::Reset(mfxVideoParam *par)
 {
     m_pMJPEGVideoDecoder->Reset();
     m_numPic = 0;
-    delete m_pCc;
-    m_pCc    = 0;
 
     m_vPar = *par;
 
@@ -1570,9 +1567,6 @@ mfxStatus VideoDECODEMJPEGBase_HW::Close(void)
     m_pMJPEGVideoDecoder->Close();
     m_numPic = 0;
     m_isOpaq = false;
-
-    delete m_pCc;
-    m_pCc    = 0;
 
     {
         std::lock_guard<std::mutex> guard(m_guard);
@@ -1725,7 +1719,6 @@ mfxStatus VideoDECODEMJPEGBase_HW::RunThread(void *params, mfxU32, mfxU32 )
             mfxSts = ((mfx_UMC_FrameAllocator_D3D_Converter *)m_FrameAllocator.get())->CheckPreparingToOutput(info->surface_out,
                                                                                                               info->dst,
                                                                                                               &m_vPar,
-                                                                                                              &m_pCc,
                                                                                                               (mfxU16)info->vppTaskID);
             if(mfxSts != MFX_TASK_DONE)
             {
@@ -1886,7 +1879,7 @@ mfxStatus VideoDECODEMJPEGBase_HW::FillEntryPoint(MFX_ENTRY_POINT *pEntryPoint, 
         ((mfx_UMC_FrameAllocator_D3D_Converter *)m_FrameAllocator.get())->SetJPEGInfo(&info);
 
         // decoding is ready. prepare to output:
-        mfxStatus mfxSts = ((mfx_UMC_FrameAllocator_D3D_Converter *)m_FrameAllocator.get())->StartPreparingToOutput(surface_out, dst, &m_vPar, &m_pCc, &taskId, m_isOpaq);
+        mfxStatus mfxSts = ((mfx_UMC_FrameAllocator_D3D_Converter *)m_FrameAllocator.get())->StartPreparingToOutput(surface_out, dst, &m_vPar, &taskId, m_isOpaq);
         if (mfxSts < MFX_ERR_NONE)
         {
             return mfxSts;
--- a/_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
+++ b/_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -28,53 +28,6 @@
 
 #if defined (MFX_ENABLE_MJPEG_VIDEO_DECODE)
 #include "mfx_vpp_jpeg_d3d9.h"
-
-mfxStatus DoHwJpegCc(VideoVppJpegD3D9 **pCc, 
-                     VideoCORE* mfxCore, 
-                     mfxFrameSurface1 *pDst, 
-                     mfxFrameSurface1 *pSrc, 
-                     const mfxVideoParam * par, 
-                     bool isD3DToSys, 
-                     mfxU16 *taskId)
-{
-    mfxStatus sts = MFX_ERR_NONE;
-
-    if (*pCc == NULL)
-    {
-        *pCc = new VideoVppJpegD3D9(mfxCore, isD3DToSys,(par->IOPattern & MFX_IOPATTERN_OUT_OPAQUE_MEMORY)?true:false);
-
-        sts = (*pCc)->Init(par);
-        MFX_CHECK_STS( sts );
-    }
-
-    sts = (*pCc)->BeginHwJpegProcessing(pSrc, pDst, taskId);
-
-    return sts;
-}
-
-mfxStatus DoHwJpegCcFw(VideoVppJpegD3D9 **pCc, 
-                       VideoCORE* mfxCore, 
-                       mfxFrameSurface1 *pDst, 
-                       mfxFrameSurface1 *pSrcTop, 
-                       mfxFrameSurface1 *pSrcBottom, 
-                       const mfxVideoParam * par, 
-                       bool isD3DToSys, 
-                       mfxU16 *taskId)
-{
-    mfxStatus sts = MFX_ERR_NONE;
-
-    if (*pCc == NULL)
-    {
-        *pCc = new VideoVppJpegD3D9(mfxCore, isD3DToSys,(par->IOPattern & MFX_IOPATTERN_OUT_OPAQUE_MEMORY)?true:false);
-
-        sts = (*pCc)->Init(par);
-        MFX_CHECK_STS( sts );
-    }
-
-    sts = (*pCc)->BeginHwJpegProcessing(pSrcTop, pSrcBottom, pDst, taskId);
-
-    return sts;
-}
 
 UMC::Status mfx_UMC_FrameAllocator_D3D_Converter::InitMfx(UMC::FrameAllocatorParams *,
                                                           VideoCORE* mfxCore,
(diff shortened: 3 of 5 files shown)


_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -180,6 +180,19 @@ mfxStatus mfx_UMC_FrameAllocator_D3D_Converter::StartPreparingToOutput(mfxFrameS
 
     mfxStatus sts = MFX_ERR_NONE;
     bool isD3DToSys = false;
+    mfxVideoParam  localParams;
+    const mfxVideoParam *par = parArg;
```
mgonchar: This is ugly construction. Since we are creating new variable of type ```mfxVideoParam``` let's change function signature to accept ```mfxVideoParam``` by value (i.e. to perform copying on function call) and then change local copy of ```mfxVideoParam``` in place

Alexandr-Konovalov: Well, it's your decision. I though that rotation is rare case, so we may want to optimize and do copy only when this is necessary. sizeof(mfxVideoParam) is 208 Bytes, so the copy is not that free.

mgonchar: You've already wasted such amount of bytes on tmp variable, so savings related only to copy. As for me it not worth it. Let's wait for some other opinion

Alexandr-Konovalov: Stack allocation is free as far as it's not touched, and `mfxVideoParam` doesn't have a constructor.

dmitry-gurulev: Construction like
```
localParams = *parArg; 
//...
par = &localParams; 
```
is acceptable while a function is short and this piece is at the beginning of function. Otherwise it might be surprising to reveal at point where `par` is used that its address was changed somewhere.

I'd rather use
```
mfxVideoParam const& par //as method arg
//...
mfxVideoParam  vp = par; //local copy of FrameInfo
// ... using of vp.XXX instead par->XXX
```
until we found that we have perf. drop because of copying.

BTW, that's why I argue against `mfxVideoParam` as one universal argument everywhere. It looks like we really don't need the whole `mfxVideoParam` here, just frame info. The only place where it is needed is `DoHwJpegCcFw` for `VideoVppJpegD3D9` initialization. The latter is odd thing for me itself - why we can't create & initialize it explicitly, at decoder's `Init` method? At `Init` we have no any conserns about how many bytes is `sizeof(mfxVideoParam)` and is it worth to avoid copying or not.

Alexandr-Konovalov: > I'd rather use
> mfxVideoParam const& par //as method arg
> //...

Sorry, I can't see why it is better than just passing `mfxVideoParam` by value, as Miroslav suggested above.

> until we found that we have perf. drop because of copying.

I bet we don't, because this is called once for a frame. But a problem with such approach is that it creates tons of very small bottlenecks, that is real pain if someone need performance improvement after all.

dmitry-gurulev: @Alexandr-Konovalov,
> why it is better than just passing mfxVideoParam by value

because it (passing structs by value) is
1. quite rare code pattern thereof looks like a typo and pushes reader to go deeper into the details to find out why it is done in that way
2. implicit as against `mfxVideoParam  vp = par;` where we show our intention clearly

> it creates tons of very small bottlenecks
>real pain if someone need performance improvement

IMO, clean design is the first until performance analyzer says that we need sacrifice it.

Let's refactor code in a way when `mfxVideoParam` is not needed anymore at that function. It makes design clearer and allow us to avoid these (very potential) bottlenecks.

mgonchar: Yes, I also vote for refactoring as best solution.

Alexandr-Konovalov: I still don't understand what's unclear with creation of duplicates only when we need them. OK, let's do it your way.

Getting rid of `DoHwJpegCc* ` etc seems right thing.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1534,6 +1534,40 @@ mfxStatus VideoDECODEMJPEGBase_HW::Init(mfxVideoParam *decPar, mfxFrameAllocRequ
...
+
+        if (m_vPar.mfx.Rotation == MFX_ROTATION_90 || m_vPar.mfx.Rotation == MFX_ROTATION_270)
+        {
+            // Frame allocation is possible inside VideoVppJpegD3D9::Init().
+            // Those frames must have width/height of target image, so the swapping.
+            std::swap(localParams.mfx.FrameInfo.Width, localParams.mfx.FrameInfo.Height);
+            std::swap(localParams.mfx.FrameInfo.CropW, localParams.mfx.FrameInfo.CropH);
+        }
+
+        m_pCc = new VideoVppJpegD3D9(core, isD3DToSys, m_vPar.IOPattern & MFX_IOPATTERN_OUT_OPAQUE_MEMORY);
```
mgonchar: according to line 1537 you need to delete the old value first?

Maybe it is time to make ```m_pCc``` a ```unique_ptr```

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1534,6 +1534,40 @@ mfxStatus VideoDECODEMJPEGBase_HW::Init(mfxVideoParam *decPar, mfxFrameAllocRequ
...
+        bool isD3DToSys = false;
+
+        if(m_vPar.IOPattern & MFX_IOPATTERN_OUT_SYSTEM_MEMORY)
+        {
+            isD3DToSys = true;
+        }
+        else if (m_vPar.IOPattern & MFX_IOPATTERN_OUT_OPAQUE_MEMORY)
+        {
+            mfxExtOpaqueSurfaceAlloc *pOpaqAlloc = (mfxExtOpaqueSurfaceAlloc *)GetExtendedBuffer(m_vPar.ExtParam, m_vPar.NumExtParam, MFX_EXTBUFF_OPAQUE_SURFACE_ALLOCATION);
+            if (!pOpaqAlloc)
```
mgonchar: MFX_CHECK

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -296,24 +213,21 @@ mfxStatus mfx_UMC_FrameAllocator_D3D_Converter::StartPreparingToOutput(mfxFrameS
...
                                                                        UMC::FrameData* in,
                                                                        const mfxVideoParam * par,
-                                                                       VideoVppJpegD3D9 **pCc,
+                                                                       VideoVppJpegD3D9 *pCc,
                                                                        mfxU16 taskId)
 {
     UMC::AutomaticUMCMutex guard(m_guard);
 
-    mfxStatus sts = (*pCc)->QueryTaskRoutine(taskId);
+    mfxStatus sts = pCc->QueryTaskRoutine(taskId);
```
mgonchar: You still have to check input pointers

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1534,6 +1534,40 @@ mfxStatus VideoDECODEMJPEGBase_HW::Init(mfxVideoParam *decPar, mfxFrameAllocRequ
...
+                return MFX_ERR_INVALID_VIDEO_PARAM;
+
+            isD3DToSys = !((pOpaqAlloc->Out.Type & MFX_MEMTYPE_SYSTEM_MEMORY) == 0);
+        }
+
+        mfxVideoParam localParams = m_vPar;
```
mgonchar: I thought that idea was to not create full copy of ```mfxVideoParam```

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
Alexandr-Konovalov: > according to line 1537 you need to delete the old value first?

On the contrary, I checked that `m_pCc` is null. Raw pointers are evil, will do.

mgonchar: right, this macro confuses me from time to time... I think we need to change it's name to something more obvious about expected value

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
Alexandr-Konovalov: I would like to clarify. Are you for 
```
-mfxStatus VideoVppJpegD3D9::Init(const mfxVideoParam *par)
+mfxStatus VideoVppJpegD3D9::Init(mfxU16 AsyncDepth, const mfxFrameInfo &FrameInfo, mfxU16 Rotation,
+                                 const mfxInfoVPP &vpp)
```
or more elegant solution possible?

dmitry-gurulev: @Alexandr-Konovalov, @mgonchar 

Maybe it makes sense to have `VideoVppJpegD3D9` as part of `mfx_UMC_FrameAllocator_D3D_Converter`? Looks like this class do nothing w/ decoding and perform only conversion that is what `mfx_UMC_FrameAllocator_D3D_Converter` is devoted to. What do u think?

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -171,75 +124,51 @@ UMC::Status mfx_UMC_FrameAllocator_D3D_Converter::InitMfx(UMC::FrameAllocatorPar
...
     mfxStatus sts = MFX_ERR_NONE;
-    bool isD3DToSys = false;
+
+    mfxHDLPair pPair;
+    if(isOpaq)
+        sts = m_pCore->GetFrameHDL(surface_work->Data.MemId, (mfxHDL*)&pPair);
+    else
+        sts = m_pCore->GetExternalFrameHDL(surface_work->Data.MemId, (mfxHDL*)&pPair);
 
     if(par->mfx.FrameInfo.PicStruct == MFX_PICSTRUCT_PROGRESSIVE)
```
dmitry-gurulev: From refactoring point of view it makes sense to extract this surface processing (obtaining memId, adjusting width/height etc.) into separate method and then call it from both progressive and interlace branches.

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
```diff
@@ -115,11 +115,11 @@ class VideoDECODEMJPEGBase_HW : public VideoDECODEMJPEGBase
 
 
 protected:
-    VideoVppJpegD3D9 *m_pCc;
-    // True if we need special VPP color conversion after decoding
-    bool   m_needVpp;
+    std::unique_ptr<VideoVppJpegD3D9> m_pCc;
```
dmitry-gurulev: I still think it should be part of `mfx_UMC_FrameAllocator_D3D_Converter`

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1526,15 +1525,51 @@ mfxStatus VideoDECODEMJPEGBase_HW::Init(mfxVideoParam *decPar, mfxFrameAllocRequ
...
     return MFX_ERR_NONE;
 }
 
 mfxStatus VideoDECODEMJPEGBase_HW::Reset(mfxVideoParam *par)
 {
     m_pMJPEGVideoDecoder->Reset();
     m_numPic = 0;
-    delete m_pCc;
-    m_pCc    = 0;
+    m_pCc.release();
```
mgonchar: why ```release```, shouldn't it be ```reset```?

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1571,8 +1606,7 @@ mfxStatus VideoDECODEMJPEGBase_HW::Close(void)
     m_numPic = 0;
     m_isOpaq = false;
 
-    delete m_pCc;
-    m_pCc    = 0;
+    m_pCc.release();
```
mgonchar: same

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
Alexandr-Konovalov: I unable to unify adjusting width/height.

_studio/mfx_lib/decode/mjpeg/include/mfx_mjpeg_dec_decode.h
Alexandr-Konovalov: Done.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -1534,6 +1534,40 @@ mfxStatus VideoDECODEMJPEGBase_HW::Init(mfxVideoParam *decPar, mfxFrameAllocRequ
...
+        {
+            isD3DToSys = true;
+        }
+        else if (m_vPar.IOPattern & MFX_IOPATTERN_OUT_OPAQUE_MEMORY)
+        {
+            mfxExtOpaqueSurfaceAlloc *pOpaqAlloc = (mfxExtOpaqueSurfaceAlloc *)GetExtendedBuffer(m_vPar.ExtParam, m_vPar.NumExtParam, MFX_EXTBUFF_OPAQUE_SURFACE_ALLOCATION);
+            if (!pOpaqAlloc)
+                return MFX_ERR_INVALID_VIDEO_PARAM;
+
+            isD3DToSys = !((pOpaqAlloc->Out.Type & MFX_MEMTYPE_SYSTEM_MEMORY) == 0);
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
Alexandr-Konovalov: Done.

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
dmitry-gurulev:
```
inline void AdjustSize(mfxFrameInfo& info, int rotation)
{
    if (!(rotation == MFX_ROTATION_0 || rotation == MFX_ROTATION_180))
        std::swap(info.CropW, info.CropH);

    if (info.PicStruct != MFX_PICSTRUCT_PROGRESSIVE)
        info.CropH /= 2;
}
///...
mfxStatus mfx_UMC_FrameAllocator_D3D_Converter::StartPreparingToOutput(mfxFrameSurface1 *surface_work, ...)
{
    ///...
    for (int i = 0; i < surface_work->Info.PicStruct != MFX_PICSTRUCT_PROGRESSIVE; ++i)
    {
        //assuming `m_frameDataInternal.GetSurface` copies `FrameInfo`
        mfxFrameSurface1 surface = m_frameDataInternal.GetSurface(in[i]->GetFrameMID()); 
        AdjustSize(surface.Info, par->mfx.Rotation);
    }
}

```

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -169,77 +122,79 @@ UMC::Status mfx_UMC_FrameAllocator_D3D_Converter::InitMfx(UMC::FrameAllocatorPar
...
 
-        mfxHDLPair pPair;
-        if(isOpaq)
-            sts = m_pCore->GetFrameHDL(surface_work->Data.MemId, (mfxHDL*)&pPair);
-        else
-            sts = m_pCore->GetExternalFrameHDL(surface_work->Data.MemId, (mfxHDL*)&pPair);
+        // if memid of in is same as memid of surface_work, this method must not be called
+        MFX_CHECK_WITH_ASSERT(!pPair.first || pPair.first != memId, MFX_ERR_UNSUPPORTED);
+
+        mfxFrameSurface1 pSrc = m_frameDataInternal.GetSurface(index);
```
dmitry-gurulev: Why `pSrc` if it's not actually a pointer?

Alexandr-Konovalov: Done.

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
```diff
@@ -169,77 +122,79 @@ UMC::Status mfx_UMC_FrameAllocator_D3D_Converter::InitMfx(UMC::FrameAllocatorPar
...
                                                                        UMC::FrameData* in,
-                                                                       const mfxVideoParam * par,
-                                                                       VideoVppJpegD3D9 **pCc,
+                                                                       const mfxVideoParam *par,
                                                                        mfxU16 *taskId,
                                                                        bool isOpaq)
 {
     UMC::AutomaticUMCMutex guard(m_guard);
 
+    if(!m_pCc)
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

_studio/shared/src/mfx_umc_mjpeg_vpp.cpp
Alexandr-Konovalov: Thanks, got the loop & array idea. I'm not sure the adjusting deserve a special function. It seems `BeginHwJpegProcessing` can be unified in this way as well.

PR #1277, opened 2019-03-20
MediaSDK Dispatcher for Windows

gkrivor:
Uploading OpenSource version of MediaSDK Dispatcher for Windows.

### DIFF
--- a/api/mfx_dispatch/windows/include/mfx_critical_section.h
+++ b/api/mfx_dispatch/windows/include/mfx_critical_section.h
@@ -1,0 +1,66 @@
+// Copyright (c) 2012-2019 Intel Corporation
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#if !defined(__MFX_CRITICAL_SECTION_H)
+#define __MFX_CRITICAL_SECTION_H
+
+#include <mfxdefs.h>
+
+namespace MFX
+{
+
+// Just set "critical section" instance to zero for initialization.
+typedef volatile mfxL32 mfxCriticalSection;
+
+// Enter the global critical section.
+void mfxEnterCriticalSection(mfxCriticalSection *pCSection);
+
+// Leave the global critical section.
+void mfxLeaveCriticalSection(mfxCriticalSection *pCSection);
+
+class MFXAutomaticCriticalSection
+{
+public:
+    // Constructor
+    explicit MFXAutomaticCriticalSection(mfxCriticalSection *pCSection)
+    {
+        m_pCSection = pCSection;
+        mfxEnterCriticalSection(m_pCSection);
+    }
+
+    // Destructor
+    ~MFXAutomaticCriticalSection()
+    {
+        mfxLeaveCriticalSection(m_pCSection);
+    }
+
+protected:
+    // Pointer to a critical section
+    mfxCriticalSection *m_pCSection;
+
+private:
+    // unimplemented by intent to make this class non-copyable
+    MFXAutomaticCriticalSection(const MFXAutomaticCriticalSection &);
+    void operator=(const MFXAutomaticCriticalSection &);
+};
+
+} // namespace MFX
+
+#endif // __MFX_CRITICAL_SECTION_H
--- a/api/mfx_dispatch/windows/include/mfx_dispatcher.h
+++ b/api/mfx_dispatch/windows/include/mfx_dispatcher.h
@@ -1,0 +1,221 @@
+// Copyright (c) 2012-2019 Intel Corporation
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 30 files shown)


api/mfx_dispatch/windows/include/mfx_dispatcher_log.h
```diff
@@ -0,0 +1,282 @@
...
+    ~FileSink()
+    {
+        if (NULL != m_hdl)
+            fclose(m_hdl);
+    }
+private:
+    FILE * m_hdl;
+    FileSink(const std::string & log_file)
+    {
+        fopen_s(&m_hdl, log_file.c_str(), "a");
```
mgonchar: on linux we don't use '_s' functions. I know that it is windows code, but maybe better to align this. @dvrogozh what do you think?

api/mfx_dispatch/windows/include/mfx_vector.h
```diff
@@ -0,0 +1,210 @@
...
+            return mRecords + mIndex;
+        }
+    };
+
+    class MFXVectorRangeError : public std::exception
+    {
+    };
+
+    template <class T>
+    class MFXVector
```
mgonchar: can we get rid of this?

api/mfx_dispatch/windows/include/mfx_dispatcher.h
```diff
@@ -0,0 +1,221 @@
...
+
+};
+
+// declare comparison operator
+inline
+bool operator == (const mfxVersion &one, const mfxVersion &two)
+{
+    return (one.Version == two.Version);
+
+} // bool operator == (const mfxVersion &one, const mfxVersion &two)
```
dmitry-gurulev: What a useless comment.

api/mfx_dispatch/windows/include/mfx_dispatcher.h
```diff
@@ -0,0 +1,221 @@
...
+
+inline
+bool operator < (const mfxVersion &one, const mfxVersion &two)
+{
+    return (one.Major == two.Major) && (one.Minor < two.Minor);
+
```
dmitry-gurulev: Kind weird impl. - what if major is not equal? As e.g., it will result false for [v1_0 < v2_0].

api/mfx_dispatch/windows/include/mfx_dispatcher_log.h
gkrivor: Should be OK for Windows

api/mfx_dispatch/windows/include/mfx_vector.h
gkrivor: No :(

api/mfx_dispatch/windows/include/mfx_dispatcher.h
gkrivor: Fixed

api/mfx_dispatch/windows/include/mfx_dispatcher.h
gkrivor: Fixed

api/mfx_dispatch/windows/include/mfx_dispatcher_log.h
dvrogozh: On Linux there is no fopen_s, but this is Windows-only code. So, you can use fopen_s in this code. However, if you are doing this, do that correctly. According to https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-s-wfopen-s?view=vs-2017, function prototype is:
```
errno_t fopen_s(
   FILE** pFile,
   const char *filename,
   const char *mode
);
```
You miss error handling. Probably you need to check it and throw exception.

gkrivor: As I understand this logic was introduced by some reason: it will continue execution, even call of fopen_s was failed (FileSink::Write handles case when m_hdl==NULL). Documentation says content of pFile will be unchanged (==NULL) in case of error.
I can only suggest explicitly set m_hdl=NULL before call of fopen_s.

api/mfx_dispatch/windows/include/mfx_vector.h
dvrogozh: @gkrivor : why?

onabiull: As we recommend embedding dispatcher into user app on Windows, we should minimize dependencies. Using std::vector would bring additional dependency from runtime libs. We can consider _optional_ standard implementation (define?) but I'm not sure if it worth it.

api/mfx_dispatch/windows/include/mfx_dispatcher_log.h
dvrogozh: 'file' is private field, left not initialized by default and we discuss constructor. Looks like if current execution will continue it will be undefined behavior.

api/mfx_dispatch/windows/include/mfx_vector.h
dvrogozh: so, we avoid std::vector and freely use std::exception just few lines above. I am not sure this makes sense to me. Wise versa can make some sense. Native-C dispatcher also might make sense. This - does not.

api/mfx_dispatch/windows/include/mfx_dispatcher_log.h
```diff
@@ -0,0 +1,282 @@
...
+#define DISPATCHER_LOG_AUTO(lvl, msg)\
+    DispatchLogBlockHelper name_from_line(__auto_log_)(lvl); name_from_line(__auto_log_).Write msg;
+
+#include <memory>
+#include <map>
+#include <list>
+#include <string>
+
+template <class T> 
+class DSSingleTone
```
dmitry-gurulev: `SingleTone` means single tone? What is this? I believe that pattern [Singleton](https://en.wikipedia.org/wiki/Singleton_pattern) was implied.

api/mfx_dispatch/windows/include/mfx_vector.h
dmitry-gurulev: We also use  `std::string DispatcherLog_GetMFXImplString(int impl)` Forbidding `std::vector` from that perspective looks inconsistent.

api/mfx_dispatch/windows/include/mfx_critical_section.h
```diff
@@ -0,0 +1,66 @@
...
+#include <mfxdefs.h>
+
+namespace MFX
+{
+
+// Just set "critical section" instance to zero for initialization.
+typedef volatile mfxL32 mfxCriticalSection;
+
+// Enter the global critical section.
+void mfxEnterCriticalSection(mfxCriticalSection *pCSection);
```
dmitry-gurulev: What does it mean 'global critical section'? `pCSection` have to be an object defined at global scope or what?

api/mfx_dispatch/windows/src/mfx_critical_section.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+mfxU32 mfxInterlockedXchg32(mfxCriticalSection *pCSection, mfxU32 value)  
+{ 
+    return _InterlockedExchange(pCSection, value);
+}
+
+void mfxEnterCriticalSection(mfxCriticalSection *pCSection)
+{
+    while (MFX_SC_IS_TAKEN == mfxInterlockedCas32(pCSection,
```
dmitry-gurulev: Do we really need polling here? What is an intention to have our own impl. of crit. section? Why not `std::mutex` or native [CRITICAL_SECTION](https://docs.microsoft.com/en-us/windows/desktop/Sync/critical-section-objects)?

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
```diff
@@ -0,0 +1,320 @@
...
+#include <string.h>
+#include <windows.h>
+#pragma warning(disable:4355)
+
+MFX_DISP_HANDLE::MFX_DISP_HANDLE(const mfxVersion requiredVersion) :
+    _mfxSession()
+    ,apiVersion(requiredVersion)
+    ,pluginHive()
+    ,pluginFactory((mfxSession)this)
+{
```
dmitry-gurulev: Inconsistency - some members are initialized through init. list whilst others - at ctor. body.

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
```diff
@@ -0,0 +1,320 @@
...
+} // MFX_DISP_HANDLE::~MFX_DISP_HANDLE(void)
+
+mfxStatus MFX_DISP_HANDLE::Close(void)
+{
+    mfxStatus mfxRes;
+
+    mfxRes = UnLoadSelectedDLL();
+
+    // the library wasn't unloaded
+    if (MFX_ERR_NONE == mfxRes)
```
dmitry-gurulev: Accord. to comment `MFX_ERR_NONE` means that libary was not unloaded. Quite odd as for me.

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
```diff
@@ -0,0 +1,320 @@
...
+        assert(hModule == (mfxModuleHandle)0);
+        DISPATCHER_LOG_BLOCK(("invoking LoadLibrary(%S)\n", pPath));
+
+        // load the DLL into the memory
+        hModule = MFX::mfx_dll_load(pPath);
+
+        if (hModule)
+        {
+            int i;
+
```
dmitry-gurulev: Looks like it needs to be at `for` scope below.

PR #1279, opened 2019-03-20
[sample_multi_transcode] Enable tiles for HEVC encoder

zoss9999:
This is changes to enable tiles for HEVC encoder in sample_multi_transcode sample.


zoss9999:
> Did you test this change? It won't work without adjustment of condition at https://github.com/Intel-Media-SDK/MediaSDK/blob/master/samples/sample_multi_transcode/src/transcode_utils.cpp#L2276

It's true. I will fix. Thanks.

mgonchar:
@zoss9999 fix history of your commits please. There should be only one commit

PR #1281, opened 2019-03-20
[HEVCe] Make TransformSkip = ON by default

asmirno1:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -2937,8 +2937,10 @@ void SetDefaults(
 
 
 #if (MFX_VERSION >= 1026)
-    if (!CO3.TransformSkip)
+    if (!CO3.TransformSkip && par.m_platform < MFX_HW_ICL)
```
mgonchar: It's wrong condition. What if user set it explicitly to off? You will change user decision. Correct way:
```
if (!CO3.TransformSkip)
    CO3.TransformSkip = ( par.m_platform < MFX_HW_ICL) ? MFX_CODINGOPTION_OFF : MFX_CODINGOPTION_ON;
```

PR #1283, opened 2019-03-21
Decorate the brc parameter in cqp mode.

hongbolv:


_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -1309,6 +1309,16 @@ namespace MPEG2EncoderHW
             }
         }
 
+        if (m_VideoParamsEx.mfxVideoParams.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
+        {
+            if (m_VideoParamsEx.mfxVideoParams.mfx.BufferSizeInKB != 0)
```
dmitryermilov: @ogolikov, @splotnikv , @lakulako , can you please review this particular part ? this code block is essential. I mean is it okay to turn on CQP_HRD mode if  a user provides non-zero BufferSizeInKB or we need to add other conditions here?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_full_hw.cpp
```diff
@@ -51,8 +51,7 @@ mfxStatus RunSeqHeader(mfxVideoParam *par, mfxU8* pBuffer, mfxU16 maxLen, mfxU16
     
     mfxU32 bitrate      = 0;
     mfxU32 buffer_size  = 0;
-
-    if (par->mfx.RateControlMethod != MFX_RATECONTROL_CQP) 
+    if ((par->mfx.RateControlMethod != MFX_RATECONTROL_CQP)||(bCqpHrdMode == true))
```
lakulako: This mode requires TargetKbps, but it is not checked to be !0. Does 0 makes sense?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_full_hw.cpp
```diff
@@ -227,12 +226,12 @@ mfxStatus FullEncode::ResetImpl()
     if (!m_pBRC)
     {
         m_pBRC = new MPEG2BRC_HW (m_pCore);
-        sts = m_pBRC->Init(&paramsEx->mfxVideoParams);
+        sts = m_pBRC->Init(paramsEx);
```
lakulako: traditional API changed

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_full_hw.cpp
```diff
@@ -312,8 +311,11 @@ mfxStatus FullEncode::GetVideoParam(mfxVideoParam *par)
 
     if (mfxExtCodingOptionSPSPPS* ext = MPEG2EncoderHW::GetExtCodingOptionsSPSPPS(par->ExtParam, par->NumExtParam))
     {
+        mfxVideoParamEx_MPEG2* paramsEx = m_pController->getVideoParamsEx();
+        bool CqpHrdMode = paramsEx->bCqpHrdMode;
         mfxU16 len = 0;
-        sts = RunSeqHeader(par, ext->SPSBuffer, ext->SPSBufSize,len);
+
+        sts = RunSeqHeader(par, ext->SPSBuffer, ext->SPSBufSize, len, CqpHrdMode);
```
lakulako: Here were no coding parameters in function args. Try without introducing them.

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
lakulako: What is the new mode? Does it requires bitrate to be provided?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -1880,8 +1891,16 @@ namespace MPEG2EncoderHW
             MFX_CHECK_UMC_STS (ret);
 
             mfxU32 bufferSizeInKB = ((brcParams.maxFrameSize + 999)/1000);
-            par->mfx.BRCParamMultiplier = (mfxU16)((bufferSizeInKB + 0x10000) / 0x10000);
```
lakulako: Not related to current commit:
What if multiplier was already set to different value?

_studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp
```diff
@@ -188,8 +188,11 @@ mfxStatus ExecuteBuffers::Init(const mfxVideoParamEx_MPEG2* par, mfxU32 funcId,
 
         mfxU32 multiplier = MFX_MAX(par->mfxVideoParams.mfx.BRCParamMultiplier, 1);
 
-        m_sps.bit_rate = (par->mfxVideoParams.mfx.RateControlMethod != MFX_RATECONTROL_CQP) ?
-                                                            par->mfxVideoParams.mfx.TargetKbps * multiplier : 0;
+        if ((par->mfxVideoParams.mfx.RateControlMethod != MFX_RATECONTROL_CQP) ||
+           (par->bCqpHrdMode == true))
+        {
+            m_sps.bit_rate = par->mfxVideoParams.mfx.TargetKbps * multiplier;
```
lakulako: TargetKbps can be 0 here. OK?

dmitryermilov:
@lakulako , the idea of this is encoder can write bitrate info in seq header if CQP mode. We **already** have the same in AVCe : https://github.com/Intel-Media-SDK/MediaSDK/blob/d2746f5194aa08ec978eec509803c64ed1753891/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L7504

lakulako:
Then write bitrate to header if it is provided irrelevant to CQP or not. Just declare it in docs.
Here is no "mode" really.

dmitryermilov:
@lakulako , yes, we plan to describe it in docs since this mode already implemented for AVCe

lakulako:
Do you agree to refuse this commit? Instead write to bitstream HRD parameters if provided even in CQP mode. Without verification their values (too small for example)? Don't introduce this new mode. Do the same for h264 implementation.

dmitryermilov:
No, I do disagree to refuse this change. The approach enabled in AVCe (and current PR) is the only way to write BRC params in headers in CQP mode. Which alternative can you suggest

lakulako:
In _studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp remove condition at 191-192 (lines in proposed code)

dmitryermilov:
If we remove condition at 191-192, how to distinguish a case when a user in CQP mode passes QPI/QPP/QPB value and a case when user initialize encode in CQP mode, doesn't provide QPI/QPP/QPB init value and provies per frame QP value?

lakulako:
How these lines help to distinguish? And why to distinguish is needed? Write the bitrate that is given. In CQP use per-frame QP if is given or initial one if not given.

hongbolv:
@dmitryermilov  Can you invite Olga to sync about this PR?

ogolikov:
Hi! 
See example on external BRC for CQP mode in 
https://git-ccr-1.devtools.intel.com/gerrit/#/c/120987/

samples/sample_common/include/brc_routines.h
samples/sample_common/src/brc_routines.cpp
samples/sample_encode/include/pipeline_encode.h
samples/sample_encode/src/pipeline_encode.cpp

hongbolv:
Let's resume the conversion. The SW BRC can't distinguish the given TargetKbps or given QPP in the union field in the CQP mode.
            union {
                mfxU16  TargetKbps;
                mfxU16  QPP;
                mfxU16  ICQQuality;
            };

Any other good suggestion?

lakulako:
Need more detailed description of the use case.

ogolikov:
SW BRC is initialized  with  TargetKbps only.
QP is passed via mfxBRCFrameCtrl

hongbolv:
> SW BRC is initialized with TargetKbps only.
> QP is passed via mfxBRCFrameCtrl

Yes. Qp is passed via mfxBRCFrameCtrl. We need to distinguish the two cases for CqpHrd mode.
1)   a user in CQP mode passes QPI/QPP/QPB init value
2)   a user initializes encode in CQP mode, doesn't provide QPI/QPP/QPB init value and provides per frame QP value?

ogolikov:
Hi all,

As I remember, it was decided to implement CqpHrd via external BRC, right? So in this case encoder is initialized in VBR or CBR mode with external BRC.

------------------------------------------------------------------------

We need to distinguish the two cases for CqpHrd mode.

  1.  a user in CQP mode passes QPI/QPP/QPB init value
  2.  a user initializes encode in CQP VBR or CBR  mode with External BRC, doesn't provide QPI/QPP/QPB init value and provides per frame QP value?

hongbolv:
Fix the issue based on the ext brc implementation. Close the PR.

PR #1286, opened 2019-03-21
To fix security issue in 'printf' function call

asaechnx:
Function 'printf' can accepts format string that may be influenced by user, causing format string vulnerability. To fix this issue a specification for character string added in 'printf' format string.


onabiull:
Just for sake of record. It **is not** security issue. It merely a workaround for a flaw in static analysis tool. In this particular case strings can't be manipulated by user. Going forward please provide better commit messages and do not copy-paste.

PR #1290, opened 2019-03-22
h264 ext brc: improve qp search utility functions

mgonchar:


ogolikov:
I think, two functions is too mutch in this case. it better  to fix this condition : 
if (qp_new < 51 && qstep_new >(QSTEP[qp_new] + QSTEP[qp_new + 1]) / 2)        qp_new++;
on
if (qp_new >0 && qstep_new <(QSTEP[qp_new -1] + QSTEP[qp_new]) / 2) qp --;
or to create one function with rounding parameter.


 
        qp_new++;

mgonchar:
@ogolikov agree about any kind of fix for this piece of code. but I like my variant with two functions more than having one rounding functions with switch around some enum indicating whether it is floor / ceil / round. I think it is ok to have them as separate functions. Let's wait for some other opinion

mgonchar:
Olga is against this change. I will upload just small fixes which are really required

dmitry-gurulev:
We already have similar QP rounding functions -
- umc_h264_brc.cpp :: Qstep2QP
- mfx_brc_common.cpp :: Qstep2QP

I don't think that introducing a new one is a good idea.

dmitry-gurulev:
IMO, we need a bunch of utility functions
- QStep2QpCeil
- QStep2QpFloor
- Qstep2QpRound

defined at common header (mfx_utils.h ?) and reused at all other places.

PR #1292, opened 2019-03-22
h264 brc: fix qp adjustment

mgonchar:


dmitry-gurulev:
@mgonchar, it's almost the same as PR#1290, isn't it? What is intention of these two commits?

dmitry-gurulev:
@mgonchar, please look at my comments for PR#1290

mgonchar:
@dmitry-gurulev this PR only fixes existing issue and not introducing any new logic. @ogolikov is strictly against adding any new functions in this code

mgonchar:
@dmitry-gurulev in general I agree with you to have unified brc utils. One issue we have is that samples and MSDK lib have different utils headers. I don't know if it is a good idea to introduce dependencies between lib and samples. But at least within lib we can unify utils usage

mgonchar:
@dmitry-gurulev in general I agree with your message. But as I said @ogolikov is against such approach and we didn't come to agreement (but we agree that code across all brcs we have should follow same priciples). So my idea was to merge fixes to existing issues (this PR) and then talk about how this code should be organized. At least we will have workable code during this time. After we have some conclusion - new PR with refactoring will be opened. What do you think about described approach?

dmitry-gurulev:
>What do you think about described approach?

OK, let's do in that way. BTW, my change requst doesn't prevent you to  merge fixes to existing issues.

PR #1164, opened 2019-02-05
18.4.2

dvrogozh:
18.4.2 scope.

PR #1297, opened 2019-03-25
18.4.2 gtest

dvrogozh:
That's 18.4 branch update w/ gtest patches only (compare to #1164)


dvrogozh:
Dropping in a preference of later releases which has these changes.

PR #1299, opened 2019-03-25
[jpegd] Drop hardcoded size limits for enabling software fallback

Alexandr-Konovalov:
Size limits are driver-dependent, they are checked in
VideoDECODEMJPEGBase_HW::CheckVPPCaps().

Issue: MDP-52482


Alexandr-Konovalov:
> This looks to be correct. However, I am surprised to see this line above:
> `if (core->GetHWType() == MFX_HW_BDW || core->GetHWType() == MFX_HW_SCL)`
> at line1056. Why we strictly check for BDW/SKL? Is that actually correct?

Is seems how platform difference between BDW-SCL/APL-KBL-etc coded. What's the problem with that?

> I was expecting to see at least some section with >= some platform to count new HW...

There are 
```
    if (core->GetHWType() == MFX_HW_APL
        || core->GetHWType() >= MFX_HW_KBL)
```
at line 1021-1022. Clearly this code lacks some `else if` instead of current `if`.

PR #1302, opened 2019-03-25
[h265e] Corrected assertion

semelianova:
has_emulation_bytes reset to zero in all cases, it can not have other values


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1791,7 +1791,7 @@ mfxStatus VAAPIEncoder::Execute(Task const & task, mfxHDLPair pair)
                     mfxU32 length = (header.bit_length + 7) / 8;
 
                     assert(mfxU32(bsDataStart + m_width*m_height - bsDataEnd) > length);
```
mgonchar: no need in this assert, since it is already included in check on line 1795

semelianova: You are right, it is duplicate assertion. I will remove it

PR #1326, opened 2019-04-02
[VP9e] Moved all hardcoded caps to HardcodeCaps function

eshiryae:


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -611,6 +611,32 @@ VAProfile ConvertGuidToVAAPIProfile(const GUID& guid)
     return VAProfileNone; /// Lowpower == OFF is not supported yet.
 }
 
+// this function is aimed to workaround all CAPS reporting problems in driver
+void HardcodeCaps(ENCODE_CAPS_VP9& caps, VideoCORE* pCore)
```
mgonchar: you have to check input pointers. I believe you don't need core here, just pass ```eMFXHWType``` instead

eshiryae: done

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -611,6 +611,30 @@ VAProfile ConvertGuidToVAAPIProfile(const GUID& guid)
     return VAProfileNone; /// Lowpower == OFF is not supported yet.
 }
 
+// this function is aimed to workaround all CAPS reporting problems in driver
+void HardcodeCaps(ENCODE_CAPS_VP9& caps, eMFXHWType platform)
+{
+    caps.CodingLimitSet = 1;
+    caps.Color420Only =  1; // See DDI
```
mgonchar: "See DDI" is a bad comment, either explain why such default is used either remove

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -611,6 +611,30 @@ VAProfile ConvertGuidToVAAPIProfile(const GUID& guid)
...
+    caps.Color420Only =  1;
+
+#if (MFX_VERSION >= 1027)
+    if (platform >= MFX_HW_ICL)
+    {
+        caps.Color420Only = 0;
+        caps.MaxEncodedBitDepth = 1; //0: 8bit, 1: 8 and 10 bit;
+        caps.NumScalablePipesMinus1 = 0; // TODO: for now driver doesn't support multiple pipes scalability
+    }
+#endif
```
mgonchar:
```
#else 
std::ignore = platform;
#endif
```
To avoid possible compiling warnings if version set to < 1027

PR #1328, opened 2019-04-02
vaapi_allocator: Wrap using RGB565 info #ifdef

Losev-Ivan:


_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -179,7 +179,9 @@ static bool isFourCCSupported(unsigned int va_fourcc)
         case VA_FOURCC_P208:
         case VA_FOURCC_P010:
         case VA_FOURCC_AYUV:
+#if defined (MFX_ENABLE_FOURCC_RGB565)
         case VA_FOURCC_RGB565:
+#endif // MFX_ENABLE_FOURCC_RGB565
```
DenWolf: Comment "// MFX_ENABLE_FOURCC_RGB565" is not needed

Losev-Ivan: Done

_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -71,10 +71,10 @@ unsigned int ConvertMfxFourccToVAFormat(mfxU32 fourcc)
         return VA_FOURCC_YV12;
     case MFX_FOURCC_AYUV:
         return VA_FOURCC_AYUV;
-#if defined (MFX_ENABLE_FOURCC_RGB565)
+#if (MFX_VERSION >= 1028)
```
mgonchar: No, dedicated define ```MFX_ENABLE_FOURCC_RGB565``` is much better since we have it

Losev-Ivan: Done

PR #1334, opened 2019-04-04
Deleted unnecessary assertion

semelianova:
There is no problem if we get in the default case. For example if LookAheadDS=4
we get valid returned value. In other cases returned value checks in next steps.


semelianova:
In one our test case we set LookAheadDS=4, in result we get MFX_ERR_NONE  as is the case where LookAheadDS=MFX_LOOKAHEAD_DS_4x . If it is not correct I think LaDSenumToFactor must return an error.

dmitry-gurulev:
@semelianova, please fix a component which doesn't check range of LookAheadDS on its input.

semelianova:
The problem is different, created new pr #1339

PR #1335, opened 2019-04-04
[H264E]  API 1.29 functionality was added

aulitin:
Parameters for BRC: repack, hrd params
Affects external BRC, La
Repacking feature is available for VA only

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
+++ b/_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -853,6 +853,10 @@ namespace MfxHwH264Encode
         mfxU32 cost;
     };
 
+    struct BRCFrameParams : mfxBRCFrameParam
+    {
+        mfxU16 picStruct;
+    };
 
     class DdiTask : public Reconstruct
     {
@@ -917,7 +921,6 @@ namespace MfxHwH264Encode
             , m_notProtected(false)
             , m_nextLayerTask(0)
             , m_repack(0)
-            , m_repackForBsDataLength(0)
             , m_fractionalQP(0)
             , m_midRaw(MID_INVALID)
             , m_midRec(MID_INVALID)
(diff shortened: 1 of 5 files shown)


_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
```diff
@@ -853,6 +853,29 @@ namespace MfxHwH264Encode
...
 
+#if !defined(MFX_EXT_BRC_DISABLE)
+    struct BRCFrameParams : mfxBRCFrameParam
+    {
+        mfxU16 picStruct;
+    };
+#else
+    struct BRCFrameParams
+    {
+        mfxU32 reserved[23];
```
dmitry-gurulev: That's not a part of public API that we need to be ABI compatible. Why do we need reserved fields here?

_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
```diff
@@ -1209,6 +1249,9 @@ namespace MfxHwH264Encode
         CmSurface2DUP *m_wsGpuImage;
         SurfaceIndex  *m_wsIdxGpuImage;
         mfxU8         *m_Yscd;
+
+        BRCFrameParams  m_brcFrameParams;
+        mfxBRCFrameCtrl m_brcFrameCtrl;
```
dmitry-gurulev: Both `mfxBRCFrameParam` & `mfxBRCFrameCtrl` are part of BRC API (mfxbrc.h). Why we re-define one as `BRCFrameParams` (in case of `MFX_EXT_BRC_DISABLE`) but don't do the same for `BRCFrameCtrl`? We either re-define both or none.

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2236,6 +2234,20 @@ mfxU32 GetMaxFrameSize(DdiTask const & task, MfxVideoParam const &video, Hrd con
         std::min(maxFrameSize_hrd,maxFrameSize):
         std::max(maxFrameSize_hrd,maxFrameSize);
 }
+void GetHRDParamFromBRC(DdiTask  & task, mfxU32 &initCpbRemoval, mfxU32 &initCpbRemovalOffset)
```
dmitry-gurulev: Make it `static` or `inline`

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -2236,6 +2234,20 @@ mfxU32 GetMaxFrameSize(DdiTask const & task, MfxVideoParam const &video, Hrd con
...
+{
+    if (task.m_brcFrameCtrl.InitialCpbRemovalDelay ||
+        task.m_brcFrameCtrl.InitialCpbRemovalOffset)
+    {
+        initCpbRemoval = task.m_brcFrameCtrl.InitialCpbRemovalDelay;
+        initCpbRemovalOffset = task.m_brcFrameCtrl.InitialCpbRemovalOffset;
+    }
+}
+void UpdateBRCParams(DdiTask  & task)
```
dmitry-gurulev: Make it `static` or `inline`

PR #1339, opened 2019-04-05
Added check valid value of LookAheadDS

semelianova:
In some cases where LookAheadDS is out of range we get MFX_ERR_NONE. For example
if LookAheadDS=4 we don't get errors because LaDSenumToFactor returns valid
value=4 as is the case MFX_LOOKAHEAD_DS_4x. But it isn't right due to
MFX_LOOKAHEAD_DS_4x=3 by enum, so we need to check LookAheadDS.

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -3409,6 +3409,12 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         par.mfx.RateControlMethod = 0;
     }
 
+    if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_4x)
+    {
+        extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_4x;
+        changed = true;
+    }
+
     if (extOpt2->LookAheadDS == MFX_LOOKAHEAD_DS_4x)
     {
         extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_2x;


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -3417,8 +3417,16 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
 
     if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_OFF)
     {
-        par.calcParam.widthLa  = mfx::align2_value((par.mfx.FrameInfo.Width  / LaDSenumToFactor(extOpt2->LookAheadDS)), 16);
-        par.calcParam.heightLa = mfx::align2_value((par.mfx.FrameInfo.Height / LaDSenumToFactor(extOpt2->LookAheadDS)), 16);
+        if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_4x)
+        {
+            extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_UNKNOWN;
+            changed = true;
```
mgonchar: It's more looking as ```unsupported``` for me. I don't see this parameter changed below which is assumed when ```UNKNOWN``` is used. I think you need either to set it off or copy logic of assignment of some value from ```SetDefaults```

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -3417,7 +3417,12 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
 
     if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_OFF)
     {
-        par.calcParam.widthLa  = mfx::align2_value((par.mfx.FrameInfo.Width  / LaDSenumToFactor(extOpt2->LookAheadDS)), 16);
+        if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_4x)
+        {
+            extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_2x;
```
mgonchar: no. If you adjust it to some valid value - it is ```changed```. So if you can make some meaningful correction it should be something:
0)
```
if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_4x)
{
    extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_2x; // maybe MFX_LOOKAHEAD_DS_4x better? User wanted maximal allowed DS
    changed = true;
}
```

If you can't adjust - you have 2 options:
1) return immediately with error which means that you can't proceed encoding anyway
2) set value to 0 (here it would be ```MFX_LOOKAHEAD_DS_UNKNOWN```) and make ```unsupported``` true

So you should choose either 0) either 2). I think 0) is better

semelianova: I prefer 0) too.
But MFX_LOOKAHEAD_DS_4x isn't better, because it is replaced with MFX_LOOKAHEAD_DS_2x
https://github.com/Intel-Media-SDK/MediaSDK/blob/f5875364dd55a2efdd69540fd059722361f528bd/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L3412
`if (extOpt2->LookAheadDS == MFX_LOOKAHEAD_DS_4x)
    {
        extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_2x;
        changed = true;
    }`

mgonchar: Then you need to reuse this logic in following way:
```
    if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_4x)
    {
        extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_4x;
        changed = true;
    }

    if (extOpt2->LookAheadDS == MFX_LOOKAHEAD_DS_4x)
    {
        extOpt2->LookAheadDS = MFX_LOOKAHEAD_DS_2x;
        changed = true;
    }

    if (extOpt2->LookAheadDS > MFX_LOOKAHEAD_DS_OFF)
    {
        par.calcParam.widthLa  = mfx::align2_value((par.mfx.FrameInfo.Width  / LaDSenumToFactor(extOpt2->LookAheadDS)), 16);
        par.calcParam.heightLa = mfx::align2_value((par.mfx.FrameInfo.Height / LaDSenumToFactor(extOpt2->LookAheadDS)), 16);
    } else
    {
        par.calcParam.widthLa  = par.mfx.FrameInfo.Width;
        par.calcParam.heightLa = par.mfx.FrameInfo.Height;
    }
```

the benefit is that if in future 4x will be supported it would require only one change in code

PR #1342, opened 2019-04-05
To get compile under Visual Studio with a project from scratch

mofo7777:


samples/sample_common/src/d3d_device.cpp
```diff
@@ -20,7 +20,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "mfx_samples_config.h"
 #include <algorithm>
 
-#if defined(WIN32) || defined(WIN64)
+#if defined(_WIN32) || defined(_WIN64)
```
dmitry-gurulev: MSDN: [Predefined Macros](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019)
> _WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.

So, it's not required to check `_WIN64`.

PR #1349, opened 2019-04-09
encode: convert status if no SW fallback

mgonchar:
change MFX_WRN_PARTIAL_ACCELERATION to MFX_ERR_UNSUPPORTED in Encode Init function if no SW fallback exist

### DIFF
--- a/_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
+++ b/_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


mgonchar:
@dmitry-gurulev, I see. We also need to change closed source part, since there is UNSUPPORTED

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -478,6 +478,19 @@ std::pair<bool, bool> check_fei(VideoCORE* core)
     return std::pair<bool,bool>(true, *feiEnabled);
 }
 
+static inline bool is_sw_fallback_supported(mfxU32 codec_id, VideoCORE* core)
```
dmitry-gurulev: AFAIR, `static inline` makes sense in C, why we need that `static` in C++?

mgonchar: @dmitry-gurulev  AFAIK this usage of ```static``` is still valid in C++ and has the same meaning as in C. If your question is why not to use of unnamed ```namespace``` to do the same (limit the lifetime and hide from linker)... well, I have no preference between these two. ```namespace``` approach seems to be more C++, I can change approach in other PR

PR #1357, opened 2019-04-12
smt: replaced MSDKEvent with std::condition_variable

mgonchar:


samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -1271,7 +1271,7 @@ mfxStatus CTranscodingPipeline::Encode()
                 // Getting next frame
                 while (MFX_ERR_MORE_SURFACE == curBuffer->GetSurface(DecExtSurface))
                 {
-                    if (MFX_ERR_NONE != curBuffer->WaitForSurfaceInsertion(MSDK_SURFACE_WAIT_INTERVAL)) {
+                    if (!curBuffer->WaitForSurfaceInsertion(MSDK_SURFACE_WAIT_INTERVAL)) {
```
dmitry-gurulev: if buffering is not allowed (`m_IsBufferingAllowed == false`) `GetSurface` returns `MFX_ERR_MORE_SURFACE` and we get into `WaitForSurfaceInsertion`. The latter will wait for `m_bInsEventHappened` that becomes true at `AddSurface` when `m_IsBufferingAllowed == true`, so we'll exit w/ `MFX_ERR_NOT_FOUND` won't 'we ?

mgonchar: That should not happen, because ```m_IsBufferingAllowed``` is set to ```false``` only in ```CancelBuffering``` which happens only after the main loop of encoding is finished (i.e. no more encoding will occur). So buffering is canceled right before application shutdown. So current code (as well as previous implementation) should work

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -4568,38 +4568,45 @@ void DecreaseReference(mfxFrameData *ptr)
...
-mfxStatus SafetySurfaceBuffer::WaitForSurfaceInsertion(mfxU32 msec)
+bool SafetySurfaceBuffer::WaitForSurfaceInsertion(mfxU32 msec)
 {
-    return pInsEvent->TimedWait(msec);
+    using namespace std::chrono;
+
+    std::unique_lock<std::mutex> lock(m_mutex);
+    if (m_InsEvent.wait_until(lock, system_clock::now() + milliseconds(msec), [this](){ return m_bInsEventHappened; }))
+    {
+        m_bInsEventHappened = false;
```
dmitry-gurulev: @mgonchar,

Can we use `m_SList.empty()` as a condition instead of introducing a new state flag  `m_bInsEventHappened`?

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -4568,38 +4568,45 @@ void DecreaseReference(mfxFrameData *ptr)
...
 
-mfxStatus SafetySurfaceBuffer::WaitForSurfaceRelease(mfxU32 msec)
+void SafetySurfaceBuffer::WaitForSurfaceRelease(mfxU32 msec)
 {
-    return pRelEvent->TimedWait(msec);
+    using namespace std::chrono;
+
+    std::unique_lock<std::mutex> lock(m_mutex);
+    if (m_RelEvent.wait_until(lock, system_clock::now() + milliseconds(msec), [this](){ return m_bRelEventHappened; }))
+        m_bRelEventHappened = false;
```
dmitry-gurulev: @mgonchar,

Can we use `m_SList.size() == m_AsyncDepth` as a condition instead of introducing a new state flag m_bRelEventHappened?

samples/sample_multi_transcode/include/pipeline_transcode.h
```diff
@@ -571,8 +571,10 @@ namespace TranscodingSample
         std::mutex                   m_mutex;
         std::list<SurfaceDescriptor> m_SList;
         bool                         m_IsBufferingAllowed;
-        MSDKEvent*                   pRelEvent;
-        MSDKEvent*                   pInsEvent;
+        std::condition_variable      m_RelEvent;
```
dvrogozh: How often we use these CVs? Windows has a bug in CV implementation which causes significant increase in CPU%... We might need to be careful with this PR...

mgonchar: I'm afraid quite often. For example they are used in 1:N scenarios for synchronization of produce/consume parts of pipelines. Which windows implementation bug do you refer?

Issue #906, opened 2018-11-16
VA RT format and fourcc are mixed in MediaSDK


xhaihao:
When creating a surface with fourcc setting to VA_FOURCC_P010, the format is expected to VA_RT_FORMAT_YUV420_10. However the format is actually set to VA_FOURCC_P010 in MSDK.

xhaihao:
Below is the related code in libmfx_allocator_vaapi.cpp

197                 attrib.value.value.i = va_fourcc;                                                                                                                                                                                    
198                 format               = va_fourcc;

egorovdanil:
The problem no longer reproduce.

PR #1360, opened 2019-04-16
[vaapi] minor update for P010 format in allocator

IlyaSergeyev:
When creating a surface with fourcc setting to VA_FOURCC_P010, the format is expected to VA_RT_FORMAT_YUV420_10

Issue #906

### DIFF
--- a/_studio/shared/src/libmfx_allocator_vaapi.cpp
+++ b/_studio/shared/src/libmfx_allocator_vaapi.cpp
@@ -129,6 +129,11 @@ static void FillSurfaceAttrs(std::vector<VASurfaceAttrib> &attrib, unsigned int 
             attrib[0].value.value.i  = VA_FOURCC_P208;
             format                   = VA_FOURCC_P208;
             break;
+#if VA_CHECK_VERSION(1,2,0)
+        case MFX_FOURCC_P010:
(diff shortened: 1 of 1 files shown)


_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -129,6 +129,9 @@ static void FillSurfaceAttrs(std::vector<VASurfaceAttrib> &attrib, unsigned int
             attrib[0].value.value.i  = VA_FOURCC_P208;
             format                   = VA_FOURCC_P208;
             break;
+        case MFX_FOURCC_P010:
+            format = VA_RT_FORMAT_YUV420_10;
```
vilichev: Do we need to add VA API guard here using VA_CHECK_VERSION?

IlyaSergeyev: > Do we need to add VA API guard here using VA_CHECK_VERSION?

We have not such guards in other places with MFX_FOURCC_P010

vilichev: I meant guard for VA_RT_FORMAT_YUV420_10 (see mfx_h265_encode_vaapi.cpp, VAAPIEncoder::CreateAuxilliaryDevice())

PR #1362, opened 2019-04-16
Ignore "Wimplicit-fallthrough" warning

Losev-Ivan:


Losev-Ivan:
@mgonchar Yes, but we have a lot of switches where this warning occur. The another solution is to add `[[clang::fallthrough]];` but I think that this is not god idea because instead of:
`case MFX_HW_CNL:`
`case MFX_HW_ICL:`
`case MFX_HW_ICL_LP:`

we will have to write:
`case MFX_HW_CNL: [[clang::fallthrough]];`
`case MFX_HW_ICL: [[clang::fallthrough]];`
`case MFX_HW_ICL_LP: [[clang::fallthrough]];`

mgonchar:
@Losev-Ivan as alternative you can mute specific compiler warning for block of code. I don't like the idea of turning this option off. Let's maybe wait for some other opinion. By the way, how many hits of such warning do we have?

Losev-Ivan:
@mgonchar approximately 43

PR #1379, opened 2019-04-19
sample_multi_transcode: enable HEVC encode with temporal layers

jonrecker:
New test suite: smt_hevcd_to_hevce_temporal_layers
Add new switches for temporal layers
Update par file parsing to handle string literals
WA for compiler error on Windows - too many if-else
  clauses in cmd-line parsing causes build error

### DIFF
--- a/samples/sample_multi_transcode/src/pipeline_transcode.cpp
+++ b/samples/sample_multi_transcode/src/pipeline_transcode.cpp
@@ -171,6 +171,9 @@ CTranscodingPipeline::CTranscodingPipeline():
     MSDK_ZERO_MEMORY(m_PluginOpaqueAlloc);
     MSDK_ZERO_MEMORY(m_PreEncOpaqueAlloc);
     MSDK_ZERO_MEMORY(m_ExtLAControl);
+    MSDK_ZERO_MEMORY(m_AvcTemporalLayers);
+    MSDK_ZERO_MEMORY(m_CodingOptionSPSPPS);
+    MSDK_ZERO_MEMORY(m_CodingOption);
     MSDK_ZERO_MEMORY(m_CodingOption2);
     MSDK_ZERO_MEMORY(m_CodingOption3);
     MSDK_ZERO_MEMORY(m_ExtHEVCParam);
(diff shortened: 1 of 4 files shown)


samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -2573,6 +2585,38 @@ MFX_IOPATTERN_IN_VIDEO_MEMORY : MFX_IOPATTERN_IN_SYSTEM_MEMORY);
...
 
+    if (pInParams->nAvcTemp)
+    {
+        if (pInParams->EncodeId == MFX_CODEC_HEVC)
+        {
+            int i;
+            for (i = 0; i < 8; i++)
+            {
+                m_AvcTemporalLayers.Layer[i].Scale = pInParams->nAvcTemporalLayers[i];
+                m_AvcTemporalLayers.BaseLayerPID = pInParams->nBaseLayerPID;
```
mgonchar: why this is in the loop?

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -2573,6 +2585,38 @@ MFX_IOPATTERN_IN_VIDEO_MEMORY : MFX_IOPATTERN_IN_SYSTEM_MEMORY);
 #endif
     }
 
+    if (pInParams->nAvcTemp)
+    {
+        if (pInParams->EncodeId == MFX_CODEC_HEVC)
+        {
+            int i;
```
mgonchar: better to move it in for loop ```for (int i = 0; ...)```

samples/sample_multi_transcode/src/pipeline_transcode.cpp
jonrecker: Fixed. Thanks.
Updated commit:  3f5bc4f

samples/sample_multi_transcode/src/pipeline_transcode.cpp
jonrecker: Fixed. Thanks.
Updated commit:  3f5bc4f

PR #1386, opened 2019-04-24
[MCTF - ICL] Removal of excess CreateQueue calls

spossos:
CreateQueue(Ex) is called at init of MCTF, but later when
the kernel has been configured for operation the code had
extra CreateQueue calls, which reset the original queue,
this is an issue on ICL because ICL requires special queue
creation by using CreateQueueEx. By removing the additional
CreateQueue calls the original queue (properly configured)
is not reset and MCTF works correctly.

Issue: MDP-48160
Test: vpp_mctf_denoise


spossos:
New pull request will be done including the fix to the copyright yer

PR #1387, opened 2019-04-24
[MCTF - ICL] Removal of excess CreateQueue calls

spossos:
CreateQueue(Ex) is called at init of MCTF, but later when
the kernel has been configured for operation the code had
extra CreateQueue calls, which reset the original queue,
this is an issue on ICL because ICL requires special queue
creation by using CreateQueueEx. By removing the additional
CreateQueue calls the original queue (properly configured)
is not reset and MCTF works correctly.

Issue: MDP-48160
Test: vpp_mctf_denoise


spossos:
Hi Oleg and Dmitry,

I am not sure why the error in windows check, as it is failing because sample_encode project has too many nested folders, but my change has nothing to do with sample_encode.

dvrogozh:
Some patch was merged into master branch which regressed Windows build. You need to wait till it will be fixed, then just rebase your change.

spossos:
Ohh ok, I will. Thanks!

spossos:
I am going to push a clean new request

PR #1388, opened 2019-04-25
Cmd line parsing fail in sample_encode.cpp with open source windows b…

mounikamandava:
…uild is fixed

PR #1389, opened 2019-04-25
[VP9e] Fix setting default value of initialDelayInKb

eshiryae:
bufferSizeInKb is used before its default values is set, so default value of initialDelayInKb isn't set correctly.


onabiull:
Windows CI failed due to known issue in samples. Going to be fixed by #1388

PR #1392, opened 2019-04-25
hevce: Enable QVBR support

otsirule:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -925,6 +928,12 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     m_caps.VCMBitRateControl =
         attrs[ idx_map[VAConfigAttribRateControl] ].value & VA_RC_VCM ? 1 : 0; //Video conference mode
+    m_caps.QVBRBRCSupport =
+#ifdef MFX_ENABLE_QVBR
```
dvrogozh: Can we, please, avoid ifdef macro in the middle of assignment? I think m_caps is zeroed initially, hence, we just can use:
```
#ifdef MFX_ENABLE_QVBR
    m_caps.QVBRBRCSupport = attrs[ idx_map[VAConfigAttribRateControl] ].value & VA_RC_QVBR ? 1 : 0;
#endif
```

otsirule: Ok

PR #1394, opened 2019-04-26
sample_decode: replace MSDKThread with std::thread

mgonchar:


samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1660,21 +1655,18 @@ mfxStatus CDecodingPipeline::RunDecoding()
...
-        if (!pDeliverThread || !m_pDeliverOutputSemaphore || !m_pDeliveredEvent) {
-            MSDK_SAFE_DELETE(pDeliverThread);
-            MSDK_SAFE_DELETE(m_pDeliverOutputSemaphore);
-            MSDK_SAFE_DELETE(m_pDeliveredEvent);
-            return MFX_ERR_MEMORY_ALLOC;
-        }
+
+        // Create a thread and let it work independently from the thread handle 
+        std::thread deliverThread(&CDecodingPipeline::DeliverLoop, this);
+        deliverThread.detach();
```
dvrogozh: hm. why we detach the thread? what is your motivation for that? Will this make memory leaks checking more complicated? IMHO, I would prefer to see honest join in the very end.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1660,21 +1655,18 @@ mfxStatus CDecodingPipeline::RunDecoding()
     bool                bErrIncompatibleVideoParams = false;
     CTimeInterval<>     decodeTimer(m_bIsCompleteFrame);
     time_t start_time = time(0);
-    MSDKThread * pDeliverThread = NULL;
 #if (MFX_VERSION >= 1025)
     mfxExtDecodeErrorReport *pDecodeErrorReport = NULL;
 #endif
 
-    if (m_eWorkMode == MODE_RENDERING) {
+    if (m_eWorkMode == MODE_RENDERING)
```
dvrogozh: please, don't change what you actually did not modify.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1963,16 +1955,14 @@ mfxStatus CDecodingPipeline::RunDecoding()
             CTimer::ConvertToSeconds(*std::min_element(m_vLatency.begin(), m_vLatency.end()))*1000);
     }
 
-    if (m_eWorkMode == MODE_RENDERING) {
+    if (m_eWorkMode == MODE_RENDERING)
```
dvrogozh: please, don't change what you actually did not modify.

samples/sample_decode/src/pipeline_decode.cpp
mgonchar: Well, ```join``` looks more similar to original code, so I redesigned it with ```join```

samples/sample_decode/src/pipeline_decode.cpp
mgonchar: fixed

samples/sample_decode/src/pipeline_decode.cpp
mgonchar: fixed

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1559,17 +1558,7 @@ mfxStatus CDecodingPipeline::DeliverLoop(void)
...
-
-unsigned int MFX_STDCALL CDecodingPipeline::DeliverThreadFunc(void* ctx)
-{
-    CDecodingPipeline* pipeline = (CDecodingPipeline*)ctx;
-
-    mfxStatus sts;
-    sts = pipeline->DeliverLoop();
-
-    return 0;
+    return;
```
dvrogozh: minor. 'return' is useless here - drop.

mgonchar: will do in next changes, I plan to get rid of of semaphores here

Issue #1398, opened 2019-04-26
AVC low power encoding fails on ubuntu 19.04


dvrogozh:
Install Ubuntu 19.04 and mediasdk coming from the OS:
```
sudo apt-get install libmfx1 libmfx-tools
sudo apt-get install libva-drm2 libva-x11-2
```

And run the following sample encode command line to encode via AVC Low Power encoder:
```
export LIBVA_DRIVER_NAME=iHD

# just to prepare some YUV input file:
wget https://fate-suite.libav.org/h264-conformance/AUD_MW_E.264
/usr/share/mfx/samples/sample_decode h264 -i AUD_MW_E.264 -o AUD_MW_E.yuv

# and now encode:
sample_encode: /build/intel-mediasdk-AzMSof/intel-mediasdk-18.4.1/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp:1422: virtual mfxStatus MfxHwH264Encode::VAAPIEncoder::CreateAuxilliaryDevice(VideoCORE*, GUID, mfxU32, mfxU32, bool): Assertion `0x00000000 == vaSts' failed.

Aborted (core dumped)
```

Please, make sure to have `intel-media-va-driver` installed instead of `intel-media-va-driver-non-free` since the latter on does not return the error on the above command line:
```
sudo apt-get install intel-media-va-driver
```

dvrogozh:
There are few issues here:
1. gcc-8.3.0 compiler on Ubuntu 19.04 aligns mediasdk internally defined GUID structure size by 8 and we compare non-initialized memory via memcmp inside operator== for GUIDs. As a result we fail to recognize that user requested LowPower encoding and try to initialize Dual Pipe which is not available on `intel-media-va-driver`
2. I suspect that in case of the full `intel-media-va-driver-non-free` we actually run Dual Pipe encoding instead of Low Power for the same reason as explained above
3. PR #1399 fixes the issue in mediasdk to compare GUIDs correctly (but it does nothing to assure that user will get Low Power if he requested it explicitly). However, there is a *second issue in media-driver* because of which Low Power encoding still won't work after #1399. Effectively, https://github.com/intel/media-driver/releases/tag/intel-media-18.4.1 release miss the following patch: https://github.com/intel/media-driver/commit/26d7d503c993f46377f6366ac076e6fc98ec179f

dvrogozh:
Filed https://github.com/intel/media-driver/issues/613 for driver part.

[...]

PR #1399, opened 2019-04-26
Make sure to compare guids correctly

dvrogozh:
Fix for #1398

On Ubuntu 19.04 compiler (gcc-8.3.0) aligns GUID stucture size which
we defined internally in mediasdk by 8 and does not zero out the
structure when allocated on stack. As a result when we use current
approach to compare guids via !memcpy(a,b,sizeof(GUID)) we actually
compare unitialized regions and mistakenly think some guids not
equal which leads to runtime errors. For example, we might fail
to enable LowPower encoding on this command line:
  sample_encode h264 -w 176 -h 144 -f 30 -cqp \
    -qpi 30 -qpp 30 -qpb 30 -qsv-ff -nv12 -i a.yuv -o a.264

### DIFF
--- a/_studio/shared/include/mfxvideo++int.h
+++ b/_studio/shared/include/mfxvideo++int.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -36,6 +36,7 @@
 #ifndef GUID_TYPE_DEFINED
 
 #include <string.h>
+#include <algorithm>
 
 typedef struct {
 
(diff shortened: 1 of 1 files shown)


_studio/shared/include/mfxvideo++int.h
```diff
@@ -57,7 +58,11 @@ typedef struct {
 
 static inline int operator==(const GUID & guidOne, const GUID & guidOther)
 {
-    return !memcmp(&guidOne, &guidOther, sizeof(GUID));
+    return
+        guidOne.Data1 == guidOther.Data1 &&
+        guidOne.Data2 == guidOther.Data2 &&
+        guidOne.Data3 == guidOther.Data3 &&
+        std::equal(guidOne.Data4, guidOne.Data4 + sizeof(guidOne.Data4), guidOther.Data4);
```
dvrogozh: This also can be fixed by:
```
#pragma pack(push, 4)
typedef struct {
...
} GUID;
#pragma pack(pop)
```
or by another way to compare Data4. Opinions?

mgonchar: I like your variant

PR #1413, opened 2019-05-13
Remove AutomaticUmcMutex

akokoshn:


akokoshn:
WIP

_studio/shared/src/libmfx_core.cpp
```diff
@@ -82,22 +82,22 @@ mfxStatus CommonCORE::API_1_19_Adapter::QueryPlatform(mfxPlatform* platform)
 
 mfxStatus CommonCORE::AllocBuffer(mfxU32 nbytes, mfxU16 type, mfxHDL *mid)
 {
-    UMC::AutomaticUMCMutex guard(m_guard);
+    std::lock_guard<std::recursive_mutex> guard(m_guard);
```
mgonchar: do we really need ```recursive_mutex``` here? Using of this kind of mutexes often indicates bad design

_studio/shared/src/mfx_umc_alloc_wrapper.cpp
```diff
@@ -1049,12 +1049,12 @@ mfxStatus mfx_UMC_FrameAllocator::PrepareToOutput(mfxFrameSurface1 *surface_work
     surface.Info.Shift = m_IsUseExternalFrames ? m_extSurfaces[index].FrameSurface->Info.Shift : m_frameDataInternal.GetSurface(index).Info.Shift;
 
     //Performance issue. We need to unlock mutex to let decoding thread run async.
-    guard.Unlock();
+    guard.unlock();
```
mgonchar: doesn't really makes much sense for me. If it is ```recursive_mutex``` it is still can be locked after this ```unlock``` call, so what is the point of unlocking it?

_studio/shared/src/mfx_umc_alloc_wrapper.cpp
```diff
@@ -1104,14 +1104,14 @@ mfxStatus   mfx_UMC_FrameAllocator_D3D::PrepareToOutput(mfxFrameSurface1 *surfac
             surface.Data.Y = 0;
             surface.Data.MemId = idx;
             //Performance issue. We need to unlock mutex to let decoding thread run async.
-            guard.Unlock();
+            guard.unlock();
```
mgonchar: same

_studio/shared/src/mfx_umc_alloc_wrapper.cpp
akokoshn: I guess this unlock not related recursive mutex lock (DoFastCopyExtended() locks CommonCore::m_guard).
In my understanding unlock need for improve performance: another thread can run PrepareToOutput during execute CM copy

mgonchar: I don't understand what you are talking about, you unlock here a ```recursive_mutex``` locked on line 958

akokoshn: right, this unlock need for improve performance: another thread can run PrepareToOutput() during execute CM copy (DoFastCopyWrapper())

mgonchar: Another thread can't run anything if this ```mutex``` is locked. Since you are using ```recursive_mutex``` after calling ```unlock``` is still can be locked and thought can't be used by another thread. That's why combination of ```recursive_mutex``` and manual ```unlock``` doesn't make sense for me

akokoshn:
resubmitted in another PR: https://github.com/Intel-Media-SDK/MediaSDK/pull/1445

PR #1425, opened 2019-05-17
Update readme

akharche:
Update readme for [#1385](https://github.com/Intel-Media-SDK/MediaSDK/pull/1385) merge


README.md
```diff
@@ -93,6 +93,7 @@ Media SDK depends on a number of packages which are identified and checked for t
 | ------ | ------ | ----------- |
 | API | master\|latest\|major.minor | Build mediasdk library with specified API. 'latest' will enable experimental features. 'master' will configure the most recent available published API (default: master). |
 | ENABLE_OPENCL | ON\|OFF | Enable OpenCL dependent code to be built (default: ON) |
+| ENABLE_X11 | ON\|OFF | Enable X11 dependent code to be built (default: OFF) |
```
dvrogozh: It is a little more complex than other options. Should be something like: "Enable X11 dependent code to be built (default: try to enable)". So, there will be runtime decision on configuration stage...

PR #1426, opened 2019-05-17
API: Fix binary api layout

mgonchar:
Fixed structures binary layout with compiler's pack pragma

### DIFF
--- a/api/include/mfxastructures.h
+++ b/api/include/mfxastructures.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -93,6 +93,7 @@ enum
     MFX_AUDIO_AAC_JOINT_STEREO=    3
 };
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxU32                CodecId;
     mfxU16                CodecProfile;
@@ -119,7 +120,9 @@ typedef struct {
         };
     };
 } mfxAudioInfoMFX;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_STRUCT_W_PTR()
 typedef struct {
     mfxU16  AsyncDepth;
     mfxU16  Protected;
@@ -129,13 +132,17 @@ typedef struct {
     mfxExtBuffer**    ExtParam;
     mfxU16            NumExtParam;
 } mfxAudioParam;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxU32  SuggestedInputSize;
     mfxU32  SuggestedOutputSize;
     mfxU32  reserved[6];
 } mfxAudioAllocRequest;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_STRUCT_W_L_TYPE()
 typedef struct {
     mfxU64  TimeStamp; /* 1/90KHz */
     mfxU16  Locked;
@@ -152,6 +159,7 @@ typedef struct {
     mfxU32  NumExtParam;
     mfxExtBuffer **ExtParam;
 } mfxAudioFrame;
+MFX_PACK_END()
 
 #ifdef __cplusplus
 }
--- a/api/include/mfxbrc.h
+++ b/api/include/mfxbrc.h
@@ -32,6 +32,7 @@ enum {
     MFX_EXTBUFF_BRC = MFX_MAKEFOURCC('E','B','R','C')
 };
 
+MFX_PACK_BEGIN_STRUCT_W_PTR()
 typedef struct {
 #if (MFX_VERSION >= 1026)
     mfxU32 reserved[23];
@@ -50,7 +51,9 @@ typedef struct {
     mfxU16 NumExtParam;
     mfxExtBuffer** ExtParam;
 } mfxBRCFrameParam;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_STRUCT_W_PTR()
 typedef struct {
     mfxI32 QpY;             // Frame-level Luma QP
 #if (MFX_VERSION >= 1029)
@@ -67,6 +70,7 @@ typedef struct {
     mfxHDL reserved2;
 #endif
 } mfxBRCFrameCtrl;
+MFX_PACK_END()
 
 /* BRCStatus */
 enum {
@@ -77,16 +81,19 @@ enum {
     MFX_BRC_PANIC_SMALL_FRAME = 4  // Coded frame is too small, no further recoding possible - required padding to MinFrameSize
 };
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxU32 MinFrameSize;    // Size in bytes, coded frame must be padded to when Status = MFX_BRC_PANIC_SMALL_FRAME
     mfxU16 BRCStatus;       // See BRCStatus enumerator
     mfxU16 reserved[25];
     mfxHDL reserved1;
 } mfxBRCFrameStatus;
+MFX_PACK_END()
 
 /* Structure contains set of callbacks to perform external bit-rate control.
 Can be attached to mfxVideoParam structure during encoder initialization.
 Turn mfxExtCodingOption2::ExtBRC option ON to make encoder use external BRC instead of native one. */
+MFX_PACK_BEGIN_STRUCT_W_PTR()
 typedef struct {
     mfxExtBuffer Header;
 
@@ -112,6 +119,7 @@ typedef struct {
 
     mfxHDL reserved1[10];
 } mfxExtBRC;
+MFX_PACK_END()
 
 #ifdef __cplusplus
 } // extern "C"
--- a/api/include/mfxcamera.h
+++ b/api/include/mfxcamera.h
@@ -52,6 +52,7 @@ typedef enum {
     MFX_CAM_GAMMA_LUT        = 0x0002,
 } mfxCamGammaParam;
 
+MFX_PACK_BEGIN_STRUCT_W_L_TYPE()
 typedef struct {
     mfxExtBuffer    Header;
     mfxU16  Mode;
@@ -62,14 +63,16 @@ typedef struct {
     mfxU16  NumPoints;
     mfxU16  GammaPoint[1024];
     mfxU16  GammaCorrected[1024];
-    mfxU32  reserved3[4]; 
+    mfxU32  reserved3[4];
 } mfxExtCamGammaCorrection;
+MFX_PACK_END()
 
 typedef enum {
     MFX_CAM_WHITE_BALANCE_MANUAL   = 0x0001,
     MFX_CAM_WHITE_BALANCE_AUTO     = 0x0002
 } mfxCamWhiteBalanceMode;
 
+MFX_PACK_BEGIN_STRUCT_W_L_TYPE()
 typedef struct {
     mfxExtBuffer    Header;
     mfxU32          Mode;
@@ -79,7 +82,9 @@ typedef struct {
     mfxF64          G1;
     mfxU32          reserved[8];
 } mfxExtCamWhiteBalance;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxExtBuffer Header;
     mfxU16        R;
@@ -90,7 +95,9 @@ typedef struct {
     mfxU16        Y;
     mfxU16        reserved[6];
 } mfxExtCamTotalColorControl;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxExtBuffer Header;
 
@@ -99,13 +106,17 @@ typedef struct {
     mfxF32       PostOffset[3];
     mfxU16       reserved[30];
 } mfxExtCamCscYuvRgb;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxExtBuffer    Header;
     mfxU16          PixelThresholdDifference;
     mfxU16          PixelCountThreshold;
 } mfxExtCamHotPixelRemoval;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxExtBuffer    Header;
     mfxU16          R;
@@ -114,19 +125,25 @@ typedef struct {
     mfxU16          G1;
     mfxU32          reserved[4];
 } mfxExtCamBlackLevelCorrection;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxU8 integer;
     mfxU8 mantissa;
 } mfxCamVignetteCorrectionElement;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxCamVignetteCorrectionElement R;
     mfxCamVignetteCorrectionElement G0;
     mfxCamVignetteCorrectionElement B;
     mfxCamVignetteCorrectionElement G1;
 } mfxCamVignetteCorrectionParam;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_STRUCT_W_PTR()
 typedef struct {
     mfxExtBuffer    Header;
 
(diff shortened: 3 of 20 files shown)


dmitry-gurulev:
@mgonchar

I'd rather use something like:
```
//some common header (mfx_utils.h ?)
#if defined(_MSC_VER)
	#define PRAGMA(x) __pragma(x)
#else
	#define PRAGMA_IMPL(x) _Pragma(#x)
#endif

#define PACK_BEGIN_X(x) PRAGMA_IMPL(pack(push, x))

#if defined(_WIN64) || defined(LINUX64) 
	#define PACK_BEGIN() PACK_BEGIN_X(8)
#elif defined(_WIN32) || defined(LINUX32)
	#define PACK_BEGIN() PACK_BEGIN_X(4)
#else
	#error Unknown packing
#endif 

#define PACK_END()      PRAGMA_IMPL(pack(pop))

//API header (e.g. mfxvideo.h)
PACK_BEGIN()
typedef struct { 
    mfxU32      reserved[4]; 
    mfxHDL      pthis; 
    mfxStatus  (MFX_CDECL  *GetHDL)   (mfxHDL pthis, mfxMemId mid, mfxHDL *handle); 
    mfxStatus  (MFX_CDECL  *Free)     (mfxHDL pthis, mfxFrameAllocResponse *response); 
} mfxFrameAllocator; 
PACK_END()
```
instead of having these
`#if defined(_WIN64) || defined(LINUX64)` at every struct definition

mgonchar:
It is a little bit more complicated with 32 / 64 bits builds. When structure contains some of the long types, i.e. ```mfxU64``` on windows and linux size is different, so definition of ```PACK_BEGIN``` from your suggestion is not so straightforward

dmitry-gurulev:
@mgonchar,

There is PACK_BEGIN_X(X) u can use it in that cases. Or maybe I didn't catch, could u give me an example?

mgonchar:
@dmitry-gurulev I mean places like this https://github.com/Intel-Media-SDK/MediaSDK/pull/1426/commits/2377e7e8fe58a09af56a8d9be15a6c85924ee44a#diff-98d5dfefa7a7d2d9ee08e8574acbe54eR207 So we need another variant of ```PACK_BEGIN_X``` for such structs

dmitry-gurulev:
@mgonchar,

So, the thing is 8-bytes (like `mfxU64`) types will be aligned by 8 with default packing on Win32, right? This might be resolved with introducing `PACK_BEGIN_MAX` -
```
#if defined(_WIN64) || defined(LINUX64) 
	#define PACK_BEGIN()     PACK_BEGIN_X(8)
	#define PACK_BEGIN_MAX() PACK_BEGIN()
	#define x64
#elif defined(_WIN32)
	#define PACK_BEGIN()     PACK_BEGIN_X(4)
	#define PACK_BEGIN_MAX() PACK_BEGIN_X(8)
#elif defined(LINUX32)
	#define PACK_BEGIN()     PACK_BEGIN_X(4)
	#define PACK_BEGIN_MAX() PACK_BEGIN()
#else
	#error Unknown packing
#endif
```
```
//API header
PACK_BEGIN_MAX()
typedef struct { 
	//...
} mfxFrameData; 
PACK_END()
```

mgonchar:
@dmitry-gurulev Hm, yes I like this one. I will try it

api/include/mfxastructures.h
```diff
@@ -119,7 +120,13 @@ typedef struct {
         };
     };
 } mfxAudioInfoMFX;
+#pragma pack(pop)
 
+#if defined(_WIN64) || defined(LINUX64)
```
dvrogozh: Wait!!! This seems to be a direct regression to mediasdk end-user experience who will require to set compiler undefined LINUX32 and/or LINUX64 properly. (I thinkw _WIN32 and _WIN64 are compiler-defined, right?)

We need to think about some other approach.

mgonchar: I've changed approach to get rid of ```LINUX32```/```LINUX64``` macro. Yes, ```_WIN32```/```_WIN64``` are compiler defined

mgonchar:
@dmitry-gurulev I think I can give better wording to names here, but let's wait for all CIs first, then I will upload one more patch

api/include/mfxdefs.h
```diff
@@ -50,6 +50,29 @@ extern "C"
 {
 #endif /* __cplusplus */
 
+#if defined(_MSC_VER)
+    #define PRAGMA_IMPL(x) __pragma(x)
+#else
+    #define PRAGMA_IMPL(x) _Pragma(#x)
+#endif
+
+#define PACK_BEGIN_X(x) PRAGMA_IMPL(pack(push, x))
```
dvrogozh: Shouldn't macro names be prefixed with MFX_ to avoid interleaving with something user-level defined?

mgonchar: Well, probably yes

mgonchar: done

api/include/mfxdefs.h
```diff
@@ -50,6 +50,29 @@ extern "C"
...
+#else
+    #define MFX_PRAGMA_IMPL(x) _Pragma(#x)
+#endif
+
+#define MFX_PACK_BEGIN_X(x) MFX_PRAGMA_IMPL(pack(push, x))
+
+#if defined(_WIN64) || defined(__LP64__)
+    #define MFX_PACK_BEGIN()     MFX_PACK_BEGIN_X(8)
+    #define MFX_PACK_BEGIN_MAX() MFX_PACK_BEGIN()
+    #define x64
```
dvrogozh: what is this x64 define about?

api/include/mfxdefs.h
```diff
@@ -50,6 +50,29 @@ extern "C"
...
+#elif defined(_WIN32) || defined(_M_IX86) && !defined(__linux__)
+    #define MFX_PACK_BEGIN()     MFX_PACK_BEGIN_X(4)
+    #define MFX_PACK_BEGIN_MAX() MFX_PACK_BEGIN_X(8)
+#elif defined(_M_IX86) && defined(__linux__) || defined(__ILP32__)
+    #define MFX_PACK_BEGIN()     MFX_PACK_BEGIN_X(4)
+    #define MFX_PACK_BEGIN_MAX() MFX_PACK_BEGIN()
```
dvrogozh: I don't intuitively understand when you are going to use MFX_PACK_BEGIN and MFX_PACK_BEGIN_MAX. Do names signify anything special? Why _MAX?

api/include/mfxdefs.h
```diff
@@ -50,6 +50,29 @@ extern "C"
...
 
+#if defined(_MSC_VER)
+    #define MFX_PRAGMA_IMPL(x) __pragma(x)
+#else
+    #define MFX_PRAGMA_IMPL(x) _Pragma(#x)
+#endif
+
+#define MFX_PACK_BEGIN_X(x) MFX_PRAGMA_IMPL(pack(push, x))
+
+#if defined(_WIN64) || defined(__LP64__)
```
dvrogozh: Do you have some links to confirm that all linux compilers define these __LP64__ and other macro defs?

api/include/mfxdefs.h
mgonchar: will remove

api/include/mfxdefs.h
mgonchar: Maybe there is a better naming. Meaning was to guarantee correct alignment of long long int / double which is different across 32 win/lin

api/include/mfxdefs.h
mgonchar: Well, I checked sources of gcc/clang and they do (unfortunately didn't find it in docs)

api/include/mfxstructures.h
```diff
@@ -2057,15 +2203,19 @@ typedef struct {
...
 typedef struct {
     mfxU16 Type;
     mfxU16 reserved1;
     mfxU32 Offset;
     mfxU32 Size;
     mfxU32 reserved[5];
 } mfxEncodedUnitInfo;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN()
```
dvrogozh: mfxU64  reserved1 - why not _MAX?

api/include/mfxstructures.h
```diff
@@ -1968,14 +2101,17 @@ enum {
...
 typedef struct {
     mfxU16  FeatureEnabled;         /* see enum SegmentFeature */
     mfxI16  QIndexDelta;
     mfxI16  LoopFilterLevelDelta;
     mfxU16  ReferenceFrame;        /* see enum VP9ReferenceFrame */
     mfxU16  reserved[12];
 } mfxVP9SegmentParam;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN()
```
dvrogozh: mfxU64          reserved1 - why not _MAX?

api/include/mfxdefs.h
```diff
@@ -50,6 +50,41 @@ extern "C"
...
+    #define MFX_PRAGMA_IMPL(x) __pragma(x)
+#else
+    #define MFX_PRAGMA_IMPL(x) _Pragma(#x)
+#endif
+
+/* The general rule for alignment is following:
+   - structures with pointers have 4/8 bytes alignment on 32/64 bit systems
+   - structures with fields of tupe mfxU64/mfxF64 (unsigned long long / double)
+     have alignment 8 bytes on 64 bit and 32 bit Windows, on Linux alignment is 4 bytes
+   - all the rest structures are 4 bytes aligned */
```
dvrogozh: It might help if you will add macro name you intend to use in each of the above 3 cases. Also, might help to highlight what you intend to do in case of structure containing both 1 and 2 from the above?

api/include/mfxfeihevc.h
```diff
@@ -106,22 +113,28 @@ typedef struct {
 
     mfxFeiHevcEncCtuCtrl *Data;
 } mfxExtFeiHevcEncCtuCtrl;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: No it shouldn't. Pointer size depends on arch, but array is not a pointer. It decays to pointer without explicit conversion, but has different type, different size and alignment

api/include/mfxstructures.h
```diff
@@ -44,8 +45,9 @@ typedef struct {
         };
     };
 } mfxFrameId;
+MFX_PACK_END()
 
-#pragma pack(push, 4)
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: Historical reasons. See original code. We've already had several align pragmas before this change. They were added to structs which hadn't had pointers, but pointers were introduced to them at some moment. It was done to keep alignment of other members to previous state without pointers (that was 4-bytes alignment)

api/include/mfxplugin.h
```diff
@@ -68,6 +68,7 @@ typedef enum {
     MFX_THREADPOLICY_PARALLEL  = 1
 } mfxThreadPolicy;
 
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: Yes, I think we can add it at least to be consistent

api/include/mfxstructures.h
```diff
@@ -257,20 +259,24 @@ typedef struct {
...
 } mfxFrameSurface1;
+MFX_PACK_END()
 
 enum {
     MFX_TIMESTAMPCALC_UNKNOWN = 0,
     MFX_TIMESTAMPCALC_TELECINE = 1,
 };
 
 /* Transcoding Info */
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: Unions have alignment equal to maximal alignment of it's elements

api/include/mfxdefs.h
```diff
@@ -50,6 +50,41 @@ extern "C"
...
+
+#if defined(_MSC_VER)
+    #define MFX_PRAGMA_IMPL(x) __pragma(x)
+#else
+    #define MFX_PRAGMA_IMPL(x) _Pragma(#x)
+#endif
+
+/* The general rule for alignment is following:
+   - structures with pointers have 4/8 bytes alignment on 32/64 bit systems
+   - structures with fields of tupe mfxU64/mfxF64 (unsigned long long / double)
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxbrc.h
```diff
@@ -32,6 +32,7 @@ enum {
     MFX_EXTBUFF_BRC = MFX_MAKEFOURCC('E','B','R','C')
 };
 
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxbrc.h
```diff
@@ -50,7 +51,9 @@ typedef struct {
     mfxU16 NumExtParam;
     mfxExtBuffer** ExtParam;
 } mfxBRCFrameParam;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxcamera.h
```diff
@@ -136,21 +153,27 @@ typedef struct {
...
 
+MFX_PACK_BEGIN_X(4)
 typedef struct {
     mfxExtBuffer    Header;
     mfxU16          Threshold;
     mfxU16          reserved[27];
 } mfxExtCamBayerDenoise;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxcamera.h
```diff
@@ -167,20 +191,25 @@ typedef enum {
...
 typedef struct{
     mfxU16 Pixel;
     mfxU16 Red;
     mfxU16 Green;
     mfxU16 Blue;
 } mfxCamFwdGammaSegment;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN()
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxcamera.h
```diff
@@ -209,13 +241,16 @@ enum {
...
+MFX_PACK_BEGIN_X(4)
 typedef struct {
     mfxU16 R;
     mfxU16 G;
     mfxU16 B;
     mfxU16 Reserved;
 } mfxCam3DLutEntry;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN()
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxcommon.h
```diff
@@ -87,6 +91,7 @@ typedef enum
 } mfxPriority;
 
 typedef struct _mfxEncryptedData mfxEncryptedData;
+MFX_PACK_BEGIN()
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxstructures.h
```diff
@@ -909,52 +926,67 @@ enum {
...
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_X(4)
 typedef struct {
     mfxExtBuffer    Header;
     mfxU16  DetailFactor;
 } mfxExtVPPDetail;
+MFX_PACK_END()
 
+MFX_PACK_BEGIN_X(4)
```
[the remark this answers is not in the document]
mgonchar: fixed

api/include/mfxstructures.h
mgonchar: fixed

api/include/mfxstructures.h
mgonchar: fixed

api/include/mfxplugin.h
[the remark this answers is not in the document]
mgonchar: Added alignment

api/include/mfxdefs.h
mgonchar: you mean something like this?
```
#define MFX_PACK_USUAL_STRUCT() MFX_PACK_BEGIN_X(4)
#define MFX_PACK_STRUCT_W_PTR() MFX_PACK_BEGIN()
#define MFX_PACK_STRUCT_L_TYPE() MFX_PACK_BEGIN_MAX()
```
(you may suggest better names)
of course we can get rid of ```MFX_PACK_BEGIN``` and ```MFX_PACK_BEGIN_MAX``` is such case

mgonchar: implemented in new patch

api/include/mfxstructures.h
```diff
@@ -1632,7 +1728,7 @@ enum {
 
 #endif
 
-#pragma pack(push, 4)
+MFX_PACK_BEGIN_USUAL_STRUCT()
```
dvrogozh: `mfxU64          GeneralConstraintFlags;`

If that's to preserve old alignment, I suggest to provide explicit comment in the code before the struct.

api/include/mfxstructures.h
```diff
@@ -44,8 +45,9 @@ typedef struct {
         };
     };
 } mfxFrameId;
+MFX_PACK_END()
 
-#pragma pack(push, 4)
+MFX_PACK_BEGIN_USUAL_STRUCT()
```
dvrogozh: `mfxU64 BufferSize;`

I suggest to have explicit comment if we need to preserve old alignment.

api/include/mfxstructures.h
mgonchar: ok

api/include/mfxstructures.h
mgonchar: ok

mgonchar:
@dvrogozh I've updated comments as you requested

PR #1430, opened 2019-05-20
Fix klocwork issue in umc_h265_va_packer_cenc.cpp

LukeZheZhu:
Fix klocwork issue. 
Issue ID: 99914
Pointer 'supplier->GetDPBList()' returned from call to function 'GetDPBList' at line 143 may be NULL and will be dereferenced at line 143


_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer_cenc.cpp
```diff
@@ -132,7 +132,9 @@ namespace UMC_HEVC_DECODER
         for(; index < rps->getNumberOfNegativePictures() + rps->getNumberOfPositivePictures() + rps->getNumberOfLongtermPictures(); index++)
         {
             int32_t poc = rps->getPOC(index);
-            H265DecoderFrame *pFrm = supplier->GetDPBList()->findLongTermRefPic(pCurrentFrame, poc, pSeqParamSet->log2_max_pic_order_cnt_lsb, !rps->getCheckLTMSBPresent(index));
+            H265DecoderFrame *pFrm = 0;
```
mgonchar: nullptr

PR #1432, opened 2019-05-21
[h264e] Report/check VBR/CQP/CBR via caps

AntonGrishin:
We can not modify ENCODE_CAPS struct, but need to add new BRC
checks in caps. This commit contains new struct (with encode_caps
inside) and code refactor to work with it

Issue: MDP-54861

### DIFF
--- a/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
+++ b/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -314,6 +314,15 @@ typedef struct tagENCODE_CAPS
     UCHAR    MaxNum_WeightedPredL1;
 } ENCODE_CAPS;
 
+// We cannot modify ENCODE_CAPS structure, but need to add new features
+struct MFX_ENCODE_CAPS
+{
+    ENCODE_CAPS ddi_caps;
+    //bitrate control capabilities provided by MSDK
+    bool CQPSupport;
+    bool CBRSupport;
+    bool VBRSupport;
+};
 
 // this enumeration is used to define the block size for intra prediction. they
 // are used as bit flags to control what block size will be checked for intra
--- a/_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
+++ b/_studio/mfx_lib/encode_hw/h264/include/mfx_h264_encode_hw_utils.h
@@ -305,7 +305,7 @@ namespace MfxHwH264Encode
         DdiTask const &       task);
 
     bool isBitstreamUpdateRequired(MfxVideoParam const & video,
-        ENCODE_CAPS caps,
+        MFX_ENCODE_CAPS caps,
         eMFXHWType platform);
     // Helper which checks number of allocated frames and auto-free.
 
@@ -406,13 +406,13 @@ namespace MfxHwH264Encode
 
     // add hwType param
     mfxStatus CheckEncodeFrameParam(
-        MfxVideoParam const & video,
-        mfxEncodeCtrl *       ctrl,
-        mfxFrameSurface1 *    surface,
-        mfxBitstream *        bs,
-        bool                  isExternalFrameAllocator,
-        ENCODE_CAPS const &   caps,
-        eMFXHWType            hwType = MFX_HW_UNKNOWN);
+        MfxVideoParam const &      video,
+        mfxEncodeCtrl *            ctrl,
+        mfxFrameSurface1 *         surface,
+        mfxBitstream *             bs,
+        bool                       isExternalFrameAllocator,
+        MFX_ENCODE_CAPS const &    caps,
+        eMFXHWType                 hwType = MFX_HW_UNKNOWN);
 
     template<typename T> void Clear(std::vector<T> & v)
     {
@@ -2095,7 +2095,7 @@ namespace MfxHwH264Encode
         MfxFrameAllocResponse   m_opaqResponse;     // Response for opaq
         MfxFrameAllocResponse   m_histogram;
 
-        ENCODE_CAPS             m_caps;
+        MFX_ENCODE_CAPS         m_caps;
         mfxStatus               m_failedStatus;
         mfxU32                  m_inputFrameType;
         mfxU32                  m_NumSlices;
@@ -2553,8 +2553,8 @@ namespace MfxHwH264Encode
         mfxU32                frameOrderInGop,
         mfxEncodeCtrl const * ctrl,
         mfxU16                intraStripeWidthInMBs,
-        SliceDivider &  divider,
-        ENCODE_CAPS     caps);
+        SliceDivider &        divider,
+        MFX_ENCODE_CAPS       caps);
 
     mfxStatus UpdateIntraRefreshWithoutIDR(
         MfxVideoParam const & oldPar,
@@ -2563,8 +2563,8 @@ namespace MfxHwH264Encode
         mfxI64                oldStartFrame,
         mfxI64 &              updatedStartFrame,
         mfxU16 &              updatedStripeWidthInMBs,
-        SliceDivider &  divider,
-        ENCODE_CAPS     caps);
+        SliceDivider &        divider,
+        MFX_ENCODE_CAPS       caps);
 
     BiFrameLocation GetBiFrameLocation(
         MfxVideoParam const & video,
@@ -2598,10 +2598,10 @@ namespace MfxHwH264Encode
         bool                  optimize);
 
     void ConfigureTask(
-        DdiTask &             task,
-        DdiTask const &       prevTask,
-        MfxVideoParam const & video,
-        ENCODE_CAPS const &   caps);
+        DdiTask &                 task,
+        DdiTask const &           prevTask,
+        MfxVideoParam const &     video,
+        MFX_ENCODE_CAPS const &   caps);
 
     mfxStatus GetNativeHandleToRawSurface(
         VideoCORE &           core,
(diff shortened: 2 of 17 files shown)


_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -314,6 +314,20 @@ typedef struct tagENCODE_CAPS
     UCHAR    MaxNum_WeightedPredL1;
 } ENCODE_CAPS;
 
+// We cannot modify ENCODE_CAPS structure, but need to add new features
+struct MFX_ENCODE_CAPS
+{
+    ENCODE_CAPS ddi_caps;
```
mgonchar: why not inherit new caps from previous? it will remove all those massive changes to introduce ```.ddi_caps``` in myriad places.
```
struct MFX_ENCODE_CAPS : ENCODE_CAPS
{
    union {
        struct {
            UINT CQPBRCSupport: 1;
            UINT CBRSupport   : 1;
            UINT VBRSupport   : 1;
            UINT              : 28;
        };
        UCHAR   CodingLimits3;
    }
};
```

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -314,6 +314,20 @@ typedef struct tagENCODE_CAPS
...
+    union {
+        struct {
+            UINT CQPBRCSupport: 1;
+            UINT CBRSupport   : 1;
+            UINT VBRSupport   : 1;
+            UINT              : 28;
+        };
+        UCHAR   CodingLimits3;
```
mgonchar: why ```UCHAR``` ? shouldn't it be ```UINT```?

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -314,6 +314,20 @@ typedef struct tagENCODE_CAPS
...
+// We cannot modify ENCODE_CAPS structure, but need to add new features
+struct MFX_ENCODE_CAPS
+{
+    ENCODE_CAPS ddi_caps;
+    union {
+        struct {
+            UINT CQPBRCSupport: 1;
+            UINT CBRSupport   : 1;
+            UINT VBRSupport   : 1;
+            UINT              : 28;
```
mgonchar: ```UINT              : 29;``` ?

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -314,6 +314,20 @@ typedef struct tagENCODE_CAPS
     UCHAR    MaxNum_WeightedPredL1;
 } ENCODE_CAPS;
 
+// We cannot modify ENCODE_CAPS structure, but need to add new features
+struct MFX_ENCODE_CAPS
+{
+    ENCODE_CAPS ddi_caps;
+    union {
+        struct {
+            UINT CQPBRCSupport: 1;
```
dmitryermilov: I think you don't need to be economical with bits here. I suggest to use bool's without union.
The rest is fine.

AntonGrishin: Done.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
AntonGrishin: Change union to bool

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
AntonGrishin: same

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -314,6 +314,14 @@ typedef struct tagENCODE_CAPS
     UCHAR    MaxNum_WeightedPredL1;
 } ENCODE_CAPS;
 
+// We cannot modify ENCODE_CAPS structure, but need to add new features
+struct MFX_ENCODE_CAPS
+{
+    ENCODE_CAPS ddi_caps;
+    bool CQPBRCSupport: 1;
+    bool CBRSupport   : 1;
+    bool VBRSupport   : 1;
```
mgonchar: no, please don't do it. do not introduce bitfields without strict requirement. Convert to simple bools:
```
    bool CQPBRCSupport;
    bool CBRSupport;
    bool VBRSupport;
```

AntonGrishin: Sorry, mistake after refactored. Done.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
AntonGrishin: First, i thought about inherit, but @dmitryermilov said me what current way more better.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
```diff
@@ -314,6 +314,14 @@ typedef struct tagENCODE_CAPS
     UCHAR    MaxNum_WeightedPredL1;
 } ENCODE_CAPS;
 
+// We cannot modify ENCODE_CAPS structure, but need to add new features
+struct MFX_ENCODE_CAPS
+{
+    ENCODE_CAPS ddi_caps;
+    bool CQPBRCSupport;
```
mgonchar: A little disproportion in names... You have ```BRC``` in first name but not in two subsequent. I suggest either remove it from first one or add to the rest to align naming style

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
mgonchar: ok

AntonGrishin: My way provide more clearly interface and separates our structure from ddi. 
But your way is better from another point, because this have no need to massive refactor.
What we must choose? Dmitry doesn`t mind about inheritance, but from my point of view it`s a matter of taste.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
AntonGrishin: CBR and VBR contains BR in name of method, but CQP not contains this one, in this way i added brc suffix. If you think what this will more clearly, i will change it.

AntonGrishin: Aligned with first.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
dmitryermilov: I'm personally ok with both: inheritance and inclusion. Advantage of inclusion is it will be clear where ddi part of caps is and where our custom part is.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
lakulako: "constant bit rate bit rate control" sounds ugly. It was clear without BRC what is it about. Common comment would be enough.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
lakulako: I would vote for inheritance. Less code, much less changes, more modern style, which was introduced for such cases.

mgonchar: I think we can do following thing:
1) Use inheritance
2) Pass caps by reference using ```ENCODE_CAPS const & hwCaps``` object everywhere when possible
3) Use ```MFX_ENCODE_CAPS``` only when required

@dmitryermilov what do you think?

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
AntonGrishin: @lakulako I am confused. You think I must revert changes (CQPBRC, CBR, VBR) and add comments?

lakulako: Don't use BRC at all and add common comment like "Bit rate control capabilities, provided by MSDK"

AntonGrishin: Done.

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
dmitryermilov: Hi @mgonchar , 
Sorry but I don't see the idea behind 2) and 3) above.  The purpose of this change is to extend current ENCODE_CAPS structure. So apparently once we extend it, we need to use a new MFX_ENCODE_CAPS (which encapsulates old  ENCODE_CAPS) everywhere.

mgonchar: Idea is to not introduce sufficient delta. I.e. most of the functions accepting caps don't need our additional fields, and if we use const ref and pass base class (even if we pass only .ddi in new notation) it will decrease delta of this patch + limit scope of usage of new fields. Honestly I just don't like huge delta which we have now and thinking about reducing it, I think we can do it even if extension will be done via inclusion

mgonchar: For 2-3) look for example here: https://github.com/Intel-Media-SDK/MediaSDK/pull/1432/files#diff-0bd236db445515bb8ccdb4a89eb6f852L220 I think that better will be to not change signature of this function, but pass ```.ddi_caps``` member inside instead. @onabiull @dmitryermilov what do you think about such simplification?

dmitryermilov: I do see your point. However I think we shouldn't be afraid of changing code here. IMO, confusion , which readers may have coming across different caps strictures is more important. Just my private opinion.

dmitryermilov:
@dvrogozh , your opinion ?

_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h
mgonchar: ok. I removed my -1 for this one since no major issues remained here

PR #1433, opened 2019-05-21
Use infinite timeout in SyncOperation in case of MFX_WRN_DEVICE_BUSY …

dmitryermilov:
…handling

SyncOperation with 1 ms timeout was a bus and caused high CPU usage in corner cases.


samples/sample_common/src/sample_utils.cpp
```diff
@@ -2466,24 +2466,28 @@ mfxStatus CH264FrameReader::PrepareNextFrame(mfxBitstream *in, mfxBitstream **ou
...
+        {
+            MSDK_TRACE_ERROR(MSDK_STRING("WaitForDeviceToBecomeFree: SyncOperation failed, sts = ") << stsSync);
+            currentStatus = MFX_ERR_ABORTED;
         }
-    } else {
-        MSDK_SLEEP(DEVICE_WAIT_TIME);
+    }
+    else
+    {
+        MSDK_SLEEP(1);
```
mgonchar: shouldn't we set ```currentStatus``` here?

dmitryermilov: probably yes

dmitryermilov:
> We need some validation results to approve the change. Did it result into any CPU% improvements?

WIP

PR #1435, opened 2019-05-22
Avoid nullptr situaton of supplier->GetDPBList() in PackPicParams fun…

LukeZheZhu:
Add an if() branch to avoid nullptr

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer_cenc.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer_cenc.cpp
@@ -129,6 +129,9 @@ namespace UMC_HEVC_DECODER
         for(; index < rps->getNumberOfNegativePictures() + rps->getNumberOfPositivePictures(); index++)
                 pocList[numRefPicSetStCurrBefore + numRefPicSetStCurrAfter++] = picParam->CurrPic.pic_order_cnt + rps->getDeltaPOC(index);
 
+        if(!(supplier->GetDPBList()))
+            throw h265_exception(UMC_ERR_FAILED);
+
         for(; index < rps->getNumberOfNegativePictures() + rps->getNumberOfPositivePictures() + rps->getNumberOfLongtermPictures(); index++)
         {
             int32_t poc = rps->getPOC(index);


_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer_cenc.cpp
```diff
@@ -129,18 +129,21 @@ namespace UMC_HEVC_DECODER
         for(; index < rps->getNumberOfNegativePictures() + rps->getNumberOfPositivePictures(); index++)
                 pocList[numRefPicSetStCurrBefore + numRefPicSetStCurrAfter++] = picParam->CurrPic.pic_order_cnt + rps->getDeltaPOC(index);
 
-        for(; index < rps->getNumberOfNegativePictures() + rps->getNumberOfPositivePictures() + rps->getNumberOfLongtermPictures(); index++)
+        if(!(supplier->GetDPBList()))
```
dmitry-gurulev: Absence of DPB (null pointer) here is a logic error u can't just silently ignore it. I'd rather throw an exception.

LukeZheZhu: Hi,
I use "throw h265_exception(UMC_ERR_NULL_PTR)" here instead of just silently ignoreing it. Is that OK?

_studio/shared/umc/codec/h265_dec/src/umc_h265_va_packer_cenc.cpp
```diff
@@ -129,6 +129,9 @@ namespace UMC_HEVC_DECODER
         for(; index < rps->getNumberOfNegativePictures() + rps->getNumberOfPositivePictures(); index++)
                 pocList[numRefPicSetStCurrBefore + numRefPicSetStCurrAfter++] = picParam->CurrPic.pic_order_cnt + rps->getDeltaPOC(index);
 
+        if(!(supplier->GetDPBList()))
+            throw h265_exception(UMC_ERR_NULL_PTR);
```
dmitry-gurulev: `UMC_ERR_NULL_PTR` is described as
> null pointer in input parameters
at `_studio\shared\umc\core\umc\include\umc_structures.h`. So, I'd rather use `UMC_ERR_FAILED` instead.

LukeZheZhu: Use UMC_ERR_FAILED replacing the UMC_ERR_NULL_PTR

PR #1438, opened 2019-05-22
Avoid uninitialized of m_payloads

LukeZheZhu:
The "m_payload" may uninitialized in constructor function SeiPayloadArray(const SeiPayloadArray & payloads) of umc_h264_au_splitter.cpp

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
@@ -40,9 +40,16 @@ SeiPayloadArray::SeiPayloadArray()
 SeiPayloadArray::SeiPayloadArray(const SeiPayloadArray & payloads)
 {
     size_t count = payloads.GetPayloadCount();
-    for (size_t i = 0; i < count; i++)
+    if (count == 0)
     {
-        AddPayload(payloads.GetPayload(i));
+        SeiPayloadArray();
+    }
+    else
+    {
+        for (size_t i = 0; i < count; i++)
+        {
+            AddPayload(payloads.GetPayload(i));
+        }
     }
 }
 


_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -40,9 +40,16 @@ SeiPayloadArray::SeiPayloadArray()
 SeiPayloadArray::SeiPayloadArray(const SeiPayloadArray & payloads)
 {
     size_t count = payloads.GetPayloadCount();
-    for (size_t i = 0; i < count; i++)
+    if (count == 0)
     {
-        AddPayload(payloads.GetPayload(i));
+        SeiPayloadArray();
```
dmitry-gurulev: Here u just create temporary `SeiPayloadArray`. Maybe u're trying delegate constructor here? Delegating constructors works in the constructor's [initialization list](https://en.cppreference.com/w/cpp/language/initializer_list#Delegating_constructor) only.

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
```diff
@@ -40,9 +40,16 @@ SeiPayloadArray::SeiPayloadArray()
 SeiPayloadArray::SeiPayloadArray(const SeiPayloadArray & payloads)
 {
     size_t count = payloads.GetPayloadCount();
-    for (size_t i = 0; i < count; i++)
+    if (count == 0)
```
mgonchar: ```m_payloads``` is a ```std::vector``` which is default constructable and well-defined in case if ```count==0```. you don't need to change anything

_studio/shared/umc/codec/h264_dec/src/umc_h264_au_splitter.cpp
LukeZheZhu: Hi @dmitry-gurulev 
I just want to call default constructor if count equals to zero.

mgonchar: @LukeZheZhu here you just create temporary nameless object which will be destroyed at the end of scope. But as I said before, in my understanding you don't need to do anything. Which real problem you are trying to solve?

LukeZheZhu: The kk reports the m_payloads will uninitialized in this function.

mgonchar: it is a ```std::vector``` it will be properly initialized by default. Static analysis reports false positive alert. No changes required

PR #1439, opened 2019-05-22
[HEVCe] Setting quality factor in QVBR case

semelianova:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -381,6 +381,7 @@ mfxStatus SetRateControl(
     VAStatus vaSts;
     VAEncMiscParameterBuffer *misc_param;
     VAEncMiscParameterRateControl *rate_param;
+    mfxExtCodingOption3 const & extOpt3 = par.m_ext.CO3;
```
mgonchar: Currently this variable used only on line 423, why don't you define it there?

semelianova: Changes are based on the same part of code in AVC. I did it so that there are not discrepancy.

mgonchar: well, sure it is not big issue, but looks like it is room for optimization for HEVC and AVC. The only thing that should be taken into account is CS codebase, maybe this variable used there.

PR #1442, opened 2019-05-22
Enabled support of I915_SAMPLE_WAIT, I915_SAMPLE_SEMA PMU in metrics_…

dmitryermilov:
…monitor

### DIFF
--- a/samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
+++ b/samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
@@ -40,10 +40,33 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <time.h>
 #include <unistd.h>
 #include <i915_drm.h>
+#include <xf86drm.h>
+
+static bool has_param(int fd, int param)
+{
+    drm_i915_getparam_t gp;
+    int tmp = 0;
+
+    memset(&gp, 0, sizeof(gp));
+    gp.value = &tmp;
+    gp.param = param;
+
+    if (drmIoctl(fd, DRM_IOCTL_I915_GETPARAM, &gp))
+    {
+        return false;
+    }
+
+    return tmp > 0;
+}
 
 static bool is_engine_config(uint64_t config)
 {
     return config < __I915_PMU_OTHER(0);
+}
+
+static uint8_t get_engine_sample(uint64_t config)
+{
+    return config & I915_PMU_SAMPLE_MASK;
 }
 
 static int get_engine_id(uint64_t config)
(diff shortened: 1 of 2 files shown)


dmitryermilov:
@dvrogozh , can you please review?

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -321,15 +321,15 @@ cttStatus CTTMetrics_PMU_Init()
     /* When adding new condigs, don't mix metrics from different groups!*/
     int configs[] = {
         /* Render engine metrics */
-        I915_PMU_ENGINE_BUSY(I915_ENGINE_CLASS_RENDER, 0), /*I915_PMU_ENGINE_WAIT(I915_ENGINE_CLASS_RENDER, 0), I915_PMU_ENGINE_SEMA(I915_ENGINE_CLASS_RENDER, 0),*/
+        I915_PMU_ENGINE_BUSY(I915_ENGINE_CLASS_RENDER, 0), I915_PMU_ENGINE_WAIT(I915_ENGINE_CLASS_RENDER, 0), I915_PMU_ENGINE_SEMA(I915_ENGINE_CLASS_RENDER, 0),
```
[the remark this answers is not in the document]
dmitryermilov: Done

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -518,13 +529,35 @@ cttStatus CTTMetrics_PMU_GetValue(unsigned int count, float* out_metric_values)
...
                 value *= 100; // we need %
+                value_sema *= 100;
+                value_wait *= 100;
+            }
+
+            value /= time;
+            if (is_sema_wait_available) {
+                value_sema /= time_sema;
+                value_wait /= time_wait;
+                value -= value_sema + value_wait; // Substract sema and wait from busy metric
```
dvrogozh: We should do that only if i915 supports preemption. Otherwise, we need to count sema and wait as a time engine is really occupied.

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -518,13 +529,35 @@ cttStatus CTTMetrics_PMU_GetValue(unsigned int count, float* out_metric_values)
...
+                value_wait = perf_elapsed(&group->metrics[I915_SAMPLE_WAIT]);
+                time_wait = (double)perf_elapsed_time(&group->metrics[I915_SAMPLE_WAIT]);
+            }
+
             if (g_ctx.metrics[metric_idx] == CTT_AVG_GT_FREQ) {
                 value *= 1000000000; // to compensate time in ns
             }
             else {
                 value *= 100; // we need %
+                value_sema *= 100;
```
mgonchar: Well, here it is possible that ```is_sema_wait_available == false```, so ```value_sema``` and ```value_wait``` will be multiple of garbage by 100. It is not a big issue in current code, but I think better to add zero initialization at declaration to avoid unpredictable values in such case

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -40,12 +40,38 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <time.h>
 #include <unistd.h>
 #include <i915_drm.h>
+#include <xf86drm.h>
+
+char I915_DRI_DIR[16] = "/dev/dri/card0";
```
dvrogozh: This will limit the usage to the 1st card being intel and limit access to privileged user. Access to render node would be more preferable I think. You may try to explore libdrm for the helper function to open the device.

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -40,12 +40,38 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    memset(&gp, 0, sizeof(gp));
+    gp.value = &tmp;
+    gp.param = param;
+
+    if (drmIoctl(fd, DRM_IOCTL_I915_GETPARAM, &gp))
+    {
+        return false;
+    }
+
+    errno = 0;
```
dvrogozh: why?

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
dvrogozh: looks done. Reopen if needed

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
dmitryermilov: debug leftover

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
dmitryermilov: Addressed

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -304,7 +347,14 @@ cttStatus CTTMetrics_PMU_Init()
     if (CTT_ERR_NONE != status)
         return status;
 
-    if (0 == perf_init(&g_ctx.pm, num_configs, configs)) {
+    g_ctx.gem_fd = drmOpenRender(128);
```
dvrogozh: That's still limited to Intel gfx being the first card on the system, right?

dmitryermilov: I just tested "drmOpen("i915", NULL);". t works

PR #1443, opened 2019-05-22
Adding ability to generate RPM packages with CMake/CPack

troyev:
Hello,

I have added an option to make a rpm with "make package" after the build. Please take a look and let me know what you think.

I looked through the contributing guidelines, and the instructions for testing look a bit out of date, so I wasn't quite sure how to add the test results. This change really shouldn't impact anything at run time.

Thanks,
Troy


CMakeLists.txt
```diff
@@ -202,3 +202,13 @@ message("  BUILD_TESTS                             : ${BUILD_TESTS}")
...
 message("*****************************************************************************")
+
+set(CPACK_COMPONENTS_ALL runtime devel)
+set(CPACK_GENERATOR "RPM")
+set(CPACK_PACKAGE_NAME "intel-mediasdk")
+set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
+set(CPACK_RPM_runtime_FILE_NAME "${CPACK_PACKAGE_NAME}-runtime.${API_VERSION}.${CMAKE_SYSTEM_PROCESSOR}.rpm")
+set(CPACK_RPM_devel_FILE_NAME "${CPACK_PACKAGE_NAME}-devel.${API_VERSION}.${CMAKE_SYSTEM_PROCESSOR}.rpm")
+set(CPACK_RPM_COMPONENT_INSTALL "ON")
+include(CPack)
```
dvrogozh: I would like to avoid introducing cpack support into the project. It often occurs that the tool is buggy and it is required to install/update cmake to the very latest version to get what you planned to. Secondly, Linux distros in any way ignore this packaging support and develop .spec or debian packaging on their own. See here debian example: https://salsa.debian.org/debian/intel-mediasdk.

troyev: Thanks for your consideration.
These sibling packages already use CPack:
https://github.com/intel/gmmlib/releases
https://github.com/intel/media-driver/blob/master/CMakeLists.txt

Would it not be consistent to add it here?

I was not able to find a site with an rpm/spec for intel-mediask for CentOS/RHEL that looks up-to-date. 
If you definitely want to keep CPack out, I can just maintain a patch or fork offline.

PR #1444, opened 2019-05-24
Add disable_presets option to sample_encode

vsidorow:


samples/sample_encode/src/sample_encode.cpp
```diff
@@ -173,7 +173,8 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage, ...)
 #endif
     msdk_printf(MSDK_STRING("Example: %s h265 -i InputYUVFile -o OutputEncodedFile -w width -h height -hw -p 2fca99749fdb49aeb121a5b63ef568f7\n"), strAppName);
     msdk_printf(MSDK_STRING("   [-preset <default,dss,conference,gaming>] - Use particular preset for encoding parameters\n"));
-    msdk_printf(MSDK_STRING("   [-pp] - Print preset parameters\n"));    msdk_printf(MSDK_STRING("\nExample: %s h265 -i InputYUVFile -o OutputEncodedFile -w width -h height -hw -p 2fca99749fdb49aeb121a5b63ef568f7\n"), strAppName);
+    msdk_printf(MSDK_STRING("   [-pp] - Print preset parameters\n"));
+    msdk_printf(MSDK_STRING("   [-disable_presets] - Do not use presets\n"));
```
mgonchar: add statement about default value:

```Do not use presets (default is false)```

vsidorow: This option does not have a default value

mgonchar: @vsidorow of course every option have default, this one have zero default, since entire struct is zeroed: https://github.com/Intel-Media-SDK/MediaSDK/blob/d051aedaff12424136684f6b6bed23706ec4a11d/samples/sample_encode/src/sample_encode.cpp#L1542

Other ways you should have added this default with newly introduced fields

PR #1445, opened 2019-05-26
Remove automatic umc mutex

akokoshn:


_studio/mfx_lib/decode/vc1/include/mfx_vc1_decode.h
```diff
@@ -222,7 +222,7 @@ class MFXVideoDECODEVC1 : public VideoDECODE, public MfxCriticalErrorHandler
     mfxU32                           m_BufOffset;
 
     // to synchronize async and sync parts
-    UMC::Mutex                       m_guard;
+    std::recursive_mutex             m_guard;
```
mgonchar: ```recursive_mutex``` is an antipatters which often indicates bad architecture design

_studio/shared/src/mfx_umc_alloc_wrapper.cpp
```diff
@@ -1104,14 +1105,14 @@ mfxStatus   mfx_UMC_FrameAllocator_D3D::PrepareToOutput(mfxFrameSurface1 *surfac
             surface.Data.Y = 0;
             surface.Data.MemId = idx;
             //Performance issue. We need to unlock mutex to let decoding thread run async.
-            guard.Unlock();
+            guard.unlock();
```
mgonchar: Again, not much sense in manual unlocks of ```recursive_mutex```

_studio/shared/umc/core/vm_plus/include/umc_mutex.h
```diff
@@ -40,28 +40,6 @@ class Mutex : public MfxMutex
     Mutex& operator=(const Mutex&);
```
mgonchar: what about ```MfxAutoMutex```, ```Mutex```, ```MfxMutex``` classes? Do we use them in some other code?

_studio/mfx_lib/decode/vc1/include/mfx_vc1_decode.h
akokoshn: Agree about bad design, but I wouldn't introduce redesign to this PR

_studio/shared/umc/core/vm_plus/include/umc_mutex.h
akokoshn: yes, we still do

_studio/mfx_lib/decode/vc1/include/mfx_vc1_decode.h
mgonchar: @akokoshn let's then split this PR to two, to have all the stuff related to ```recursive_mutex``` separetely

_studio/shared/src/mfx_umc_alloc_wrapper.cpp
dmitry-gurulev: @mgonchar,

But it still makes sense - when guard is locked only once it enables copying w/o lock.

mgonchar: @dmitry-gurulev yes, but it doesn't give any hint on why it is required. So it might be locked (if some other routine of this thread captured this mutex) or unlocked at this point. If copying works only when mutex is unlocked - we have a problem, if copying doesn't care about state of the mutex then I want to understand how possibility of that mutex to be unlocked helps to perform copy

dmitry-gurulev: @mgonchar,

AFAIR, copying itself doesn't care about state, unlocking just allow other threads do their work not waiting when copying finishes.

mgonchar: @dmitry-gurulev so this is kind of optimization that might happen if ```mutex``` ownership was exclusive. Well, better to look at this architecture closer, because such places looks suspicious

akokoshn:
WIP (waiting validation)

Issue #1446, opened 2019-05-27
[MediaSDK/tools] metrics_monitor build failed on Ubuntu 16.04.06


zhoujd:
It looks like following commits may cause MSDK build failed on Ubuntu 16.04.6
https://github.com/Intel-Media-SDK/MediaSDK/commit/11a171e6b4efd9675a845691e0756e13eee6496f 
https://github.com/Intel-Media-SDK/MediaSDK/commit/846c00ab3490d88b5421b8ed82fd614d22999afb

The libdrm version is **2.4.0** on Ubuntu 16.04.6 with 'apt install', but from the comments
Use PMU definitions from libdrm

libdrm >= **2.4.92** is required

**DMFX_BUILD_INFO="\"Linux-4.15.0-47-generic 4.15.0-47-generic | GNU 5.4.0 | glibc 2.23\""**
Following is build log - 

18:45:46,081 INFO  - cd /opt/media/build/msdk/__cmake_build/samples/metrics_monitor && /usr/bin/c++ -DLINUX -DLINUX32 -DLINUX64 -DMFX_API_VERSION=\"1.30\" -DMFX_BUILD_INFO="\"Linux-4.15.0-47-generic 4.15.0-47-generic | GNU 5.4.0 | glibc 2.23\"" -DMFX_GIT_COMMIT=\"\" -DMSDK_BUILD=\"\" -DUNIX -D_FILE_OFFSET_BITS=64 -D__USE_LARGEFILE64 -Dcttmetrics_EXPORTS -I/opt/media/build/msdk/api/mediasdk_structures -I/opt/media/build/msdk/api/include -I/opt/media/build/msdk/__cmake_build -I/opt/media/build/msdk/samples/metrics_monitor/include -msse4.2 -std=c++11 -m64 -g -g3 -pipe -fPIC -Wall -Wformat -Wformat-security -Wnon-virtual-dtor -Wno-deprecated-declarations -Wno-unknown-pragmas -Wno-unused -O3 -DNDEBUG -fPIC -I/opt/media/install/include -I/usr/include/libdrm -o CMakeFiles/cttmetrics.dir/src/cttmetrics_i915_pmu.cpp.o -c /opt/media/build/msdk/samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
[...]

PR #1450, opened 2019-05-27
[metrics_monitor] Use VERSION_GREATER for version checking

dmitryermilov:
Fixes: #1446

cmake-3.6 doesn't respect a minimal version specified for a component.
It doesn't work properly for constructions like:
pkg_check_modules(PKG_AAA AAA>=1.2.3)
if( PKG_AAA_FOUND)
endif()

So let's use this instead as a workaround:
if( ${PKG_XXX_VERSION} VERSION_GREATER 1.2.2 )

### DIFF
--- a/samples/metrics_monitor/CMakeLists.txt
+++ b/samples/metrics_monitor/CMakeLists.txt
@@ -1,12 +1,9 @@
 # metrics_monitor library
 
-pkg_check_modules(PKG_LIBDRM libdrm>=2.4.92)
+if( PKG_LIBDRM_FOUND AND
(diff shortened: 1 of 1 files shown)


dmitryermilov:
I believe it's a bug because the same "pkg_check_modules(PKG_LIBDRM libdrm>=2.4.92)" works fine  in 3.12

dvrogozh:
We now clearly have 2 pkg_check_modules lines for the same output variable. This can be "implementation defined" and change from version to version. Regardless, it is not the best practice to obscure higher level variable. Please, remove second pkg_check_modules.

PR #1452, opened 2019-05-28
[VP9D] syntax error handling politics - skip frames

IlyaSergeyev:
DecodeFrameCheck returns MFX_ERR_MORE_DATA instead of
UNDEFINED_BEHAVIOR, so app skip broken frames and keep decoding
stream

Test: decode Broken_VP9_FC* streams

### DIFF
--- a/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
+++ b/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
@@ -35,6 +35,7 @@
 #include "mfx_vpx_dec_common.h"
 #include <libmfx_core_vaapi.h>
 
+#define MFX_ERR_INVALID_STREAM ((mfxStatus)UMC::UMC_ERR_INVALID_STREAM)
 
 using namespace UMC_VP9_DECODER;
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -35,6 +35,7 @@
 #include "mfx_vpx_dec_common.h"
 #include <libmfx_core_vaapi.h>
 
+#define MFX_ERR_INVALID_STREAM ((mfxStatus)UMC::UMC_ERR_INVALID_STREAM)
```
mgonchar: I vote for exposure of this status to public API

IlyaSergeyev: Thank you

dmitry-gurulev: I don't understand why `VideoDECODEVP9_HW::DecodeFrameHeader` can't return `UMC::Status` and we do need this hack.

IlyaSergeyev: Just for make code a little bit more readable

PR #1455, opened 2019-05-29
Add limite to prevent second arg of vaMapBuffer function from out of …

LukeZheZhu:
…range

Limit buffer id less than 58

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -1765,6 +1765,7 @@ mfxStatus VAAPIEncoder::Execute(Task const & task, mfxHDLPair pair)
         codedBuffer = m_bsQueue[idxBs].surface;
         {
            MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaMapBuffer");
+           MFX_CHECK(codedBuffer <= VABufferTypeMax, MFX_ERR_UNSUPPORTED);
            vaSts = vaMapBuffer(
               m_vaDisplay,
               codedBuffer,
@@ -1787,10 +1788,12 @@ mfxStatus VAAPIEncoder::Execute(Task const & task, mfxHDLPair pair)
             for (int i = 0; i < num_buffers; i += 2)
             {
                 void *pBufferHeader;
+                MFX_CHECK(buf_id[i] <= VABufferTypeMax, MFX_ERR_UNSUPPORTED);
                 vaSts = vaMapBuffer(m_vaDisplay, buf_id[i], &pBufferHeader);
                 MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
                 void *pData;
+                MFX_CHECK(buf_id[i + 1] <= VABufferTypeMax, MFX_ERR_UNSUPPORTED);
                 vaSts = vaMapBuffer(m_vaDisplay, buf_id[i + 1], &pData);
                 MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1765,6 +1765,8 @@ mfxStatus VAAPIEncoder::Execute(Task const & task, mfxHDLPair pair)
         codedBuffer = m_bsQueue[idxBs].surface;
         {
            MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaMapBuffer");
+           if (codedBuffer > VABufferTypeMax)
```
mgonchar: please use ```MFX_CHECK``` macro:
```MFX_CHECK(codedBuffer <= VABufferTypeMax, MFX_ERR_UNSUPPORTED);```

File: _studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp [hunk 2]
```diff
@@ -1787,10 +1789,14 @@ mfxStatus VAAPIEncoder::Execute(Task const & task, mfxHDLPair pair)
             for (int i = 0; i < num_buffers; i += 2)
             {
                 void *pBufferHeader;
+                if (buf_id[i] > VABufferTypeMax)
```
mgonchar: same

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1787,10 +1789,14 @@ mfxStatus VAAPIEncoder::Execute(Task const & task, mfxHDLPair pair)
...
+                    return MFX_ERR_UNSUPPORTED;
                 vaSts = vaMapBuffer(m_vaDisplay, buf_id[i], &pBufferHeader);
                 MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
                 void *pData;
+                if (buf_id[i + 1] > VABufferTypeMax)
```
mgonchar: same

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1909,6 +1915,8 @@ mfxStatus VAAPIEncoder::QueryStatus(Task & task)
 
                 {
                     MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaMapBuffer");
+                    if (codedBuffer > VABufferTypeMax)
```
mgonchar: same

LukeZheZhu:
> Do we really need this change?
> 
> Driver has checks in `vaMapBuffer`
> 
> https://github.com/intel/media-driver/blob/efbb30c652efdf8efcdd6f20e1a40e0cfc22fa3d/media_driver/linux/common/ddi/media_libva.cpp#L2580

Code scan reports this issue. Maybe add a check here can let this function safer. Double check is always better than one. Furthermore, maybe this check can let the user easy to find mistakes, instead of tracing to the media driver level.
So what's your opinion?

mgonchar:
@LukeZheZhu I don't like the idea of introducing this check for several reasons:
1) It will confuse other developers, which will wonder why this check exists only here. If you want to introduce new check you have to change entire code base. And I don't want to overcomplicate developers experience, if this check is unnecessary you don't need to introduce it. 
2) Performance reasons. Since we use ```vaMapBuffer``` quite often additional checks will introduce excessive CPU work
3) Double check is good if value can change between checks. But here is another case, so double check for value that can't change is just unoptimal code :)

LukeZheZhu:
> @LukeZheZhu I don't like the idea of introducing this check for several reasons:
> 
> 1. It will confuse other developers, which will wonder why this check exists only here. If you want to introduce new check you have to change entire code base. And I don't want to overcomplicate developers experience, if this check is unnecessary you don't need to introduce it.
> 2. Performance reasons. Since we use `vaMapBuffer` quite often additional checks will introduce excessive CPU work
> 3. Double check is good if value can change between checks. But here is another case, so double check for value that can't change is just unoptimal code :)

@mgonchar 
I totally understand your point. I will discuss this issue with PM and code scan reporter to see if they can silence the warning of this situation.
Let's keep in touch.
Thanks

PR #1456, opened 2019-05-29
Set fei to false in libmfxsw_encode.cpp to avoid uninitialize problem

LukeZheZhu:


LukeZheZhu:
> Absolutely no sense in this change. This variable is initializrd on next line

Yeah. I absolutely agree with you. But code scan reports this problem. It there anything we can do to let code scan shut up?

mgonchar:
We can mark this hit as ```Not a problem``` in static analysis tool web interface and this will mute the hit

PR #1457, opened 2019-05-29
initialize the minLog2TileColumns and maxLog2TileColumn

LukeZheZhu:


LukeZheZhu:
> `minLog2TileColumns`, `maxLog2TileColumns` are output parameters of `GetTileNBits` function and will be filled inside. No explicit initialization required for this code

Code scan reports this issue. We will discuss it with PM. Thanks

PR #1458, opened 2019-05-29
[hevce] Report/check VBR/CQP/CBR/ICQ support via caps

AntonGrishin:
We can not modify ENCODE_CAPS_HEVC struct, but need to add new BRC
checks in caps. This commit contains new struct (with encode_caps_hevc
inside) and code refactor to work with it

Issue: MDP-53880

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/include/hevce_ddi_main.h
+++ b/_studio/mfx_lib/encode_hw/h265/include/hevce_ddi_main.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -112,3 +112,13 @@ typedef struct tagENCODE_CAPS_HEVC
     UINT     reserved32bits3;
 } ENCODE_CAPS_HEVC;
 
+//We cannot modify original struct, but we need to add new checks.
+struct MFX_ENCODE_CAPS_HEVC
+{
+    ENCODE_CAPS_HEVC ddi_caps;
+    //brc caps provided by msdk
+    bool CQPSupport;
+    bool CBRSupport;
+    bool VBRSupport;
+    bool ICQSupport;
+};
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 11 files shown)


_studio/mfx_lib/encode_hw/h265/include/hevce_ddi_main.h
```diff
@@ -112,3 +112,13 @@ typedef struct tagENCODE_CAPS_HEVC
     UINT     reserved32bits3;
 } ENCODE_CAPS_HEVC;
 
+//We cannot modify original struct, but we need to add new checks.
+struct MFX_ENCODE_CAPS_HEVC
+{
+    ENCODE_CAPS_HEVC ddi_caps;
```
otsirule: In my opinion, it would be better if you extend structure ENCODE_CAPS_HEVC, not include it in another one: struct MFX_ENCODE_CAPS_HEVC : ENCODE_CAPS_HEVC {...} In that case, you wouldn't need to change all existing caps with duplicated naming: m_caps.ddi_caps

dmitryermilov: We can't. It's from DDI.

AntonGrishin: @otsirule Yes, but with inclusion it will be clear where our custom part of caps, and where ddi part.
Also we discussed about this in [changes for h264](https://github.com/Intel-Media-SDK/MediaSDK/pull/1432), and I think we must stick to one code style.

otsirule: @dmitryermilov , we can. As Anton mentioned, you've discussed this in another pull request. 
I can't approve these changes, because find them excessive and it's difficult to work with such code. But I won't block it.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
```diff
@@ -83,58 +83,58 @@ DriverEncoder* CreatePlatformH265Encoder(VideoCORE* core, ENCODER_TYPE type)
 }
 
 // this function is aimed to workaround all CAPS reporting problems in mainline driver
-mfxStatus HardcodeCaps(ENCODE_CAPS_HEVC& caps, VideoCORE* core)
+mfxStatus HardcodeCaps(MFX_ENCODE_CAPS_HEVC& caps, VideoCORE* core)
```
onabiull: Do we need to use new type everywhere? This specific function looks like good candidate for leaving just ENCODE_CAPS_HEVC.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -830,7 +830,7 @@ bool CheckTU(mfxU8 support, mfxU16& tu)
 }
 
 mfxU32 GetDefaultLCUSize(MfxVideoParam const & par,
-                         ENCODE_CAPS_HEVC const & hwCaps)
+                         MFX_ENCODE_CAPS_HEVC const & hwCaps)
```
onabiull: And this...

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -926,7 +926,7 @@ mfxU16 GetMaxNumRef(MfxVideoParam &par, bool bForward)
 #endif // (MFX_VERSION >= 1025)
 
 #ifdef MFX_ENABLE_HEVCE_ROI
-mfxStatus CheckAndFixRoi(MfxVideoParam  const & par, ENCODE_CAPS_HEVC const & caps, mfxExtEncoderROI *ROI, bool &bROIViaMBQP)
+mfxStatus CheckAndFixRoi(MfxVideoParam  const & par, MFX_ENCODE_CAPS_HEVC const & caps, mfxExtEncoderROI *ROI, bool &bROIViaMBQP)
```
onabiull: And this...

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
AntonGrishin: I agree with next changes (and mb FixDirtyRect), but in this function we can hardcode caps which doesnt report by driver, including new checks.

PR #1462, opened 2019-05-30
Add limit to prevent tl.Layer from out of range

LukeZheZhu:
Code scan reporter and PM ask me to submit a patch for this kk issue.

### DIFF
--- a/_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
+++ b/_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
@@ -776,7 +776,7 @@ constexpr auto NUM_OF_SUPPORTED_EXT_BUFFERS = 7; // mfxExtVP9Param, mfxExtOpaque
         mfxU16 i = 0;
         mfxExtVP9TemporalLayers const & tl = GetExtBufferRef(par);
 
-        if (par.m_numLayers > 0)
+        if (par.m_numLayers > 0 && par.m_numLayers < 8)
         {
             mfxU32 maxScale = tl.Layer[par.m_numLayers - 1].FrameRateScale;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
```diff
@@ -776,7 +776,7 @@ constexpr auto NUM_OF_SUPPORTED_EXT_BUFFERS = 7; // mfxExtVP9Param, mfxExtOpaque
         mfxU16 i = 0;
         mfxExtVP9TemporalLayers const & tl = GetExtBufferRef(par);
 
-        if (par.m_numLayers > 0)
+        if (par.m_numLayers > 0 && par.m_numLayers < 8)
```
mgonchar: Did you check that this situation is possible? I think those parameters are adjusted in functions like ```CheckVideoParam``` and condition ```par.m_numLayers > 0``` just signalize us that temporal layers are used.

Anyway I think this function is wrong place for such adjustment. It should be done in another place

TianmiChen: @mgonchar , we checked that this parameters are not checked in CheckVideoParam(). For this critical klocwork issue, we have to fix it. If Zhe's solution is not good, could you please share one solution for that? 

Thanks,

dmitry-gurulev: 1. I'd rather use something like `std::extent` to derive upper limit of `par.m_numLayers` instead of hardcoded `8`
2. This limit is precondition for `CalcTemporalLayerIndex` and we need to somehow report about violation (assertion at least, not just silently do nothing and return `0`

PR #1473, opened 2019-06-07
[hevce] Fix RoI/DirtyRect behavior

AntonGrishin:
Issue: MDP-53548

Change-Id: I9fcc877b9e31b9e51c94d59de76ffc645616032d
I333a72d61b212df04fc27f9038114a048233c255

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
@@ -925,10 +925,32 @@ mfxU16 GetMaxNumRef(MfxVideoParam &par, bool bForward)
 
 #endif // (MFX_VERSION >= 1025)
 
+#if defined (MFX_ENABLE_HEVCE_ROI) || defined (MFX_ENABLE_HEVCE_DIRTY_RECT)
+mfxStatus CheckAndFixRect(RectData *rect, MfxVideoParam const & par, ENCODE_CAPS_HEVC const & caps) {
+    mfxU32 changed = 0;
+
+    mfxU32 blockSize = 1 << (caps.BlockSize + 3);
+    changed += CheckRange(rect->Left,   0u, mfxU32(par.mfx.FrameInfo.Width));
+    changed += CheckRange(rect->Right,  0u, mfxU32(par.mfx.FrameInfo.Width));
+    changed += CheckRange(rect->Top,    0u, mfxU32(par.mfx.FrameInfo.Height));
+    changed += CheckRange(rect->Bottom, 0u, mfxU32(par.mfx.FrameInfo.Height));
+    // align to BlockSize size
+    changed += AlignDown(rect->Left, blockSize);
+    changed += AlignDown(rect->Top,  blockSize);
+    changed += AlignUp(rect->Right,  blockSize);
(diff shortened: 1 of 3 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -925,10 +925,32 @@ mfxU16 GetMaxNumRef(MfxVideoParam &par, bool bForward)
...
+    changed += CheckRange(rect->Right, mfxU32(0), mfxU32(par.mfx.FrameInfo.Width));
+    changed += CheckRange(rect->Top, mfxU32(0), mfxU32(par.mfx.FrameInfo.Height));
+    changed += CheckRange(rect->Bottom, mfxU32(0), mfxU32(par.mfx.FrameInfo.Height));
+    // align to BlockSize size
+    changed += AlignDown(rect->Left, blockSize);
+    changed += AlignDown(rect->Top, blockSize);
+    changed += AlignUp(rect->Right, blockSize);
+    changed += AlignUp(rect->Bottom, blockSize);
+
+    if (changed)
```
mgonchar: Use ```MFX_CHECK``` in such cases

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -925,10 +925,32 @@ mfxU16 GetMaxNumRef(MfxVideoParam &par, bool bForward)
 
 #endif // (MFX_VERSION >= 1025)
 
+#if defined (MFX_ENABLE_HEVCE_ROI) || defined (MFX_ENABLE_HEVCE_DIRTY_RECT)
+mfxStatus CheckAndFixRect(RectData *rect, MfxVideoParam const & par, ENCODE_CAPS_HEVC const & caps) {
+    mfxU32 changed = 0;
+
+    mfxU32 blockSize = 1 << (caps.BlockSize + 3);
+    changed += CheckRange(rect->Left, mfxU32(0), mfxU32(par.mfx.FrameInfo.Width));
```
mgonchar: ```0u``` is unsigned int literal.
and please add alignment
```
     changed += CheckRange(rect->Left,  0u, mfxU32(par.mfx.FrameInfo.Width));
     changed += CheckRange(rect->Right, 0u, mfxU32(par.mfx.FrameInfo.Width)); 
```

AntonGrishin: Done.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
AntonGrishin: In my opinion, mfx_check good idea if we check false condition, but in this case (!changed, MFX_WRN...) may confuse another developers.

mgonchar: main idea of macro like ```MFX_CHECK``` is not to simplify code writing but enabling traces (you can switch on additional define and see where the error/wrn status arises). So I generally agree that using of macro with such condition may look strange, but ability to trace the error to origin without going to deep debug is worth it in my opinion

PR #1479, opened 2019-06-13
Add Android Q compile support

LukeZheZhu:


mgonchar:
I think we need a feedback from @dvrogozh here

PR #1483, opened 2019-06-18
samples: added bitstream resource management class

mgonchar:


mgonchar:
@dmitry-gurulev it is not ```std::vector``` only because vector is heavier class, ```std::unique_ptr``` with default deleter should be comparable to regular pointer in size

mgonchar:
Replaced ```unique_ptr``` with ```vector``` to workaround Win iterator check settings (it complained previously on ```std::copy``` function)

samples/sample_common/include/sample_utils.h
```diff
@@ -495,6 +495,53 @@ mfxF64 CalculateFrameRate(mfxU32 nFrameRateExtN, mfxU32 nFrameRateExtD);
...
+
+    mfxBitstreamWrapper & operator= (mfxBitstreamWrapper bs_wrapper)
+    {
+        std::swap(*this, bs_wrapper);
+
+        return *this;
+    }
+
+    mfxBitstreamWrapper(mfxBitstreamWrapper && bs_wrapper)           = default;
+    ~mfxBitstreamWrapper()                                           = default;
```
dmitry-gurulev: What is a reason to define dtor?

samples/sample_common/include/sample_utils.h
```diff
@@ -495,6 +495,53 @@ mfxF64 CalculateFrameRate(mfxU32 nFrameRateExtN, mfxU32 nFrameRateExtD);
...
+        m_data = bs_wrapper.m_data;
+
+        static_cast<mfxBitstream&>(*this) = static_cast<const mfxBitstream&>(bs_wrapper);
+
+        Data = m_data.data();
+    }
+
+    mfxBitstreamWrapper & operator= (mfxBitstreamWrapper bs_wrapper)
+    {
+        std::swap(*this, bs_wrapper);
```
dmitry-gurulev: `swap` will use `operator=` isn't it?
Looks like we need either to define this operator as `operator= (mfxBitstreamWrapper const& bs_wrapper)` and have `operator= (mfxBitstreamWrapper&& bs_wrapper) = default` or swap class' members one by one (like `swap(this->m_data,  bs_wrapper.m_data)` or define our own `swap` for _mfxBitstreamWrapper_ class

samples/sample_common/include/sample_utils.h
```diff
@@ -495,6 +495,53 @@ mfxF64 CalculateFrameRate(mfxU32 nFrameRateExtN, mfxU32 nFrameRateExtD);
...
+
+        return *this;
+    }
+
+    mfxBitstreamWrapper(mfxBitstreamWrapper && bs_wrapper)           = default;
+    ~mfxBitstreamWrapper()                                           = default;
+
+    void Extend(mfxU32 n_bytes)
```
dmitry-gurulev: Method's name & its implementation implies that we have no way to reduce bitstream size. Do we really need this constrain?

samples/sample_common/include/sample_utils.h
mgonchar: just to be consistent

samples/sample_common/include/sample_utils.h
```diff
@@ -495,6 +495,53 @@ mfxF64 CalculateFrameRate(mfxU32 nFrameRateExtN, mfxU32 nFrameRateExtD);
...
+        : mfxBitstream()
+    {
+        Extend(n_bytes);
+    }
+
+    mfxBitstreamWrapper(const mfxBitstreamWrapper & bs_wrapper)
+    {
+        m_data = bs_wrapper.m_data;
+
+        static_cast<mfxBitstream&>(*this) = static_cast<const mfxBitstream&>(bs_wrapper);
```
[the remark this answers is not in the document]
mgonchar: yes, looks better

samples/sample_common/include/sample_utils.h
mgonchar: I think it should use move assignment operator... will check it out

dmitry-gurulev: @mgonchar,

Yes, it should. But u define your own copy assignment operator - AFAIR this prevents implicitly declared move assignment being generated.

samples/sample_common/include/sample_utils.h
mgonchar: I think yes. I just constrained to existing usage. Do you know any real use case, when we need to shrink used bitstream? It might be enhanced to shrink and forget all data before ```DataOffset``` but other ways we will possibly lose bitstream data

mgonchar:
@dmitry-gurulev could you review updated version?

samples/sample_common/include/sample_utils.h
mgonchar: yes, you are right. Right solution here would be to implement copy-swap idiom via several swaps like:
```
mfxBitstreamWrapper & operator= (mfxBitstreamWrapper bs_wrapper)
{
    std::swap(static_cast<mfxBitstream&>(*this), static_cast<const mfxBitstream&>(bs_wrapper));
    std::swap(m_data, bs_wrapper.m_data);
}
```

But as for me better to default move assignment and implement fair copy assignment here

samples/sample_common/include/sample_utils.h
```diff
@@ -495,6 +495,54 @@ mfxF64 CalculateFrameRate(mfxU32 nFrameRateExtN, mfxU32 nFrameRateExtD);
...
+    {}
+
+    mfxBitstreamWrapper(mfxU32 n_bytes)
+        : mfxBitstream()
+    {
+        Extend(n_bytes);
+    }
+
+    mfxBitstreamWrapper(const mfxBitstreamWrapper & bs_wrapper)
+    : mfxBitstream(bs_wrapper)
```
dmitry-gurulev: Please fix alignment (indentation)

samples/sample_common/include/sample_utils.h
```diff
@@ -495,6 +495,54 @@ mfxF64 CalculateFrameRate(mfxU32 nFrameRateExtN, mfxU32 nFrameRateExtD);
...
+    {
+        Data = m_data.data();
+    }
+
+    mfxBitstreamWrapper & operator= (const mfxBitstreamWrapper& bs_wrapper)
+    {
+        static_cast<mfxBitstream&>(*this) = static_cast<const mfxBitstream&>(bs_wrapper);
+        m_data = bs_wrapper.m_data;
+        Data   = m_data.data();
+
```
dmitry-gurulev: W/ that implementation we need to have protection against self-assignment.

samples/sample_common/include/sample_utils.h
dmitry-gurulev: > But as for me better to default move assignment and implement fair copy assignment here

copy-swap gives two benefits - exception safety (since swap is noexcept) and protection against self-assignment. So, I'd rather use copy-swap.

samples/sample_common/include/sample_utils.h
mgonchar: Why? It wouldn't harm the data. I don't like protections like:
```
if (this != &bs_wrapper)
{
...
}
```
because it is optimization for rare case which would actually slow down each assignment

samples/sample_common/include/sample_utils.h
mgonchar: agree

samples/sample_common/include/sample_utils.h
mgonchar: Well, exception safety of copy-swap is exactly same as for fair copy assignment (both of them can throw ```std::bad_alloc``` at allocation of new chunk of memory). But for copy-swap we lose defaulted move assignment, and we have temporary object for move assignment, which we don't have in case of default move assignment. Now what related to copy assignment, as I said we remain the same level of exception safety in this particular case and also we won't have temporary object. My point to introduce copy-swap was to reduce code duplication and for ```std::unique_ptr``` design, which I used at first time, it was critical, because it involved low-level memory copying, which I didn't want to duplicate) But with ```std::vector``` and such complicate (same amount of code lines as fair assignment) copy-swap assignment operator I don't see any benefits in it for this particular case. Also I like this idiom, but as for me it doesn't fit good this particular case

dmitry-gurulev: > exception safety of copy-swap is exactly same as for fair copy assignment

In my understanding, no, it's not the same. For copy-swap your object either successfully copied or u get an exception and object remain at the same state and still could be used. For this "fair copy assignment" your object might be at the intermediate sate - `mfxBitstream` subobject was updated and points to the memory that actually belongs to the another object. The first invariant is - `Data` points to the memory that is owned by object's vector (`m_data` member). When we have `bad_alloc` during `m_data = bs_wrapper.m_data` this variant is violated and object can't be used anymore.

I'd rather do something like
```
mfxBitstreamWrapper& operator=(mfxBitstreamWrapper  const& other)
{
    mfxBitstreamWrapper tmp(other);
    using std::swap;
    swap(tmp.m_data, this->m_data);
    swap(static_cast<mfxBitstream&>(tmp), static_cast<mfxBitstream&>(*this));

    return *this;
}

mfxBitstreamWrapper & operator= (mfxBitstreamWrapper&&) = default; 
```

mgonchar: Agree, we can't get more than basic exception safety guarantee here without intermediate copy. I also think we can save some tmp copies (which arises inside ```std::swap``` like this):
```
mfxBitstreamWrapper& operator=(mfxBitstreamWrapper  const& other)
{
    mfxBitstreamWrapper tmp(other);

    m_data = std::move(tmp.m_data);
    static_cast<mfxBitstream&>(*this) = static_cast<mfxBitstream&>(tmp);

    return *this;
}

mfxBitstreamWrapper & operator= (mfxBitstreamWrapper&&) = default; 
```

dmitry-gurulev: For me, `swap` is more idiomatic. Anyway, current variant is quite OK, so let's move forward.

PR #1490, opened 2019-06-24
h265e: prevent P-slices for HEVC VDEnc

vilichev:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1008,6 +1007,7 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     m_caps.ddi_caps.IntraRefreshBlockUnitSize = 2;
     m_caps.ddi_caps.TileSupport = (attrs[idx_map[VAConfigAttribEncTileSupport]].value == 1);
 
+    m_caps.PSliceSupport = IsOn(par.mfx.LowPower) ? 0 : 1;
```
dmitryermilov: This condition looks controversial. Why are there nothing about platforms? I mean should regular P frames work on all platforms ?

vilichev: @dmitryermilov It seems there are no restrictions for currently supported platforms

PR #1494, opened 2019-06-25
[jpegd] Enable SFC for CSC to ARGB during jpeg 420 decode.

Alexandr-Konovalov:
Off dy default (2-step conversion via internal VPP is used), enabled via
MFX_EXTBUFF_DEC_VIDEO_PROCESSING in prameter's extended buffer.

Issue: MDP-57617

### DIFF
--- a/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
@@ -216,10 +216,12 @@ mfxStatus VideoDECODEMJPEG::Init(mfxVideoParam *par)
 
         VideoDECODEMJPEGBase_HW * dec = new VideoDECODEMJPEGBase_HW;
         decoder.reset(dec);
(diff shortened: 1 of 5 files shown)


alariono:
Need to handle new Ext buffer Query() method

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -680,6 +700,15 @@ mfxStatus VideoDECODEMJPEG::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
         (*surface_out)->Info.CropX = 0;
         (*surface_out)->Info.CropY = 0;
 
+        mfxExtDecVideoProcessing * videoProcessing = (mfxExtDecVideoProcessing *)GetExtendedBuffer(m_vFirstPar.ExtParam, m_vFirstPar.NumExtParam, MFX_EXTBUFF_DEC_VIDEO_PROCESSING);
```
alariono: Handle this Ext buffer in Query() method

Alexandr-Konovalov: Done.

PR #1495, opened 2019-06-25
[AVCe] Added interface to disable 8x8TransformMode

Losev-Ivan:
Change-Id: I984d8b83d7890dd1721c36e91dc19d141475def1

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -3323,6 +3323,13 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         }
     }
 
+    if ((par.mfx.CodecProfile != MFX_PROFILE_UNKNOWN) && (par.mfx.CodecProfile != MFX_PROFILE_AVC_HIGH)
+        && IsOn(extDdi->Transform8x8Mode))
+    {
+        unsupported = true;
+        extDdi->Transform8x8Mode = MFX_CODINGOPTION_UNKNOWN;
+    }
+
     if (par.calcParam.cqpHrdMode)
     {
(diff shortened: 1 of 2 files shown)


_studio/shared/include/mfx_ext_buffers.h
```diff
@@ -79,6 +79,7 @@ typedef struct {
 
     mfxU16 DisablePSubMBPartition;  // tri-state, default depends on Profile and Level
     mfxU16 DisableBSubMBPartition;  // tri-state, default depends on Profile and Level
+    mfxU16 Transform8x8Mode;        // tri-state: 0, MFX_CODINGOPTION_OFF, MFX_CODINGOPTION_ON
```
stasefremov: Please don't just insert in the middle. It will break binary compatibility with older versions of apps/tools which use this buffer. Please replace reserved field instead.

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -3323,6 +3323,12 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         }
     }
 
+    if ((par.mfx.CodecProfile != MFX_PROFILE_AVC_HIGH) && IsOn(extDdi->Transform8x8Mode))
```
stasefremov: It's incorrect. If Profile is not set by app explicitly (i.e. CodecProfile == 0), enabling transform 8x8 will always cause error. Please check that CodecProfile is set (not 0) before checking that it's equal to MFX_PROFILE_AVC_HIGH.

Losev-Ivan: Done

Losev-Ivan:
mfxExtCodingOptionDDI buffer is not a part of API

PR #1496, opened 2019-06-25
Fix comparison operators for plugin guids

mgonchar:
Avoid usage of low-level memory comparison in favor of operator ==

### DIFF
--- a/_studio/shared/include/mfx_utils.h
+++ b/_studio/shared/include/mfx_utils.h
@@ -24,6 +24,7 @@
 #include "mfx_config.h"
 
 #include "mfxstructures.h"
+#include "mfxplugin.h"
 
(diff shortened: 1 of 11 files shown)


_studio/shared/include/mfx_utils.h
```diff
@@ -174,4 +175,9 @@ inline mfxStatus CheckAndDestroyVAbuffer(VADisplay display, VABufferID & buffer_
 }
 #endif
 
+bool operator==(mfxPluginUID const& l, mfxPluginUID const& r)
```
dvrogozh: this fails to build. You need `inline` I guess (or static inline).

mgonchar: yes, looks like so

PR #1499, opened 2019-06-27
utils: modify comparison operator

mgonchar:
Use explicit per-field comparison instead of whole struct memcmp

### DIFF
--- a/_studio/shared/include/mfx_utils.h
+++ b/_studio/shared/include/mfx_utils.h
@@ -178,14 +178,17 @@ inline mfxStatus CheckAndDestroyVAbuffer(VADisplay display, VABufferID & buffer_
 #define MFX_EQ_FIELD(Field) l.Field == r.Field
 #define MFX_EQ_ARRAY(Array, Num) std::equal(l.Array, l.Array + Num, r.Array)
 
+#define MFX_DECL_OPERATOR_NOT_EQ(Name)                      \
+static inline bool operator!=(Name const& l, Name const& r) \
+{                                                           \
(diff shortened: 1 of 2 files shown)


_studio/shared/include/mfx_utils.h
```diff
@@ -175,6 +175,9 @@ inline mfxStatus CheckAndDestroyVAbuffer(VADisplay display, VABufferID & buffer_
 }
 #endif
 
+#define MFX_CHECK_EQUALITY_FIELD(Field) l.Field == r.Field
```
dvrogozh: I think that's too difficult macro name. Maybe MFX_CHECK_FIELDS_EQ or just MFX_FIELDS_EQ (mind plural).

mgonchar: @dvrogozh Well, I'm ok to shorten the name. But I think clarification "FIELDS" should go last, because we have similar macro for "ARRAY". By the way, why plural? This one macro for one field only

mgonchar:
@dmitry-gurulev not all cmp operators can be written through ```tuple``` trick. For example we have in our code base SPS comparison operator which checks existence of HRD params and then decide whether to compare them or not. So we need flexibility to control per-field comparison. I plan to heavily use those macro in such cases, to avoid copy-paste mistakes

_studio/shared/include/mfx_utils.h
mgonchar: @dvrogozh how about ```MFX_EQ_FIELD``` and ```MFX_EQ_ARRAY``` ?

PR #1502, opened 2019-06-28
[h264e] Fix wrong initialCpbRemovalDelay rounding

pgribov:
Wrong rounding may cause HDR overflow in some streams


pgribov:
Need to add fix

PR #1507, opened 2019-07-02
[smt] Enable per session FPS report

AntonGrishin:
Issue: MDP-57883

### DIFF
--- a/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
+++ b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
@@ -583,21 +583,24 @@ mfxStatus Launcher::ProcessResult()
 
     for (mfxU32 i = 0; i < m_pThreadContextArray.size(); i++)
     {
-        mfxStatus _sts = m_pThreadContextArray[i]->transcodingSts;
+        mfxStatus transcodingSts = m_pThreadContextArray[i]->transcodingSts;
+        mfxF64 workTime = m_pThreadContextArray[i]->working_time;
+        mfxU32 framesNum = m_pThreadContextArray[i]->numTransFrames;
 
         if (!FinalSts)
-            FinalSts = _sts;
+            FinalSts = transcodingSts;
 
-        msdk_string SessionStsStr = _sts ? msdk_string(MSDK_STRING("FAILED"))
+        msdk_string SessionStsStr = transcodingSts ? msdk_string(MSDK_STRING("FAILED"))
             : msdk_string((MSDK_STRING("PASSED")));
 
         msdk_stringstream ss;
         ss << MSDK_STRING("*** session ") << i
            << MSDK_STRING(" [") << m_pThreadContextArray[i]->pPipeline->GetSessionText()
            << MSDK_STRING("] ") << SessionStsStr <<MSDK_STRING(" (")
-           << StatusToString(_sts) << MSDK_STRING(") ")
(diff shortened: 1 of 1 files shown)


samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -584,6 +584,8 @@ mfxStatus Launcher::ProcessResult()
     for (mfxU32 i = 0; i < m_pThreadContextArray.size(); i++)
     {
         mfxStatus _sts = m_pThreadContextArray[i]->transcodingSts;
+        mfxF64 _workTime = m_pThreadContextArray[i]->working_time;
```
mgonchar: Usage of variables starting with underscore is a bad practice. It is not recommended since there are some rules which underscore variables are reserved and used by system. See here https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

I'd suggest removing underscore from here

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -596,8 +598,9 @@ mfxStatus Launcher::ProcessResult()
            << MSDK_STRING(" [") << m_pThreadContextArray[i]->pPipeline->GetSessionText()
            << MSDK_STRING("] ") << SessionStsStr <<MSDK_STRING(" (")
            << StatusToString(_sts) << MSDK_STRING(") ")
-           << m_pThreadContextArray[i]->working_time << MSDK_STRING(" sec, ")
-           << m_pThreadContextArray[i]->numTransFrames << MSDK_STRING(" frames")
+           << _workTime << MSDK_STRING(" sec, ")
+           << _framesNum << MSDK_STRING(" frames, ") 
+           << std::fixed << std::setprecision(3) << _framesNum / _workTime << MSDK_STRING(" fps")
```
mgonchar: It won't work, ```_framesNum``` is integer, so result of ```_framesNum / _workTime``` is integer too. Fix it by casting: ```mfxF64(_framesNum) / _workTime```

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
AntonGrishin: Done. Also change "_sts" to "sts_"

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
AntonGrishin: Why? _workTime already float value. If either the divisor or the dividend is a floating-point number, a floating-point divide is executed.
Also I checked it before (got float) :)

mgonchar: Oh, my bad, I missed that ```workTime``` is ```float```

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -583,21 +583,24 @@ mfxStatus Launcher::ProcessResult()
 
     for (mfxU32 i = 0; i < m_pThreadContextArray.size(); i++)
     {
-        mfxStatus _sts = m_pThreadContextArray[i]->transcodingSts;
+        mfxStatus sts_ = m_pThreadContextArray[i]->transcodingSts;
```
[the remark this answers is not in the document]
AntonGrishin: Changed

PR #1509, opened 2019-07-02
[hevce] Use MFX_TASK_THREADING_INTRA

dmitryermilov:
"DEDICATED" reduces density for multisession workloads in join mode.

MDP-57311


dmitryermilov:
You're right. Thanks !

PR #1516, opened 2019-07-04
[sample_encode] Replaced mfxBitstream with mfxBitstreamWrapper

dmitryermilov:

### DIFF
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -1644,51 +1644,7 @@ mfxU32 GCD(mfxU32 a, mfxU32 b)
     return b1;
 }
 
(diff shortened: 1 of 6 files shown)


samples/sample_common/src/sample_utils.cpp
```diff
@@ -1644,51 +1644,7 @@ mfxU32 GCD(mfxU32 a, mfxU32 b)
     return b1;
 }
 
-mfxStatus DARtoPAR(mfxU32 darw, mfxU32 darh, mfxU32 w, mfxU32 h, mfxU16 *pparw, mfxU16 *pparh)
```
mgonchar: Shouldn't it be different PR?

dmitryermilov: Nooo. Consider it a cleanup:)

PR #1525, opened 2019-07-08
vp9e: added operator == for mfxVP9SegmentParam

mgonchar:
Removed memcmp for comparison for mfxVP9SegmentParam struct

### DIFF
--- a/_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
+++ b/_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
@@ -31,6 +31,14 @@
 #include "mfx_enc_common.h"
 #include "assert.h"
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_utils.h
```diff
@@ -906,29 +914,12 @@ inline bool CompareSegmentMaps(mfxExtVP9Segmentation const & first, mfxExtVP9Seg
...
-    bool equal = false;
-
-    if (first.NumSegments == second.NumSegments)
-    {
-        mfxU8 equalParams = 0;
-        for (mfxU8 i = 0; i < first.NumSegments; i++)
-        {
-            if (0 == memcmp(&first.Segment[i], &second.Segment[i], sizeof(mfxVP9SegmentParam)))
-            {
-                equalParams++;
```
DmitryMalishev: Why would you stop comparing and return FALSE on the very first non-equal param? No need for equalParams var in this case.

PR #1527, opened 2019-07-08
vpp: added operator != to SignalInfo struct

mgonchar:


_studio/shared/include/mfx_vpp_interface.h
```diff
@@ -239,6 +239,13 @@ namespace MfxHwVideoProcessing
             bool   enabled;
             mfxU16 TransferMatrix;
             mfxU16 NominalRange;
+
+            bool operator!=(const SignalInfo & other)
```
dmitry-gurulev: `bool operator!=(const SignalInfo & other) const`

mgonchar: done

PR #1533, opened 2019-07-12
asg-hevc: actualized copyright year in files

semelianova:


mgonchar:
Slightly changed name of PR to underline what exactly was changed in source files

PR #1537, opened 2019-07-17
vpp: Set unsupported sts when offset between Y/UV not aligned to pitch

### LINKED ISSUE #1497: [qsv_vpp][mirror]garbage in chroma occurs if there are paddings between Y and UV
Apply patch in ffmpeg:
1. allocate continuous memory for qsv vpp:
https://patchwork.ffmpeg.org/patch/13405/
2. add support for transpose:
https://patchwork.ffmpeg.org/patch/13595/

Run cmdline:
ffmpeg -init_hw_device qsv=foo -filter_hw_device foo -v debug -f rawvideo -pix_fmt yuv420p -s:v 352x288 -i MOBILE_352x288_30_orig_01.yuv -vf 'hwupload=extra_hw_frames=64,format=qsv,vpp_qsv=transpose=vflip' -vframes 10 -f rawvideo -pix_fmt yuv420p -y ./transpose.yuv

if we modify in the patch and don't remove the plane_padding, there will be garbage in chroma field.
With plane_padding removed in this hard-coded way, the output image is good.

 /* remove plane_padding introduced by av_frame_get_buffer */
+        for (i = 1; i < 4; i++) {
+            if (out_frame->frame->data[i])
+                out_frame->frame->data[i] -= i * 128;
+        }

Analysis:

For vpp/mirror, CmCopy is used.

For the garbage caused by paddings, the UVOffset in MSDK is calculated by UV-Y, and then it will be calculated  based on pitch
height_stride = (Data.UV – Data.Y) /Data.Pitch,
https://github.com/Intel-Media-SDK/MediaSDK/blob/8fc7b41eaa0a8a20d55ac06438def8dbe1a1df86/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp#L3468
MSDK passed row based (pitch based) UVOffset to CM.
https://github.com/Intel-Media-SDK/MediaSDK/blob/8fc7b41eaa0a8a20d55ac06438def8dbe1a1df86/_studio/shared/src/cm_mem_copy.cpp#L1528

Summary:
So if padding is introduced, it won’t break the pipeline, but will assign the **incorrect address** to the pointer of Data.UV.
And that’s the reason for garbage in chroma filed in mirror.

AntonGrishin:
fixes #1497

### DIFF
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -3317,7 +3317,9 @@ mfxStatus VideoVPPHW::PreWorkInputSurface(std::vector<ExtSurface> & surfQueue)
                     dstTempSurface.Data.MemId = &dstHandle;
 
                     mfxI64 verticalPitch = (mfxI64)(srcTempSurface.Data.UV - srcTempSurface.Data.Y);
-                    verticalPitch = (verticalPitch % srcTempSurface.Data.Pitch)? 0 : verticalPitch / srcTempSurface.Data.Pitch;
+                    // offset beetween Y and UV must be align to pitch
+                    MFX_CHECK(!(verticalPitch % srcTempSurface.Data.Pitch), MFX_ERR_UNSUPPORTED);
+                    verticalPitch /= srcTempSurface.Data.Pitch;
                     mfxU32 srcPitch = srcTempSurface.Data.PitchLow + ((mfxU32)srcTempSurface.Data.PitchHigh << 16);
 
                     sts = m_pCmCopy->CopyMirrorSystemToVideoMemory(dstHandle.first, 0, srcTempSurface.Data.Y, srcPitch, (mfxU32)verticalPitch, roi, MFX_FOURCC_NV12);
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -3317,7 +3317,9 @@ mfxStatus VideoVPPHW::PreWorkInputSurface(std::vector<ExtSurface> & surfQueue)
                     dstTempSurface.Data.MemId = &dstHandle;
 
                     mfxI64 verticalPitch = (mfxI64)(srcTempSurface.Data.UV - srcTempSurface.Data.Y);
-                    verticalPitch = (verticalPitch % srcTempSurface.Data.Pitch)? 0 : verticalPitch / srcTempSurface.Data.Pitch;
+                    // offset beetween Y and UV must be align to pitch
```
mgonchar: Do we have such requirement in our docs?

dmitryermilov: I think it's a subject for release notes..

dvrogozh: Is that an actual requirement or CM feature request?

AntonGrishin: I think this is feature request, but as mentioned in issue discussion, re-writing of cm kernel will require significant efforts.

PR #1540, opened 2019-07-17
[VPP] Corrected FourCC support constraints

semelianova:
Corrected FourCC support checks, added FourCC support dependence on platform
and VPP filter

### DIFF
--- a/_studio/mfx_lib/vpp/include/mfx_vpp_defs.h
+++ b/_studio/mfx_lib/vpp/include/mfx_vpp_defs.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -96,7 +96,8 @@
 #define VPP_ALIGN32(X) (((mfxU32)((X)+31)) & (~ (mfxU32)31))
 
 // names of VPP smart filters. internal using only
-enum {
+enum
+{
     MFX_EXTBUFF_VPP_CSC               =   MFX_MAKEFOURCC('C','S','C','F'),//COLOR SPACE CONVERSION FILTER
     MFX_EXTBUFF_VPP_RESIZE            =   MFX_MAKEFOURCC('R','S','Z','F'),
     MFX_EXTBUFF_VPP_DI                =   MFX_MAKEFOURCC('S','D','I','F'),//STANDARD DEINTERLACE FILTER (60i->30p)
@@ -114,25 +115,26 @@ enum {
     MFX_EXTBUFF_VPP_LSHIFT_OUT      =   MFX_MAKEFOURCC('L','S','F','O'),
 };
 
-typedef enum {
+enum mfxRequestType
+{
     MFX_REQUEST_FROM_VPP_CHECK    = 0x0001,
     MFX_REQUEST_FROM_VPP_PROCESS = 0x0002
+};
 
-} mfxRequestType;
-
-enum {
+enum
+{
     VPP_IN  = 0x00,
     VPP_OUT = 0x01
 };
 
-typedef enum {
+enum PicStructMode
+{
     PASS_THROUGH_PICSTRUCT_MODE   = 0x0001,
     DYNAMIC_DI_PICSTRUCT_MODE     = 0x0004,
     ERR_PICSTRUCT_MODE            = 0x0000
+};
 
-} PicStructMode;
-
-typedef enum
+enum mfxGamutMode
 {
     // gamut compression
     GAMUT_COMPRESS_BASE_MODE     = 1,
@@ -146,16 +148,16 @@ typedef enum
 
     // invalid mode
     GAMUT_INVALID_MODE  = 100
+};
 
-} mfxGamutMode;
 
-
-typedef enum
+enum MirroringPositions
 {
     MIRROR_INPUT   = 0x0001,
     MIRROR_OUTPUT  = 0x0002,
     MIRROR_WO_EXEC = 0x0004
-} MirroringPositions;
+};
+
 
 #endif // __MFX_VPP_DEFS_H
 #endif // MFX_ENABLE_VPP
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <assert.h>
 #include <limits>
+#include <stdexcept>
 #include "math.h"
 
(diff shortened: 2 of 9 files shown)


_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
```diff
@@ -1016,22 +989,24 @@ mfxStatus VideoVPPBase::Query(VideoCORE * core, mfxVideoParam *in, mfxVideoParam
...
 #endif
             out->vpp.Out.FourCC != MFX_FOURCC_P010 &&
             out->vpp.Out.FourCC != MFX_FOURCC_P210 &&
 #if (MFX_VERSION >= 1027)
             out->vpp.Out.FourCC != MFX_FOURCC_Y210 &&
             out->vpp.Out.FourCC != MFX_FOURCC_Y410 &&
 #endif
-            out->vpp.Out.FourCC != MFX_FOURCC_AYUV &&
-            out->vpp.Out.FourCC != MFX_FOURCC_A2RGB10 )
+            out->vpp.In.FourCC != MFX_FOURCC_AYUV &&
```
mgonchar: this is intentional correction? comment on line 991 says this block is dedicated to OUT params

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -434,27 +434,39 @@ mfxStatus VAAPIVideoProcessing::QueryCapabilities(mfxVppCaps& caps)
 
     caps.uFieldWeavingControl = 1;
 
+    VAAPIVideoCORE* hwCore = dynamic_cast<VAAPIVideoCORE*>(m_core);
+    eMFXHWType hwType = hwCore->GetHWType();
```
mgonchar: check for ```nullptr``` before dereferencing, ```dynamic_cast``` might fail

_studio/mfx_lib/vpp/include/mfx_vpp_defs.h
```diff
@@ -157,6 +158,575 @@ typedef enum
     MIRROR_WO_EXEC = 0x0004
 } MirroringPositions;
 
+typedef enum
+{
```
mgonchar: also, I believe you need to add a comment before declaration of this enums, that order here matters. So no one would change the order of enum entries

_studio/mfx_lib/vpp/include/mfx_vpp_defs.h
```diff
@@ -157,6 +158,575 @@ typedef enum
...
+    {MFX_FOURCC_P016, P016},
+    {MFX_FOURCC_Y216, Y216},
+    {MFX_FOURCC_Y416, Y416}
+#endif
+};
+
+const mfxStatus VPPSupportTable[NumGens][NumFilters][NumFormats][NumFormats] =
+{
+    {//gen6
+        {//Color Format Conversion
```
mgonchar: looks cool! :)

_studio/mfx_lib/vpp/src/mfx_vpp_sw_core.cpp
semelianova: It is my mistake due to inattention; corrected

semelianova:
Moved the support table and some necessary definitions in separate file for use it in tests without another needless information

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -6549,5 +6736,22 @@ mfxStatus GetWorstSts(mfxStatus sts1, mfxStatus sts2)
     return (it1 < it2) ? *it1 : *it2;
 } // mfxStatus GetWorstSts(mfxStatus sts1, mfxStatus sts2)
 
+PlatformGen GetPlatformGen(eMFXHWType type)
+{
+    if (type < MFX_HW_SCL)
+        return GEN6;
+    else if (type < MFX_HW_ICL)
```
mgonchar: no need in ```else``` statement

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5267,6 +5286,19 @@ mfxStatus ConfigureExecuteParams(
             case MFX_EXTBUFF_VPP_DEINTERLACING:
             case MFX_EXTBUFF_VPP_DI:
             {
+                try
+                {
+                    mfxStatus support_status = CheckFilterSupport(GetPlatformGen(hwType),
+                        DEINTERLACING,
+                        MFXFourCC2VPPSupportTableFourCC.at(videoParam.vpp.In.FourCC),
```
mgonchar: ok, looks like you have this code at each switch, maybe we can consolidate this try-catch before ```switch``` ?

semelianova: Yes, we can convert MFXFourCC to format which is used in CheckFilterSupport function and write it to some variable before _for loop_ and consolidate all loop in try-catch

mgonchar: Why do you need try-catch for all loop? You only need to place ```.at()``` calls inside try-catch block

semelianova: `VPPSupportTableFourCC inFourCC = MFXFourCC2VPPSupportTableFourCC.at(videoParam.vpp.In.FourCC);`
I want do this before loop. If I close `try` I can't use `inFourCC` in loop. If we catch exception we don't need to run loop

mgonchar: Workaround is simple :)

```
VPPSupportTableFourCC inFourCC, outFourCC;
try
{
    inFourCC  = MFXFourCC2VPPSupportTableFourCC.at(videoParam.vpp.In.FourCC);
    outFourCC = MFXFourCC2VPPSupportTableFourCC.at(videoParam.vpp.Out.FourCC);
}
catch (const std::out_of_range & ex)
{
    MFX_RETURN(MFX_ERR_INVALID_VIDEO_PARAM); // didn't find input or output FourCC in VPPSupportTableFourCC
}

for (mfxU32 j = 0; j < pipelineList.size(); j += 1)
{
...
}
```

semelianova: Yes, it is also possible, looks prettier, thanks

semelianova: Can't do so because I added `VPPExtBuf2VPPSupportTableFilter` map to convert filters and it is used in many cases in loop `VPPExtBuf2VPPSupportTableFilter.at(curFilterId)`

dmitryermilov:
@semelianova , is it actual ?

semelianova:
Need to change driver ddi and get supported info from driver

PR #1544, opened 2019-07-24
[ASG-HEVC] Fix to build if not the latest mfx_version is used

semelianova:
Move inclusion test_processor.h above MFX_VERSION check to define global
identifiers in any case regardless of version. In other case project may not
compile.

### DIFF
--- a/tools/asg-hevc/include/test_processor.h
+++ b/tools/asg-hevc/include/test_processor.h
@@ -22,13 +22,13 @@
 #define __ASG_HEVC_TEST_PROCESSOR_H__
 
 #include "mfxvideo.h"
+#include "sample_defs.h"
 
 #if MFX_VERSION >= MFX_VERSION_NEXT
 
 #include <fstream>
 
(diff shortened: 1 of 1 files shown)


semelianova:
"msdk_char" is undefined in main when MFX_VERSION < MFX_VERSION_NEXT

mgonchar:
Ok, if so I'd prefer to get rid of ```msdk_char``` in this project and replace it with ```wchar_t```

semelianova:
We need to replace a lot of msdk aliases to replace msdk_char with wchar_t (msdk_printf, msdk_strcmp and others). It will affect the samples too. It needs a separate commit. Also it won't fix asg-hevc buildю Moving test_processor.h inclusion is inevitably.

mgonchar:
Ok, let's merge this one as is. But could you create an issue to clean up ```char``` related code?

semelianova:
I can do it, but it won't happen so fast.

PR #1551, opened 2019-07-26
Create solutions to build tools on Windows

semelianova:


dmitry-gurulev:
@semelianova 
why don't we use cmake to generate all required projects/solutions?

dmitryermilov:
@semelianova  ,please reply to questions reviewers posted

semelianova:
> why don't we use cmake to generate all required projects/solutions?

Now all windows components are built by VS solutions. Cmake is used for Linux only.

tools/AllTools.sln
```diff
@@ -0,0 +1,61 @@
+﻿
```
dvrogozh: Why do we need tools dedicated solution? We just deleted the one dedicated solution for tutorials. I vote for the single solution entry point - AllBuild.sln

semelianova: Deleted tools dedicated solution

tools/asg-hevc/asg-hevc.sln
```diff
@@ -0,0 +1,41 @@
+﻿
```
dvrogozh: @dmitryermilov : do we need this sln? I would drop it as well.

tools/bs_parser_hevc/bs_parser_hevc.sln
```diff
@@ -0,0 +1,31 @@
+﻿
```
dvrogozh: same

tools/bs_parser_hevc/tools/hevc_fei_extractor/hevc_fei_extractor.sln
```diff
@@ -0,0 +1,41 @@
+﻿
```
dvrogozh: same

tools/asg-hevc/asg-hevc.sln
dmitryermilov: Well, it's controversial. We have separate solution for samples. Following this logic we need to drop them in samples as well.

dvrogozh: These are not samples. If we will merge them in we will need to support them. Will we? For example, we don't even CI test these solutions: we use on AllBuild.sln

dmitryermilov: Why samples require different approach ?

dvrogozh: They don't, but their solution files existed before. If I would make a decision, I would delete them all, including individual solutions for samples and tutorials.

dmitryermilov: I see. Makes sense to me.

PR #1553, opened 2019-07-29
sample_decode: avoid looping forever

xhaihao:
Don't read the next frame and parse bit stream any more when EOS has
been reached, otherwise it will result in looping forever

### DIFF
--- a/samples/sample_decode/src/pipeline_decode.cpp
+++ b/samples/sample_decode/src/pipeline_decode.cpp
@@ -606,6 +606,19 @@ mfxStatus CDecodingPipeline::InitMfxParams(sInputParams *pParams)
         }
         if (MFX_ERR_MORE_DATA == sts)
         {
+            /* The next calling of m_FileReader->ReadNextFrame will set
+             * pBS->DataFlag to 0 at first and return MFX_ERR_MORE_DATA
+             * if EOS in input stream has been reached. The next calling
+             * of m_pmfxDEC->DecodeHeader will return MFX_ERR_MORE_DATA
+             * again, which will result in looping forever. So checking
+             * whether EOS has been reached before m_FileReader->ReadNextFrame
+             * may work-around the looping forever issue.
+             *
+             * TODO: introduce early exit in ReadNextFrame if EOS is detected.
+             */
+            if (m_mfxBS.DataFlag & MFX_BITSTREAM_EOS)
+                break;
+
             if (m_mfxBS.MaxLength == m_mfxBS.DataLength)
             {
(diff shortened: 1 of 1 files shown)


samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -647,6 +647,9 @@ mfxStatus CDecodingPipeline::InitMfxParams(sInputParams *pParams)
         }
         if (MFX_ERR_MORE_DATA == sts)
         {
+            if (m_mfxBS.DataFlag & MFX_BITSTREAM_EOS)
+                break;
```
mgonchar: I think we should put this ```break``` on line 664 instead

xhaihao: is it possible that m_FileReader->ReadNextFrame() returns MFX_ERR_MORE_DATA but m_mfxBS has valid data?

mgonchar: Hm, I looked at the code and currently I don't understand how it will fall into forever loop?
In my undenderstanding the workflow is following
1) ```DecodeHeader``` return ```ERR_MORE_DATA```
2) ```ReadNextFrame``` will return ```ERR_MORE_DATA``` again
3) ```EOS``` set
4) ```continue```
5) repeat 1)
6) repeat 2)
7) as ```EOS``` already set, return error

xhaihao: EOS is cleared in `ReadNextFrame`.

mgonchar: ok, now I see. the problem is that each overload of ```ReadNextFrame``` rewrites state of ```pBS->DataFlag```. In my understanding the correct fix would be to introduce early exit in these function if they detect ```EOS``` in input stream. But such change may affect our validation, so it should go through some smoke val first. I'm going to treat your fix as WA for such problem and I will remove my -1. Also if you are in hurry with this fix, I'm ok to merge it but with TODO comment in code that this is WA and description of proper fix. Other ways I would prefer to fix it in ```ReadNextFrame```, but that will take some time

xhaihao: I agree my fix is a workaround for the issue and the better fix is to introduce early exit in each `ReadNextFrame`, however what I need is a workable tool and the WA is simpler. I added some comment about the WA but I'm fine if you won't merge this WA.

xhaihao:
Closed my PR because Miroslav has a better solution.

PR #1557, opened 2019-07-29
Moved sample_decode to ExtBufHolder usage

dmitryermilov:

### DIFF
--- a/samples/sample_common/include/sample_utils.h
+++ b/samples/sample_common/include/sample_utils.h
@@ -37,6 +37,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "mfxbrc.h"
 #include "mfxfei.h"
 #include "mfxfeihevc.h"
+#include "mfxmvc.h"
 
 #include "vm/strings_defs.h"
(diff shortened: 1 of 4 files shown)


samples/sample_common/include/sample_utils.h
```diff
@@ -119,110 +113,407 @@ bool IsDecodeCodecSupported(mfxU32 codecFormat);
...
+    };
+
+    std::vector<mfxExtBuffer*> m_ext_buf;
+};
+
+typedef ExtBufHolder<mfxVideoParam> MfxVideoParamsWrapper;
+typedef ExtBufHolder<mfxEncodeCtrl> mfxEncodeCtrlWrap;
+typedef ExtBufHolder<mfxInitParam>  mfxInitParamlWrap;
+typedef ExtBufHolder<mfxBitstream> _mfxBitstreamWrapper;
+
```
onabiull: As I see we don't use _mfxBitstreamWrapper anywhere but in these two lines. Maybe we can just squash them?

dmitryermilov: Done

PR #1559, opened 2019-07-30
set h264 encode default fps to 30

fuweitax:
if input frame rate is unreasonable, fix it to a feasible value (30).

Signed-off-by: Fuwei Tang <fuweix.tang@intel.com>

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -2991,15 +2991,16 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         }
     }
 
+    // if input frame rate is unreasonable, fix it to a feasible value (30.0)
     if ((par.mfx.FrameInfo.FrameRateExtN == 0) !=
         (par.mfx.FrameInfo.FrameRateExtD == 0))
     {
         if (extBits->SPSBuffer && !IsOff(extOpt3->TimingInfoPresent))
             return Error(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
 
-        unsupported = true;
-        par.mfx.FrameInfo.FrameRateExtN = 0;
-        par.mfx.FrameInfo.FrameRateExtD = 0;
+        changed = true;
+        par.mfx.FrameInfo.FrameRateExtN = 30;
+        par.mfx.FrameInfo.FrameRateExtD = 1;
     }
 
     if (!IsOff(extOpt3->TimingInfoPresent) &&


fuweitax:
I saw 30 is set as the default fps if fps info is missed in sample_encode and sample_multi_transcode, so how about setting this default value in MediaSDK?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -2991,15 +2991,15 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
...
     if ((par.mfx.FrameInfo.FrameRateExtN == 0) !=
         (par.mfx.FrameInfo.FrameRateExtD == 0))
     {
         if (extBits->SPSBuffer && !IsOff(extOpt3->TimingInfoPresent))
             return Error(MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
 
-        unsupported = true;
-        par.mfx.FrameInfo.FrameRateExtN = 0;
-        par.mfx.FrameInfo.FrameRateExtD = 0;
+        par.mfx.FrameInfo.FrameRateExtN = 30;
```
stasefremov: Default values for parameters MUST be set in the function SetDefaults(). CheckVideoParamQueryLike() is only check and fix of parameters only.

fuweitax: Thanks for your comment, @stasefremov . I have refined this PR. If FrameRateExtN * FrameRateExtD = 0, it's treated as an unreasonable input frame rate, then we can fix it to a feasible value (30.0) like sample_multi_transcode does. What we do here in CheckVideoParamQueryLike() matches the function of it.

fuweitax: Hi, @stasefremov , does it make sense for you?

stasefremov: This way is worse. Query must not touch uninitialized parameters. This implementation will change (N=0,D=0) to (N=30, D=1) which shouldn't happen.
I could agree that we can treat (N=30,D=0) as incorrect value (not as initialized N and not initialized D). And just "fix" it in Query. But if both N and D are 0, default value should be definitely set in Init (i.e. in SetDefaults()), but not in Query.
Your previous logic may be applicable ( e.g. fix (N=153,D=0) to (N=30,D=1) in Query). But I still have 2 related concerns:
1) In MSDK we usually try to check strictly at initialization stage, and check loosely in runtime. I.e. in Init we usually return errors if user explicitly provides input which we can't fix with good confidence. And in runtime we usually return warnings to not interrupt coding session w/o very hard reason. I can't say that we can fix something like (N=0, D=332) with good confidence. We treat it as significant error in app's logic, and don't allow app start encoding w/o fixing frame rate. It's the logic behind this check. 
2) Even if we change this logic in AVC, we need to make same changes in all encoders (MJPEG, MPEG2, HEVC, VP9) because we need to keep them consistent.

fuweitax: Thank you for the explanation, @stasefremov. I have refined it. And for your concerns:
1. In some cases, (e.g. N = 0, but D != 0) this input may not be incorrect, it may mean that's variable framerate or unknown, for these cases, why not fixing it to a feasible value in MSDK to keep coding running, but throwing a warning? It may improve faults tolerant capability and robustness. Do you think so?
2. Indeed it is. I will make other patches for other encoders if this one is accepted.

stasefremov: Do I understand it right that the reason for this change is demand to pass some particular types of partially defined frame rates like (N=0,D=1)? If you can summ up all cases that (by your opinion) should be accepted by MSDK w/o error, we can call them "reasonable" and fix in Query. And still reject weird combinations like (771, 0) or (0, 943). What do you think?
It's always difficult to decide where is the border between "reasonable" and "not reasonable". Looks like here we have real life examples of reasonable partially defined framerates. We can use this knowledge to set the border and define two MSDK behaviors.

Also, I'll not be available for 2 weeks. Adding Evgeniy Starov to support the discussion.

fuweitax: Sorry for late reply and thank you for your patient explanation, @stasefremov . Yes, just like what you said, we encountered some cases with partially defined frame rates. we have some h264 streams without VUI info in SPS, then when doing transcoding with gstreamer-msdk, it can't calculate the framerate, gstreamer will set the default value 0/1 which is valid in gstreamer and means variable framerate or unknown, but in MSDK, it can't be supported.
What do you think should be done about these cases?

dmitryermilov:
Stanislav Efremov is OOO until Aug, 26

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
fuweitax: Hi, @stasefremov , what do you think of this?

PR #1561, opened 2019-08-01
[core] Add device id for EHL

Alexandr-Konovalov:
Issue: MDP-58421

### DIFF
--- a/api/include/mfxcommon.h
+++ b/api/include/mfxcommon.h
@@ -1,15 +1,15 @@
 // Copyright (c) 2018-2019 Intel Corporation
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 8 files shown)


api/include/mfxcommon.h
```diff
@@ -174,7 +174,8 @@ enum {
     MFX_PLATFORM_CANNONLAKE     = 20,
 #endif
 #if (MFX_VERSION >= 1027)
-    MFX_PLATFORM_ICELAKE = 30,
+    MFX_PLATFORM_ICELAKE        = 30,
+    MFX_PLATFORM_ELKHARTLAKE    = 33,
```
dvrogozh: I guess this should go to the next API version? 1.31 or 1.30 if it is not yet released. @onabiull  - ?

onabiull: Right now we have 1.30 as latest API. So any changes should target at least 1.31, but better to omit exact version and use just NEXT.

But. I believe we should not version enum expansions. Change is too small and safe from backward compatibility point of view. 

So, please remove that guarding ifdef from this piece of code.

Alexandr-Konovalov:
> Please put MFX_

@onabiull Could you please clarify?

api/include/mfxcommon.h
Alexandr-Konovalov: @onabiull Done.

PR #1562, opened 2019-08-02
Allow a larger surface for mpeg2/jpeg encoding

xhaihao:
Like other CODECs, we may allow a larger surface for mpeg2/jpeg
encoding. If do so, we may use the same way to allocate surfaces for all
CODECs in user application / library.

### DIFF
--- a/_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_hw.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -1031,8 +1031,8 @@ mfxStatus MFXVideoENCODEMJPEG_HW::CheckEncodeFrameParam(
             MFX_CHECK(surface->Data.Y != 0 || isExternalFrameAllocator, MFX_ERR_UNDEFINED_BEHAVIOR);
         }
 
-        if (surface->Info.Width != m_vParam.mfx.FrameInfo.Width || surface->Info.Height != m_vParam.mfx.FrameInfo.Height)
-            sts = MFX_WRN_INCOMPATIBLE_VIDEO_PARAM;
+        MFX_CHECK(surface->Info.Width >= m_vParam.mfx.FrameInfo.Width, MFX_ERR_INVALID_VIDEO_PARAM);
+        MFX_CHECK(surface->Info.Height >= m_vParam.mfx.FrameInfo.Height, MFX_ERR_INVALID_VIDEO_PARAM);
     }
     else
     {
--- a/_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
@@ -1675,8 +1675,9 @@ namespace MPEG2EncoderHW
             {
                 return MFX_ERR_UNDEFINED_BEHAVIOR;
             }
-            bWarning = bWarning || (!m_InputSurfaces.CheckInputFrame(&surface->Info));
 
+            MFX_CHECK(surface->Info.Width >= m_VideoParamsEx.mfxVideoParams.mfx.FrameInfo.Width, MFX_ERR_INVALID_VIDEO_PARAM);
+            MFX_CHECK(surface->Info.Height >= m_VideoParamsEx.mfxVideoParams.mfx.FrameInfo.Height, MFX_ERR_INVALID_VIDEO_PARAM);
             MFX_CHECK(surface->Info.FourCC == MFX_FOURCC_NV12, MFX_ERR_UNDEFINED_BEHAVIOR);
 
             if (surface->Data.Y)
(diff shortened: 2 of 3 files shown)


dmitryermilov:
No concern from my side about the idea of the change but please fix the build. Also need a review from @stasefremov

dmitryermilov:
The script expects 2019 year in updated file. Currently it's 2018. Please update

xhaihao:
I think it should be done in another patch.

dmitryermilov:
> I think it should be done in another patch.

No need. Feel free to update current one.

xhaihao:
Ok, I updated the patch, thanks for your comment.

dmitryermilov:
Thanks! 
@stasefremov, your turn:)

xhaihao:
@dvrogozh I adjusted the stride in gst-msdk (https://github.com/intel-media-ci/gst-plugins-bad/pull/6) , which works well for decoding, vpp and h264/h265/vp9 encoding, however it broke mpeg2/jpeg encoding.  Without this fix, I have to use different stride in gst-msdk for mpeg2/jpeg encoding.

xhaihao:
@stasefremov any comment? A MR for gst-msdk (https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/merge_requests/519)  is blocked by this fix.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_hw.cpp
```diff
@@ -1031,8 +1031,8 @@ mfxStatus MFXVideoENCODEMJPEG_HW::CheckEncodeFrameParam(
             MFX_CHECK(surface->Data.Y != 0 || isExternalFrameAllocator, MFX_ERR_UNDEFINED_BEHAVIOR);
         }
```
stasefremov: Does this commit pass behavior tests for JPEG, MPEG2?

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_hw.cpp
```diff
@@ -1031,8 +1031,8 @@ mfxStatus MFXVideoENCODEMJPEG_HW::CheckEncodeFrameParam(
...
 
-        if (surface->Info.Width != m_vParam.mfx.FrameInfo.Width || surface->Info.Height != m_vParam.mfx.FrameInfo.Height)
-            sts = MFX_WRN_INCOMPATIBLE_VIDEO_PARAM;
+        MFX_CHECK(surface->Info.Width >= m_vParam.mfx.FrameInfo.Width, MFX_ERR_INVALID_VIDEO_PARAM);
```
stasefremov: Warning changed to error. Looks like a typo.

xhaihao: @stasefremov  I think it would be better to align the behavior with other CODECs, E.g. for H264,  MFX_ERR_INVALID_VIDEO_PARAM is returned when the condition is false (https://github.com/Intel-Media-SDK/MediaSDK/blob/f3b12a7ce55568be212a42a36257de2abc8150f2/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils.cpp#L3021-L3022)

dmitryermilov: I agree with Haihao

xhaihao: It would be better to return MFX_ERR_INVALID_VIDEO_PARAM for jpeg too when the condition is false. If doing so, we may use the same error handling path in user application.  How do you think? @dmitryermilov @stasefremov

xhaihao: In addition, I don't think it is suitable to return MFX_WRN_INCOMPATIBLE_VIDEO_PARAM when the surface size is smaller than the video frame size.

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -1675,7 +1675,9 @@ namespace MPEG2EncoderHW
             {
                 return MFX_ERR_UNDEFINED_BEHAVIOR;
             }
-            bWarning = bWarning || (!m_InputSurfaces.CheckInputFrame(&surface->Info));
+            bWarning = bWarning ||
```
gkrivor: And regarding previous discussion about codec alignment - it should be an error instead of warning.

gkrivor: And looks like best way is change handling of bWarning at the end of the method.

xhaihao: @gkrivor I updated the patch but didn't change the handling of bWarning because it is related to  many parameter checks.  Could you please review the patch again?

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -1675,7 +1675,9 @@ namespace MPEG2EncoderHW
             {
                 return MFX_ERR_UNDEFINED_BEHAVIOR;
             }
-            bWarning = bWarning || (!m_InputSurfaces.CheckInputFrame(&surface->Info));
```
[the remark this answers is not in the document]
xhaihao: Done.

gkrivor:
Sorry, owner of this task is on sick leave, need more time for internal preparations, I hope we finish it next week

xhaihao:
Any progress on this PR?

gkrivor:
Sorry, no, task owner is still unavailable

Alexandr-Konovalov:
@gkrivor @onabiull Internal preparations were done.

xhaihao:
Thanks all for merging the patch.

PR #1565, opened 2019-08-05
Change internal and samples allocator for align pointers to pagesize

### LINKED ISSUE #1546: Internal allocator and samples don't align pointers to page size
MSDK internal allocator and samples don't align pointers to page size. Basically this is requirement of I915_GEM_USERPTR which is needed to make video memory from system memory. Another USERPTR requirement is memory size shall be page aligned.

AntonGrishin:
Samples and MediaSDK doesn't align pointers to page size. But I915_GEM_USERPTR
requirement is memory size and pointer shall be page aligned.

Fixes #1546

### DIFF
--- a/_studio/shared/src/libmfx_allocator.cpp
+++ b/_studio/shared/src/libmfx_allocator.cpp
@@ -1,15 +1,15 @@
-// Copyright (c) 2017-2018 Intel Corporation
-// 
+// Copyright (c) 2017-2019 Intel Corporation
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 // FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -24,6 +24,9 @@
 
 #include "mfx_utils.h"
 #include "mfx_common.h"
+#if defined (MFX_VA_LINUX)
+#include "unistd.h"
+#endif
 
 #define ALIGN32(X) (((mfxU32)((X)+31)) & (~ (mfxU32)31))
 #define ID_BUFFER MFX_MAKEFOURCC('B','U','F','F')
@@ -41,7 +44,14 @@ mfxStatus mfxDefaultAllocator::AllocBuffer(mfxHDL pthis, mfxU32 nbytes, mfxU16 t
     if(!mid)
         return MFX_ERR_NULL_PTR;
     mfxU32 header_size = ALIGN32(sizeof(BufferStruct));
-    mfxU8 *buffer_ptr=(mfxU8 *)malloc(header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
+
+//MSVS still has no aligned_alloc support, because
+//it incompatible with their implementation of free()
+#if defined (MFX_VA_LINUX)
+    mfxU8 *buffer_ptr = (mfxU8 *)aligned_alloc(sysconf(_SC_PAGESIZE), header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
+#else
+    mfxU8 *buffer_ptr = (mfxU8 *)malloc(header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
+#endif
 
     if (!buffer_ptr)
         return MFX_ERR_MEMORY_ALLOC;
--- a/samples/sample_common/src/sysmem_allocator.cpp
+++ b/samples/sample_common/src/sysmem_allocator.cpp
@@ -392,10 +392,19 @@ mfxStatus SysMemBufferAllocator::AllocBuffer(mfxU32 nbytes, mfxU16 type, mfxMemI
         return MFX_ERR_UNSUPPORTED;
 
     mfxU32 header_size = MSDK_ALIGN32(sizeof(sBuffer));
-    mfxU8 *buffer_ptr = (mfxU8 *)calloc(header_size + nbytes + 32, 1);
+
+//MSVS still has no aligned_alloc support, because
+//it incompatible with their implementation of free()
+#if defined (LINUX32) || defined (LINUX64)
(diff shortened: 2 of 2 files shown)


samples/sample_common/src/sysmem_allocator.cpp
```diff
@@ -392,7 +392,11 @@ mfxStatus SysMemBufferAllocator::AllocBuffer(mfxU32 nbytes, mfxU16 type, mfxMemI
         return MFX_ERR_UNSUPPORTED;
 
     mfxU32 header_size = MSDK_ALIGN32(sizeof(sBuffer));
+#if defined (LINUX32) || defined (LINUX64)
+    mfxU8 *buffer_ptr = (mfxU8 *)aligned_alloc(sysconf(_SC_PAGESIZE), header_size + nbytes + 32);
```
dmitryermilov: Need to zero memory to align with Windows

_studio/shared/src/libmfx_allocator.cpp
```diff
@@ -41,7 +44,11 @@ mfxStatus mfxDefaultAllocator::AllocBuffer(mfxHDL pthis, mfxU32 nbytes, mfxU16 t
     if(!mid)
         return MFX_ERR_NULL_PTR;
     mfxU32 header_size = ALIGN32(sizeof(BufferStruct));
-    mfxU8 *buffer_ptr=(mfxU8 *)malloc(header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
+#if defined (MFX_VA_LINUX)
+    mfxU8 *buffer_ptr = (mfxU8 *)aligned_alloc(sysconf(_SC_PAGESIZE), header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
```
dmitryermilov: Need to zero memory to align with Windows

dmitryermilov: Please also add a comment in code why we don't do the same alignment for windows

samples/sample_common/src/sysmem_allocator.cpp
AntonGrishin: Changed calloc to malloc and set zero memory below (like in libmfx_allocator).

_studio/shared/src/libmfx_allocator.cpp
AntonGrishin: Returned malloc for non-linux case.

AntonGrishin: Done.

mgonchar: why don't we use ```std::align``` instead? I believe best way would be to refactor this class in base of ```std::aligned_storage```

samples/sample_common/src/sysmem_allocator.cpp
dvrogozh: I thought Windows gas _aligned_alloc and _aligned_free pair. Can we use them and introduce some high level macro  covering both windows and linux? Like:
```
#ifdef _WIN32
#define ALIGNED_ALLOC(X, S) _aligned_malloc(X, S)
#define ALIGNED_ALLOC(P) _aligned_free(P)
#else 
#define ALIGNED_ALLOC(X, S) aligned_malloc(X, S)
#define ALIGNED_ALLOC(P) free(X, S)
#endif 

```
Or we can try follow @mgonchar proposal to use std.

samples/sample_common/src/sysmem_allocator.cpp
```diff
@@ -392,11 +392,20 @@ mfxStatus SysMemBufferAllocator::AllocBuffer(mfxU32 nbytes, mfxU16 type, mfxMemI
...
 
     mfxU32 header_size = MSDK_ALIGN32(sizeof(sBuffer));
-    mfxU8 *buffer_ptr = (mfxU8 *)calloc(header_size + nbytes + 32, 1);
+
+//MSVS still has no aligned_alloc support, because
+//it incompatible with their implementation of free()
+#if defined (LINUX32) || defined (LINUX64)
+    mfxU8 *buffer_ptr = (mfxU8 *)aligned_alloc(sysconf(_SC_PAGESIZE), header_size + nbytes + 32);
+#else
+    mfxU8 *buffer_ptr = (mfxU8 *)malloc(header_size + nbytes + 32);
```
dvrogozh: hmmm. What is the header_size? Actually, the critical point is: where pointer to the surface data starts? This pointer should be aligned by page boundary. If it is not (but header is) - this will NOT do.

_studio/shared/src/libmfx_allocator.cpp
dvrogozh: Ditto in samples change. Windows has _algined_malloc and _aligned_free pair. Introduce a wrapping macro which uses appropriate version or use std as @mgonchar suggested.

_studio/shared/src/libmfx_allocator.cpp
```diff
@@ -41,7 +44,14 @@ mfxStatus mfxDefaultAllocator::AllocBuffer(mfxHDL pthis, mfxU32 nbytes, mfxU16 t
...
         return MFX_ERR_NULL_PTR;
     mfxU32 header_size = ALIGN32(sizeof(BufferStruct));
-    mfxU8 *buffer_ptr=(mfxU8 *)malloc(header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
+
+//MSVS still has no aligned_alloc support, because
+//it incompatible with their implementation of free()
+#if defined (MFX_VA_LINUX)
+    mfxU8 *buffer_ptr = (mfxU8 *)aligned_alloc(sysconf(_SC_PAGESIZE), header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
+#else
+    mfxU8 *buffer_ptr = (mfxU8 *)malloc(header_size + nbytes + DEFAULT_ALIGNMENT_SIZE);
```
dvrogozh: Where pointer to surface data will start? Will it be at page boundary?

samples/sample_common/src/sysmem_allocator.cpp
dmitryermilov: Good catch! @AntonGrishin , can you please check if we need this header at all? Better to  get rid of it.

_studio/shared/src/libmfx_allocator.cpp
AntonGrishin: @dvrogozh, yes, you are right. But align_alloc also std function, but currently not implemented in windows.  I will try to investigate both ways which you suggest for choose the best one.

mgonchar: @AntonGrishin unfortunately ```std::aligned_alloc``` is part of C++17 which is not supported in MSDK project. Otherwise it would have solved our problem. ```_algined_malloc``` which is suggested by @dvrogozh is properly supported by MSVC, see [here ](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2019)

AntonGrishin:
@dvrogozh @dmitryermilov  I tested performance (for mirroring via CM) with and without this changes, and FPS in both cases was quite similar.

samples/sample_common/src/sysmem_allocator.cpp
mgonchar: Also, let's follow our [coding guide](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/Coding_guidelines.md#memory-management) and switch to c++ cast at least in new code :)

PR #1566, opened 2019-08-05
Added support of multiadapters

mgonchar:
- Added new API to dispatcher
- Added information about adapter type to mfxPlatformInfo
- Added support of new API to sample_encode, sample_decode, sample_vpp

### DIFF
--- a/api/include/mfxadapter.h
+++ b/api/include/mfxadapter.h
@@ -1,0 +1,39 @@
+// Copyright (c) 2019 Intel Corporation
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+#ifndef __MFXADAPTER_H__
+#define __MFXADAPTER_H__
+
+#include "mfxstructures.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+mfxStatus MFXQueryAdapters(mfxComponentInfo* input_info, mfxAdaptersInfo* adapters);
+mfxStatus MFXQueryAdaptersDecode(mfxBitstream* bitstream, mfxU32 codec_id, mfxAdaptersInfo* adapters);
+mfxStatus MFXQueryAdaptersNumber(mfxU32* num_adapters);
+#ifdef __cplusplus
+} // extern "C"
+#endif
+
+#endif // __MFXADAPTER_H__
+#endif
--- a/api/include/mfxstructures.h
+++ b/api/include/mfxstructures.h
@@ -2269,6 +2269,48 @@ MFX_PACK_END()
 
 #endif
 
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+/* Multi-adapters Querying structs */
+typedef enum
+{
+    MFX_COMPONENT_ENCODE = 1,
+    MFX_COMPONENT_DECODE = 2,
+    MFX_COMPONENT_VPP    = 3
+} mfxComponentType;
+
+MFX_PACK_BEGIN_STRUCT_W_PTR()
+typedef struct
+{
+    mfxComponentType Type;
+    mfxVideoParam    Requirements;
+
+    mfxU16           reserved[4];
+} mfxComponentInfo;
+MFX_PACK_END()
+
+/* Adapter description */
+MFX_PACK_BEGIN_USUAL_STRUCT()
+typedef struct
+{
+    mfxPlatform Platform;
+    mfxU32      Number;
+
+    mfxU16      reserved[14];
+} mfxAdapterInfo;
+MFX_PACK_END()
+
+MFX_PACK_BEGIN_STRUCT_W_PTR()
+typedef struct
+{
+    mfxAdapterInfo * Adapters;
+    mfxU32           NumAlloc;
+    mfxU32           NumActual;
+
+    mfxU16           reserved[4];
+} mfxAdaptersInfo;
+MFX_PACK_END()
+
+#endif
 #ifdef __cplusplus
 } // extern "C"
 #endif
--- a/api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
+++ b/api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
@@ -26,6 +26,13 @@
 
 #include <string.h>
 #include <windows.h>
+
+#include "mfx_dxva2_device.h"
+#include "mfxvideo++.h"
+#include "mfx_vector.h"
+#include "mfxadapter.h"
+#include <algorithm>
+
 #pragma warning(disable:4355)
 
 MFX_DISP_HANDLE::MFX_DISP_HANDLE(const mfxVersion requiredVersion) :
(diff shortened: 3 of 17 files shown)


api/mfx_dispatch/windows/include/mfx_dispatcher.h
```diff
@@ -218,4 +218,76 @@ FUNCTION_DESCRIPTION APIFunc[eVideoFuncTotal];
 
 extern const
 FUNCTION_DESCRIPTION APIAudioFunc[eAudioFuncTotal];
+
+
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
```
dvrogozh: Why this API is made _windows_ dispatcher specific. I think it makes sense to consider declaring it on mediasdk API level otherwise it is not quite clear whether this API is official or not?

api/mfx_dispatch/windows/include/mfx_dispatcher.h
```diff
@@ -218,4 +218,76 @@ FUNCTION_DESCRIPTION APIFunc[eVideoFuncTotal];
...
+    mfxAdapterInfo * Adapters;
+    mfxU32           NumAlloc;
+    mfxU32           NumActual;
+
+    mfxU16           reserved[4];
+};
+
+mfxStatus MFXQueryAdapters(mfxComponentInfo* input_info, mfxAdaptersInfo* adapters);
+mfxStatus MFXQueryAdaptersDecode(mfxBitstream* bitstream, mfxU32 codec_id, mfxAdaptersInfo* adapters);
+mfxStatus MFXQueryAdaptersNumber(mfxU32* num_adapters);
```
dvrogozh: This API follows approach: I have something, do you support that? That's a question not everyone bothers to actually ever ask, hence lower number of adoptions than there could be:). Alternative approach which I believe suites better is: give me what you support and I will check whether I can use it?

From the perspective of the 2nd API approach there would be 1 or 2 primary functions:
```
mfxStatus MFXQueryAdaptersNumber(mfxU32* num_adapters);
mfxStatus MFXQueryAdaptersInfo(mfxU32 num_adapters, mfxAdapterInfo* adapters);
```

After that we can have any number of helper functions which utilize the above query API and search thru it.

Mind that with this approach you immediately get opportunity to write very simple sample application which queries all adapters and reports back their caps.

mgonchar: I'm absolutely agree with you. But we have to define capabilities somehow. We need to define some general enough struct that can serve as description for supported features. It might happen that such struct would be giant and thus not easy to manipulate. And the second and most hard thing - we need to define those caps. Actually current API is extendable to support such approach also. We can add pointer to capabilities to ```mfxAdapterInfo```. So we can use it as you suggested in following way via proposed API: pass ```nullptr``` instead of requirements and attach capabilities struct to each ```mfxAdapterInfo```. Then MSDK lib will fill those structs with information about supported features for every adapter. I think we may introduce such feature later after merge of this API, because it will take some time to design proper caps struct and implementing such reporting mechanism.

api/mfx_dispatch/windows/include/mfx_dispatcher.h
mgonchar: Currently we have only windows implementation, so we will have those defines for some time anyway. It is logical that those functions should be implemented on dispatcher layer (or in some separate lib operating on same layer than dispatcher), so technically it is not part of MSDK library.

onabiull: From precedent point of view, doc/mediasdk-man.pdf already define API in terms of lib+dispatcher. I.e. MFXInit would behave slightly differently w/o dispatcher. Absence of Linux implementation is a clear gap.

mgonchar: exposed to MSDK API

api/include/mfxstructures.h
```diff
@@ -2269,6 +2269,48 @@ MFX_PACK_END()
 
 #endif
 
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+/* Multi-adapters Querying structs */
+typedef enum
+{
+    MFX_ENCODE_COMPONENT = 1,
```
dvrogozh: From my perspective `MFX_COMPONENT_<name>` style will be better.

api/include/mfxstructures.h
```diff
@@ -2269,6 +2269,48 @@ MFX_PACK_END()
...
+    MFX_DECODE_COMPONENT = 2,
+    MFX_VPP_COMPONENT    = 3
+} mfxComponentType;
```
dvrogozh: how FEI is classified?

api/include/mfxstructures.h
```diff
@@ -2269,6 +2269,48 @@ MFX_PACK_END()
...
+
+MFX_PACK_BEGIN_STRUCT_W_PTR()
+typedef struct
+{
+    mfxComponentType Type;
```
dvrogozh: hm... We have few reserved fields in mfxVideoParam. Maybe we will just add `mfxComponentType` in there and avoid defining mfxComponentInfo altogether?

api/include/mfxstructures.h
```diff
@@ -2269,6 +2269,48 @@ MFX_PACK_END()
...
+MFX_PACK_END()
+
+/* Adapter description */
+MFX_PACK_BEGIN_USUAL_STRUCT()
+typedef struct
+{
+    mfxPlatform Platform;
+    mfxU32      Number;
+
+    mfxU16      reserved[14];
```
dvrogozh: we might need adapter type on Linux to distinguish render nodes and cards

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
```diff
@@ -318,3 +325,292 @@ mfxStatus MFX_DISP_HANDLE::UnLoadSelectedDLL(void)
     return mfxRes;
 
 } // mfxStatus MFX_DISP_HANDLE::UnLoadSelectedDLL(void)
+
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
+static mfxStatus InitDummySession(mfxU32 adapter_n, MFXVideoSession & dummy_session)
```
dvrogozh: errgghh.... I do NOT like creating any dummy sessions in the dispatcher for the reason that they are not cheap. They create scheduler and fully qualified core. If that's needed to support adapter query via bitstream I would better sacrifice this function for now.

api/include/mfxadapter.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "mfxstructures.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+mfxStatus MFXQueryAdapters(mfxComponentInfo* input_info, mfxAdaptersInfo* adapters);
+mfxStatus MFXQueryAdaptersDecode(mfxBitstream* bitstream, mfxU32 codec_id, mfxAdaptersInfo* adapters);
+mfxStatus MFXQueryAdaptersNumber(mfxU32* num_adapters);
```
dvrogozh: Per my opinion this API is over complex. For example, 2 structs with almost the same name (mfxAdapterSInfo) confuse. I would this:
```
mfxU32 MFXQueryAdaptersNumber();

// num_adapters size of the adapters array allocated by the user
// functions return number of match adapters within adapters array
mfxU32 MFXQueryAdapters(mfxComponentInfo* input_info, mfxAdapterInfo* adapters, mfxU32 num_adapters);
mfxU32 MFXQueryAdaptersDecode(mfxBitstream* bitstream, mfxU32 codec_id, mfxAdapterInfo* adapters, mfxU32 num_adapters);
```

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
dvrogozh: To solve that I would suggest to implement MFXQueryAdapters* functions directly inside mediasdk and teach dispatcher to collect this info directly from msdk libs and return aggregated info.

mgonchar:
@dvrogozh documentation added

api/include/mfxstructures.h
mgonchar: Well, maybe

api/include/mfxstructures.h
mgonchar: FEI is ENCODE + extension buffer (attached to ```mfxVideoParam``` on Init). On HEVC currently it is separate plugin, but it is possible to initialize it same way (with ext buffer) as AVC. We just need to add such buffer to API for HEVC.

mgonchar: But for HEVC FEI scope of supported configurations is absolutely same as for regular HEVC encoder, so it is not critical issue.

api/include/mfxstructures.h
mgonchar: Yep, those reserved fields might me reused for Linux support

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
mgonchar: Dummy sessions used in each function. No difference for bitstream in this

api/include/mfxstructures.h
mgonchar: No, I don't like this idea. We widely use ```mfxVideoParam``` struct in other parts of API and design of this struct assumes that it is component agnostic

api/include/mfxadapter.h
mgonchar: But we also report back internal statuses, which is aligned with MSDK API, where we always return statuses

doc/mediasdk-man.md
```diff
@@ -5671,11 +5672,14 @@ The `mfxPlatform` structure contains information about hardware platform.
 --- | ---
 `CodeName` | Intel® microarchitecture code name. See the [PlatformCodeName](#PlatformCodeName) enumerator for a list of possible values.
 `DeviceId` | Reserved.
+`MediaAdapterType` | Description of Intel Gen Graphics adapter type. See the [mfxMediaAdapterType](#mfxMediaAdapterType) enumerator for a list of possible values.
 
 **Change History**
 
 This structure is available since SDK API 1.19.
 
+The SDK API **TBD** adds `MediaAdapterType` field.
```
dvrogozh: @onabiull : how we will handle these in releases if API won't land in the release? I would suggest to make API changes in separate commits which we can specially mark somehow and just revert commits in release branch.

api/include/mfxadapter.h
dvrogozh: This is not an argument to me. Imho, "we always did that" is not kind of argument I would accept. I still am persistent with my opinion: API is over complex.

api/include/mfxstructures.h
dvrogozh: That's another badly designed API. mfxVideoParam has union inside _without_ any way to understand which union path is to be followed from the data stored in the structure. IMHO, mfxComponentType should have been initially in this union from the msdk birth day!

doc/mediasdk-man.md
```diff
@@ -1161,6 +1165,125 @@ Assume the pipeline is **A->B->C**. The application synchronizes on sync point *
 
 ## <a id='hardware_acceleration'>Working with hardware acceleration</a>
 
+### <a id='Working_with_multiple_intel_media_devices'>Working with multiple Intel media devices</a>
+
+If your system has multiple Intel Gen Graphics adapters you may need hints on which adapter suits better to process some particular workload. The SDK provides API to help with adapter selection. See Example 9 that showcases workload processing initialization on discrete adapter.
```
dvrogozh: hard to read....

"Media SDK provides helper API to select best suitable adapter for your workload based on passed workload description. See Example 9 that showcases workload initialization on discrete adapter." (remove word "processing" from example line)

doc/mediasdk-man.md
```diff
@@ -1161,6 +1165,125 @@ Assume the pipeline is **A->B->C**. The application synchronizes on sync point *
...
+if (idx_d == adapters.Adapters + adapters.NumActual)
+{
+    printf("Warning: No dGfx detected on machine\n");
+    return -1;
+}
+
+mfxU32 idx = static_cast<mfxU32>(std::distance(adapters.Adapters, idx));
+
+// Choose correct implementation for discrete adapter
+switch (adapters.Adapters[idx].Number)
```
dvrogozh: hm. And why API does not return impl directly:)?

doc/mediasdk-man.md
```diff
@@ -1161,6 +1165,125 @@ Assume the pipeline is **A->B->C**. The application synchronizes on sync point *
...
+    impl = MFX_IMPL_HARDWARE_ANY;
+    break;
+}
+
+// Initialize mfxSession in regular way with obtained implementation
+```
+
+As you see in Example 9, after obtaining adapter list with [MFXQueryAdapters](#MFXQueryAdapters) further initialization of [mfxSession](#mfxSession) is performed in regular way. Particular adapter choice is performed with [MFX_IMPL_HARDWARE](#mfxIMPL)**...**[MFX_IMPL_HARDWARE4](#mfxIMPL) values of [mfxIMPL](#mfxIMPL) (see [Multiple Monitors](#Multiple_monitors) section for details).
+
+Example 10 showcases usage of [MFXQueryAdapters](#MFXQueryAdapters) for querying best suitable adapter for particular encode workload processing (see [MFXQueryAdapters](#MFXQueryAdapters) description for adapter priority rules).
```
dvrogozh: remove "processing"

doc/mediasdk-man.md
```diff
@@ -1469,6 +1592,97 @@ Global functions initialize and de-initialize the SDK library and perform query
...
+
+**Syntax**
+
+[mfxStatus](#mfxStatus) `MFXQueryAdapters(mfxComponentInfo* input_info, mfxAdaptersInfo* adapters);`
+
+**Parameters**
+
+| | |
+--- | ---
+`input_info` | Pointer to workload description. See [mfxComponentInfo](#mfxComponentInfo) description for details.
```
dvrogozh: Can I pass input_info=nullptr? What I will get?

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
dvrogozh: There will be difference for Linux dispatcher which is tiny and I don't want to make it bigger.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -141,6 +141,142 @@ CDecodingPipeline::~CDecodingPipeline()
     Close();
 }
 
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
+mfxU32 CDecodingPipeline::GetPreferredAdapterNum(const mfxAdaptersInfo & adapters, const sInputParams & params)
```
dvrogozh: Why this function is in decoder sample? Why not in utilities?

dvrogozh: Do we actually need WIN protection here and in most other places? API is common between Windows and Linux...

samples/sample_decode/src/sample_decode.cpp
```diff
@@ -62,6 +62,10 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage)
     msdk_printf(MSDK_STRING("   [-di bob/adi]             - enable deinterlacing BOB/ADI\n"));
 #if (MFX_VERSION >= 1025)
     msdk_printf(MSDK_STRING("   [-d]                      - enable decode error report\n"));
+#endif
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
```
dvrogozh: this

samples/sample_decode/src/sample_decode.cpp
```diff
@@ -416,6 +419,16 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
         {
             pParams->bSoftRobustFlag = true;
         }
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
```
dvrogozh: and this are likely the only 2 places which need WIN protection since on Linux we don't have API implemented (yet)

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -1350,12 +1352,181 @@ mfxStatus CEncodingPipeline::InitFileWriters(sInputParams *pParams)
...
+    case MFX_FOURCC_Y416:
+        return 12;
+        break;
+#endif
+    default:
+        return 0;
+    }
+}
+
+mfxU32 CEncodingPipeline::GetPreferredAdapterNum(const mfxAdaptersInfo & adapters, const sInputParams & params)
```
dvrogozh: "he's once again my guest..."

samples/sample_vpp/src/sample_vpp_utils.cpp
```diff
@@ -414,6 +426,107 @@ mfxStatus InitParamsVPP(mfxVideoParam* pParams, sInputParams* pInParams, mfxU32
 
 /* ******************************************************************* */
 
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
+mfxU32 GetPreferredAdapterNum(const mfxAdaptersInfo & adapters, const sInputParams & params)
```
dvrogozh: emmmm.....

doc/mediasdk-man.md
onabiull: My plan is to strip all \*\*TBD\*\* sections from documents in release branch. Probably in the same commit with version change. After all there are just dozen or so of them.

api/include/mfxstructures.h
mgonchar: Well, maybe, but MSDK uses different approach for component type resolution. We use different functions calls, i.e. resolution happens on level of function name

api/include/mfxadapter.h
mgonchar: Well, from other side your approach hiding errors. What if something will go wrong? You suggest to return 0 in such case (no suitable adapters), but how user will know the reason of this? Either it is workload that can't be handled, or some internal error occured

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
mgonchar: About implementing it in lib. How do you suppose to retrieve information from adapter? This information becomes available at init time. How should we redesign MSDK to connect unitialized library to adapter? Currently dispatcher loads lib during first init call (that's why there are dummy sessions and inits here), but how to bypass them, I don't see an easy way now, which will be better that proposed solution

doc/mediasdk-man.md
mgonchar: So, manual editing of docs?) Shouldn't we hide hints for automatic cut inside md comments for TBD sections?

doc/mediasdk-man.md
mgonchar: Because impl type supports up to 4 adapters, in future we will use ext buffer to support more adapters

doc/mediasdk-man.md
mgonchar: ok, will do later

doc/mediasdk-man.md
mgonchar: list of adapters. If I forgot to mention it in doc, I will add

samples/sample_decode/src/pipeline_decode.cpp
mgonchar: since it is implementation, yes we need. I can think about putting it to utility

samples/sample_vpp/src/sample_vpp_utils.cpp
mgonchar: these are different input parameters, but they have same name. We shouldn't bring this to utility, since we don't want to put giant structs of input parameters description to common code

samples/sample_encode/src/pipeline_encode.cpp
mgonchar: "It's not what it seems to be..."

samples/sample_decode/src/sample_decode.cpp
mgonchar: No, because setting these two parameters to ```false``` doesn't mean bypassing new API calls. By default we will query best suitable adapter

samples/sample_vpp/src/sample_vpp_utils.cpp
dvrogozh: I got impression that you use just 1, maybe 2 values from input parameters. If so, you can change function prototype and pass only what you need making it possible to move function to utilities library.

mgonchar: Not sure if it makes sense. In this case it would be something like:
```
common_header.h
...
mfxU32 GetPreferredAdapterNum(const mfxAdaptersInfo & adapters, const mfxVideoParam& params)
```

But in each sample:
```
sample.cpp
...
void FillParameters(const sInputParams & input_params, mfxVideoParam& output_params)
```

Doesn't look much better

doc/mediasdk-man.md
onabiull: Right now - yes. We can think about employing html-style comments (they won't be rendered), but I not sure that in current scheme - 4 releases per year, it would worth investments.

doc/mediasdk-man.md
mgonchar: done

doc/mediasdk-man.md
mgonchar: done

doc/mediasdk-man.md
mgonchar: Doc updated

onabiull:
As we discussed offline, looks like accepting this in current state, are best option. Still there are clearly much room for improvement :)

PR #1567, opened 2019-08-05
[mpeg2d] Frame doubling/tripling cases handling

dmitryermilov:

### DIFF
--- a/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
+++ b/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


dmitryermilov:
@asmirno1 , please don't forget to pick it up

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -1044,6 +1044,18 @@ namespace UMC_MPEG2_DECODER
                 m_localFrameTime += (m_localDeltaFrameTime / 2);
             }
             break;
+        case DPS_FRAME_DOUBLING:
+            if (m_params.lFlags & UMC::FLAG_VDEC_TELECINE_PTS)
+            {
+                m_localFrameTime += m_localDeltaFrameTime;
+            }
+            break;
+        case DPS_FRAME_DOUBLING:
```
mgonchar: I believe it should be different case

dmitryermilov: Thanks!

PR #1570, opened 2019-08-06
VP9dec: the driver may support up to 8K video

xhaihao:
See https://github.com/intel/media-driver/blob/master/docs/media_features.md#supported-decoding-output-format-and-max-resolution

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -180,12 +180,12 @@ namespace MFX_VPX_Utility
             if (!p_in->mfx.FrameInfo.ChromaFormat && !(!p_in->mfx.FrameInfo.FourCC && !p_in->mfx.FrameInfo.ChromaFormat))
                 sts = MFX_ERR_UNSUPPORTED;
 
-            if (p_in->mfx.FrameInfo.Width % 16 == 0 && p_in->mfx.FrameInfo.Width <= GetMaxWidth(codecId))
+            if (p_in->mfx.FrameInfo.Width % 16 == 0)
                 p_out->mfx.FrameInfo.Width = p_in->mfx.FrameInfo.Width;
             else
                 sts = MFX_ERR_UNSUPPORTED;
 
-            if (p_in->mfx.FrameInfo.Height % 16 == 0 && p_in->mfx.FrameInfo.Height <= GetMaxHeight(codecId))
+            if (p_in->mfx.FrameInfo.Height % 16 == 0)
                 p_out->mfx.FrameInfo.Height = p_in->mfx.FrameInfo.Height;
             else
                 sts = MFX_ERR_UNSUPPORTED;


mgonchar:
@xhaihao according to your table max output for BXT/APL is still 4k. Don't we need to introduce platform check as well?

xhaihao:
Per my understanding, the driver should check whether the resolution is supported, so it is not needed to check the resolution in MediaSDK.

mgonchar:
This function is used in our ```Query``` function to tell user if his setting is supported. In case of BXT/APL we will report incorrect status to user for resolution >4k. Destination of these functions to report maximum width/height on user's platform, but not a theoretical width he might be able to process if he would be using our latest graphics. I think correct solution is to add platform as additional parameter to these functions.

Also, according to your table it seems that other values in these functions need to be adjusted as well.

dmitryermilov:
Hi @xhaihao , I suggest the following:
* Can you please remove this hardcoded functional at all
* And replace them with https://github.com/Intel-Media-SDK/MediaSDK/blob/ac53ba9d2c7597697ab523e33f37c61103515e7d/_studio/shared/src/libmfx_core_vaapi.cpp#L1279

xhaihao:
@dmitryermilov  Per my understanding, `VAAPIVideoCORE::IsGuidSupported` is for vaapi only, Is it reasonable to call `IsGuidSupported` in the `Query` function?

dmitryermilov:
> Per my understanding, `VAAPIVideoCORE::IsGuidSupported` is for vaapi only

Basically yes. What is your concern here?

> Is it reasonable to call `IsGuidSupported` in the `Query` function?

Yes, Please refer e.g. to AVC https://github.com/Intel-Media-SDK/MediaSDK/blob/76d894191b74b43b8de9cf598fd86a2b1c6e3e45/_studio/shared/umc/codec/h264_dec/src/umc_h264_mfx_supplier.cpp#L529 or MPEG https://github.com/Intel-Media-SDK/MediaSDK/blob/fcb7e6bf4c6b3acbc2698ecb45926be6ac8cd911/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_utils.cpp#L60

xhaihao:
@dmitryermilov @mgonchar Actually IsGuidSupported has been called in `Query`

In VideoDECODEVP9_HW::Query
https://github.com/Intel-Media-SDK/MediaSDK/blob/5b23796ab993771718fec3b997c3f4a5c98c6c4c/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp#L615

In CheckHardwareSupport
https://github.com/Intel-Media-SDK/MediaSDK/blob/5b23796ab993771718fec3b997c3f4a5c98c6c4c/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp#L83-L84

So, how about removing the lines below from MFX_VPX_Utility::Query

https://github.com/Intel-Media-SDK/MediaSDK/blob/5b23796ab993771718fec3b997c3f4a5c98c6c4c/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp#L183-L191

Note: MFX_VPX_Utility::Query is called only for vp9 decoding, so removing the lines above doesn't impact other codecs

dmitry-gurulev:
@xhaihao 
> So, how about removing the lines below from MFX_VPX_Utility::Query

We can remove max resolution checks (mfx.FrameInfo.Width <= GetMaxWidth) but need to leave checks of alignment 'as-is'.

xhaihao:
@dmitryermilov @dmitry-gurulev @mgonchar I updated the patch, could you please review it again?

onabiull:
Looks test trigger lost on CI side...

PR #1571, opened 2019-08-06
vp9dec: fix vp9 qindex

xhaihao:
Refer to libvpx code
https://github.com/webmproject/libvpx/blob/e46820c693d0dc01685841cfc704c16dbac48e80/vp9/common/vp9_quant_common.c#L196-L206

### DIFF
--- a/_studio/shared/umc/codec/vp9_dec/src/umc_vp9_utils.cpp
+++ b/_studio/shared/umc/codec/vp9_dec/src/umc_vp9_utils.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


_studio/shared/umc/codec/vp9_dec/src/umc_vp9_utils.cpp
```diff
@@ -282,10 +282,9 @@ namespace UMC_VP9_DECODER
         else
         {
             const int32_t data = GetSegData(seg, segmentId, UMC_VP9_DECODER::SEG_LVL_ALT_Q);
-            return
-                 seg.absDelta == SEGMENT_ABSDATA ?
-                 data :  // Abs value
-                 clamp(baseQIndex + data, 0, MAXQ);  // Delta value
+            const int seg_qindex =
```
mgonchar: Please, align style and use ```int32_t``` type

xhaihao: Thanks for your comment, I updated the patch to use `int32_t`

PR #1574, opened 2019-08-07
doc: coding guide added

mgonchar:
Added coding guide document. Violation of rules from this document may be a reason to request change on uploaded PR. Any input on improving this document is welcome!


doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+```
+-   Prefer switch operator to giant else-if statements.
+
+2.  Keep code at level of 0 compiler warning build.
+
+3.  Do not rely on side effects of your code.
+
+**Code comments**
+
+4.  Do not leave dead commented code. It often will be barriers in source base forever. Either leave meaningful comment why this code is commented (i.e. determine the condition when this code can be switched on again) otherwise delete it at all. Traces of deleted code will remain in git history.
```
dvrogozh: > It often will be barriers 
What does that mean? Rephrase.

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+-   Use ```MFX_CHECK``` / ```MFX_RETURN``` / ```MFX_CHECK_STS``` macro for ```mfxStatus``` checks. It allows to turn on tracing.
+
+    Hint: Treat these macros like an ```assert``` -  expression have to be evaluated to true and no action is taken in that case, otherwise error is returned.
+
+    Do not put complex conditions into these macros it may reduce readability, if you need such conditions use the following approach:
+```
+    bool mem_pattern_valid = (par->IOPattern & MFX_IOPATTERN_OUT_VIDEO_MEMORY) && (par->IOPattern & MFX_IOPATTERN_OUT_SYSTEM_MEMORY));
+    MFX_CHECK(mem_pattern_valid, MFX_ERR_INVALID_VIDEO_PARAM);
+```
+    Do not put function calls inside these macro, it may lead to double call.
```
dvrogozh: Something is wring with formatting here (visible when rendered: https://github.com/mgonchar/MediaSDK/blob/introduce_guidelines/doc/coding_guide/Coding_guidelines.md). Probably you need to add empty lines before/after coding blocks.

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+```
+-   Include guards must be used for every header.
+
+-   Include guards are preferable to ```#pragma once```, since the last one is not standardized. But all popular compilers supports it. Don't mix these two approaches, choose one.
+
+-   Stop (!!) inventing/using maximum/minimum macro. We already have ```IPP_MAX```, ```MFX_MAX```, ```UMC_MAX```, ```Max```. Use ```std::max```/```min``` instead.
+
+7.  Do not overuse conditional compilation.
+
+-   Do not leave dead code under ```#if 0```. Use comments to hide it from compiler (do not forget to verbose your intention in related comment).
```
dvrogozh: you had dead code comment above. move there?

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+
+8.  Use macro when extend public API / functionality.
+
+-   Fix binary layout of newly defined structs with ```MFX_PACK_BEGIN_USUAL_STRUCT``` / ```MFX_PACK_BEGIN_STRUCT_W_PTR``` macro (they are expanded to proper ```#pragma pack``` directives) to guarantee compatibility with different compilers.
```
dvrogozh: Is that related to API support or library code support? If API support maybe it worths absolutely separate section in the document.

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+    : field()
+    {}
+```
+-   Do not use ```memcmp``` for comparison of UDT. Even for PODs compiler may fill padding between fields with random garbage which will lead to incorrect comparison result.
+
+-   ```memset```/```memcopy``` non-POD structs will lead to unpredictable fails.
+
+-   Do not use ```memcpy``` if pointers may overlap, use ```memove``` instead
+
+-   ```std::copy``` is better to use for arrays copying. It is safer, it could be faster, because it could benefit from knowing data block size, ```memcpy``` operates with typeless ```void*```. And it works correctly with overlapping blocks (mind the polarity of overlap, maybe you will need ```std::copy_backward```).
```
dvrogozh: it is not clear that we recommend std::copy/fill whenever possible. It lost in the middle of the section.

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+        int x, y;
+    } Point;
+
+    // C++ way
+    struct
+    {
+        int x, y;
+    } Point;
+```
+-   Be sure to know rule of three and rule of five. See picture below
```
dvrogozh: external link?

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+
+-   Misusage or overusage of references. References are better to use instead of pointers, it allows to skip the pointer check. Some developers consider ```const``` reference as a way to avoid undesired copying of input parameters to functions. The common rule here is that it is very cheap (faster than to dereference a pointer) to copy anything with size less than a cache line (about 64 bytes!).
+
+-   Use prefix ```++i``` operator. Almost all increments which developers use are actually prefix increments. Use postfix ```i++``` only if you need a side effect of postfix ```i++```. This may introduce sufficient difference for some untrivial iterators.
+
+-   Always add ```default:``` in switch operator.
+
+-   If you allocate some buffer during work with C-style strings, do not forget to count terminal ```'\0'``` symbol.
+
+Feel free to improve this document. Bring more examples and rules. But please try to keep size of document small enough. This document is for wide audience of developers, so avoid insufficient details and rare cases.
```
dvrogozh: Good, but... Sometimes it is too wordy. Hence, too big. I barely made it to line ~100 and at line ~130 I almost stopped looking scrolling down hoping that my eye will catch something. Try to be more precise and better structured... Maybe something like:
1. NO dead code. You don't need it - delete it!
     - Do NOT left commented out code like:
```
//if (x > 5) {
//    printf("DEBUG: x=%d", x);
//}
```

      - Do NOT left code under defines which are never used, typically `#if 0`:

```
#if 0
if (x > 5) {
    printf("DEBUG: x=%d", x);
}
#endif 
```

/sorry for bad formatting of the second block/

Explore whether markdown has some Expand sections. In that way you can hide some explanation details of "why?" you claim this requirement.

doc/coding_guide/Coding_guidelines.md
```diff
@@ -0,0 +1,650 @@
...
+
+This document introduces code guide which should highlight some most common issues and bad practices existing in MSDK library's code. This document doesn't cover questions related to comments, code alignment, spaces/tabs and naming conventions. It focuses on code, logic and implementation.
+
+These tips shouldn't be considered as strict rules, they are just good piece of advice on how to implement your solution in modern C++. But please be sure to know recommended way before implementing something different. Violations of these rules must be committed consciously, other ways better solution is most likely to be in these tips.
+
+**General rules.**
+
+1.  Do not overcomplicate your code, every single block which makes some work should be as local as it possible. Person who reads the code should receive minimal context to understand the logic.
+
+-   Keep variables definition local, i.e. as close to usage as possible.
```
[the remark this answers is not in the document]
mgonchar: will do

doc/coding_guide/Coding_guidelines.md
mgonchar: yes, will rephrase

doc/coding_guide/Coding_guidelines.md
mgonchar: tried to fix all the formatting in next patch

doc/coding_guide/Coding_guidelines.md
mgonchar: maybe better to add a cross-link

doc/coding_guide/Coding_guidelines.md
mgonchar: will rephrase

doc/coding_guide/Coding_guidelines.md
mgonchar: will add

doc/coding_guide/Coding_guidelines.md
mgonchar: Yes, I agree that there is still a lot of work to do. Better wording, better structuring, better formatting, more examples... Anyway I think we should proceed with review and merge and apply incremental changes after PR is merged. Because we need to have guidelines document, and I think current state is good enough

PR #1577, opened 2019-08-09
h264dec: update the calculation for DPB size

xhaihao:
Sometimes the DPB size returned from CalculateDPBSize is less than max_num_ref_frames, which will result in corrupted output. In this fix, we use 696320 to calculate DPB size again when level_idc reaches the supported maximium level however dpbSize is still less than num_ref_frames

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
+++ b/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
@@ -674,6 +674,7 @@ private:
 inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t height, uint32_t num_ref_frames)
 {
     uint32_t dpbSize;
+    auto constexpr INTERNAL_MAX_LEVEL = H264VideoDecoderParams::H264_LEVEL_MAX + 1;
 
     num_ref_frames = MFX_MIN(16, num_ref_frames);
 
@@ -722,7 +723,13 @@ inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t heig
             MaxDPBMbs = 184320;
             break;
         default:
-            MaxDPBMbs = 184320; //get as much as we may
+            // We don't support level greater than 5.2 but
+            // relax resolution constrains up to 4K, hence
+            // use value 696320 which is from level 6+ for
+            // the calculation of the DPB size when level_idc
+            // reaches 5.1+ but dpbSize is still less
+            // than num_ref_frames
+            MaxDPBMbs = 696320;
             break;
         }
 
@@ -734,7 +741,7 @@ inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t heig
         uint32_t dpbLevel = MaxDPBMbs*256 / (width * height);
         dpbSize = MFX_MIN(16, dpbLevel);
 
-        if (num_ref_frames <= dpbSize || level_idc == H264VideoDecoderParams::H264_LEVEL_MAX)
+        if (num_ref_frames <= dpbSize || level_idc == INTERNAL_MAX_LEVEL)
             break;
 
         switch (level_idc) // increase level_idc
@@ -771,14 +778,21 @@ inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t heig
         case H264VideoDecoderParams::H264_LEVEL_5:
             level_idc = H264VideoDecoderParams::H264_LEVEL_51;
             break;
+
+        // Set level_idc to INTERNAL_MAX_LEVEL so that MaxDPBMbs = 696320
+        // can be used to calculate the DPB size.
         case H264VideoDecoderParams::H264_LEVEL_51:
         case H264VideoDecoderParams::H264_LEVEL_52:
-            level_idc = H264VideoDecoderParams::H264_LEVEL_MAX;
+            level_idc = INTERNAL_MAX_LEVEL;
             break;
         default:
             throw h264_exception(UMC_ERR_FAILED);
         }
     }
+
+    // Restore level_idc to H264_LEVEL_MAX
+    if (level_idc == INTERNAL_MAX_LEVEL)
+      level_idc = H264VideoDecoderParams::H264_LEVEL_MAX;
 
     return dpbSize;
 }


dmitry-gurulev:
@xhaihao 

Level 6 is not only about max. resolution it also relaxes MVs lengthes. HW has a restriction for max. vertical MV length (MaxVmvR) = 512 samples (level 5.2). If there were MVs which exceed this limit we would have artifacts at output frame.

xhaihao:
@dmitry-gurulev without this fix, we would have artifacts for a 3840x3840 video if the DPB size were greater than 3. So the situation with this fix is better than the situation without this fix. 

E.g. currently gstreamer-vaapi may decode a 3840x3840 AVC video correctly however gst-msdk fails to decode the same video due to the wrong DBP size in MSDK.

dmitryermilov:
> E.g. currently gstreamer-vaapi may decode a 3840x3840 AVC video correctly however gst-msdk fails to decode the same video due to the wrong DBP size in MSDK.

@xhaihao , does gstreamer-vaapi guarantee conformance for any level6 streams?

dmitry-gurulev:
@xhaihao 

Actually, 3840x3840 exceeds 5.2 level limits, so it doesn't matter how much is calculated DPB size. So, such stream must have `level_idc >= 6.0` and it should be interpreted as unsupported by MSDK.

xhaihao:
@dmitry-gurulev I agree with you that 3840x3840 exceeds 5.2 level limits, but we can't guarantee that each element in the stream follows the spec, e.g. the level is 5.2 in my testing. Actually `CalculateDPBSize` can increase `level_idc` to get the right DPB size.

https://github.com/Intel-Media-SDK/MediaSDK/blob/449f49b08cd8a6acd2fdcb3f0e3f667de086dc9a/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h#L740-L780

xhaihao:
@dmitryermilov No, because the media driver supports up to 4096x4096 video for H.264/AVC
https://github.com/intel/media-driver/blob/master/docs/media_features.md#supported-decoding-output-format-and-max-resolution

dmitry-gurulev:
@xhaihao 
> we can't guarantee that each element in the stream follows the spec

In that case (corrupted stream) it is OK that result would be corrupted. And that's not OK if we say that we support level 6+, however we don't. And the problem here is that we even can't report an error - no curruption flags will be set for corrupted frame.

As a WA for such case we may introduce a special mode, like IGNORE_LEVEL_CONSTRAINS at initialization step and **user** may set it if he really wants to ignore  potential issues w/ output.

>the level is 5.2 in my testing

As WA we could **internally** increase DPB size, but don't increase `level_idc`.

>driver supports up to 4096x4096 video for H.264/AVC

This is because HW limitation for **frame resolution** is 4096x4096. But it doesn't imply that level 6+ could be supported.

xhaihao:
@dmitry-gurulev the value of max_num_ref_frames is right in the stream, so other decoder may decode this stream correctly. I updated this patch to return the max value between the calculated value and max_num_ref_frames when level idc reaches the maximum value. Could you please review it again when you get a chance?

lizhong1008:
> As a WA for such case we may introduce a special mode, like IGNORE_LEVEL_CONSTRAINS at initialization step and **user** may set it if he really wants to ignore potential issues w/ output.

I  belive current status are:
1.  Level 6.0+ can't be fully supported by HW. 
2.  Some level 6.0+ clip can be decoding well by HW, and some clips can't.

Based on such status,  I think add an decoding flag such as IGNORE_LEVEL_CONSTRAINS to ingore MB level limation is a good idea (but please don't forget to check some other paramters in frame/slice level 
 limation which deson't support by HW, such as resouliton larger than 4096x4096), and also expose level 6.0+ in MSDK API.

_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
```diff
@@ -734,7 +734,10 @@ inline int32_t CalculateDPBSize(uint8_t & level_idc, int32_t width, int32_t heig
         uint32_t dpbLevel = MaxDPBMbs*256 / (width * height);
         dpbSize = MFX_MIN(16, dpbLevel);
 
-        if (num_ref_frames <= dpbSize || level_idc == H264VideoDecoderParams::H264_LEVEL_MAX)
+        if (level_idc == H264VideoDecoderParams::H264_LEVEL_MAX)
+          dpbSize = MFX_MAX(dpbSize, num_ref_frames);
```
mgonchar: I would ask to use ```std::max``` instead of macro. I know that this approach mimic existing style, but I want to get #925 merged someday, so it's better to not introduce potential conflicts

xhaihao: Thanks for your comment, I updated the patch.

xhaihao:
> I'd rather slightly change `level_idc` switch's default label as
> `default: MaxDPBMbs = 696320; //get as much as we may `

This can't fix the issue for me.
1) the input level is 5.2,  so MaxDPBMbs=184320
2) resolution is 3840x3840, so dpbSize=3
3) `num_ref_frames <= dpbSize` is false however `level_idc == H264VideoDecoderParams::H264_LEVEL_MAX` is true, so it returns dpbSize(3)
4) even I change the condition in 3) so it may increase level idc. Note the supported max level idc is 5.2, hence MaxDPBMbs is still 184320.

In addition, level_idc is a in/out parameter, so we can't set it to a value which is greater than `H264_LEVEL_52`

So changing the default label only doesn't work. If you'd like to change the MaxDPBMbs in the default label, I may do other changes as well. 

> Reasons is - max_num_ref_frames must not exceed MaxDpbFrames but might be not equal to it. In other words, it is possible to have a stream w/ max_num_ref_frames = N but requires DPB size = N + m due to some tricky reordering.

My changing doesn't change the original logic. it will return dpbSize (N+M) if the condition of `dpbSize > num_ref_frames` is true

dmitry-gurulev:
@xhaihao 
> num_ref_frames <= dpbSize is false however level_idc == H264VideoDecoderParams::H264_LEVEL_MAX is true, so it returns dpbSize(3)

OK, got it. That's because we leave  H264_LEVEL_MAX to be equal to H264_LEVEL_52.

> My changing doesn't change the original logic. it will return dpbSize (N+M) if the condition of dpbSize > num_ref_frames is true

Yes, it doesn't, however it has to. It returns `dpbSize` if `dpbSize > num_ref_frames` but calculated `dpbSize` might be incorrect for some streams that actually exceed level 5.2 constrains.

xhaihao:
I updated the patch to use 696320 for DPB size calculation when level_idc reaches the supported maximum level but dpbSize is still less than num_ref_frames. please help to review this patch again.

dmitry-gurulev:
@xhaihao, 

Looks OK, but I'd rather define local constant like
`auto constexpr INTERNAL_MAX_LEVEL = H264VideoDecoderParams::H264_LEVEL_MAX + 1;`
and then use it in place of `level_idc == (H264VideoDecoderParams::H264_LEVEL_MAX + 1)` and others.

Also, it makes sense to add a comment like
`we don't support level greater than 5.2 but relaxes resolution constrains up to 4K`
Moreover, such explanation must be added to documentation, maybe to release notes. @onabiull, what do u think?

PS if we relax constrains only up to 4K, it makes sense to calculate `MaxDPBMbs` accordingly rather then use value  `696320` which is from level 6+

xhaihao:
@dmitry-gurulev thanks for your comment, I updated the patch and added comment about MaxDPBMbs = 696320.

xhaihao:
@dvrogozh @onabiull @varistar Could you please review this patch?  It needs approval from code owner.

IlyaSergeyev:
I think it's a sort of workaround, but till stream will not using profile level 6 features (except 4K) it is safe.

dmitryermilov:
> @IlyaSergeyev , once merged please don't forget to port to close source repo.

@IlyaSergeyev , you forgot :( pity.

PR #1580, opened 2019-08-12
Dev/enable build time tests

dmitry-gurulev:
test: enable build-time unit testing

- move 'googletest' into 'contrib' folder
- move unit tests (mfx_dispatch) into 'tests\unit\suites' folder
- add 'UNIT_TEST_SUITES' target that runs unit tests at build time
  for target 'all'


tests/unit/CMakeLists.txt
```diff
@@ -18,6 +18,17 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-if (BUILD_DISPATCHER)
+if(BUILD_DISPATCHER)
```
dvrogozh: Can you, please, describe in the commit message how this is supposed to work? Please, list cmake configuration command and make command(s) to run tests.

tests/unit/CMakeLists.txt
```diff
@@ -18,6 +18,17 @@
...
+add_custom_target( test_suites
+  ALL
+  DEPENDS ${UNIT_TEST_SUITES}
+)
+
+add_custom_command(TARGET test_suites
+  COMMENT "Run tests"
+  POST_BUILD
+  COMMAND ctest ARGS --output-on-failure
+)
```
dvrogozh: Basically, currently the following scheme works:
```
cmake -DBUILD_TESTS=ON ...
make -j8    # builds all targets including test targets, this is require step for the next command
make test  # run all tests
```
On the output of the last command you will see:
```
# make test
Running tests...
Test project /home/dvrogozh/git/github/msdk/build
    Start 1: run_mfx_dispatch_test
1/1 Test #1: run_mfx_dispatch_test ............   Passed    0.02 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.03 sec
```

What are you trying to achieve different from the above?

I would note that I am against running any tests by default within major make command. I strongly believe this should be a separate optional step.

tests/unit/suites/mfx_dispatch/linux/CMakeLists.txt
```diff
@@ -78,4 +87,7 @@ if(TARGET gmock)
   endif()
 endif()
 
-set_property(TEST run_mfx_dispatch_test PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${LIBRARY_PATH}")
+set_property(TEST run_mfx_dispatch_test
```
dvrogozh: Well, I do not have a preference on this vs. prev. coding style and on the location of the gtest and/or tests themselves. However, moving single tiny test around without adding new stuff hardly makes any sense...

tests/unit/CMakeLists.txt
dmitryermilov: > I would note that I am against running any tests by default within major make command. I strongly believe this should be a separate optional step.

What are objective reasons for it? An optional step means a user may not even know that test exists.

dvrogozh: They will know if we will document it. And package maintainers and those who build for production usually ask this question: how to test whatever I've just built? So, they eventually will get to know that we have tests and how to run them.

Tests are separate logical step. I prefer to keep things separated whenever possible. That's simpler to use and support. At least for me.

When you tight tests to build step, that's a burden to developers in the same way as forcing -Werror by default. It is not unusual to add some code which violates something during debug process. Consider media driver experience we depend on. They almost immediately got an issue to provide an option to disable automated tests from the build.

dmitry-gurulev: @dvrogozh,

>Tests are separate logical step. I prefer to keep things separated whenever possible. That's simpler to use and support. At least for me.

I'm actually agree w/ your point, I believe such testing need to be a separate step and be an essential part of CI. Despite this, now we have a target that we committed to implement - `Enable unit tests for Media SDK on build`.

>What are you trying to achieve different from the above?

The main reason for this feature - since this is automatic step, it ensures that a developer won't forget to run tests. The second, u (as developer) don't need to deep dive into documentation (and many developers won't) to understand which tests we have and how to run them. The third thing, if developer writes a test that significantly slows building process - automatic test running makes him re-work the test.

Anyway, I don't think that this is a real burden because developer can easily do `make mfxhw64` (or any other specific target) which doesn't run tests. Moreover, such `make <certain target>` is quite usual thing during debugging process, in opposite to `make all`  . The second, even if this (build-time test) feature is enabled it still will be an optional step because`BUILD_TESTS` is `OFF` by default. The last - I could introduce an additional `cmake` level option like `ENABLE_RUNNING_BUILD_TIME_TEST`

tests/unit/suites/mfx_dispatch/linux/CMakeLists.txt
dmitry-gurulev: @dvrogozh,

I've tried to somehow align style over that certain file since I've touched it, because currently we have a mess, that was an intention. I will revert these changes If u insist.

dmitry-gurulev:
Since [test: enable build time tests](https://github.com/Intel-Media-SDK/MediaSDK/pull/1580/commits/57aee76d38460ec70e631e556d0fa8ee06e4c102) looks to have different meaning about if it is needed or not, this  PR was split into different PRs to unblock dependent work:

1. [test: move GTest to 'contrib' folder](https://github.com/Intel-Media-SDK/MediaSDK/pull/1582)
2. [test: use 'tests/unit/suites' folder](https://github.com/Intel-Media-SDK/MediaSDK/pull/1583)

dvrogozh:
Is it still relevant?

dmitry-gurulev:
@dvrogozh,

Probably not, but let's wait until EOW.

dmitry-gurulev:
It was decided that having UTs as obligatory CI step plus possibility for manual running (via `make test`) is enough.

Issue #1579, opened 2019-08-11
HEVC encode fails with QVBR rate-control method


ishmael1985:
I am facing issues with HEVC QVBR encoding, reproduced as follows:
```
ishmael@ishmael-ThinkPad-P50:~/msdk_stack/msdk/build/__bin/release$ ./sample_encode h265 -i /home/ishmael/Videos/720p.yuv -o 30_se.h265 -nv12 -w 1280 -h 720 -f 30 -b 2500 -hw -vaapi -n 600 -qvbr 26
libva info: VA-API version 1.6.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'iHD'
libva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_6
libva info: va_openDriver() returns 0
Encoding Sample Version 8.4.27.

Input file format	NV12
Output video		HEVC
Source picture:
	Resolution	1280x720
	Crop X,Y,W,H	0,0,1280,720
Destination picture:
	Resolution	1280x720
	Crop X,Y,W,H	0,0,1280,720
Frame rate	30.00
Bit rate(Kbps)	2500
Gop size	0
Ref dist	0
Ref number	0
Idr Interval	0
Target usage	balanced
Memory type	vaapi
Media SDK impl		hw
Media SDK version	1.30

Processing started

[ERROR], sts=MFX_ERR_ABORTED(-12), SynchronizeFirstTask, SyncOperation failed at /home/ishmael/msdk_stack/msdk/samples/sample_encode/src/pipeline_encode.cpp:157

[ERROR], sts=MFX_ERR_ABORTED(-12), GetFreeTask, m_TaskPool.SynchronizeFirstTask failed at /home/ishmael/msdk_stack/msdk/samples/sample_encode/src/pipeline_encode.cpp:1987

[ERROR], sts=MFX_ERR_ABORTED(-12), Run, m_pmfxENC->EncodeFrameAsync failed at /home/ishmael/msdk_stack/msdk/samples/sample_encode/src/pipeline_encode.cpp:2204

[ERROR], sts=MFX_ERR_ABORTED(-12), main, pPipeline->Run failed at /home/ishmael/msdk_stack/msdk/samples/sample_encode/src/sample_encode.cpp:1587
Frame number: 0
```
[...]

PR #1581, opened 2019-08-13
Fix HEVC QVBR support at least on SKL

onabiull:
Fixes: #1579


ishmael1985:
Confirmed the fix works on my Skylake system as well :)

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -397,7 +397,7 @@ mfxStatus SetRateControl(
         rate_param->ICQ_quality_factor = par.mfx.ICQQuality;
 #ifdef MFX_ENABLE_QVBR
     else if (par.mfx.RateControlMethod == MFX_RATECONTROL_QVBR)
-        rate_param->quality_factor = extOpt3.QVBRQuality;
+        rate_param->ICQ_quality_factor = extOpt3.QVBRQuality;
```
dvrogozh: hmmm. Why not quality_factor? There is a note for the ICQ_quality_factor: "this value will be deprecated in future, to use quality_factor instead of it." Do we need a follow up with the driver team on that?

onabiull: It looks like we promoted this change ahead of driver. There are https://github.com/intel/media-driver/pull/632 which replace `ICQ_quality_factor` with `quality_factor`.

PR #1584, opened 2019-08-15
dispatcher, win: fixed incorrect status check

mgonchar:


api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
```diff
@@ -60,7 +60,7 @@ mfxStatus MFX_DISP_HANDLE::Close(void)
     mfxRes = UnLoadSelectedDLL();
 
     // the library wasn't unloaded
-    if (MFX_ERR_NONE == mfxRes)
+    if (MFX_ERR_NONE != mfxRes)
     {
         implType = MFX_LIB_SOFTWARE;
```
onabiull: Are you sure that this handling would be correct?
We get here only in case of MFX_ERR_UNDEFINED_BEHAVIOR returned either from MFXClose or from mfx_dll_free.

mgonchar: Well, in other case I don't understand what this code is trying to do :)

Actually, this is backmerge from CS aithored by @gkrivor 

Did I get you right, that you suggest just to remove this whole block?

gkrivor: Originally, we discussed it with our Media Expert and decided to correct  behavior.

PR #1585, opened 2019-08-15
sample_decode: improve EOS flag handling

mgonchar:
- Encapsulates logic of EOS flag handling to helper bitstream reader classes
- Fixes problem with possible forever loop in sample_decode

### DIFF
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -622,6 +622,12 @@ mfxStatus CSmplBitstreamReader::Init(const msdk_char *strFileName)
     return MFX_ERR_NONE;
 }
 
+#define CHECK_SET_EOS(pBitstream)                  \
+    if (feof(m_fSource))                           \
+    {                                              \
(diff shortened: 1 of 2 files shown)


samples/sample_common/src/sample_utils.cpp
```diff
@@ -622,18 +622,28 @@ mfxStatus CSmplBitstreamReader::Init(const msdk_char *strFileName)
...
 mfxStatus CSmplBitstreamReader::ReadNextFrame(mfxBitstream *pBS)
 {
     if (!m_bInited)
         return MFX_ERR_NOT_INITIALIZED;
 
     MSDK_CHECK_POINTER(pBS, MFX_ERR_NULL_PTR);
 
-    mfxU32 nBytesRead = 0;
+    // Not enough memory to read new chunk of data
+    if (pBS->MaxLength - pBS->DataLength == 0)
```
dmitryermilov: maybe just "==" ?

mgonchar: yes, maybe

mgonchar: done

PR #1587, opened 2019-08-16
H264dec: reset DPB size if needed.

xhaihao:
A stream may contain multiple (SPS, PPS) parameters at the beginning,
use one (SPS, PPS) as active parameters then swith to another (SPS,
PPS) at next IDR picture. Different (SPS, PPS) pairs may require different
DPB sizes, so resetting DPB size in the view is needed when adding a
IDR slice which requires a new DPB buffer to a view.

Sample clip:
https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/uploads/34603b58fd2e804308d6dc1441b88ec2/JTBC_E72.170422.720p-NEXT_short.mp4

Test command:
gst-launch-1.0 filesrc location=JTBC_E72.170422.720p-NEXT_short.mp4 ! qtdemux \
! h264parse ! msdkh264dec ! glimagesink

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
@@ -1090,7 +1090,25 @@ ViewItem & MVC_Extension::AllocateAndInitializeView(H264Slice * slice)
     }
     ViewItem * view = FindView(slice->GetSliceHeader()->nal_ext.mvc.view_id);
     if (view)
+    {
+        const UMC_H264_DECODER::H264SliceHeader *slice_hdr = slice->GetSliceHeader();
+
+        if (!slice_hdr->IdrPicFlag)
+            return *view;
+
+        H264SeqParamSet *pSps = const_cast<H264SeqParamSet*>(slice->m_pSeqParamSet);
+        int32_t dpbsize = CalculateDPBSize(m_level_idc ? m_level_idc : pSps->level_idc,
+                                           pSps->frame_width_in_mbs * 16,
+                                           pSps->frame_height_in_mbs * 16,
+                                           pSps->num_ref_frames);
+
+        dpbsize = pSps->vui.max_dec_frame_buffering ? pSps->vui.max_dec_frame_buffering : dpbsize;
+
+        if (dpbsize != view->GetDPBList(0)->GetDPBSize())
+            view->SetDPBSize(const_cast<H264SeqParamSet*>(slice->m_pSeqParamSet), m_level_idc);
+
         return *view;
+    }
 
     Status umcRes = AllocateView(slice->GetSliceHeader()->nal_ext.mvc.view_id);
     if (UMC_OK != umcRes)


xhaihao:
@eponomar rebased, thx.

_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
```diff
@@ -1090,7 +1090,25 @@ ViewItem & MVC_Extension::AllocateAndInitializeView(H264Slice * slice)
...
     ViewItem * view = FindView(slice->GetSliceHeader()->nal_ext.mvc.view_id);
     if (view)
+    {
+        const UMC_H264_DECODER::H264SliceHeader *slice_hdr = slice->GetSliceHeader();
+
+        if (!slice_hdr->IdrPicFlag)
+            return *view;
+
+        H264SeqParamSet *pSps = const_cast<H264SeqParamSet*>(slice->m_pSeqParamSet);
+        int32_t dpbsize = CalculateDPBSize(m_level_idc ? m_level_idc : pSps->level_idc,
```
IlyaSergeyev: Why this code block is here, but not in SetDPBSize()?

xhaihao: SetDPBSize are called at different place and I am not sure it needs such change for all.

IlyaSergeyev:
I agree with Dmitry, but we have not a suitable error code for this. What we can use here - UMC_ERR_INVALID_PARAMS or UMC_ERR_NOT_ENOUGH_DATA?

xhaihao:
MSDK doesn't take sps::max_num_ref_frames into account when calculating the suggested number of frames
https://github.com/Intel-Media-SDK/MediaSDK/blob/690341fa0e85adaed1c31c8402cf5776a98fd100/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp#L966

In addition, user application may creates frames more than the suggested number.

xhaihao:
Currently user application gets MFX_WRN_DEVICE_BUSY when calling MFXVideoDECODE_DecodeFrameAsync() for this case, According to https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md, user application can do nothing but call MFXVideoDECODE_DecodeFrameAsync() again.

xhaihao:
I am not familiar with the error handling path in MSDK, so I will close this PR and file a MSDK issue instead if you finally think an error is required for this case.

dmitry-gurulev:
@xhaihao
> doesn't take sps::max_num_ref_frames into account

That's because it doesn't define DPB size that might be bigger. Spec. states only that `max_num_ref_frames` must be **in range** [0, MaxDpbFrames]

@IlyaSergeyev
>but we have not a suitable error code for this

`UMC_NTF_NEW_RESOLUTION` -> `MFX_ERR_INCOMPATIBLE_VIDEO_PARAM`, see https://github.com/Intel-Media-SDK/MediaSDK/blob/690341fa0e85adaed1c31c8402cf5776a98fd100/_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp#L2422

xhaihao:
@dmitry-gurulev  Did you mean check slice header in TaskSupplier::DecodeHeaders? 

BTW all sps/pps units are placed at the beginning in this special clip, so there is no any sps/pps units when calling DecodeHeaders() for the 2nd IDR picture, hence it won't return  UMC_NTF_NEW_RESOLUTION

dmitry-gurulev:
@xhaihao

>all sps/pps units are placed at the beginning

Yes, I see. My note was just about this -
>we have not a suitable error code for this

xhaihao:
> That's because it doesn't define DPB size that might be bigger. Spec. states only that `max_num_ref_frames` must be **in range** [0, MaxDpbFrames]

So in this case the number of the required frame is large enough and user application needn't allocate additional frames, hence an error is not needed.

The root cause for this issue is the internal DPB size is too small for the stream when decoding the 2nd IDR slice because `max_num_ref_frames` is used to limit the DPB size for internal usage. 


> > all sps/pps units are placed at the beginning
> 
> Yes, I see. My note was just about this -

However it is hard for user application to resolve the issue even if MSDK returns MFX_ERR_INCOMPATIBLE_VIDEO_PARAM to user application. According to the sample code (https://github.com/Intel-Media-SDK/MediaSDK/blob/e79dba293985adedcc3d5a021f96c8a8f194622d/samples/sample_decode/src/sample_decode.cpp#L723-L734), user application should reset the decoder, which means closing the current decoder, re-initializing  a new decoder, then decode the 2nd IDR slice with the new decoder again but the bitstream data doesn't contain SPS/PPS units

dmitry-gurulev:
@xhaihao,

>So in this case the number of the required frame is large enough and user application needn't allocate additional frames, hence an error is not needed.

A level, resolution or `VUI :: max_dec_frame_buffering` could be changed by new SPS, therefore DPB size could be changed, thus the error is needed.

>However it is hard for user application to resolve the issue even if MSDK returns MFX_ERR_INCOMPATIBLE_VIDEO_PARAM to user application.

OK, it is. I think we need to investigate and find a common solution, this looks like a gap in current implementation. Maybe VPx like approach w/ `MFX_ERR_REALLOC_SURFACE` is an escape.

Could u please clarify, what is your point here - do u suggest to have WA for this very certain stream?

xhaihao:
> A level, resolution or `VUI :: max_dec_frame_buffering` could be changed by new SPS, therefore DPB size could be changed, thus the error is needed.

`VUI :: max_dec_frame_buffering` is not taken into account when calculating the suggested DPB size in VideoDECODEH264::QueryIOSurfInternal. 

I agree with you that the DBP size could be changed, but for this case, level idc and resolution is not changed when a new SPS/PPS is activated, and the number of the allocated frames is large enough for the whole stream. Maybe we can add more check when decoding the 2nd IDR picture to make sure no additional frames are required from user application. 

> Could u please clarify, what is your point here - do u suggest to have WA for this very certain stream?

Yes, Resolve it in MSDK because MSDK have all information.

PR #1588, opened 2019-08-16
Fix of HEVC encoded picture width/height update

ndomnina:
In case of joint use of HEVC SAO and VPP rotate, parameters PicWidthInLumaSamples
and PicHeightInLumaSamples are not updated for the changed resolution, so we
get MFX_ERR_INVALID_VIDEO_PARAM. Patch fixes this.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
@@ -653,10 +653,15 @@ namespace ExtBuffer
 
             return false;
         }
-        if (buf.PicWidthInLumaSamples == 0)
-            buf.PicWidthInLumaSamples  = align2_value(par.mfx.FrameInfo.CropW > 0 ? (mfxU16)(par.mfx.FrameInfo.CropW + par.mfx.FrameInfo.CropX)  : par.mfx.FrameInfo.Width,  codedPicAlignment);
 
-        if (buf.PicHeightInLumaSamples == 0)
+        bool needUpdate = false;
+        if (buf.PicWidthInLumaSamples > par.mfx.FrameInfo.Width || buf.PicHeightInLumaSamples > par.mfx.FrameInfo.Height)
+            needUpdate = true;
+
+        if (buf.PicWidthInLumaSamples == 0 || needUpdate)
+            buf.PicWidthInLumaSamples = align2_value(par.mfx.FrameInfo.CropW > 0 ? (mfxU16)(par.mfx.FrameInfo.CropW + par.mfx.FrameInfo.CropX)  : par.mfx.FrameInfo.Width,  codedPicAlignment);
+
+        if (buf.PicHeightInLumaSamples == 0 || needUpdate)
             buf.PicHeightInLumaSamples = align2_value(par.mfx.FrameInfo.CropH > 0 ? (mfxU16)(par.mfx.FrameInfo.CropH + par.mfx.FrameInfo.CropY)  : par.mfx.FrameInfo.Height, codedPicAlignment);
 
         return true;


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -653,11 +653,9 @@ namespace ExtBuffer
 
             return false;
         }
-        if (buf.PicWidthInLumaSamples == 0)
-            buf.PicWidthInLumaSamples  = align2_value(par.mfx.FrameInfo.CropW > 0 ? (mfxU16)(par.mfx.FrameInfo.CropW + par.mfx.FrameInfo.CropX)  : par.mfx.FrameInfo.Width,  codedPicAlignment);
 
-        if (buf.PicHeightInLumaSamples == 0)
-            buf.PicHeightInLumaSamples = align2_value(par.mfx.FrameInfo.CropH > 0 ? (mfxU16)(par.mfx.FrameInfo.CropH + par.mfx.FrameInfo.CropY)  : par.mfx.FrameInfo.Height, codedPicAlignment);
+        buf.PicWidthInLumaSamples  = align2_value(par.mfx.FrameInfo.CropW > 0 ? (mfxU16)(par.mfx.FrameInfo.CropW + par.mfx.FrameInfo.CropX)  : par.mfx.FrameInfo.Width,  codedPicAlignment);
```
mgonchar: Are sure that this is correct fix? In my understanding you will overwrite data provided by user in buffer ```mfxExtHEVCParam```, dig into line 649 for details

ndomnina: Yes, indeed. I fixed this, now PicHeightInLumaSamples and PicWidthInLumaSamples will be updated only when they contain incorrect data.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -653,10 +653,15 @@ namespace ExtBuffer
 
             return false;
         }
-        if (buf.PicWidthInLumaSamples == 0)
-            buf.PicWidthInLumaSamples  = align2_value(par.mfx.FrameInfo.CropW > 0 ? (mfxU16)(par.mfx.FrameInfo.CropW + par.mfx.FrameInfo.CropX)  : par.mfx.FrameInfo.Width,  codedPicAlignment);
 
-        if (buf.PicHeightInLumaSamples == 0)
+        bool needUpdate = false;
+        if (buf.PicWidthInLumaSamples > par.mfx.FrameInfo.Width || buf.PicHeightInLumaSamples > par.mfx.FrameInfo.Height)
```
mgonchar: Ok, so user requested size bigger than in initialization, right? If so, we should return with error, or it turns out that we silently ignore buffer passed by user... Or why should we update ```PicWidthInLumaSamples``` but not ```mfx.FrameInfo.Width```. Still unclear for me

ndomnina: mfx.FrameInfo.Width is updated correctly when rotated, while PicWidthInLumaSamples has initial value and we get an error in CheckVideoParam(). Having the initial resolution of the stream 1920x1080, we get 1088x1080 after this function instead of 1088x1920, because the correct value of par.mfx.FrameInfo.Height is compared to the incorrect value of PicHeightInLumaSamples, and then par.mfx.FrameInfo.Height equals PicHeightInLumaSamples.

mgonchar: well, if so I believe you need to update values at the same time when ```mfx.FrameInfo.Width```. Where in code this adjustment happens?

ndomnina: This happens in the app. You're right, it's better to update it there. I'll close this pull request.

Issue #836, opened 2018-10-29
What is minimal required gtest/gmock version?


dvrogozh:
I was trying to build mediasdk on CentOS 7 with:
`cmake -DBUILD_TESTS=ON -DRUN_TESTS=ON`
And used system available versions of gtest/gmock. I did not get any configuration issues, but got compile issues:
```
In file included from /usr/include/gmock/gmock-generated-function-mockers.h:43:0,
                 from /usr/include/gmock/gmock.h:61,
                 from /home/dvrogozh/git/github/MediaSDK/tests/mfx_dispatch/linux/mfx_dispatch_test_mocks.h:26,
                 from /home/dvrogozh/git/github/MediaSDK/tests/mfx_dispatch/linux/mfx_dispatch_test_main.h:25,
                 from /home/dvrogozh/git/github/MediaSDK/tests/mfx_dispatch/linux/mfx_dispatch_test_cases_plugins.cpp:21:
/home/dvrogozh/git/github/MediaSDK/tests/mfx_dispatch/linux/mfx_dispatch_test_fixtures.h: In member function ‘void DispatcherPluginsTest::SetupForGoodLibInit(MockCallObj&)’:
/home/dvrogozh/git/github/MediaSDK/tests/mfx_dispatch/linux/mfx_dispatch_test_fixtures.h:91:9: error: invalid use of member function ‘testing::internal::MockSpec<void*(const char*, int)>& MockCallObj::gmock_dlopen(const testing::Matcher<const char*>&, const testing::Matcher<int>&)’ (did you forget the ‘()’ ?)
         EXPECT_CALL(mock, dlopen).Times(AtLeast(1)).WillRepeatedly(Return(MOCK_DLOPEN_HANDLE));
         ^
/home/dvrogozh/git/github/MediaSDK/tests/mfx_dispatch/linux/mfx_dispatch_test_fixtures.h:92:9: error: invalid use of member function ‘testing::internal::MockSpec<void*(void*, const char*)>& MockCallObj::gmock_dlsym(const testing::Matcher<void*>&, const testing::Matcher<const char*>&)’ (did you forget the ‘()’ ?)
         EXPECT_CALL(mock, dlsym).Times(AtLeast(1)).WillRepeatedly(Invoke(&mock, &MockCallObj::EmulateAPI));
         ^
[...]

PR #1589, opened 2019-08-20
test: fix GTest CI issues

dmitry-gurulev:
Having different projects (vcxproj) w/ the same project GUID produces issues at CI. Old projects (2005/2010) were dropped to resolve ambiguity.


dmitry-gurulev:
@dvrogozh,

1. I've used exactly the same copy as we have at GitHub
2. My point is certainly same - we have to use 3d party dependencies as-is without any modifications. 

@gkrivor, could u please explain current CI issues w/ googletest to Dmitry R.?

mgonchar:
@dvrogozh I agree with you here. @gkrivor I understand your point, but this is third-party projects. We only mess things up and trying to fix somebody else's code. I believe we should completely whitelist and ignore non-critical problem in third-party code. That would be much easier for us just to keep clean copy of external repo. At least for this problem we can limit rules related to vcxproj files only for our codebase

gkrivor:
Let's go from other point of view. How you will use existing solutions in our code? It requires to upgrade compiler version/windows sdk version.

In this case I have a suggestion: mandatory forbid any changes in any thirdparty code.

Then I suggest @dmitry-gurulev to do not use projects/solutions/cmake/non-code-things from this folder and then we may add exclusion rule for whole folder: projects will be ignored, but usage of projects will be also forbidden in repo. In case of copying one of existing project - it GUID should be regenerated, because it will be new project/solution.

Is it looks like a solution?

mgonchar:
@gkrivor well, absence of VS 2017/2019 solution may be considered as a problem. But again, let's treat third-party code as not our's code, i.e. we do not fix this until it is not critical / security bugs and deleting some project files looks like unnecessary action.

gkrivor:
@mgonchar, current CI and my requirement - do not introduce any duplication/misunderstandings. How it will be achieved - we have many choices. One of that - removing unused entities.

Currently, we are blocking @dmitry-gurulev work by no clear decision about 3rd party code.

I can help with fixing CI in case we will not introduce duplication. If we will not use "meta"-objects from 3rd party - I don't have any concerns with adding exceptions. In case projects will be used in our solutions - I don't see any difference between upgrading projects/regenerating GUIDs and removing projects.

dvrogozh:
@gkrivor : contrib/googletest is a 3rd party code taken as-is for convenience. It is taken as-is to simplify its support: basically we are not responsible for this code and are not going to modify it, we just intend to use it. If update will be needed it will be taken in full. 

I don't understand your talk about support of VS 2005. If mediasdk on Windows does not support it any more and you want to drop that support - that's totally ok, do that if that's your intent. But how this relates to gtest copy?! Just don't use VS 2005 project from gtrest and use more recent one, like https://github.com/Intel-Media-SDK/MediaSDK/tree/master/contrib/googletest/googlemock/msvc/2015. We do NOT advertise using and installing of our own gtest copy. Instead people should go to google original repository for that. This copy is for mediasdk use only for self-testing and nothing more.

This is not a problem which actually needs any solving since there is no problem.

gkrivor:
@dvrogozh, please forget about support. It is side-effect. Original issue - duplicated GUIDs in projects. It should be solved.

As I mentioned before - one of way is removing unused projects.

I heard only one motivation - we don't want to touch 3rd party components. If we will use any of existing projects in our solution - we have to upgrade them, this action will change the files. What's the difference with removing? No difference.

I suggested alternate solution - mandatory rule to do not change anything in 3rd party (even folder structure) - it is reasonable. In this case - @dmitry-gurulev may copy project to our codebase and regenerate GUID.

Do you agree with such approach? Do you have any other solutions?

dvrogozh:
You are saying that GUIDs are duplicated. Can you, please, point me to what is actually duplicated? Please, provide links to project files and post GUID here.

gkrivor:
@dvrogozh,

https://github.com/Intel-Media-SDK/MediaSDK/blob/48a4add341ba7e91dabcb5796aa01707971d167b/contrib/googletest/googlemock/msvc/2015/gmock.vcxproj#L22
https://github.com/Intel-Media-SDK/MediaSDK/blob/48a4add341ba7e91dabcb5796aa01707971d167b/contrib/googletest/googlemock/msvc/2005/gmock.vcproj#L6
https://github.com/Intel-Media-SDK/MediaSDK/blob/48a4add341ba7e91dabcb5796aa01707971d167b/contrib/googletest/googlemock/msvc/2010/gmock.vcxproj#L22

dvrogozh:
ok. And how/why this managed to affect CI?

Where is the solution which is being built? Do we managed to include few different gtest project files in the same solution?

I honestly don't understand why this spark any problem. My understanding is that we just use only one of these projects not all 3 at once.

gkrivor:
We have "CI check for unique GUID's" in repos. This requirement is broken by googletest.

I don't want to block Dmitry's work. So I vote for a strong rule. If rule have any exceptions - then this PR might be merged as exception. If not - let's work with that as "totally untouchable".

Anyway, do you plan "sometimes" change forked 3rd party source, "if needed"? If yes, why we cannot change it now, due to our requirements to not have duplicates?

If you don't want to remove files - I may suggest to regenerate GUIDs, another way we will still have a conflict because entities will have duplicates in OpenSource repo.

If we still have some misunderstanding of this situation - I would like suggest to make a call/meeting.

dvrogozh:
From the development stand point I don't see an issue having the same GUID for the mutually excluding projects. That's exactly the case. No one will ever use 2 gtest VS projects targeting different VS versions in the same solution. I understand that the intent of the CI check was to make sure we don't reuse same guid for different not mutually excluding projects within mediasdk, i.e. that we did not copy project and forgot to change guid. However, I think that we will die if we will start to apply our CI checks for the external projects and address such failures patching external projects. I think that we might need to patch external project only if there is runtime or build failure which we can't fix with the project update to the most recent version. In this case we will need to report bug against this project and might temporary apply our own fix.

So, in this particular case my recommendation is to avoid merging in this PR and request update of CI check to exclude contrib/googletest folder from it.

gkrivor:
@dvrogozh, we shouldn't apply CI rules for external repos. I agree.

I really appreciate you keep "your garden" in order and you do it awesome. We are trying do similar things.

And I would like support your point of "read-only 3rd party code", but I don't understand: it should be really read-only, or it should be read-only by mood.

Actually, this change was created as a workaround for fail from existing CI check. You had duplicated googletest. And all of you missed/skipped it, but CI check has found it. But I see disagreement with that.

GUID is Globally Unique IDentifier. Globally. Unique. Project**GUID**. Represents non-identical entities. And here is an issue in original googletest - identifiers are not unique. Isn't it? My original suggestion was: regenerate GUID's, if you still want to keep duplication of code.

But if you decided to have more gogletests in repos - I would like keep an order in projects/solutions while we don't have better check. To be able understand: which project should be used in selected solution. To automatically prevent a blind deleting/renaming/moving projects/solutions and to prevent appearing a lost projects/solutions.

And as I said face2face to @dmitry-gurulev: this is partial solution, anyway we need to regenerate GUID's in remained folder, otherwise CI check will fail. Because last ones weren't changed: as I understand they will be changed in a future commit.

@dmitry-gurulev is waiting for final solution. You don't want to change "3rd party code". We don't want to keep duplicates.

@dvrogozh , @mgonchar could you please think about solution from this point?

If it will be solved by removing duplicates and will be regenerated only for one folder - it will be applicable for me. I don't think about this code as "3rd party" because it was copied into our repo and became part of it. And we have to fix existing issue with non-unique GUID's. You agree to fix existing issues even in 3rd party code if it won't be fixed in original source. It is an issue. But it is not applicable for you because change... is too big? Am I right?

In this case **I suggest** to make a smaller change: regenerate GUID's. Change in source will be much smaller and will be easy traceable in case of any update.

Even we will close our eyes on duplication, we will need to solve requirement with updating solutions to new version. How we will solve it? Inventing yet another workaround to keep "3rd party code" formally non-changed? Why did we forked it in this case? To introduce some new requirements for ourselves? Or to be able to change it, if needed?

I only ask to not block our colleague by long discussions. If folder should be excluded - ok, but go til the end: provide alternate solution to automatically prevent duplications and allow easy moving/linking/renaming.

Let's sync if needed.

dvrogozh:
I do not try to block anyone and/or keep a long discussion. But I see a PR which needs a review and I see that I don't understand the nature of the problem neither from the commit message nor from the existing comments. Hence I ask questions. This is a review process. I am trying to help find the best solution.

If we will consider support of contrib/googletest from the perspective that it should be unchanged upstream copy of googletest, then it seems that there was some step out from this idiom. I mean that there were some other change(s) which already got merged into the repo: when I look into upstream googletest I simply don't see any VS projects, only bare cmake - see https://github.com/google/googletest/tree/master/googlemock, there is no msvc folder as we have here: https://github.com/Intel-Media-SDK/MediaSDK/tree/master/contrib/googletest/googlemock/msvc

**How this did happen and when (give link to closed PR, please)?** It seems that I missed some earlier PR otherwise this issue would spark earlier.

And if I understand correctly, gtest is intended to have only cmake build system both on Windows and Linux. If so, **we need to consider how we build gtest on Windows for the mediasdk needs and where we place VS projects if we pursue this way?** From my perspective they need to be outside of contrib/googletest. Maybe just directly in contrib.

dmitry-gurulev:
@gkrivor,
>Sorry, I have to recall my approval because of updated information from @dvrogozh.

So, u suggest me to wait me until we switch to [googletest](url)'s master branch?

@dvrogozh,
Switching to master would be the best option, but AFAIR we're using of googletest's release 1.8.1 for some reasons, see
https://github.com/Intel-Media-SDK/MediaSDK/pull/842
https://github.com/Intel-Media-SDK/MediaSDK/pull/856
https://github.com/Intel-Media-SDK/MediaSDK/issues/836

mgonchar:
@gkrivor as for me, I just vote for completely read only third-party (the only exception is security/critical bugfixes, which are not merged to third-party projects yet for some reason)

gkrivor:
@dvrogozh @mgonchar Could we review two approaches?
1. remove msvc/* folders as part of preparing to moving on new version/master (it already removed in googletest master)
2. update googletest to the master

dvrogozh:
master branch might be unstable. I strongly suggest to take latest stable release and document tag name from which we've taken code.

Here, on GitHub I don't see any CI checks over GUID which fail...

Anyhow, do we have any VS projects in the current code base on the GitHub (already merged) which depend on gtest? In other words, do we have at least 1 gtest based test which can be built on Windows?

PR #1598, opened 2019-08-27
[SMT] Add different adapter types support

romka0017:


romka0017:
> I don't see change in help section. New parameters should be listed in help print

Thank you. Done.

Issue #1597, opened 2019-08-26
Convert all manuals to web-friendly format (like .md)


dvrogozh:
@onabiull : thank you for converting major mediasdk manual into "developer readable format" :) https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md. When we can expect other docs converted from .pdf?

PR #1599, opened 2019-08-27
Convert manuals to markdown

onabiull:
Fixes: #1597


onabiull:
HEVC_FEI_overview most likely would remain as it is for now. Not sure, that converting it worth investments. And unlike other manuals it more resembles finalized article. 
I'll check out asg-hevc-man tomorrow.

PR #1604, opened 2019-08-29
[mpeg2d] Fix for a stream with invalid temporal_reference count

IlyaSergeyev:
Correct submitting of first I-frame top part

testcase: tractor_422_1920x1080i_50f.mpg

### DIFF
--- a/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
+++ b/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
@@ -848,12 +848,20 @@ namespace UMC_MPEG2_DECODER
         return frame;
     }
 
-    static
-    bool IsFieldOfOneFrame(const MPEG2PictureHeader& picHdr, const MPEG2PictureCodingExtension& picExtHdr,
-                           const MPEG2PictureHeader & newPictureHdr, const MPEG2PictureCodingExtension& newPicExtHdr)
+    bool MPEG2Decoder::IsFieldOfCurrentFrame() const
     {
-        if (picHdr.temporal_reference != newPictureHdr.temporal_reference)  // 6.3.9
-            return false;
+        const auto firstFrameSlice = m_currFrame->GetAU(0)->GetSlice(0);
+        const auto picHdr = firstFrameSlice->GetPicHeader();
+        const auto picExtHdr = firstFrameSlice->GetPicExtHeader();
+        const auto newPicHdr = *m_currHeaders.picHdr.get();
+        const auto newPicExtHdr = *m_currHeaders.picExtHdr.get();
+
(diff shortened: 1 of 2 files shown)


_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -874,12 +883,9 @@ namespace UMC_MPEG2_DECODER
             return true; // Full frame
         }
 
-        // field picture
-        const auto firstFrameSlice = m_currFrame->GetAU(0)->GetSlice(0);
-        const auto newPicHdr = *m_currHeaders.picHdr.get();
         const auto newPicExtHdr = *m_currHeaders.picExtHdr.get();
 
-        if (IsFieldOfOneFrame(firstFrameSlice->GetPicHeader(), firstFrameSlice->GetPicExtHeader(), newPicHdr, newPicExtHdr))
+        if (IsFieldOfSameFrame())
```
dmitryermilov: I know I previously wrote that member function is ok. Thinking more, now I can't say so. "Is field of one frame" - what field and what frame? Previously it was clear from arg list

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -848,12 +848,21 @@ namespace UMC_MPEG2_DECODER
...
     {
-        if (picHdr.temporal_reference != newPictureHdr.temporal_reference)  // 6.3.9
-            return false;
+        const auto firstFrameSlice = m_currFrame->GetAU(0)->GetSlice(0);
+        const auto picHdr = firstFrameSlice->GetPicHeader();
+        const auto picExtHdr = firstFrameSlice->GetPicExtHeader();
+        const auto newPicHdr = *m_currHeaders.picHdr.get();
+        const auto newPicExtHdr = *m_currHeaders.picExtHdr.get();
+
+        // for I-frame we don't need to check if temporal refs are eq
```
dmitryermilov: need to highlighting that this is WA.  "for I-frame we don't need to check" - this is wrong, we need to check!

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
IlyaSergeyev: Maybe we need to use a different name?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
IlyaSergeyev: ok

PR #1606, opened 2019-08-29
samples: do not abuse public API to reach Wayland device

mgonchar:
Remove custom defined ```mfxHandleType``` for Wayland support: ```HANDLE_WAYLAND_DRIVER```

Use direct API instead

### DIFF
--- a/samples/sample_common/include/vaapi_device.h
+++ b/samples/sample_common/include/vaapi_device.h
@@ -130,8 +130,6 @@ private:
 
 class Wayland;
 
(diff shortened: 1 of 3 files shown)


mgonchar:
@dvrogozh no, I missed that in SMT ```hwdev``` is a smart pointer. Now should work

dvrogozh:
Looks good to me, but can anyone try this out?

alariono:
Can somebody run sample_decode with "-rwld" on Wayland?
I would like to make sure what all works smoothly

PR #1609, opened 2019-08-30
[HEVCe][SW BRC] Set 2 references by default in the field case

pgribov:
Two references is used for second field by default


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -3657,13 +3657,13 @@ void ConfigureTask(
     if (isB)
     {
         mfxI32 layer = par.isBPyramid() ? clamp<mfxI32>(task.m_level - 1, 0, 7) : 0;
-        task.m_numRefActive[0] = (mfxU8)CO3.NumRefActiveBL0[layer];
+        task.m_numRefActive[0] = Max<mfxU8>((mfxU8)CO3.NumRefActiveBL0[layer], (task.m_secondField ? 2 : 1));
```
mgonchar: Please, do not use custom ```max```/```min```, switch to ```std::max``` instead, [see](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/Coding_guidelines.md#macro) coding guide

PR #1619, opened 2019-09-03
[mpeg2d] Invalid slices handling

asmirno1:
1) Handling case when vertical_position of slice is out of range
2) Recalculating numberMBsInSlice of prev slice if two
slices are located at the same row

### DIFF
--- a/_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
+++ b/_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
@@ -187,6 +187,7 @@ namespace UMC_MPEG2_DECODER
 
         // Action on new picture
         virtual bool OnNewPicture(); // returns true on full frame
+        virtual void EliminateSliceErrors(MPEG2DecoderFrame& frame, uint8_t fieldIndex);
         virtual UMC::Status CompletePicture(MPEG2DecoderFrame& frame, uint8_t fieldIndex);
         bool IsFieldOfCurrentFrame() const;
 
--- a/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
+++ b/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2019 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -19,6 +19,7 @@
 // SOFTWARE.
 
 #include "umc_defs.h"
+#include "mfx_utils.h"
 #if defined (MFX_ENABLE_MPEG2_VIDEO_DECODE)
 
 #include "umc_mpeg2_defs.h"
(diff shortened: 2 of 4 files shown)


asmirno1:
> I don't see any call of EliminateSliceErrors function ...

Missed this part. Fixed.

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -187,6 +187,7 @@ namespace UMC_MPEG2_DECODER
 
         // Action on new picture
         virtual bool OnNewPicture(); // returns true on full frame
+        virtual UMC::Status EliminateSliceErrors(MPEG2DecoderFrame& frame, uint8_t fieldIndex);
```
dmitryermilov: Return status doesn't seem to be needed here

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
```diff
@@ -295,6 +296,15 @@ namespace UMC_MPEG2_DECODER
         if (CheckBSLeft())
             throw mpeg2_exception(UMC::UMC_ERR_INVALID_STREAM);
 
+        uint32_t width_in_MBs = (uint32_t)(mfx::align2_value<uint32_t>(slice.GetSeqHeader().horizontal_size_value, 16) / 16);
```
mgonchar: I think explicit cast is not needed, just use unsigned literals:
```
uint32_t width_in_MBs = mfx::align2_value<uint32_t>(slice.GetSeqHeader().horizontal_size_value, 16) / 16u;
```

asmirno1: Done

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -903,6 +903,32 @@ namespace UMC_MPEG2_DECODER
...
+            auto slice = frameInfo.GetSlice(sliceNum);
+            auto sliceHeader = slice->GetSliceHeader();
+
+            // Slice may cover just a part of a row (not full row), this is indicated by macroblockAddressIncrement
+            // Looping over slices, we heed to check if next slice has macroblockAddressIncrement > 0,
+            // this means that we need to recalculate numberMBsInSlice for previous slice
+            if (sliceNum > 0 && sliceHeader.macroblockAddressIncrement > 0)
+            {
+                auto prevSlice = frameInfo.GetSlice(sliceNum - 1); //take previous
+                auto& prevSliceHeader = prevSlice->GetSliceHeader();
```
asmirno1: Added "&" in new patch set. Previously, without it, a copy of prevSliceHeader was created, but we need to change prevSliceHeader.numberMBsInSlice value in line 926. So now it is changing correctly.

uartie:
Does the buildbot/test include mpeg2 decode cases?  Because this patchset breaks every mpeg2 decode case in https://github.com/intel/vaapi-fits.

uartie:
The regression caused by this patch is clearly visible with the following command:
```
$ GST_VAAPI_ALL_DRIVERS=1 LIBVA_DRIVER_NAME=iHD gst-launch-1.0 -vf \
  videotestsrc horizontal-speed=2 ! msdkmpeg2enc ! msdkmpeg2dec ! vaapisink
```

uartie:
artifacts

uartie:
regression

dmitryermilov:
Thank you. And thanks again for reporting it so fast:)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_bitstream.cpp
```diff
@@ -295,6 +296,15 @@ namespace UMC_MPEG2_DECODER
         if (CheckBSLeft())
             throw mpeg2_exception(UMC::UMC_ERR_INVALID_STREAM);
 
+        uint32_t width_in_MBs  = mfx::align2_value<uint32_t>(slice.GetSeqHeader().horizontal_size_value, 16) / 16u;
+        uint32_t height_in_MBs = mfx::align2_value<uint32_t>(slice.GetSeqHeader().vertical_size_value, 16) / 16u;
+
+        // invalid slice - skipping it
+        if (slice.sliceHeader.slice_vertical_position >= height_in_MBs)
```
uartie: Should this be `if (slice.sliceHeader.slice_vertical_position > height_in_MBs)` instead?  It seems to get rid of the regression I reported.

asmirno1: Yep, I confirm that's a bug.

PR #1623, opened 2019-09-04
[samples] Added encoding in encoded order

pgribov:
Added encoding in encoded order for sample_encode.
In this case encoder takes frame info for each frame from external text file.

### DIFF
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
     return MFX_ERR_NONE;
 }
 
+QPFileReader::QPFileReader()
+{
+    m_nframes = 0;
+    m_cur_frame_num = 0;
+}
+
+QPFileReaderStatus QPFileReader::Read(const msdk_string &strFileName, mfxU32 codecid)
+{
+    msdk_fstream ifs;
+    ifs.open(strFileName, msdk_fstream::in);
+
+    if (ifs.is_open())
+    {
+        QPFileReaderFrameInfo frame_info;
+        msdk_string frame_type{};
+
+        if (!(ifs >> m_nframes))
+            return QPFILEREADER_WRONG_FILE;
+
+        m_frame_info.reserve(m_nframes);
+
+        if (codecid == MFX_CODEC_AVC)
+        {
+            while (ifs >> frame_info.display_order_num && ifs >> frame_info.QP && ifs >> frame_type)
+            {
+                if (frame_type == MSDK_STRING("IDR_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_I | MFX_FRAMETYPE_IDR | MFX_FRAMETYPE_REF;
+                else if (frame_type == MSDK_STRING("P_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_P | MFX_FRAMETYPE_REF;
+                else if (frame_type == MSDK_STRING("B"))
+                    frame_info.frame_type = MFX_FRAMETYPE_B;
+                else if (frame_type == MSDK_STRING("B_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_B | MFX_FRAMETYPE_REF;
+                else
+                    return QPFILEREADER_WRONG_FILE;
+                m_frame_info.push_back(frame_info);
+            }
+        }
+        else if (codecid == MFX_CODEC_HEVC)
+        {
+            while (ifs >> frame_info.display_order_num && ifs >> frame_info.QP && ifs >> frame_type)
+            {
+                if (frame_type == MSDK_STRING("IDR_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_IDR | MFX_FRAMETYPE_I | MFX_FRAMETYPE_REF;
+                else if (frame_type == MSDK_STRING("I_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_I | MFX_FRAMETYPE_REF;
+                else if (frame_type == MSDK_STRING("P_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_P | MFX_FRAMETYPE_REF;
+                else if (frame_type == MSDK_STRING("P"))
+                    frame_info.frame_type = MFX_FRAMETYPE_P;
+                else if (frame_type == MSDK_STRING("B_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_B | MFX_FRAMETYPE_REF;
+                else if (frame_type == MSDK_STRING("B"))
+                    frame_info.frame_type = MFX_FRAMETYPE_B;
+                else
+                    return QPFILEREADER_WRONG_FILE;
+                m_frame_info.push_back(frame_info);
+            }
+        }
+        else return QPFILEREADER_CODEC_UNSUPPORTED;
+
+        if (m_frame_info.size() < m_nframes)
+            return QPFILEREADER_WRONG_FILE;
+
+        for (const QPFileReaderFrameInfo &finfo : m_frame_info)
+            if (finfo.QP < 1 || finfo.QP > 51)
+                return QPFILEREADER_QP_OUT_OF_RANGE;
+    }
+    else return QPFILEREADER_FILE_NOT_OPEN;
+
+    return QPFILEREADER_ERR_NONE;
+}
+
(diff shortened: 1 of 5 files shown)


samples/sample_common/src/sample_utils.cpp
```diff
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
     return MFX_ERR_NONE;
 }
 
+QPFileReader::QPFileReader()
```
mgonchar: I'd better use in place defaults instead of explicit trivial constructor
```
private:
    mfxU32                                  m_nframes       = 0;
    mfxU32                                  m_cur_frame_num = 0;
    std::vector<QPFileReaderFrameInfo>      m_frame_info;
```

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+
+QPFileReaderStatus QPFileReader::Read(const msdk_string &strFileName, mfxU32 codecid)
+{
+    msdk_fstream ifs;
+    ifs.open(strFileName, msdk_fstream::in);
+
+    if (ifs.is_open())
+    {
+        QPFileReaderFrameInfo frame_info;
+        msdk_string frame_type{};
```
mgonchar: can we use ```std::string``` instead?

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+        if (!(ifs >> m_nframes))
+            return QPFILEREADER_WRONG_FILE;
+
+        m_frame_info.reserve(m_nframes);
+
+        if (codecid == MFX_CODEC_AVC)
+        {
+            while (ifs >> frame_info.display_order_num && ifs >> frame_info.QP && ifs >> frame_type)
+            {
+                if (frame_type == MSDK_STRING("IDR_REF"))
```
mgonchar: this should be separate function:
```mfxU16 QPFileReader::StringToFrameType```

File: samples/sample_common/src/sample_utils.cpp [hunk 4]
```diff
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
 
+QPFileReader::QPFileReader()
+{
+    m_nframes = 0;
+    m_cur_frame_num = 0;
+}
+
+QPFileReaderStatus QPFileReader::Read(const msdk_string &strFileName, mfxU32 codecid)
+{
+    msdk_fstream ifs;
```
mgonchar: Please, switch to ```std``` streams

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+                else if (frame_type == MSDK_STRING("B"))
+                    frame_info.frame_type = MFX_FRAMETYPE_B;
+                else if (frame_type == MSDK_STRING("B_REF"))
+                    frame_info.frame_type = MFX_FRAMETYPE_B | MFX_FRAMETYPE_REF;
+                else
+                    return QPFILEREADER_WRONG_FILE;
+                m_frame_info.push_back(frame_info);
+            }
+        }
+        else if (codecid == MFX_CODEC_HEVC)
```
mgonchar: why there is such difference with AVC case?

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1213,6 +1213,109 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+                m_frame_info.push_back(frame_info);
+            }
+        }
+        else return QPFILEREADER_CODEC_UNSUPPORTED;
+
+        if (m_frame_info.size() < m_nframes)
+            return QPFILEREADER_WRONG_FILE;
+
+        for (const QPFileReaderFrameInfo &finfo : m_frame_info)
+            if (finfo.QP < 1 || finfo.QP > 51)
```
mgonchar: I'd better to check it immidiately after reading

mgonchar: what about QP == 0 handling? in HEVC standard it is valid value... We had some problems with handling it internally. Anyway, need to clarify this

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1310,6 +1413,69 @@ mfxU16 GetFreeSurface(mfxFrameSurface1* pSurfacesPool, mfxU16 nPoolSize)
     return idx;
 }
 
+mfxU16 GetFreeSurfaceIndexEO(mfxFrameSurface1 *pSurfacesPool, mfxU16 nPoolSize)
+{
+    if (pSurfacesPool)
+    {
+        for (mfxU16 i = 0; i < nPoolSize; i++)
```
mgonchar: this is ```std::find_if```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -462,7 +462,44 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+
+        QPFileReaderStatus sts = m_QPFileReader.Read(pInParams->strQPFilePath, pInParams->CodecId);
+
+        if (QPFILEREADER_ERR_NONE != sts)
+        {
+            if (QPFILEREADER_FILE_NOT_OPEN == sts)
+            {
+                msdk_printf(MSDK_STRING("ERROR: failed to open file contains frame parameters\n"));
+            }
+            else if (QPFILEREADER_WRONG_FILE == sts)
```
mgonchar: Please, do not write giant ```if-else``` blocks like this, this is perfectly managed by ```switch```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -462,7 +462,44 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+        {
+            msdk_printf(MSDK_STRING("ERROR: invalid path to file with frame parameters or -qpfile <path> not specified\n"));
+            return MFX_ERR_NOT_INITIALIZED;
+        }
+
+        m_mfxEncParams.mfx.EncodedOrder = 1;
+
+        QPFileReaderStatus sts = m_QPFileReader.Read(pInParams->strQPFilePath, pInParams->CodecId);
+
+        if (QPFILEREADER_ERR_NONE != sts)
```
mgonchar: this should be wrapped to separate function:
```
std::string FileStatusToString(QPFileReaderStatus sts)
{
switch(sts)
{
case QPFILEREADER_FILE_NOT_OPEN:
    return "ERROR: failed to open file contains frame parameters\n";
...
default:
    return "ERROR: invalid value of QPFileReaderStatus\n";
}
}
```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -462,7 +462,44 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+                return MFX_ERR_UNSUPPORTED;
+            }
+            else if (QPFILEREADER_QP_OUT_OF_RANGE == sts)
+            {
+                msdk_printf(MSDK_STRING("ERROR: some of the QP value from file is out of range (1-51)\n"));
+            }
+            return MFX_ERR_NOT_INITIALIZED;
+        }
+    }
+    else
```
mgonchar: I'd better to get rid of this ```else``` part somehow. For example:
```
m_mfxEncParams.mfx.EncodedOrder = pInParams->UseEncodedOrder ? 1 : 0; // binary flag, 0 signals encoder to take frames in display order
```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2085,10 +2175,38 @@ mfxStatus CEncodingPipeline::Run()
...
-                sts = LoadNextFrame(pSurf);
+                if (0 == m_mfxEncParams.mfx.EncodedOrder)
+                {
+                    pSurf->Info.FrameId.ViewId = currViewNum;
+                    sts = LoadNextFrame(pSurf);
+                }
+                else
+                {
+                    // find surface for encoding
+                    pSurf = NULL;
```
mgonchar: please, use ```nullptr```

PR #1625, opened 2019-09-06
Fix BRC reset

semelianova:
Move enabling/disabling BRC reset flag to separate function from SetRateControl.
Enable this flag at Reset and disable at the end of the execute function to call hw reset while rendering.

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -173,7 +173,6 @@ mfxStatus SetRateControl(
     VADisplay         vaDisplay,
     VAContextID       vaContextEncode,
     VABufferID &      rateParamBuf_id,
-    bool              isBrcResetRequired,
     MFX_ENCODE_CAPS & caps)
 {
     VAStatus vaSts;
@@ -240,7 +239,6 @@ mfxStatus SetRateControl(
     //  MBBRC control
     // Control VA_RC_MB 0: default, 1: enable, 2: disable, other: reserved
     rate_param->rc_flags.bits.mb_rate_control = mbbrc & 0xf;
-    rate_param->rc_flags.bits.reset = isBrcResetRequired;
 
     {
         MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_EXTCALL, "vaUnmapBuffer");
@@ -248,6 +246,32 @@ mfxStatus SetRateControl(
     }
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
+    return MFX_ERR_NONE;
+}
+
+mfxStatus SetBrcResetRequired(
+    VADisplay             vaDisplay,
+    VABufferID            rateParamBuf_id,
+    bool                  isBrcResetRequired)
+{
+    if (rateParamBuf_id == VA_INVALID_ID)
+        return MFX_ERR_NONE; // if buf wasn't created
+
+    VAStatus vaSts;
+    VAEncMiscParameterBuffer *misc_param;
+    VAEncMiscParameterRateControl *rate_param;
+
+    vaSts = vaMapBuffer(vaDisplay,
+                        rateParamBuf_id,
+                        (void **)&misc_param);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    rate_param = (VAEncMiscParameterRateControl *)misc_param->data;
+    rate_param->rc_flags.bits.reset = isBrcResetRequired;
+
+    vaSts = vaUnmapBuffer(vaDisplay,
+                          rateParamBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
     return MFX_ERR_NONE;
 }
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -251,6 +249,38 @@ mfxStatus SetRateControl(
...
+    VADisplay             vaDisplay,
+    VAContextID           vaContextEncode,
+    VABufferID &          rateParamBuf_id,
+    bool                  isBrcResetRequired)
+{
+    VAStatus vaSts;
+    VAEncMiscParameterBuffer *misc_param;
+    VAEncMiscParameterRateControl *rate_param;
+
+    mfxStatus mfxSts = CheckAndDestroyVAbuffer(vaDisplay, rateParamBuf_id);
```
mgonchar: please, don't do it. You need to accept id of buffer and map / unmap it (and change the desired value between)

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -251,6 +249,38 @@ mfxStatus SetRateControl(
...
+                           &rateParamBuf_id);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    vaSts = vaMapBuffer(vaDisplay,
+                        rateParamBuf_id,
+                        (void **)&misc_param);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    rate_param = (VAEncMiscParameterRateControl *)misc_param->data;
+    rate_param->rc_flags.bits.reset = isBrcResetRequired;
```
mgonchar: I believe, you need to unmap buffer after work

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
semelianova: Thanks, got my mistake

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -251,6 +249,26 @@ mfxStatus SetRateControl(
...
 
+mfxStatus SetBrcResetRequired(
+    VADisplay             vaDisplay,
+    VABufferID &          rateParamBuf_id,
+    bool                  isBrcResetRequired)
+{
+    VAEncMiscParameterBuffer *misc_param;
+    VAEncMiscParameterRateControl *rate_param;
+
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaMapBuffer(vaDisplay,
```
mgonchar: please, don't put function call inside macro, it could lead to double function calls. see [coding guide](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/Coding_guidelines.md#macro)

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -251,6 +249,26 @@ mfxStatus SetRateControl(
...
+                                               rateParamBuf_id,
+                                               (void **)&misc_param),
+                          MFX_ERR_DEVICE_FAILED);
+
+    rate_param = (VAEncMiscParameterRateControl *)misc_param->data;
+    rate_param->rc_flags.bits.reset = isBrcResetRequired;
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaUnmapBuffer(vaDisplay, rateParamBuf_id),
```
mgonchar: same

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1859,10 +1880,11 @@ mfxStatus VAAPIEncoder::Reset(MfxVideoParam const & par)
...
-    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetRateControl(par, m_mbbrc, 0, 0, m_vaDisplay, m_vaContextEncode, m_rateParamBufferId, isBrcResetRequired, m_caps), MFX_ERR_DEVICE_FAILED);
+    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetRateControl(par, m_mbbrc, 0, 0, m_vaDisplay, m_vaContextEncode, m_rateParamBufferId, m_caps),             MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetFrameRate(par, m_vaDisplay, m_vaContextEncode, m_frameRateId),                                            MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityLevel(par, m_vaDisplay, m_vaContextEncode, m_qualityLevelId),                                      MFX_ERR_DEVICE_FAILED);
     MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetQualityParams(par, m_vaDisplay, m_vaContextEncode, m_qualityParamsId),                                    MFX_ERR_DEVICE_FAILED);
+    MFX_CHECK_WITH_ASSERT(MFX_ERR_NONE == SetBrcResetRequired(m_vaDisplay, m_rateParamBufferId, isBrcResetRequired),                                   MFX_ERR_DEVICE_FAILED);
```
dmitryermilov: what if application calls encode->Init() and right after it calls encode->Reset()? It seems it won't work because m_rateParamBufferId is not created at Reset() time

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -251,6 +249,29 @@ mfxStatus SetRateControl(
...
+mfxStatus SetBrcResetRequired(
+    VADisplay             vaDisplay,
+    VABufferID &          rateParamBuf_id,
+    bool                  isBrcResetRequired)
+{
+    VAStatus vaSts;
+    VAEncMiscParameterBuffer *misc_param;
+    VAEncMiscParameterRateControl *rate_param;
+
+    vaSts = vaMapBuffer(vaDisplay,
```
dmitryermilov: Need to check if rateParamBuf_id was created ?
Like:
mfxStatus SetBrcResetRequired(
...
if (rateParamBuf_id != VA_INVALID_ID)
{
    vaMapBuffer()
    rate_param->rc_flags.bits.reset = isBrcResetRequired;
    vaUnmap()
}
}

mgonchar: Good catch. It should be something:
```
MFX_CHECK(rateParamBuf_id != VA_INVALID_ID, MFX_ERR_INVALID_HANDLE);
```

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -251,6 +249,29 @@ mfxStatus SetRateControl(
     return MFX_ERR_NONE;
 }
 
+mfxStatus SetBrcResetRequired(
+    VADisplay             vaDisplay,
+    VABufferID &          rateParamBuf_id,
```
mgonchar: No need to pass by reference here, pass by value

dmitryermilov: > Good catch. It should be something:
> 
> MFX_CHECK(rateParamBuf_id != VA_INVALID_ID, MFX_ERR_INVALID_HANDLE);

@mgonchar , IMO - no, we shouldn't return an error. If buffer was not created,   just don't do anything here.

mgonchar: @dmitryermilov ok, if so, it should be:
```
if (rateParamBuf_id == VA_INVALID_ID)
    return MFX_ERR_NONE;
```

just to avoid putting body of function inside ```if``` statement

dmitryermilov: yes..

Issue #1622, opened 2019-09-04
sample_decode failed using drm render


MemeTao:
./sample_decode h264 -i ./input.h264 -hw -vaapi -calc_latency -async 1 -rdrm
libva info: VA-API version 1.6.0
libva info: va_getDriverName() returns 1
libva info: User requested driver 'iHD'
libva info: Trying to open /usr/local/lib/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_6
get chip id failed: -1 [13]
param: 4, val: 0
libva error: /usr/local/lib/dri/iHD_drv_video.so init failed
libva info: va_openDriver() returns 18
------------------------------------------------------------------------------------------------------------------------
1. what's the difference between -r and -rdrw?  '-r' work well on my pc.
2. x-windows is not necessary in 'DRM' mode?
3 'DRM' mode perform better?

NaveenKumar-Intel:
@MemeTao,
I can able to reproduce this issue on CL platform. 
Based on the analysis, this issue is due to the DRM mode selection when "rdrm" parameter is used with sample_decode command line. 
when "rdrm" param is used, it creates **"/dev/dri/card0"** GPU node where as it creates **"/dev/dri/renderD128"** in general case.
When "rdrm" param is used, it specifies **pParams->libvaBackend = MFX_LIBVA_DRM_MODESET**. When we change this param  to **MFX_LIBVA_DRM**, the issue is not reproducing and decoding just works fine.

Moreover, for **Gen GPU architecture**, it always supports **/dev/dri/renderD128** node not **/dev/dri/card0**.
Please find reference in the source code: file /media-driver/media_driver/linux/common/ddi/media_libva_util.h
**#define DEVICE_NAME "/dev/dri/renderD128"   // For Gen, it is always /dev/dri/renderD128 node**

Thanks,
NaveenKumar

dvrogozh:
drm rendering mode (-rdrm) is the one which should work _without_ any graphical server (like X server or Wayland) running. So, basically that's direct framebuffer rendering when you have display, but work on a server-like installation without X running. For that reason it opens /dev/dri/card0 since it performs mode setting and actual real drm authentication rather than x11 authentication. Mind: if X server or Wayland is running -rdrm was not supposed to work at all. Instead you should use -r.

-r is usual rendering mode under X Server. 

@MemeTao, @NaveenKumar-Intel : did you try -rdrm under X server running?

[...]

PR #1627, opened 2019-09-11
fix for rendering with drm

NaveenKumar-Intel:
Observing problem while using rdrm with both Sample_Decode and sample_multi_transcode applications. when rdrm is used, it is intenally invoking card0 as a GPU render node which is creating a problem when memory allocation.

### DIFF
--- a/samples/sample_decode/src/sample_decode.cpp
+++ b/samples/sample_decode/src/sample_decode.cpp
@@ -258,7 +258,6 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
         {
             pParams->memType = D3D9_MEMORY;
             pParams->mode = MODE_RENDERING;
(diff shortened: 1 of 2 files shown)


samples/sample_decode/src/sample_decode.cpp
```diff
@@ -258,7 +258,6 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
         {
             pParams->memType = D3D9_MEMORY;
             pParams->mode = MODE_RENDERING;
-            pParams->libvaBackend = MFX_LIBVA_DRM_MODESET;
```
dvrogozh: See answer in #1622. I am not sure that what you are doing is actually a fix.

NaveenKumar-Intel:
As @dvrogozh suggested in #1622 , -rdrm should not be used when x11 server and wayland are running. So PR raised makes no use. Hence closing it.

PR #1646, opened 2019-09-18
core: added proper device id reporting

mgonchar:


doc/mediasdk-man.md
```diff
@@ -5670,11 +5670,12 @@ The `mfxPlatform` structure contains information about hardware platform.
 | | |
 --- | ---
 `CodeName` | Intel® microarchitecture code name. See the [PlatformCodeName](#PlatformCodeName) enumerator for a list of possible values.
-`DeviceId` | Reserved.
+`DeviceId` | Unique identifier of graphics device.
 
 **Change History**
 
 This structure is available since SDK API 1.19.
+Starting with SDK API **TBD** `DeviceId` field is properly filled.
```
dvrogozh: I am not sure that this worth mentioning in API change history. To me that's a good point to mention in the release change log. I suggest to remove this line. @onabiull : what do you think?

onabiull: There are other similar wordings in this document, when changes do not introduce new structures/defines, but change existing behavior. I'm ok with that line.

dvrogozh: Unfortunately github does not have a way to zero a review. So, I set +1, but it is actually 0. I don't want to block the merge, but I am neutral (inclining to dislike) the decision.

PR #1649, opened 2019-09-18
[sample_vpp] Remove unused code

AntonGrishin:
This part of MCTF code didnt work, because bAttachMctfBuffer didnt set in code below.


mgonchar:
Looks ok. But I have several questions. It was added here for some purpose. What tools do we use to test MCTF (smt / transcoder)? And how hard would be to fully enable support of MCTF to this sample if we decide to do that?

dmitryermilov:
> It was added here for some purpose. And how hard would be to fully enable support of MCTF to this sample if we decide to do that?
Yes. But the problem is the efforts apparently were not completed. Even command line parsing doesn't have per frame MCTF setting support. Now we can't finish it without understand what original idea was ... IMHO, better to remove this code.

SMT, by the way, has handling of per frame MCTF parameters.

AntonGrishin:
@mgonchar btw, sample_vpp support MCTF 
https://github.com/Intel-Media-SDK/MediaSDK/blob/60211e10218d56e8f2e134ad5202f1da925d263e/samples/sample_vpp/src/sample_vpp_config.cpp#L96

mgonchar:
@AntonGrishin but only with default parameters?

AntonGrishin:
@mgonchar in block, which I provide we set params which we spicified.

PR #1653, opened 2019-09-20
[mpeg2d] fix regression in slice handling

uartie:
Fix a regression introduced in commit 4f328187b7cf
(pull-request #1619) that caused output artifacts
in bottom row of decoded mpeg2 files.


uartie:
@asmirno1 @dmitryermilov

dmitryermilov:
Thank you very much, @uartie !

PR #1657, opened 2019-09-20
delete multi methods

egorovdanil:


mgonchar:
@dvrogozh we need them in only one of the child classes. So original solution was: hey, we need it in child class, let's add this to virtual table of parent class. Current solution: Let's remove it from virtual table at all, since we don't actually use pointers on base class

mgonchar:
@dvrogozh I looked closer to this code and find out that we actually do need these functions inside vtable. Looks that we can't actually get rid of them without redesigning the whole app

PR #1659, opened 2019-09-20
clang8 warnings fix: unused command line arguments

egorovdanil:


builder/FindGlobals.cmake
```diff
@@ -36,25 +36,26 @@ if( Linux OR Darwin )
 
   include( GNUInstallDirs )
 
-  add_definitions(-DUNIX)
+  append( "-DUNIX" CMAKE_FLAGS )
```
dvrogozh: I have much simpler solution: just deleted the line:). It is not actually used anywhere in sources:
```
[dvrogozh@dvrscl msdk]$ fgrep -rsn UNIX
_studio/shared/umc/core/vm_plus/include/sys/mfx_vm++_pthread_unix.h:21:#ifndef __MFX_VMPLUSPLUS_PTHREAD_UNIX_H__
_studio/shared/umc/core/vm_plus/include/sys/mfx_vm++_pthread_unix.h:22:#define __MFX_VMPLUSPLUS_PTHREAD_UNIX_H__
_studio/shared/umc/core/vm_plus/include/sys/mfx_vm++_pthread_unix.h:33:#endif //__MFX_VMPLUSPLUS_PTHREAD_UNIX_H__
api/mfx_dispatch/linux/CMakeLists.txt:59:add_definitions(-DUNIX)
builder/FindGlobals.cmake:39:  add_definitions(-DUNIX)
builder/FindGlobals.cmake:190:elseif(UNIX)
builder/FindGlobals.cmake:213:if(NOT ERROR AND UNIX)
builder/FindGlobals.cmake:229:if(UNIX AND MFX_GLIBC)
contrib/googletest/googlemock/test/gmock_output_test.py:58:  """Changes all Windows/Mac line endings in s to UNIX line endings."""
contrib/googletest/googletest/include/gtest/gtest.h:1280:  // UNIX epoch.
contrib/googletest/googletest/include/gtest/internal/gtest-port.h:232://                    Extended Regular Expression syntax on UNIX-like platforms
contrib/googletest/googletest/src/gtest-internal-inl.h:566:  // UNIX epoch.
contrib/googletest/googletest/src/gtest-internal-inl.h:904:  // UNIX epoch.
contrib/googletest/googletest/src/gtest.cc:4568:// UNIX epoch.
contrib/googletest/googletest/test/googletest-output-test.py:86:  """Changes all Windows/Mac line endings in s to UNIX line endings."""
[dvrogozh@dvrscl msdk]$
```

And even if it were used, we should have used compiler standard identification instead I think like `__linux__`.

contrib/ipp/CMakeLists.txt
```diff
@@ -30,11 +30,18 @@ set( SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src" )
 # Optimized for processors with Intel SSE4.2
 
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
-  set( defs "-D_Y8 -D_ARCH_EM64T" )
+  append( "-D_Y8" CMAKE_C_FLAGS )
```
dvrogozh: don't use append, please. The following code will be better I believe:
```
if(CMAKE_SIZEOF_VOID_P EQUAL 8)
    set( ipp_defs "-D_Y8 -D_ARCH_EM64T" )
else()
    set( ipp_defs "-D_P8 -D_ARCH_IA32" )
endif()
list(APPEND CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ipp_defs}")
```
mind change of `defs` to `ipp_defs` since msdk build system reacts on `defs` by itself.

contrib/ipp/CMakeLists.txt
```diff
@@ -46,3 +53,6 @@ list( APPEND sources ${srcs})
 enable_language(C ASM)
 set( CMAKE_ASM_SOURCE_FILE_EXTENSIONS s )
 make_library( ipp none static )
+
+SET( ASM_OPTIONS "-x assembler-with-cpp")
+SET( CMAKE_ASM_FLAGS "${CFLAGS} ${ASM_OPTIONS}" )
```
dvrogozh: hm. Why we did not have this before anywhere in the code?

Also:
1. You don't need separately set ASM_OPTIONS I believe, you can do that within same CMAKE_ASM_FLAGS line
2. You actually override user-specified asm flags, you need to use:
list(APPEND CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -x assembler-with-cpp")
3. Why you add CFLAGS here?

builder/FindGlobals.cmake
```diff
@@ -91,8 +92,8 @@ if( Linux OR Darwin )
   set(c_warnings "-Wall -Wformat -Wformat-security")
   set(cxx_warnings "${c_warnings} -Wnon-virtual-dtor")
 
-  set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -pipe -fPIC ${c_warnings} ${no_warnings}")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pipe -fPIC ${cxx_warnings} ${no_warnings}")
+  set(CMAKE_C_FLAGS   "${CMAKE_FLAGS} ${CMAKE_C_FLAGS} -pipe -fPIC ${c_warnings} ${no_warnings}")
```
dvrogozh: Can you, please, avoid using CMAKE_FLAGS name for the variable? I suspect that this name fakes cmake native variable while actually cmake never defines such and uses CMAKE_<LANG>_FLAGS type of things. Please, set language specific flags only, i.e. CMAKE_C_FLAGS and CMAKE_CXX_FLAGS. To avoid setting everything twice, use temporary variable like:
set(common_flags -DUNIX)
```
list(APPEND common_flags -D__USE_LARGEFILE64)
...
set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -pipe -fPIC ${c_warnings} ${no_warnings} ${common_flags} ")
set(CMAKE_CXX_FLAGS   "${CMAKE_CXX_FLAGS} -pipe -fPIC ${c_warnings} ${no_warnings} ${common_flags} ")
```

Also, please, avoid using self-defined append function, use cmake native set and list. Our self definition does not actually makes much sense - it does not simplify anything, just obscures what's going on.

builder/FindGlobals.cmake
```diff
@@ -36,25 +36,26 @@ if( Linux OR Darwin )
...
 
   if( Linux )
-    add_definitions(-D__USE_LARGEFILE64 -D_FILE_OFFSET_BITS=64)
+    append( "-D__USE_LARGEFILE64" CMAKE_FLAGS )
+    append( "-D_FILE_OFFSET_BITS=64" CMAKE_FLAGS )
 
-    add_definitions(-DLINUX)
-    add_definitions(-DLINUX32)
+    append( "-DLINUX" CMAKE_FLAGS )
+    append( "-DLINUX32" CMAKE_FLAGS )
```
dvrogozh: we should have used compiler standard macro defs instead like `__linux__`, but this will be separate exercise, don't do that within this PR

contrib/ipp/CMakeLists.txt
egorovdanil: Done. I thought that if we combine C and ASM, then we need some C flags for the ASM, but it seems they are not really needed

builder/FindGlobals.cmake
egorovdanil: done

contrib/ipp/CMakeLists.txt
egorovdanil: done

builder/FindGlobals.cmake
egorovdanil: done

builder/FindGlobals.cmake
```diff
@@ -35,26 +35,25 @@ if( Linux OR Darwin )
   endif()
 
   include( GNUInstallDirs )
-
-  add_definitions(-DUNIX)
-
+set( common_flags "${common_flags} -DUNIX")
```
dvrogozh: alignment...

builder/FindGlobals.cmake
```diff
@@ -231,6 +230,12 @@ if(UNIX AND MFX_GLIBC)
 endif()
 
 git_describe( git_commit )
-add_definitions( -DMFX_BUILD_INFO=\"${BUILD_INFO}\" )
-add_definitions( -DMFX_API_VERSION=\"${API_VER_MODIF}\" )
-add_definitions( -DMFX_GIT_COMMIT=\"${git_commit}\" )
+
+set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMFX_BUILD_INFO=\"\\\"${BUILD_INFO}\"\\\"" )
```
dvrogozh: I think you can again define flags for temporary variable and when add it to the flags. This will be shorter and more readable:
```
set(version_flags -DMFX_BUILD_INFO=\"\\\"${BUILD_INFO}\\\"\"")
...
set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${version_flags}" )
set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${version_flags}" )
```

contrib/ipp/CMakeLists.txt
```diff
@@ -46,3 +48,5 @@ list( APPEND sources ${srcs})
 enable_language(C ASM)
 set( CMAKE_ASM_SOURCE_FILE_EXTENSIONS s )
 make_library( ipp none static )
+
+set( CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -x assembler-with-cpp" )
```
dvrogozh: can you, please, move this line above and place right before make_library? I prefer to see settings before target definition.

But again, why we actually need `-x assembler-with-cpp`: I don't see we used this flag before?

builder/FindGlobals.cmake
```diff
@@ -79,20 +78,20 @@ if( Linux OR Darwin )
...
     set(no_warnings "-Wno-deprecated-declarations -Wno-unknown-pragmas -Wno-unused")
   endif()
-
+ 
   set(c_warnings "-Wall -Wformat -Wformat-security")
   set(cxx_warnings "${c_warnings} -Wnon-virtual-dtor")
 
-  set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -pipe -fPIC ${c_warnings} ${no_warnings}")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pipe -fPIC ${cxx_warnings} ${no_warnings}")
+  set(CMAKE_C_FLAGS   "${common_flags} ${CMAKE_C_FLAGS} -pipe -fPIC ${c_warnings} ${no_warnings}")
```
dvrogozh: add in the very end, please. let's use the scheme:
`set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}   <new stuff>")`

contrib/ipp/CMakeLists.txt
egorovdanil: I could not remove one flag -DNDEBUG. This flag tells armclang that the source assembly file requires preprocessing. With this flag it stops throwing a warning

egorovdanil: Updated. `-DNDEBUG` is added by CMake so it can be disabled by `set( CMAKE_ASM_EXTENSIONS OFF )`

builder/FindGlobals.cmake
egorovdanil: done

builder/FindGlobals.cmake
egorovdanil: done

builder/FindGlobals.cmake
```diff
@@ -79,20 +78,20 @@ if( Linux OR Darwin )
     set( git_commit "" )
     git_describe( git_commit )
 
-    add_definitions( -DMSDK_BUILD=\"$ENV{BUILD_NUMBER}\")
+    set( common_flags "${common_flags} -DMSDK_BUILD=\\\"$ENV{BUILD_NUMBER}\\\"" )
```
mgonchar: why so many slashes? :)

egorovdanil: Because we need to get `-DMSDK_BUILD="\"<string>"\"`

PR #1660, opened 2019-09-20
clang8 warnings fix: unknown pragmas

egorovdanil:
unknown pragma warnings fix

### DIFF
--- a/samples/sample_common/include/time_statistics.h
+++ b/samples/sample_common/include/time_statistics.h
@@ -26,7 +26,6 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include <vector>
 #include <stdio.h>
 
-#pragma warning(disable:4100)
 
 class CTimeStatisticsReal
 {
@@ -184,7 +183,7 @@ public:
         return frequency;
     }
 
-    static mfxF64 ConvertToSeconds(msdk_tick elapsed)
+    static mfxF64 ConvertToSeconds(msdk_tick /*elapsed*/)
     {
         return 0;
     }
@@ -215,7 +214,7 @@ public:
 
     inline void TurnOffDumping() {}
 
-    inline void PrintStatistics(const msdk_char* prefix)
+    inline void PrintStatistics(const msdk_char* /*prefix*/)
     {
     }
 
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -33,7 +33,6 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include "mfxjpeg.h"
 #include "mfxvp8.h"
 
-#pragma warning( disable : 4748 )
 
 msdk_tick CTimer::frequency = 0;
 msdk_tick CTimeStatisticsReal::frequency = 0;
--- a/samples/sample_common/src/sysmem_allocator.cpp
+++ b/samples/sample_common/src/sysmem_allocator.cpp
@@ -24,7 +24,6 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #define ID_BUFFER MFX_MAKEFOURCC('B','U','F','F')
 #define ID_FRAME  MFX_MAKEFOURCC('F','R','M','E')
 
-#pragma warning(disable : 4100)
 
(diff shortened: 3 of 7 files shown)


samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -3046,7 +3046,8 @@ mfxStatus CTranscodingPipeline::InitVppMfxParams(sInputParams *pInParams)
             break;
         default:
             assert(0);
-            MSDK_CHECK_STATUS(MFX_ERR_UNSUPPORTED, "Unexpected encoder FourCC");
+			sts = MFX_ERR_UNSUPPORTED;
+            MSDK_CHECK_STATUS(sts, "Unexpected encoder FourCC");
```
mgonchar: Sometimes we want to use such constructions, so you just introduce unnecessary workaround to fool the compiler. I don't like this particular change

dvrogozh: yes, this one looks ugly. Which warning you have tried to fix? I don't quite see why compiler cries out....

BTW, please, fix your editor settings: we try to avoid using tabs and I think they squeezed thru in line 3849.

egorovdanil: I have tried to fix C4127 warning: conditional expression is constant

dvrogozh: I see. @mgonchar : do you have an idea how to fix it by some macro change? According to https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=vs-2019 compiler will skip the warning for some trivial expressions like `if (1)` or if for const expression. Maybe something like that:

```
#define MSDK_CHECK_STATUS(X, MSG) \
{ \
    const bool _if_expr = (X) < MFX_ERR_NONE;
    if (_if_expr) \
    {MSDK_PRINT_RET_MSG(X, MSG); return X;} \
}
```

If I read MSFT description correctly, this should work starting from VS2015 update 3. Which VS we use in github mediasdk?

mgonchar: I would say it's better to use ```constexpr```:
```
#define MSDK_CHECK_STATUS(X, MSG) \
{ \
    constexpr bool sts_is_error = (X) < MFX_ERR_NONE; \
    if (sts_is_error) \
    {MSDK_PRINT_RET_MSG(X, MSG); return X;} \
}
```

According to .sln files in repo it is VS15

egorovdanil: With `constexpr` compiler gives` C2131` error: `expression did not evaluate to a constant`, so I used `const bool`

samples/sample_common/include/sample_defs.h
```diff
@@ -145,7 +145,7 @@ enum LibVABackend
 #define MSDK_CHECK_NOT_EQUAL(P, X, ERR)          {if ((X) != (P)) {msdk_stringstream tmpStr3;tmpStr3<<MSDK_STRING(#X)<<MSDK_STRING("!=")<<MSDK_STRING(#P)<<MSDK_STRING(" error"); \
     MSDK_PRINT_RET_MSG(ERR, tmpStr3.str().c_str()); return ERR;}}
 
-#define MSDK_CHECK_STATUS(X, MSG)                {if ((X) < MFX_ERR_NONE) {MSDK_PRINT_RET_MSG(X, MSG); return X;}}
+#define MSDK_CHECK_STATUS(X, MSG)                {const bool sts_is_error = (X) < MFX_ERR_NONE; if (sts_is_error) {MSDK_PRINT_RET_MSG(X, MSG); return X;}}
```
dvrogozh: @egorovdanil : we talked offline w/ @onabiull and @mgonchar. We don't like this workaround either. Let's shut up MSFT compiler warning with conditional if to avoid clang cry out. I.e.:
```
#ifdef _MSC_VER
#pragma warning(disable: XXX)
#endif 
```

egorovdanil: done

PR #1662, opened 2019-09-20
A new approach of MFX_WRN_DEVICE_BUSY handling

dmitryermilov:
If MFX_WRN_DEVICE_BUSY is returned, it shall be paired with a valid syncPoint.
That syncPoint is not necessarily binded with actuall output data.
The purpose of the syncPoint is provide applications a way to make polling-free waiting.
Applications, of course, can choose either to wait or perform some different job.

### DIFF
--- a/_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
+++ b/_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
 } // mfxStatus mfxSchedulerCore::DoWork()
 
 
+mfxStatus mfxSchedulerCore::FindOldestActiveSyncPoint(const void *pOwner, mfxSyncPoint *pSyncPoint)
+{
+    if (nullptr == pOwner || nullptr == pSyncPoint)
+    {
+        return MFX_ERR_NULL_PTR;
+    }
 
+    mfxTaskHandle handle;
+    handle.handle = 0;
+
+    mfxU32 priority;
+    // enter protected section
+    {
+        std::lock_guard<std::mutex> guard(m_guard);
+
+        for (priority = 0; priority < MFX_PRIORITY_NUMBER; priority += 1)
+        {
+            int type;
+
+            for (type = MFX_TYPE_HARDWARE; type <= MFX_TYPE_SOFTWARE; type += 1)
+            {
+                MFX_SCHEDULER_TASK *pTask = m_pTasks[priority][type];
+
+                // run over the tasks with particular priority
+                while (pTask)
+                {
+                    // return the syncPoint associated with the oldest working task which belongs to the provided owner
+                    if ((pTask->param.task.pOwner == pOwner) &&
+                        (MFX_ERR_NONE != pTask->opRes))
+                    {
+                        handle.taskID = pTask->taskID;
+                        handle.jobID = pTask->jobID;
+                        *pSyncPoint = (mfxSyncPoint) handle.handle;
+
+                        return MFX_ERR_NONE;
+                    }
+
+                    // advance the task pointer
+                    pTask = pTask->pNext;
+                }
+            }
+        }
+    }
+
+    return MFX_ERR_NOT_FOUND;
+}
 
--- a/_studio/mfx_lib/shared/include/mfx_interface_scheduler.h
+++ b/_studio/mfx_lib/shared/include/mfx_interface_scheduler.h
@@ -1,15 +1,15 @@
-// Copyright (c) 2018 Intel Corporation
-// 
+// Copyright (c) 2018-2019 Intel Corporation
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 // FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -87,6 +87,10 @@ public:
     // Add a new task to the scheduler. Threads start processing task immediately.
     virtual
     mfxStatus AddTask(const MFX_TASK &task, mfxSyncPoint *pSyncPoint) = 0;
+
+    // Find the oldest active syncPoint for a specified owner
+    virtual
+    mfxStatus FindOldestActiveSyncPoint(const void *pOwner, mfxSyncPoint *pSyncPoint) = 0;
 
     // Make synchronization, wait until task is done.
     virtual
--- a/_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
+++ b/_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
@@ -491,74 +491,88 @@ mfxStatus MFXVideoDECODE_DecodeFrameAsync(mfxSession session, mfxBitstream *bs, 
 
     MFX_CHECK(session, MFX_ERR_INVALID_HANDLE);
     MFX_CHECK(session->m_pScheduler, MFX_ERR_NOT_INITIALIZED);
-    MFX_CHECK(session->m_pDECODE.get(), MFX_ERR_NOT_INITIALIZED);
+    MFX_CHECK(session->m_pDECODE, MFX_ERR_NOT_INITIALIZED);
     MFX_CHECK_NULL_PTR1(syncp);
     MFX_CHECK_NULL_PTR1(surface_out);
 
     try
     {
-        mfxSyncPoint syncPoint = NULL;
-        MFX_TASK task;
-
-        // Wait for the bit stream
-        mfxRes = session->m_pScheduler->WaitForDependencyResolved(bs);
-        MFX_CHECK_STS(mfxRes);
-
-        // reset the sync point
-        *syncp = NULL;
-        *surface_out = NULL;
-
-        memset(&task, 0, sizeof(MFX_TASK));
-        mfxRes = session->m_pDECODE->DecodeFrameCheck(bs, surface_work, surface_out, &task.entryPoint);
-        MFX_CHECK(mfxRes >= 0 || MFX_ERR_MORE_DATA_SUBMIT_TASK == static_cast<int>(mfxRes), mfxRes);
-
-        // source data is OK, go forward
-        if (task.entryPoint.pRoutine)
+        do
         {
-            mfxStatus mfxAddRes;
+            mfxSyncPoint syncPoint = NULL;
+            MFX_TASK task;
 
-            task.pOwner = session->m_pDECODE.get();
-            task.priority = session->m_priority;
-            task.threadingPolicy = session->m_pDECODE->GetThreadingPolicy();
-            // fill dependencies
-            task.pSrc[0] = *surface_out;
-            task.pDst[0] = *surface_out;
-            // this is wa to remove external task dependency for HEVC SW decode plugin.
-            // need only because SW HEVC decode is pseudo
+            // Wait for the bit stream
+            mfxRes = session->m_pScheduler->WaitForDependencyResolved(bs);
+            MFX_CHECK_STS(mfxRes);
+
+            // reset the sync point
+            *syncp = NULL;
+            *surface_out = NULL;
+
+            memset(&task, 0, sizeof(MFX_TASK));
+            mfxRes = session->m_pDECODE->DecodeFrameCheck(bs, surface_work, surface_out, &task.entryPoint);
+            MFX_CHECK(mfxRes >= 0 || MFX_ERR_MORE_DATA_SUBMIT_TASK == mfxRes, mfxRes);
+
+            // source data is OK, go forward
+            if (task.entryPoint.pRoutine)
             {
-                mfxPlugin plugin;
-                mfxPluginParam par;
-                if (session->m_plgDec.get())
+                mfxStatus mfxAddRes;
+
+                task.pOwner = session->m_pDECODE.get();
+                task.priority = session->m_priority;
+                task.threadingPolicy = session->m_pDECODE->GetThreadingPolicy();
+                // fill dependencies
+                task.pSrc[0] = *surface_out;
+                task.pDst[0] = *surface_out;
+                // this is wa to remove external task dependency for HEVC SW decode plugin.
+                // need only because SW HEVC decode is pseudo
                 {
-                    session->m_plgDec.get()->GetPlugin(plugin);
-                    MFX_CHECK_STS(plugin.GetPluginParam(plugin.pthis, &par));
-                    if (MFX_PLUGINID_HEVCD_SW == par.PluginUID)
+                    mfxPlugin plugin;
+                    mfxPluginParam par;
+                    if (session->m_plgDec)
                     {
-                        task.pDst[0] = 0;
+                        session->m_plgDec->GetPlugin(plugin);
+                        MFX_CHECK_STS(plugin.GetPluginParam(plugin.pthis, &par));
+                        if (MFX_PLUGINID_HEVCD_SW == par.PluginUID)
+                        {
+                            task.pDst[0] = 0;
+                        }
                     }
                 }
+
+    #ifdef MFX_TRACE_ENABLE
+                task.nParentId = MFX_AUTO_TRACE_GETID();
+                task.nTaskId = MFX::CreateUniqId() + MFX_TRACE_ID_DECODE;
+    #endif
+
+                // register input and call the task
+                mfxAddRes = session->m_pScheduler->AddTask(task, &syncPoint);
+                MFX_CHECK_STS(mfxAddRes);
             }
 
-#ifdef MFX_TRACE_ENABLE
-            task.nParentId = MFX_AUTO_TRACE_GETID();
-            task.nTaskId = MFX::CreateUniqId() + MFX_TRACE_ID_DECODE;
-#endif
+            if (MFX_ERR_MORE_DATA_SUBMIT_TASK == static_cast<int>(mfxRes))
+            {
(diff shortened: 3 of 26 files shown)


_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -813,147 +813,160 @@ mfxStatus MFXVideoENCODE_EncodeFrameAsync(mfxSession session, mfxEncodeCtrl *ctr
...
-                task.nTaskId = MFX::CreateUniqId() + MFX_TRACE_ID_ENCODE2;
-#endif
-                // register input and call the task
-                MFX_CHECK_STS(session->m_pScheduler->AddTask(task, &syncPoint));
-            }
+                mfxStatus mfxFindRes = session->m_pScheduler->FindActiveSyncPoint(session->m_pENCODE.get(), &syncPoint);
+                // If we can't find any active SyncPoint's it means that next call to MFXVideoDECODE_DecodeFrameAsync should already accept a task.
+                // Let's repeat the call right here instead of making application do it.
+                if (MFX_ERR_NOT_FOUND == mfxFindRes)
+                     continue;
```
eistarov: It is better to use some counter here(only one repeat-call) to avoid infinite loop **inside** ...FrameAsync() if something works wrong in components logic.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -813,147 +813,160 @@ mfxStatus MFXVideoENCODE_EncodeFrameAsync(mfxSession session, mfxEncodeCtrl *ctr
...
-        mfxRes = MFX_ERR_UNKNOWN;
-    }
+            // return pointer to synchronization point
+            *syncp = syncPoint;
+        }
+        // handle error(s)
+        catch(...)
+        {
+            // set the default error value
+            mfxRes = MFX_ERR_UNKNOWN;
```
eistarov: I think for this case we should not continue the loop.

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dmitryermilov: Yes, this is a good point. I didn't do it because wan't sure what this max counter should be.

dmitryermilov: Hmm, perhaps the right answer is not "1" but AsyncDepth. I mean a component can hold num tasks not more than AsyncDepth. It means we need to set this the counter  to AsyncDepth. What do you mean?

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dmitryermilov: Good catch. It's a bug. Thank you

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
eistarov: Yeah, AsyncDepth should work.
Though i'm not sure how you can to get it from component on this level.

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
 } // mfxStatus mfxSchedulerCore::DoWork()
 
 
+mfxStatus mfxSchedulerCore::FindActiveSyncPoint(const void *pOwner, mfxSyncPoint *pSyncPoint)
+{
+    if (nullptr == pOwner || nullptr == pSyncPoint)
```
mgonchar: Can we use macro like ```MFX_CHECK``` here?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
...
+    {
+        return MFX_ERR_NULL_PTR;
+    }
 
+    mfxTaskHandle handle;
+    handle.handle = 0;
+
+    mfxU32 priority;
```
mgonchar: why don't define it in ```for```-loop statement

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
...
+    // enter protected section
+    {
+        std::lock_guard<std::mutex> guard(m_guard);
+
+        for (priority = 0; priority < MFX_PRIORITY_NUMBER; priority += 1)
+        {
+            int type;
```
mgonchar: same

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
...
+
+            for (type = MFX_TYPE_HARDWARE; type <= MFX_TYPE_SOFTWARE; type += 1)
+            {
+                MFX_SCHEDULER_TASK *pTask = m_pTasks[priority][type];
+
+                // run over the tasks with particular priority
+                while (pTask)
```
mgonchar: as for me, ```for```-loop looks more clear:
```
for (MFX_SCHEDULER_TASK *pTask = m_pTasks[priority][type]; pTask; pTask = pTask->pNext)
```

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
```diff
@@ -495,77 +495,91 @@ mfxStatus MFXVideoDECODE_DecodeFrameAsync(mfxSession session, mfxBitstream *bs,
...
+                task.pDst[0] = *surface_out;
+                // this is wa to remove external task dependency for HEVC SW decode plugin.
+                // need only because SW HEVC decode is pseudo
                 {
-                    session->m_plgDec.get()->GetPlugin(plugin);
-                    MFX_CHECK_STS(plugin.GetPluginParam(plugin.pthis, &par));
-                    if (MFX_PLUGINID_HEVCD_SW == par.PluginUID)
+                    mfxPlugin plugin;
+                    mfxPluginParam par;
+                    if (session->m_plgDec.get())
```
mgonchar: no need to call ```.get()```

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -813,147 +813,160 @@ mfxStatus MFXVideoENCODE_EncodeFrameAsync(mfxSession session, mfxEncodeCtrl *ctr
...
+                                                        bs,
+                                                        &reordered_surface,
+                                                        &internal_params,
+                                                        entryPoints,
+                                                        numEntryPoints);
+            // source data is OK, go forward
+            if ((MFX_ERR_NONE == mfxRes) ||
+                (MFX_WRN_INCOMPATIBLE_VIDEO_PARAM == mfxRes) ||
+                (MFX_WRN_OUT_OF_RANGE == mfxRes) ||
+                // WHAT IS IT??? IT SHOULD BE REMOVED
```
mgonchar: maybe, let's fix this one?)
1) remove comment
2) fix the code
3) write something more meaningful instead

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
```diff
@@ -495,77 +495,91 @@ mfxStatus MFXVideoDECODE_DecodeFrameAsync(mfxSession session, mfxBitstream *bs,
...
+                // register input and call the task
+                mfxAddRes = session->m_pScheduler->AddTask(task, &syncPoint);
+                MFX_CHECK_STS(mfxAddRes);
             }
 
-#ifdef MFX_TRACE_ENABLE
-            task.nParentId = MFX_AUTO_TRACE_GETID();
-            task.nTaskId = MFX::CreateUniqId() + MFX_TRACE_ID_DECODE;
-#endif
+            if (MFX_ERR_MORE_DATA_SUBMIT_TASK == static_cast<int>(mfxRes))
```
mgonchar: why do we need any casts? ```MFX_ERR_MORE_DATA_SUBMIT_TASK``` is a member of ```mfxStatus``` enum as I see

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
```diff
@@ -495,77 +495,91 @@ mfxStatus MFXVideoDECODE_DecodeFrameAsync(mfxSession session, mfxBitstream *bs,
...
+            }
 
-            // register input and call the task
-            mfxAddRes = session->m_pScheduler->AddTask(task, &syncPoint);
-            MFX_CHECK_STS(mfxAddRes);
-        }
+            if (MFX_WRN_DEVICE_BUSY == mfxRes)
+            {
+                mfxStatus mfxFindRes = session->m_pScheduler->FindActiveSyncPoint(session->m_pDECODE.get(), &syncPoint);
+                // If we can't find any active SyncPoint's it means that next call to MFXVideoDECODE_DecodeFrameAsync should already accept a task.
```
dmitry-gurulev: When AVC decoder returns `MFX_ERR_MORE_DATA_SUBMIT_TASK` isn't the case of decoder has no active tasks and next call of 'DecodeFrameAsync' will just return the same status?

dmitry-gurulev: Aha, I've recalled that we create 'virtual' (w/o output) task in that case, so the next call of `FindActiveSyncPoint` will find that sync. point. Anyway, I'd rather use sync. point that is returned by `AddTask` instead of invoking `DecodeFrameCheck` + `FindActiveSyncPoint` over again.

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
...
+                {
+                    // return the syncPoint associated with the oldest working task which belongs to the provided owner
+                    if ((pTask->param.task.pOwner == pOwner) &&
+                        (MFX_WRN_IN_EXECUTION == pTask->opRes))
```
dmitry-gurulev: Are we sure that actual tasks have only `MFX_WRN_IN_EXECUTION` status?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dmitryermilov: not used in scheduler. So I don't want to introduce it,

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dmitryermilov: it's a current scheduler code style

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
dmitryermilov: IMHO this is wrong. What if AddTask wasn't called?

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
dmitryermilov: ok

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
dmitryermilov: definitely not in this commit..

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dmitryermilov: It looks like 3 types of statuses are possible:
MFX_ERR_NONE, MFX_WRN_IN_EXECUTION and the rest(means errors). Should we return failed task here ? I think the answer is yes. @dvrogozh , what do you think?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dmitryermilov: what about "type <= MFX_TYPE_SOFTWARE" check for this loop?

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
...
+
+            for (type = MFX_TYPE_HARDWARE; type <= MFX_TYPE_SOFTWARE; type += 1)
+            {
+                MFX_SCHEDULER_TASK *pTask = m_pTasks[priority][type];
+
+                // run over the tasks with particular priority
+                while (pTask)
+                {
+                    // return the syncPoint associated with the oldest working task which belongs to the provided owner
+                    if ((pTask->param.task.pOwner == pOwner) &&
```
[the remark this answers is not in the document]
dmitryermilov: yes. because it's a queue.

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -683,5 +683,50 @@ mfxStatus mfxSchedulerCore::DoWork()
...
 
+    mfxTaskHandle handle;
+    handle.handle = 0;
+
+    mfxU32 priority;
+    // enter protected section
+    {
+        std::lock_guard<std::mutex> guard(m_guard);
+
+        for (priority = 0; priority < MFX_PRIORITY_NUMBER; priority += 1)
```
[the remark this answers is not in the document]
dmitryermilov: Ok. Thank you. I'll do it in another commit.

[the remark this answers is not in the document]
dmitryermilov: https://github.com/Intel-Media-SDK/MediaSDK/pull/1665

_studio/mfx_lib/shared/include/mfx_interface_scheduler.h
```diff
@@ -88,6 +88,10 @@ class MFXIScheduler : public MFXIUnknown
     virtual
     mfxStatus AddTask(const MFX_TASK &task, mfxSyncPoint *pSyncPoint) = 0;
 
+    // Find the oldest active syncPoint for a specified owner
+    virtual
+    mfxStatus FindOldestActiveSyncPoint(const void *pOwner, mfxSyncPoint *pSyncPoint) = 0;
```
dvrogozh: I hate myself to say that, but we can't add it here. I afraid `MFXIScheduler3` is coming...

_studio/mfx_lib/shared/src/libmfxsw_decode.cpp
```diff
@@ -491,74 +491,88 @@ mfxStatus MFXVideoDECODE_DecodeFrameAsync(mfxSession session, mfxBitstream *bs,
...
-            // register input and call the task
-            mfxAddRes = session->m_pScheduler->AddTask(task, &syncPoint);
-            MFX_CHECK_STS(mfxAddRes);
-        }
+            if (MFX_WRN_DEVICE_BUSY == mfxRes)
+            {
+                mfxStatus mfxFindRes = session->m_pScheduler->FindOldestActiveSyncPoint(session->m_pDECODE.get(), &syncPoint);
+                // If we can't find any active SyncPoint's it means that next call to MFXVideoDECODE_DecodeFrameAsync should already accept a task.
+                // Let's repeat the call right here instead of making application do it.
+                if (MFX_ERR_NOT_FOUND == mfxFindRes)
```
dvrogozh: You have `do {...} while(false);` loop. So, calling `continue` won't make a trick: you will go to the next iteration, but loop condition will be checked first, so you will exit the loop instead of trying it once again.  I think you need:
```
while(1) {
    ...
    if (MFX_ERR_NOT_FOUND == mfxFindRes)
        continue;
    ...
    break;
}
```
kind of construction.

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -303,6 +303,7 @@ mfxStatus sTask::Close()
 
 mfxStatus sTask::WriteBitstream()
 {
+	//printf("WriteBitstream Length %d\n", mfxBS.DataLength);
```
dvrogozh: debug printf - remove

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -889,7 +890,7 @@ mfxStatus CEncodingPipeline::AllocFrames()
         return MFX_ERR_MEMORY_ALLOC;
 
     // The number of surfaces shared by vpp output and encode input.
-    nEncSurfNum = EncRequest.NumFrameSuggested;
+    nEncSurfNum = EncRequest.NumFrameSuggested + 5;
```
dvrogozh: debug

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -1902,7 +1903,7 @@ mfxStatus CEncodingPipeline::ResetMFXComponents(sInputParams* pParams)
     }
 
     mfxU32 nEncodedDataBufferSize = m_mfxEncParams.mfx.FrameInfo.Width * m_mfxEncParams.mfx.FrameInfo.Height * 4;
-    sts = m_TaskPool.Init(&m_mfxSession, m_FileWriters.first, m_mfxEncParams.AsyncDepth, nEncodedDataBufferSize, m_FileWriters.second);
+    sts = m_TaskPool.Init(&m_mfxSession, m_FileWriters.first, 45, nEncodedDataBufferSize, m_FileWriters.second);
```
dvrogozh: debug

samples/sample_vpp/src/sample_vpp.cpp
```diff
@@ -713,24 +713,24 @@ int main(int argc, msdk_char *argv[])
...
                 mfxFrameSurface1 * out_surface = nullptr;
-                sts = frameProcessor.pmfxVPP->RunFrameVPPAsyncEx(
-                    pInSurf[nInStreamInd],
-                    pWorkSurf,
-                    //pExtData,
-                    &out_surface,
-                    &syncPoint);
-
-                while(MFX_WRN_DEVICE_BUSY == sts)
+                for (;;)
```
dvrogozh: can we change this place to `for(;;)` in a separate PR, please?

dmitryermilov: https://github.com/Intel-Media-SDK/MediaSDK/pull/1696/files

dvrogozh:
@dmitryermilov : when we can expect 2nd iteration?

PR #1665, opened 2019-09-23
[scheduler] Reduced code duplication

dmitryermilov:

### DIFF
--- a/_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
+++ b/_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
@@ -1,15 +1,15 @@
-// Copyright (c) 2018 Intel Corporation
-// 
+// Copyright (c) 2018-2019 Intel Corporation
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 // FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -25,6 +25,7 @@
 
 #include <mfx_scheduler_core_thread.h>
 #include <mfx_scheduler_core_handle.h>
+#include <mfx_scheduler_core_task.h>
 
 #include <mfx_task.h>
 
@@ -152,9 +153,9 @@ public:
     virtual
     mfxStatus WaitForDependencyResolved(const void *pDependency);
 
-    // Wait until task(s) of specified owner become complete or unattended
+    // Wait until all tasks of specified owner become complete or unattended
     virtual
-    mfxStatus WaitForTaskCompletion(const void *pOwner);
+    mfxStatus WaitForAllTasksCompletion(const void *pOwner);
 
     // Reset 'waiting' status for tasks of specified owner
     virtual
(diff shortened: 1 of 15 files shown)


_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
```diff
@@ -312,6 +313,31 @@ class mfxSchedulerCore : public MFXIScheduler2
     inline MFX_SCHEDULER_THREAD_CONTEXT* GetThreadCtx(mfxU32 thread_id)
     { return &m_pThreadCtx[thread_id]; }
 
+    // Invokes functor 'bool F(MFX_SCHEDULER_TASK*)' for every valid task that returns 'false' to continue iteration or 'true' to stop it.
+    template <typename F>
+    void ForEachTask(F&& f)
+    {
+        mfxU32 priority;
```
mgonchar: minor: can move all of these definitions in place to make code more compact

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -362,63 +361,50 @@ mfxStatus mfxSchedulerCore::WaitForTaskCompletion(const void *pOwner)
...
-        {
-            std::lock_guard<std::mutex> guard(m_guard);
-            int priority;
+        mfxTaskHandle waitHandle;
+        std::lock_guard<std::mutex> guard(m_guard);
 
-            priority = MFX_PRIORITY_HIGH;
-            do
+        ForEachTask(
+            [&](MFX_SCHEDULER_TASK* task)
```
mgonchar: looks like too big closure. I believe you need only ```tasks&```

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -362,63 +361,50 @@ mfxStatus mfxSchedulerCore::WaitForTaskCompletion(const void *pOwner)
...
-                            break;
-                        }
-
-                        // get the next task
-                        pTask = pTask->pNext;
-                    }
+                    waitHandle.taskID = task->taskID;
+                    waitHandle.jobID = task->jobID;
+
+                    tasks.emplace_back(waitHandle);
```
mgonchar: well, it should be
```
tasks.emplace_back(task->taskID, task->jobID);
```

Maybe you would need to write constructor for ```mfxTaskHandle```

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
```diff
@@ -362,63 +361,50 @@ mfxStatus mfxSchedulerCore::WaitForTaskCompletion(const void *pOwner)
...
+    auto handle = tasks.begin();
+    while (!tasks.empty())
+    {
+        if (tasks.end() == handle)
+            handle = tasks.begin();
 
-            // leave the protected section
+        // wait for a while
+        // for a while and infinite are different things
+        if (MFX_WRN_IN_EXECUTION == Synchronize(*handle, MFX_TIME_TO_WAIT)) // Still working
```
mgonchar: Is it possible to stuck here forever? With some hanged task returning ```MFX_WRN_IN_EXECUTION``` over and over

_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
```diff
@@ -312,6 +313,31 @@ class mfxSchedulerCore : public MFXIScheduler2
     inline MFX_SCHEDULER_THREAD_CONTEXT* GetThreadCtx(mfxU32 thread_id)
     { return &m_pThreadCtx[thread_id]; }
 
+    // Invokes functor 'bool F(MFX_SCHEDULER_TASK*)' for every valid task that returns 'false' to continue iteration or 'true' to stop it.
+    template <typename F>
+    void ForEachTask(F&& f)
```
dvrogozh: I like the idea, but I don't like template name since I expect from it to iterate for really all the tasks in the queue while you have boolean functor which signals the stop. Can you think about rename? I am honestly not sure in how I would name it. Also, I would suggest to have 2 templates one is the same you have here and another which is real for_each:
```
template <typename F> // bool F();
void ForEachTaskWhile(F&& f) {}  // that's the one you have in PR

template <typename F> // void F();
void ForEachTask(F&& f) {
    return ForEachTaskWhile([] { F(); return false;} );
}
```

If you will follow "*While" name, consider to revert boolean meaning for `true` to continue to iterate like in `do {} while(true)` syntax, i.e. "For Each Task While Predicate returns True".

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dmitryermilov: and pOwner

dmitryermilov: ok. will slightly change

_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h
dmitryermilov: ok

_studio/mfx_lib/scheduler/src/mfx_scheduler_core_ischeduler.cpp
dmitryermilov: I think yes. but it was original logic. I don't work to change behaviour here

PR #1678, opened 2019-09-26
clang-8 warnings fix

egorovdanil:


dvrogozh:
> We can remove Cmake flags from ASM if we explicitly specify the compiler for ASM as clang-8

How? Post example in commit message where you remove this flag.

tools/asg-hevc/include/frame_processor.h
```diff
@@ -149,7 +149,14 @@ class FrameProcessor
     mfxU8 GetDefaultWeightedPredSample(mfxI32 predSampleLX);
     mfxU8 GetDefaultWeightedPredSample(mfxI32 predSampleL0, mfxI32 predSampleL1);
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-private-field"
+#endif
```
dvrogozh: Why we just can't remove m_FrameType?

egorovdanil:
I have added a `-DCMAKE_ASM_COMPILER=clang-8` flag to the command line.

tools/asg-hevc/include/frame_processor.h
egorovdanil: I found a variable that was used with the same name, but this variable is of a different class. So now I have removed `m_FrameType`

dvrogozh: build fails...

dvrogozh:
> I have added a -DCMAKE_ASM_COMPILER=clang-8 flag to the command line.

Add this to the _commit_ message, please where you modify CMakeLists.txt. That's a good catch by the way.

tools/asg-hevc/include/frame_processor.h
```diff
@@ -149,8 +149,11 @@ class FrameProcessor
     mfxU8 GetDefaultWeightedPredSample(mfxI32 predSampleLX);
     mfxU8 GetDefaultWeightedPredSample(mfxI32 predSampleL0, mfxI32 predSampleL1);
 
+<<<<<<< Updated upstream
```
dvrogozh: git left garbage!

contrib/ipp/CMakeLists.txt
```diff
@@ -46,7 +46,6 @@ endif()
 
 list( APPEND sources ${srcs})
 enable_language( C ASM )
-set( CMAKE_ASM_EXTENSIONS OFF )
```
egorovdanil: -DCMAKE_ASM_COMPILER=clang-8 flag delete CMake flags

samples/sample_plugins/rotate_opencl/src/sample_opencl_plugin.cpp
```diff
@@ -72,20 +68,17 @@ mfxStatus Rotate::PluginInit(mfxCoreInterface *core)
     m_pmfxCore = new mfxCoreInterface;
     MSDK_CHECK_POINTER(m_pmfxCore, MFX_ERR_MEMORY_ALLOC);
     *m_pmfxCore = *core;
-    mfxCoreParam par = {0};
+    mfxCoreParam par = {{0}};
```
mgonchar: just use empty initializer list:
```
 mfxCoreParam par = {};
```

egorovdanil: done

contrib/ipp/CMakeLists.txt
dvrogozh: I still don't see this in commit message.

egorovdanil: Now it's done

PR #1712, opened 2019-10-07
[sample_encode] Set SyncOperation timeout via commandline

Alexandr-Konovalov:


gkrivor:
First of all - using some "Magic Numbers" isn't good (I mean pre-defined MSDK_WAIT_INTERVAL).

It might be used by customers for:
1. Verifying: is hang caused by not enough syncop timeout
2. Trying to improve performance (if possible) by using different timeout values
3. Control performance stability in tests by using some pre-defined timeout values

dmitryermilov:
> First of all - using some "Magic Numbers" isn't good (I mean pre-defined MSDK_WAIT_INTERVAL).
> 
> It might be used by customers for:
> 
> 1. Verifying: is hang caused by not enough syncop timeout

There is already console messages for it.
> 2. Trying to improve performance (if possible) by using different timeout values

Hmmm. Which way it can improve performance ? Please give particular examples.
> 3. Control performance stability in tests by using some pre-defined timeout values

I don't follow. Which way you can control performance using different timeout values?

PR #1722, opened 2019-10-11
set ENABLE_MAX_NUM_REORDER_FRAMES_OUTPUT=true by default

ruanhongfu:
to enable feature with output decoded frames without latency regarding
SPS.VUI.max_num_reorder_frames

Signed-off-by: Ruan, Hongfu <hongfu.ruan@intel.com>


DenWolf:
I agree with the code change, but before setting CR+1 and merging this patch I suggest to double check functionality and make manual testing of the change.
I already shared instruction about testing with you. If all will be OK (I think it will be :) – I’ll set CR+1 ASAP

PR #1723, opened 2019-10-11
[H265] Don't change CF, if profile/entrypoint unsupported

semelianova:
Return MFX_ERR_INVALID_VIDEO_PARAM status, if profile/entrypoint unsupported
instead of replacing requested color format to get supported guid. Return
MFX_ERR_INVALID_VIDEO_PARAM because it is called in Init.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
@@ -35,8 +35,6 @@ GUID GetGUID(MfxVideoParam const & par)
 {
     GUID guid = DXVA2_Intel_Encode_HEVC_Main;
 
-    mfxU16 cFamily = IsOn(par.mfx.LowPower);
-
     mfxU16 bdId = 0, cfId = 0;
 
 #if (MFX_VERSION >= 1027)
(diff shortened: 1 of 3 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
```diff
@@ -45,12 +45,20 @@ GUID GetGUID(MfxVideoParam const & par)
...
+        {
+            cfId = 0; // platforms below ICL do not support Main422/Main444 profile, using Main instead.
+        }
+        else
+        {
+            if (cFamily == 0 && cfId > 1)
+                cfId = 1; // ICL+ platforms do not support Main444 profile without VDEnc
+            if (cFamily > 0 && cfId == 1)
+                cfId = 0; // ICL+ platforms do not support Main422 profile on VDEnc
+        }
```
AntonGrishin: Can you please clarify why we need change color format?
I mean that if user ask for 422 encode with lowpower, why we don`t say  `MFX_ERR_UNSUPPORTED` in query, instead of change color format to 420?

semelianova: If we don't change it we can't create device and can't reach CheckVideoParam func

saosipov: I would agree with Anton. This part of code is to choose guid from input params.
We need this guid only to check if it is in the list of guids supported on the device.
If it is there we can create corresponding VideoDecoder.
Is it correct?
If yes then we will create incorrect VedeoDecoder with your change.

semelianova: I see two ways:
1. replace color format and further send it to query CheckVideoParam
2. don't change here but do like this commit https://github.com/Intel-Media-SDK/MediaSDK/commit/fd76d781ba3bc6f874a7c43b96ada291921c565f in CreateAccelerationService func to handle returned status

dmitryermilov: @saosipov , please reply

dmitryermilov: @saosipov , ping

Issue #10, opened 2017-06-14
Feature request: Remove dependency on proprietary libVA fork


vanvugt:
According to the readme:
"Intel Media SDK depends on a special version of LibVA which comes with Intel Media Server Studio installation and is not in upstream, so this version is not compatible with the LibVA/driver available at 01org."

So building this open source project is not yet possible with a purely open source environment (public upstream VA-API). It would be very helpful to desktop Linux distributions if future versions of MSDK could build without any proprietary downloads from intel.com. Just depend on upstream libVA, at least as an option.

onabiull:
Compatibility with public upstream LibVA definitely on our radar. I can't give any strong commitments, but you may expect that MediaSDK will be compatible with opensource LibVA at some point in future.

onabiull:
Could we close this one?

vanvugt:
I don't think so, unless it has been fixed?

onabiull:
Latest master depends on opensource LibVA: [https://github.com/01org/libva/commit/df544cd5a31e54d4cbd33a391795a8747ddaf789](url), so I guess yes, it is fixed :)

vanvugt:
Sounds good, but commit df544cd5a31e54d4cbd33a391795a8747ddaf789 does not appear to be related to this issue.

Is it fixed by some different commit?

onabiull:
It fixed in Media SDK commits [1340575fffa](https://github.com/Intel-Media-SDK/MediaSDK/commit/1340575fffa39ba0ff1f58ed762f7d6fde672f9c) and [934c36b9d0f](https://github.com/Intel-Media-SDK/MediaSDK/commit/934c36b9d0f01af04de23c35d63b5916ee7b3102). Essentially we switched from proprietary LibVA to publicly-available LibVA 2.1. Moreover appropriate LibVA user mode driver also became open source, see [https://github.com/intel/media-driver](url).

vanvugt:
That sounds right. Yes, sounds closed.

PR #1725, opened 2019-10-12
Reallocate frame (system memory or vaapi surface) for VP9 DRC decoding.

### LINKED ISSUE #1554: vp9 decoder supports increasing resolution on non-keyframe incorrectly
command line:
Sample_decode vp9 -i input.ivf -o output.yuv

MSDK returns MFX_ERR_INCOMPATIBLE_VIDEO_PARAM or MFX_ERR_REALLOC_SURFACE when the resolution is changed from lower to higher

https://github.com/Intel-Media-SDK/MediaSDK/blob/1160bfa30d6b7c6658f9c7189d31254df94393d5/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp#L1020-L1024

And Sample_decode doesn't handle MFX_ERR_REALLOC_SURFACE, or calls Pipeline.ResetDecoder() for MFX_ERR_INCOMPATIBLE_VIDEO_PARAM, which will result in wrong result on non-keyframe because the existing references are deleted by Pipeline.ResetDecoder()

https://github.com/Intel-Media-SDK/MediaSDK/blob/1160bfa30d6b7c6658f9c7189d31254df94393d5/samples/sample_decode/src/sample_decode.cpp#L709-L743

wangyan42164:
Fixes #1554.


xhaihao:
Does this patch support external allocator?

wangyan-intel:
> Does this patch support external allocator?

VaapiFrameAllocator is run by external allocator path in pipeline of sample decoder. 
I am not sure if it is your asked? 
Or you can provide your gst/ffmpeg command for reference.

wangyan-intel:
> Does this patch support external allocator? Note both ffmpeg-qsv and gst-msdk set an external frame allocator by using MFXVideoCORE_SetFrameAllocator.

I think ffmpeg-qsv and gst-msdk may need support reallocate function and handling of MFX_ERR_REALLOC_SURFACE like sample_decode.

wangyan42164:
Please provide gst-msdk/ffmpeg-qsv command line to check. Thanks.

xhaihao:
> > Does this patch support external allocator? Note both ffmpeg-qsv and gst-msdk set an external frame allocator by using MFXVideoCORE_SetFrameAllocator.
> 
> I think ffmpeg-qsv and gst-msdk may need support reallocate function and handling of MFX_ERR_REALLOC_SURFACE like sample_decode.

Got it, let's validate sample code firstly.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1884,6 +1904,36 @@ mfxStatus CDecodingPipeline::RunDecoding()
...
                 pBitstream = NULL;
+                sts = MFX_ERR_NONE;
+                continue;
+            } else if (MFX_ERR_REALLOC_SURFACE == sts) {
+                mfxVideoParam param;
+                m_pmfxDEC->GetVideoParam(&param);
+
+                m_pCurrentFreeSurface->frame.Info.CropW = param.mfx.FrameInfo.CropW;
+                m_pCurrentFreeSurface->frame.Info.CropH = param.mfx.FrameInfo.CropH;
+                m_YUV_Width = MSDK_ALIGN16(std::max(param.mfx.FrameInfo.Width, (mfxU16)m_YUV_Width));
```
IlyaSergeyev: Use MSDK_MAX instead of std::max here to avoid windows issues.

samples/sample_common/src/d3d11_allocator.cpp
```diff
@@ -374,6 +374,16 @@ mfxStatus D3D11FrameAllocator::ReleaseResponse(mfxFrameAllocResponse *response)
...
     return MFX_ERR_NONE;
 }
+
+mfxStatus D3D11FrameAllocator::ReallocImpl(mfxMemId /*mid*/, const mfxFrameInfo *info, mfxU16 /*memType*/, mfxMemId *midOut)
+{
+    if (!info || !midOut)
+      return MFX_ERR_NULL_PTR;
+
+    //TODO: Need add implementation in the future.
+    return MFX_ERR_NONE;
```
IlyaSergeyev: MFX_ERR_UNSUPPORTED

samples/sample_common/src/d3d_allocator.cpp
```diff
@@ -317,6 +317,15 @@ mfxStatus D3DFrameAllocator::ReleaseResponse(mfxFrameAllocResponse *response)
     return sts;
 }
 
+mfxStatus D3DFrameAllocator::ReallocImpl(mfxMemId /*mid*/, const mfxFrameInfo *info, mfxU16 /*memType*/, mfxMemId *midOut)
+{
+    if (!info || !midOut)
+      return MFX_ERR_NULL_PTR;
+
+    //TODO: Need add implementation in the future.
+    return MFX_ERR_NONE;
```
IlyaSergeyev: MFX_ERR_UNSUPPORTED

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will change it.

samples/sample_common/src/d3d11_allocator.cpp
wangyan42164: Sure. I will change it.

samples/sample_common/src/d3d_allocator.cpp
wangyan42164: Sure. I will change it.

IlyaSergeyev:
Sorry, MSDK_MAX was depricated. So, the right way is modify windows project files with NOMINMAX

wangyan-intel:
> Sorry, MSDK_MAX was depricated. So, the right way is modify windows project files with NOMINMAX

Sure. I will roll back. Thanks

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1884,6 +1904,36 @@ mfxStatus CDecodingPipeline::RunDecoding()
...
+                ReallocSurface(inMid, &m_pCurrentFreeSurface->frame.Info, m_pCurrentFreeSurface->frame.Data.MemType, &outMid);
+
+                // We have to lock surface in case of system memory allocator to have right pitches in Data
+                // (that can be needed for INTERNAL->SYS copy operations)
+                // Better way is locking surface in decoder, but we don't know real memid inside decoders (in case of sysmem)
+                // It's a good point for improvement
+                if (m_pCurrentFreeSurface->frame.Data.MemType & MFX_MEMTYPE_SYSTEM_MEMORY)
+                {
+                    m_pGeneralAllocator->Lock(m_pGeneralAllocator->pthis, outMid, &(m_pCurrentFreeSurface->frame.Data));
+                    m_pGeneralAllocator->Unlock(m_pGeneralAllocator->pthis, outMid, NULL);
```
dvrogozh: Looks weird. You carry out information out of lock/unlock scope. I feel this should be done in some other way.

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -141,6 +141,102 @@ mfxStatus vaapiFrameAllocator::Close()
     return BaseFrameAllocator::Close();
 }
 
+mfxStatus vaapiFrameAllocator::ReallocImpl(mfxMemId mid, const mfxFrameInfo *info, mfxU16 memType, mfxMemId *midOut)
```
dmitryermilov: it's almost a copy/past of AllocImpl(). I mean code duplication. It should be avoided. 
I'd suggest to define a common part of AllocImpl and ReallocImpl and move it to a separate private function.

samples/sample_common/src/sysmem_allocator.cpp
```diff
@@ -241,71 +241,129 @@ mfxStatus SysMemFrameAllocator::CheckRequestType(mfxFrameAllocRequest *request)
...
     case MFX_FOURCC_NV12:
         nbytes = Width2*Height2 + (Width2>>1)*(Height2>>1) + (Width2>>1)*(Height2>>1);
         break;
-    case MFX_FOURCC_NV16:
-        nbytes = Width2*Height2 + (Width2>>1)*(Height2) + (Width2>>1)*(Height2);
-        break;
-#if (MFX_VERSION >= 1028)
-    case MFX_FOURCC_RGB565:
-        nbytes = 2*Width2*Height2;
+    case MFX_FOURCC_P010:
```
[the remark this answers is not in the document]
wangyan42164: Just the order of switch branch is changed. I don't change it. I will restore it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: It is necessary to calculate the offset of Y/U/V address. If we have better solution, I can change it. Thanks.

samples/sample_common/src/vaapi_allocator.cpp
wangyan42164: I can move the part of get VA FOURCC into a common private function. Thanks.

wangyan42164:
After introduce #1728, I can find windows build is passed. But now it display some OpenCL warning as error. It seems not to be related with this PR.

IlyaSergeyev:
> After introduce #1728, I can find windows build is passed. But now it display some OpenCL warning as error. It seems not to be related with this PR.

I see
```
pipeline_decode.cpp(1917): error C2220: warning treated as error - no 'object' file generated 
pipeline_decode.cpp(1917): warning C4244: '=': conversion from 'mfxU32' to 'mfxU16', possible loss of data 
pipeline_decode.cpp(1918): warning C4244: '=': conversion from 'mfxU32' to 'mfxU16', possible loss of data

```
in the log

wangyan-intel:
> > After introduce #1728, I can find windows build is passed. But now it display some OpenCL warning as error. It seems not to be related with this PR.
> 
> I see
> 
> ```
> pipeline_decode.cpp(1917): error C2220: warning treated as error - no 'object' file generated 
> pipeline_decode.cpp(1917): warning C4244: '=': conversion from 'mfxU32' to 'mfxU16', possible loss of data 
> pipeline_decode.cpp(1918): warning C4244: '=': conversion from 'mfxU32' to 'mfxU16', possible loss of data
> ```
> 
> in the log

Thanks for your comments.
I have changed it.
Windows build seems OK now.

samples/sample_decode/src/pipeline_decode.cpp
dvrogozh: I still don't see why this is needed. If you use external allocator (even with system memory), then you _must_ call Lock before using the memory and Unlock once you are done. This is what mediasdk library should do. If it is doing something else and tries directly access system memory - it's just a bug in mediasdk.

Can you, please, remove this code and provide details where you will see the crash?

wangyan-intel:
I will double check it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan-intel: I will check it. Thanks

wangyan42164: CDecodingPipeline::AllocFrames() will call LockFrame to calculate UV offset and pitch after create surface.
So after reallocate the surface, we also need call LockFrame() to calculate them. 
May it be better to move this calling into CDecodingPipeline::ReallocSurface()? If yes, I can change it.

wangyan42164: I also check the possibility to call LockFrame in Decoder. But VideoDECODEVP9_HW::m_core only keep internal memory ID and corresponding allocator reference. So it seems that it cannot handle the MemId from sample_decode.

wangyan42164: Attach the full back trace when crash for reference:
Thread 2 "sample_decode" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff515e700 (LWP 18140)]
0x00007ffff54119dd in _mm_storeu_si128(long long __vector(2)*, long long __vector(2)) (__B=..., __P=0x7ffff7fd7000) at /usr/lib/gcc/x86_64-redhat-linux/8/include/emmintrin.h:720
720	  *__P = __B;
Missing separate debuginfos, use: dnf debuginfo-install libX11-1.6.5-7.fc28.x86_64 libXau-1.0.8-11.fc28.x86_64 libXext-1.3.3-8.fc28.x86_64 libXfixes-5.0.3-5.fc28.x86_64 libgcc-8.2.1-4.fc28.x86_64 libpciaccess-0.13.4-8.fc28.x86_64 libstdc++-8.1.1-5.fc28.x86_64 libxcb-1.13-1.fc28.x86_64
(gdb) bt
#0  0x00007ffff54119dd in _mm_storeu_si128(long long __vector(2)*, long long __vector(2)) (__B=..., __P=0x7ffff7fd7000) at /usr/lib/gcc/x86_64-redhat-linux/8/include/emmintrin.h:720
#1  copyVideoToSys (src=0x7ffff7fb9480 <error: Cannot access memory at address 0x7ffff7fb9480>, dst=0x7ffff7fd6fe0 "", width=544)
    at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/include/fast_copy.h:64
#2  0x00007ffff5412c1d in mfxCopyRect<unsigned char> (pSrc=0x7ffff7fb9480 <error: Cannot access memory at address 0x7ffff7fb9480>, srcStep=640, pDst=0x7ffff7fd6fe0 "", dstStep=576, roiSize=..., flag=2)
    at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/include/fast_copy.h:214
#3  0x00007ffff54123e9 in FastCopy::Copy (pDst=0x7ffff7fb90a0 <error: Cannot access memory at address 0x7ffff7fb90a0>, dstPitch=576, pSrc=0x7ffff7f98000 <error: Cannot access memory at address 0x7ffff7f98000>, 
    srcPitch=640, roi=..., flag=2) at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/include/fast_copy.h:250
#4  0x00007ffff540fa65 in CoreDoSWFastCopy (pDst=0x7ffff515d720, pSrc=0x7ffff515d7e0, copyFlag=2) at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/src/libmfx_core.cpp:1321
#5  0x00007ffff541e653 in VAAPIVideoCORE::DoFastCopyExtended (this=0x5555557ec470, pDst=0x7ffff515d720, pSrc=0x7ffff515d7e0)
    at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/src/libmfx_core_vaapi.cpp:1178
#6  0x00007ffff541e010 in VAAPIVideoCORE::DoFastCopyWrapper (this=0x5555557ec470, pDst=0x555555826bc0, dstMemType=66, pSrc=0x7ffff515d9a0, srcMemType=17)
    at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/src/libmfx_core_vaapi.cpp:1032
#7  0x00007ffff542628e in mfx_UMC_FrameAllocator_D3D::PrepareToOutput (this=0x55555581c8c0, surface_work=0x555555826bc0, index=2, isOpaq=false)
    at /home/yanwang/open_media_workspace/MediaSDK/_studio/shared/src/mfx_umc_alloc_wrapper.cpp:1108
#8  0x00007ffff5501ed2 in VP9DECODERoutine (p_state=0x555555851000) at /home/yanwang/open_media_workspace/MediaSDK/_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp:852
#9  0x00007ffff53ac94c in mfxSchedulerCore::call_pRoutine (this=0x5555557ec940, call=...) at /home/yanwang/open_media_workspace/MediaSDK/_studio/mfx_lib/scheduler/include/mfx_scheduler_core.h:500
#10 0x00007ffff53ac687 in mfxSchedulerCore::ThreadProc (this=0x5555557ec940, pContext=0x555555806c88)
    at /home/yanwang/open_media_workspace/MediaSDK/_studio/mfx_lib/scheduler/src/mfx_scheduler_core_thread.cpp:80

dvrogozh: Ok, so, if I read code correctly you actually should have called LockFrame here:

https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/_studio/shared/src/libmfx_core_vaapi.cpp#L987

Can you, please, check:
1) Was it actually called or not? If not - we likely use cached pointers from the previous allocation.
2) If it was called, did we somehow get pointers from prev. allocation? I am not sure why this could happen though....

wangyan42164: Sure. I will check it. Thanks.

dmitryermilov:
> I still don't see why this is needed. If you use external allocator (even with system memory), then you must call Lock before using the memory and Unlock once you are done. This is what mediasdk library should do. If it is doing something else and tries directly access system memory - it's just a bug in mediasdk.
> 
> Can you, please, remove this code and provide details where you will see the crash?

sample_decode "locks" surfaces in system memory once at Init():
https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/samples/sample_decode/src/pipeline_decode.cpp#L1250
https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/samples/sample_decode/src/pipeline_decode.cpp#L1267

IMO this is a a bug!

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: I have investigated it and updated this PR. About your questions, I add my comments in the following: 

> Ok, so, if I read code correctly you actually should have called LockFrame here:
> 
> https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/_studio/shared/src/libmfx_core_vaapi.cpp#L987
> 
> Can you, please, check:
> 
> 1. Was it actually called or not? If not - we likely use cached pointers from the previous allocation.
Currently it isn't called because of dstPtr != nullptr. CDecodingPipeline::AllocFrames() will call LockFrame after create surface.
> 2. If it was called, did we somehow get pointers from prev. allocation? I am not sure why this could happen though....
After I removed LockFrame in CDecodingPipeline::AllocFrames(), this can be called but no available external frame allocator because sample_decode will not set external frame allocator when using system memory in CDecodingPipeline::CreateAllocator(). 
https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/samples/sample_decode/src/pipeline_decode.cpp#L1386.
After I added it in the latest update of this PR, this issue is fixed.

wangyan42164: Currently the only possible issue is that VideoDECODEVP9_HW::Init() will try to create internal frames by external frame allocator (system memory allocator) but failed because it wants to create MFX_MEMTYPE_DXVA2_DECODER_TARGET surface for VAAPI acceleration. After it get MFX_ERR_UNSUPPORTED from system memory allocator, it will fall back to use default allocator in the following line. So it may be OK.
https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/_studio/shared/src/libmfx_core_vaapi.cpp#L584

wangyan42164:
> > I still don't see why this is needed. If you use external allocator (even with system memory), then you must call Lock before using the memory and Unlock once you are done. This is what mediasdk library should do. If it is doing something else and tries directly access system memory - it's just a bug in mediasdk.
> > Can you, please, remove this code and provide details where you will see the crash?
> 
> sample_decode "locks" surfaces in system memory once at Init():
> https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/samples/sample_decode/src/pipeline_decode.cpp#L1250
> 
> 
> https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/samples/sample_decode/src/pipeline_decode.cpp#L1267
> 
> IMO this is a a bug!

Yes. I have replied @dvrogozh. I have removed this Lock and works fine after set external frame allocator. Thanks.

dvrogozh:
> IMO this is a a bug!

Well. I would say that this is awfully misleading implementation! Basically there are 2 ways to work with mediasdk on system memory:
1. Create and set external allocator. In this case memory access is allowed only between Lock()/Unlock() calls
2. Do not set external allocator at all and pass mfxFrameSurface1 with the correct data pointers and other meta-data (pitch) which describes system memory

Option 2 can be implemented via ever-locked frame, i.e. you create memory with some allocator, then immediately lock it and use forever. That's what our sample is doing. Would I ever recommend such approach? Definitely NO! If there is an allocator - use allocator. Our sample looks awfully strange from this perspective: we create _valid_ mediasdk external allocator, but we don't setup it. If someone will really have custom system memory allocator, then crappy things might start to happen without regular Lock/Unlock if mediasdk is paired with some other component.

I believe we need to revise approach on how we work with memory in our samples. In the scope of the current PR we have 2 options:
1. Continue current approach from this perspective @wangyan42164 is actually almost correct, but at least Unlock should be removed as far as I understand
2. Start to use external allocator and proper Lock/Unlock

I strongly vote for the 2nd option. @dmitryermilov, how do you think?

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1884,6 +1897,23 @@ mfxStatus CDecodingPipeline::RunDecoding()
...
+
+                m_pCurrentFreeSurface->frame.Info.CropW = param.mfx.FrameInfo.CropW;
+                m_pCurrentFreeSurface->frame.Info.CropH = param.mfx.FrameInfo.CropH;
+                m_YUV_Width = MSDK_ALIGN16(std::max(param.mfx.FrameInfo.Width, (mfxU16)m_YUV_Width));
+                m_YUV_Height = MSDK_ALIGN16(std::max(param.mfx.FrameInfo.Height, (mfxU16)m_YUV_Height));
+                m_pCurrentFreeSurface->frame.Info.Width = m_YUV_Width;
+                m_pCurrentFreeSurface->frame.Info.Height = m_YUV_Height;
+
+                inMid = m_pCurrentFreeSurface->frame.Data.MemId;
+                ReallocSurface(inMid, &m_pCurrentFreeSurface->frame.Info, m_pCurrentFreeSurface->frame.Data.MemType, &outMid);
```
dvrogozh: 1. error check is missing
2. Rename the function and call as: `ReallocCurrentSurface(param.mfx.FrameInfo)`

```
mfxStatus ReallocCurrentSurface(mfxFrameInfo& info)
{
...
}
```

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1384,6 +1391,10 @@ mfxStatus CDecodingPipeline::CreateAllocator()
             MSDK_CHECK_STATUS(sts, "m_hwdev->GetHandle failed");
             sts = m_mfxSession.SetHandle(MFX_HANDLE_VA_DISPLAY, va_dpy);
             MSDK_CHECK_STATUS(sts, "m_mfxSession.SetHandle failed");
+            sts = m_mfxSession.SetFrameAllocator(m_pGeneralAllocator);
```
dvrogozh: I think this should be moved out of #ifdef LIBVA_SUPPORT and done always for the system memory path.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1384,6 +1391,10 @@ mfxStatus CDecodingPipeline::CreateAllocator()
...
+            MSDK_CHECK_STATUS(sts, "m_mfxSession.SetFrameAllocator failed");
+
+            m_bExternalAlloc = true;
```
dvrogozh: Look into comments on lines 1400-1406 below - they become not relevant considering the change you are making here. Please, remove these comment entirely.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1884,6 +1897,23 @@ mfxStatus CDecodingPipeline::RunDecoding()
                 bErrIncompatibleVideoParams = true;
                 // need to go to the buffering loop prior to reset procedure
                 pBitstream = NULL;
+                sts = MFX_ERR_NONE;
+                continue;
+            } else if (MFX_ERR_REALLOC_SURFACE == sts) {
+                mfxVideoParam param;
+                m_pmfxDEC->GetVideoParam(&param);
```
dvrogozh: You never update m_mfxVideoParams. Why?

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1884,6 +1897,23 @@ mfxStatus CDecodingPipeline::RunDecoding()
...
+
+                m_pCurrentFreeSurface->frame.Info.CropW = param.mfx.FrameInfo.CropW;
+                m_pCurrentFreeSurface->frame.Info.CropH = param.mfx.FrameInfo.CropH;
+                m_YUV_Width = MSDK_ALIGN16(std::max(param.mfx.FrameInfo.Width, (mfxU16)m_YUV_Width));
```
dvrogozh: Why do you actually need m_YUV_Width and m_YUV_Height as class members? Can we write something like:
```
} else if (MFX_ERR_REALLOC_SURFACE == sts) {
    mfxVideoParam param{};
    m_pmfxDEC->GetVideoParam(&param);

    m_mfxVideoParams.mfx.FrameInfo.Width = MSDK_ALIGN16(std::max(m_mfxVideoParams.mfx.FrameInfo.Width, param.mfx.FrameInfo.Width));
    m_mfxVideoParams.mfx.FrameInfo.Height = MSDK_ALIGN16(std::max(m_mfxVideoParams.mfx.FrameInfo.Height, param.mfx.FrameInfo.Height));

    ....
}
```

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1884,6 +1897,23 @@ mfxStatus CDecodingPipeline::RunDecoding()
                 bErrIncompatibleVideoParams = true;
                 // need to go to the buffering loop prior to reset procedure
                 pBitstream = NULL;
+                sts = MFX_ERR_NONE;
+                continue;
+            } else if (MFX_ERR_REALLOC_SURFACE == sts) {
+                mfxVideoParam param;
+                m_pmfxDEC->GetVideoParam(&param);
+
+                m_pCurrentFreeSurface->frame.Info.CropW = param.mfx.FrameInfo.CropW;
```
dvrogozh: I think the update of `m_pCurrentFreeSurface` info should be done _inside_ `ReallocSurface()` function.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1701,6 +1712,8 @@ mfxStatus CDecodingPipeline::RunDecoding()
     CTimeInterval<>     decodeTimer(m_bIsCompleteFrame);
     time_t              start_time = time(0);
     std::thread         deliverThread;
+    mfxMemId            inMid = nullptr;
+    mfxMemId            outMid = nullptr;
```
dvrogozh: why do you need these declarations here? they look to be limited scope ones. Move below to the scope where you actually use them.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will change it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will change it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will remove the comment entirely. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will change it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: I will change it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will change it. Thanks

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I will change it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: In the updated latest version, m_mfxVideoParams will be updated in mfxStatus ReallocCurrentSurface(mfxFrameInfo& info). Thanks.

wangyan42164:
@dvrogozh I have refined this PR based on your latest comments. Could you please review it? Thanks.

dmitryermilov:
> > IMO this is a a bug!
> 
> Well. I would say that this is awfully misleading implementation! Basically there are 2 ways to work with mediasdk on system memory:
> 
> 1. Create and set external allocator. In this case memory access is allowed only between Lock()/Unlock() calls
> 2. Do not set external allocator at all and pass mfxFrameSurface1 with the correct data pointers and other meta-data (pitch) which describes system memory
> 
> Option 2 can be implemented via ever-locked frame, i.e. you create memory with some allocator, then immediately lock it and use forever. That's what our sample is doing. Would I ever recommend such approach? Definitely NO! If there is an allocator - use allocator. Our sample looks awfully strange from this perspective: we create _valid_ mediasdk external allocator, but we don't setup it. If someone will really have custom system memory allocator, then crappy things might start to happen without regular Lock/Unlock if mediasdk is paired with some other component.
> 
> I believe we need to revise approach on how we work with memory in our samples. In the scope of the current PR we have 2 options:
> 
> 1. Continue current approach from this perspective @wangyan42164 is actually almost correct, but at least Unlock should be removed as far as I understand
> 2. Start to use external allocator and proper Lock/Unlock
> 
> I strongly vote for the 2nd option. @dmitryermilov, how do you think?

Ideally (according to the man and from clarify of sample_decode code perspective) option №2 is better. But it most likely won't work because of such code:
https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/_studio/shared/src/libmfx_core_vaapi.cpp#L1110
https://github.com/Intel-Media-SDK/MediaSDK/blob/45b78442f7da338e69b2774cd448047c1c8b2055/_studio/shared/src/libmfx_core_vaapi.cpp#L1143

Yes, we can fix it as well but may break legacy behavior for 3rd party apps.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1886,6 +1904,17 @@ mfxStatus CDecodingPipeline::RunDecoding()
                 pBitstream = NULL;
                 sts = MFX_ERR_NONE;
                 continue;
+            } else if (MFX_ERR_REALLOC_SURFACE == sts) {
+                mfxVideoParam param;
```
dvrogozh: this needs to be initialized, add {} like this:
`mfxVideoParam param{};`

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1886,6 +1904,17 @@ mfxStatus CDecodingPipeline::RunDecoding()
...
+                m_pmfxDEC->GetVideoParam(&param);
```
dvrogozh: add error check for GetVideoParam

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I have added it. Thanks.

samples/sample_decode/src/pipeline_decode.cpp
wangyan42164: Sure. I have added it. Thanks.

wangyan42164:
@dvrogozh I have refined this PR based on your latest comments. Could you please review it? Thanks.

dmitryermilov:
> +1 from code review perspective. However, I would recommend to extract a change to setup allocator for system memory to the separate commit within this PR - this is a risky change, better to have it separately.
> 
> @dmitryermilov : this change requires full range of testing. Can you assure we give it a try?
> 
> @wangyan42164 : can you, please, verify that this change works on the range of sample decode modes? Please, take the VP9 stream and try:
> 
> 1. decoding + VP
> 2. onscreen rendering

Seems there are no regressions in sample_decode test suites.

wangyan42164:
> +1 from code review perspective. However, I would recommend to extract a change to setup allocator for system memory to the separate commit within this PR - this is a risky change, better to have it separately.
> 
> @dmitryermilov : this change requires full range of testing. Can you assure we give it a try?
> 
> @wangyan42164 : can you, please, verify that this change works on the range of sample decode modes? Please, take the VP9 stream and try:
> 
> 1. decoding + VP
> 2. onscreen rendering
I tried them on my KBL platform without error. I will separate this PR into 2 ones and submit them again. Thanks.

wangyan42164:
@dvrogozh @dmitryermilov I have separated this PR into 2 commits. Could you please review them. Thanks.

wangyan42164:
@dvrogozh @dmitryermilov I find my separation of 2 commits loses one line. Now I have updated it and tested it locally again successfully.  Sorry for inconvenience. Could you please review them again? Thanks.

PR #1728, opened 2019-10-14
[sample_decode] allow to use std::min/max in windows

IlyaSergeyev:


samples/sample_decode/sample_decode.vcxproj
```diff
@@ -94,7 +94,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(ProjectDir)\include;$(ProjectDir)\..\..\api\include;$(INTELMEDIASDKROOT)\include;$(ProjectDir)\..\sample_common\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>_WIN32;WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>NOMINMAX;_WIN32;WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
dvrogozh: @mgonchar : I thought you did that just a week ago?

mgonchar: @dvrogozh yes. I believe this is some leftover which doesn't affect CI for some reason

dvrogozh: That's `Debug|Win32`. I wonder which build configurations we test in CI? @eponomar : can you comment?

IlyaSergeyev: These defines was not needed for sample_decode in current state, because there are no std::min / std::max in sample_decode source. But for future commits (like https://github.com/Intel-Media-SDK/MediaSDK/pull/1725) it will be needed. Maybe we should merge these requests?

eponomar: > That's `Debug|Win32`. I wonder which build configurations we test in CI? @eponomar : can you comment?

We don't have any tests for windows in open source.

dvrogozh: That's not about the test. That's about the build. Which Windows build configurations we support in CI?

dvrogozh: @IlyaSergeyev : I generally prefer to avoid non-motivated changes which are not needed in the current state of the project. From this perspective, if  #1725 needs this change, then #1725 should introduce this change. I suggest to ping owner of #1725 and ask him to cherry-pick your commit.

wangyan-intel: I can introduce this change into #1725. Thanks

wangyan42164: I have introduced this change into #1725. Windows build passed. Thanks

IlyaSergeyev: @dvrogozh Agree, it's better to introduce these changes in #1725

PR #1731, opened 2019-10-15
[vpp] Removed shallow copy

dmitryermilov:
VPP shall not make shallow copy of ExtParam's from opaque
to native(real) surfaces. If app decides to change the opaque surface,
internal native surface can't be changed automatically.


dmitryermilov:
This is a regression caused by https://github.com/Intel-Media-SDK/MediaSDK/commit/9b5b240803c9a4ecfd3791a541f1753f799b7e8e/_studio/mfx_lib/vpp/include/mfx_vpp_main.h

PR #1734, opened 2019-10-17
sample_encode: Added encoding in qpfile mode

pgribov:
sample_encode reads file with information for each frame (display order, QP,
frame type) and encodes stream in encoded order using this readed information.

### DIFF
--- a/samples/sample_common/include/sample_utils.h
+++ b/samples/sample_common/include/sample_utils.h
@@ -28,6 +28,7 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #include <stdexcept>
 #include <mutex>
 #include <algorithm>
+#include <fstream>
 
 #include "mfxstructures.h"
 #include "mfxvideo.h"
@@ -109,6 +110,107 @@ enum ExtBRCType {
     EXTBRC_ON,
     EXTBRC_IMPLICIT
 };
+
+namespace QPFile {
+
+    enum ReaderStatus
+    {
+        READER_ERR_NONE,
+        READER_ERR_NOT_INITIALIZED,
+        READER_ERR_CODEC_UNSUPPORTED,
+        READER_ERR_FILE_NOT_OPEN,
+        READER_ERR_INCORRECT_FILE
+    };
+
+    struct FrameInfo
+    {
+        mfxU32 displayOrder;
+        mfxU16 QP;
+        mfxU16 frameType;
+    };
+
+    // QPFile::Reader reads QP and frame type per frame in encoding order
+    // from external text file (for encoding in qpfile mode)
+    class Reader
+    {
+    public:
+        mfxStatus Read(const msdk_string& strFileName, mfxU32 codecid);
+        void ResetState();
+
+        mfxU32 GetCurrentEncodedOrder() const;
+        mfxU32 GetCurrentDisplayOrder() const;
+        mfxU16 GetCurrentQP() const;
+        mfxU16 GetCurrentFrameType() const;
+        mfxU32 GetFramesNum() const;
+        void NextFrame();
+        std::string GetErrorMessage() const;
+
+    private:
+        void ResetState(ReaderStatus set_sts);
+
+        ReaderStatus            m_ReaderSts   = READER_ERR_NOT_INITIALIZED;
+        mfxU32                  m_nFrames     = std::numeric_limits<mfxU32>::max();
+        mfxU32                  m_CurFrameNum = std::numeric_limits<mfxU32>::max();
+        std::vector<FrameInfo>  m_FrameVals {};
+    };
+
+    inline bool get_line(std::ifstream& ifs, std::string& line)
+    {
+        std::getline(ifs, line, '\n');
+        if (!line.empty() && line.back() == '\r')
+            line.pop_back();
+        return !ifs.fail();
+    }
+    inline size_t find_nth(const std::string& str, size_t pos, const std::string& needle, mfxU32 nth)
+    {
+        size_t found_pos = str.find(needle, pos);
+        for(; nth != 0 && std::string::npos != found_pos; --nth)
+            found_pos = str.find(needle, found_pos + 1);
+        return found_pos;
+    }
+    inline mfxU16 StringToFrameType(std::string str)
+    {
+        if ("IDR_REF" == str) return MFX_FRAMETYPE_I | MFX_FRAMETYPE_IDR | MFX_FRAMETYPE_REF;
+        else if ("I_REF" == str) return MFX_FRAMETYPE_I | MFX_FRAMETYPE_REF;
+        else if ("P_REF" == str) return MFX_FRAMETYPE_P | MFX_FRAMETYPE_REF;
+        else if ("P" == str) return MFX_FRAMETYPE_P;
+        else if ("B_REF" == str) return MFX_FRAMETYPE_B | MFX_FRAMETYPE_REF;
+        else if ("B" == str) return MFX_FRAMETYPE_B;
+        else return MFX_FRAMETYPE_UNKNOWN;
+    }
+    inline std::string ReaderStatusToString(ReaderStatus sts)
+    {
+        switch (sts)
+        {
+        case READER_ERR_NOT_INITIALIZED:
+            return std::string("reader not initialized (qpfile has not yet read the file)\n");
+        case READER_ERR_FILE_NOT_OPEN:
+            return std::string("failed to open file contains frame parameters (check provided path in -qpfile <path>)\n");
+        case READER_ERR_INCORRECT_FILE:
+            return std::string("incorrect file with frame parameters\n");
+        case READER_ERR_CODEC_UNSUPPORTED:
+            return std::string("codecs, except h264 and h265, are not supported\n");
+        default:
+            return std::string();
+        }
+    }
+    inline mfxU32 ReadDisplayOrder(const std::string& line)
+    {
+        return std::stoi(line.substr(0, find_nth(line, 0, ",", 0)));
+    }
+    inline mfxU16 ReadQP(const std::string& line)
+    {
+        size_t pos = find_nth(line, 0, ",", 0) + 1;
+        return static_cast<mfxU16>(std::stoi(line.substr(pos, find_nth(line, 0, ",", 1) - pos)));
+    }
+    inline mfxU16 ReadFrameType(const std::string& line)
+    {
+        size_t pos = find_nth(line, 0, ",", 1) + 1;
+        return StringToFrameType(line.substr(pos, line.length() - pos));
+    }
+}
+
+mfxStatus GetFrameLength(mfxU16 width, mfxU16 height, mfxU32 ColorFormat, mfxU32 &length);
 
 bool IsDecodeCodecSupported(mfxU32 codecFormat);
 bool IsEncodeCodecSupported(mfxU32 codecFormat);
@@ -677,6 +779,7 @@ public :
 
     virtual void Close();
     virtual mfxStatus Init(std::list<msdk_string> inputs, mfxU32 ColorFormat, bool shouldShiftP010=false);
+    virtual mfxStatus SkipNframesFromBeginning(mfxU16 w, mfxU16 h, mfxU32 viewId, mfxU32 nframes);
     virtual mfxStatus LoadNextFrame(mfxFrameSurface1* pSurface);
     virtual void Reset();
     mfxU32 m_ColorFormat; // color format of input YUV data, YUV420 or NV12
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -21,7 +21,6 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 
 #include <math.h>
 #include <iostream>
-#include <fstream>
 #include <algorithm>
 #include <map>
 
@@ -66,6 +65,29 @@ mfxStatus CopyBitstream2(mfxBitstream *dest, mfxBitstream *src)
     return MFX_ERR_NONE;
 }
 
+mfxStatus GetFrameLength(mfxU16 width, mfxU16 height, mfxU32 ColorFormat, mfxU32 &length)
+{
+    switch (ColorFormat)
+    {
+    case MFX_FOURCC_NV12:
+    case MFX_FOURCC_I420:
+        length = 3 * width * height / 2;
+        break;
+    case MFX_FOURCC_YUY2:
+        length = 2 * width * height;
+        break;
+    case MFX_FOURCC_RGB4:
+        length = 4 * width * height;
+        break;
+    case MFX_FOURCC_P010:
+        length = 3 * width * height;
+        break;
+    default:
+        return MFX_ERR_UNSUPPORTED;
+    }
+
+    return MFX_ERR_NONE;
+}
 
 CSmplYUVReader::CSmplYUVReader()
 {
@@ -150,6 +172,23 @@ void CSmplYUVReader::Reset()
     {
         fseek(m_files[i], 0, SEEK_SET);
     }
+}
+
+mfxStatus CSmplYUVReader::SkipNframesFromBeginning(mfxU16 w, mfxU16 h, mfxU32 viewId, mfxU32 nframes)
+{
+    // change file position for read from beginning to "frameLength * nframes".
+    mfxU32 frameLength;
+
+    if (MFX_ERR_NONE != GetFrameLength(w, h, m_ColorFormat, frameLength))
+    {
+        msdk_printf(MSDK_STRING("Input color format %s is unsupported in qpfile mode\n"), ColorFormatToStr(m_ColorFormat));
+        return MFX_ERR_UNSUPPORTED;
+    }
+
+    if (0 != fseek(m_files[viewId], frameLength * nframes, SEEK_SET))
+        return MFX_ERR_MORE_DATA;
+
+    return MFX_ERR_NONE;
 }
 
 mfxStatus CSmplYUVReader::LoadNextFrame(mfxFrameSurface1* pSurface)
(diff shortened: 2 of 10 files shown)


samples/sample_common/include/sample_utils.h
```diff
@@ -110,6 +111,106 @@ enum ExtBRCType {
     EXTBRC_IMPLICIT
 };
 
+namespace QPFile {
+#   ifdef max
```
mgonchar: why do we need it? maybe you need to add ```NOMINMAX``` to proj file?

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -981,6 +986,17 @@ mfxStatus CEncodingPipeline::AllocFrames()
         }
     }
 
+    // prepare mfxEncodeCtrl array for encoder if qpfile mode is enabled
+    if (m_mfxEncParams.mfx.EncodedOrder)
+    {
+        m_pEncCtrls = new mfxEncodeCtrlWrap[m_EncResponse.NumFrameActual];
```
mgonchar: can we use ```std::vector``` here instead?

samples/sample_common/include/sample_utils.h
pgribov: It seems NOMINMAX have been already added (or where can i add it?). If i remove #undef, max macro overlaps std::numeric_limits::max.

samples/sample_encode/src/pipeline_encode.cpp
pgribov: replaced to std::vector

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2406,12 +2438,33 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
...
 }
 
+mfxStatus CEncodingPipeline::LoadNextControl(mfxEncodeCtrlWrap*& pCtrl, mfxU32 encSurfIdx)
+{
+    mfxStatus sts = MFX_ERR_NONE;
+
+    try
+    {
+        pCtrl = &m_EncCtrls[encSurfIdx];
+        pCtrl->QP = m_QPFileReader.GetCurrentQP();
```
mgonchar: you have to check pointer first

pgribov:
proposed changes made

dmitryermilov:
@pgribov , please address Miroslav's comments

pgribov:
@mgonchar

samples/sample_common/src/sample_utils.cpp
```diff
@@ -66,6 +65,71 @@ mfxStatus CopyBitstream2(mfxBitstream *dest, mfxBitstream *src)
...
+    case MFX_FOURCC_YUY2:
+    case MFX_FOURCC_RGB4:
+    case MFX_FOURCC_BGR4:
+    case MFX_FOURCC_AYUV:
+    case MFX_FOURCC_A2RGB10:
+#if (MFX_VERSION >= 1027)
+    case MFX_FOURCC_Y210:
+    case MFX_FOURCC_Y410:
+#endif 
+        switch (ColorFormat)
```
mgonchar: from common sense this function should use only one fourcc, which reflects color format of input stream

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1270,6 +1343,87 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+{
+    m_reader_sts = READER_ERR_NONE;
+    m_cur_frame_num = 0;
+
+    if (codecid != MFX_CODEC_AVC && codecid != MFX_CODEC_HEVC)
+    {
+        ResetState(READER_ERR_CODEC_UNSUPPORTED);
+        return MFX_ERR_NOT_INITIALIZED;
+    }
+    else
```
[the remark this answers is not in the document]
pgribov: removed

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1270,6 +1343,87 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+    {
+        std::ifstream ifs(strFileName, msdk_fstream::in);
+        if (!ifs.is_open())
+        {
+            ResetState(READER_ERR_FILE_NOT_OPEN);
+            return MFX_ERR_NOT_INITIALIZED;
+        }
+        else
```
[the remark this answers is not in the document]
pgribov: removed

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1270,6 +1343,87 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+                m_frame_vals.reserve(m_nframes);
+
+                while (QPFile::get_line(ifs, line))
+                {
+                    frame_info.display_order_num = QPFile::ReadDisplayOrder(line);
+                    frame_info.QP = QPFile::ReadQP(line);
+                    frame_info.frame_type = QPFile::ReadFrameType(line);
+                    if ( frame_info.display_order_num < 0 ||
+                         frame_info.display_order_num > m_nframes ||
+                         frame_info.QP < 0 || frame_info.QP > 51 ||
```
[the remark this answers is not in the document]
pgribov: changed

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -981,6 +986,17 @@ mfxStatus CEncodingPipeline::AllocFrames()
         }
     }
 
+    // prepare mfxEncodeCtrl array for encoder if qpfile mode is enabled
+    if (m_mfxEncParams.mfx.EncodedOrder)
+    {
+        m_EncCtrls.resize(m_EncResponse.NumFrameActual);
+        for (mfxU32 i = 0; i < m_EncResponse.NumFrameActual; ++i)
```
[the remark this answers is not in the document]
pgribov: changed

samples/sample_common/src/sample_utils.cpp
pgribov: i checked this moment, for correct reading reader need to know both input and encode fourcc (m_ColorFormat is FileInputFourCC, pInfo.FourCC in surface is EncodeFourCC)

mgonchar: you need both for proper placing data to surface. You don't need to know fourcc of surface to calculate size of frame in file. This function only calculate frame size

samples/sample_common/src/sample_utils.cpp
```diff
@@ -66,6 +65,33 @@ mfxStatus CopyBitstream2(mfxBitstream *dest, mfxBitstream *src)
     return MFX_ERR_NONE;
 }
 
+mfxStatus GetFrameLength(mfxFrameSurface1* const pSurface, mfxU32 ColorFormat, mfxU32 &length)
```
mgonchar: much better than was!

Still has a problem. Surface has no relation to file on disk, so you need to set up following signature:
```
mfxStatus GetFrameLength(mfxU32 width, mfxU32 height, mfxU32 ColorFormat, mfxU32 &length)
```

samples/sample_common/src/sample_utils.cpp
```diff
@@ -66,6 +65,33 @@ mfxStatus CopyBitstream2(mfxBitstream *dest, mfxBitstream *src)
...
+{
+    const mfxFrameInfo& info = pSurface->Info;
+    mfxU16 w = info.CropW ? info.CropW : info.Width;
+    mfxU16 h = info.CropH ? info.CropH : info.Height;
+
+    switch (ColorFormat)
+    {
+    case MFX_FOURCC_NV12:
+    case MFX_FOURCC_I420:
+        length = mfxU32(1.5 * w * h);
```
mgonchar: I suggest to get rid of conversions
```
length = 3 * w * h / 2;
```

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,7 +178,7 @@ void CSmplYUVReader::Reset()
     }
 }
 
-mfxStatus CSmplYUVReader::LoadNextFrame(mfxFrameSurface1* pSurface)
+mfxStatus CSmplYUVReader::LoadNextFrame(mfxFrameSurface1* pSurface, mfxU32 displayOrder/*= std::numeric_limits<mfxU32>::max()*/)
```
mgonchar: why have you commented default? leftover from debug?

samples/sample_common/src/sample_utils.cpp
```diff
@@ -184,6 +210,19 @@ mfxStatus CSmplYUVReader::LoadNextFrame(mfxFrameSurface1* pSurface)
...
+    // change file position for read to "frameLength * displayOrder"
+    if (displayOrder != std::numeric_limits<mfxU32>::max())
+    {
+        mfxU32 frameLength;
+        if (MFX_ERR_NONE != GetFrameLength(pSurface, m_ColorFormat, frameLength))
+        {
+            msdk_printf(MSDK_STRING("Input color format %s is unsupported in qpfile mode\n"), ColorFormatToStr(m_ColorFormat));
+            return MFX_ERR_UNSUPPORTED;
+        }
+        fseek(m_files[vid], frameLength * displayOrder, SEEK_SET);
```
mgonchar: you have to check return value

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1270,6 +1309,83 @@ mfxStatus CSmplYUVWriter::WriteNextFrameI420(mfxFrameSurface1 *pSurface)
...
+            }
+            m_FrameVals.push_back(frameInfo);
+        }
+        if (m_FrameVals.size() < m_nFrames)
+        {
+            ResetState(READER_ERR_INCORRECT_FILE);
+            return MFX_ERR_NOT_INITIALIZED;
+        }
+    }
+    catch (std::exception& /*e*/)
```
mgonchar: are you trying to catch some specific exception?

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2408,12 +2440,31 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
...
     m_nFramesRead++;
 
     return sts;
 }
 
+mfxStatus CEncodingPipeline::LoadNextControl(mfxEncodeCtrlWrap*& pCtrl, mfxU32 encSurfIdx)
+{
+    try
+    {
+        if (!pCtrl) return MFX_ERR_NULL_PTR;
```
mgonchar: now it useless check, since you are setting control in any way on the next line

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2408,12 +2440,31 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
...
+        pCtrl = &m_EncCtrls[encSurfIdx];
+        pCtrl->QP = m_QPFileReader.GetCurrentQP();
+        pCtrl->FrameType = m_QPFileReader.GetCurrentFrameType();
+        m_QPFileReader.NextFrame();
+    }
+    catch (std::exception& /*e*/)
```
mgonchar: same question about exception

samples/sample_common/src/sample_utils.cpp
pgribov: changed

samples/sample_common/src/sample_utils.cpp
pgribov: changed

samples/sample_common/src/sample_utils.cpp
pgribov: removed

samples/sample_common/src/sample_utils.cpp
pgribov: changed

samples/sample_common/src/sample_utils.cpp
pgribov: there may be many possible exceptions, i think it is better to leave the common exception

samples/sample_encode/src/pipeline_encode.cpp
pgribov: removed it, possible errors will be catched by catch

samples/sample_encode/src/pipeline_encode.cpp
pgribov: common case

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,7 +174,7 @@ void CSmplYUVReader::Reset()
     }
 }
 
-mfxStatus CSmplYUVReader::LoadNextFrame(mfxFrameSurface1* pSurface)
+mfxStatus CSmplYUVReader::LoadNextFrame(mfxFrameSurface1* pSurface, mfxU32 displayOrder)
```
mgonchar: now you are breaking backward compatibility for all samples. Are you sure that every sample is prepared to deal with this function taking 2 parameters instead of 1?

samples/sample_common/src/sample_utils.cpp
mgonchar: Every sample has it's own try/catch block on top level. I don't see any point in local exception handling. Putting try/catch block in every function where exception may occur is not a correct thing as for me. I'd say that you need to add exception handling only if you need to process all errors locally at current level. It is not definitely this case

samples/sample_encode/src/pipeline_encode.cpp
mgonchar: then remove. no need to introduce unnecessary exception handling at this level

samples/sample_common/src/sample_utils.cpp
pgribov: created another function for change position in file

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,6 +174,23 @@ void CSmplYUVReader::Reset()
...
+    mfxU32 frameLength;
+    if (MFX_ERR_NONE != GetFrameLength(w, h, m_ColorFormat, frameLength))
+    {
+        msdk_printf(MSDK_STRING("Input color format %s is unsupported in qpfile mode\n"), ColorFormatToStr(m_ColorFormat));
+        return MFX_ERR_UNSUPPORTED;
+    }
+    if (0 != fseek(m_files[info.FrameId.ViewId], frameLength * displayOrder, SEEK_SET))
+        return MFX_ERR_MORE_DATA;
+    return MFX_ERR_NONE;
+}
```
pgribov: @mgonchar added separate function for changing current reading position in file

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2388,14 +2416,20 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
             mfxStatus sts1 = m_pMFXAllocator->Lock(m_pMFXAllocator->pthis, pSurf->Data.MemId, &(pSurf->Data));
             MSDK_CHECK_STATUS(sts1, "m_pMFXAllocator->Lock failed");
 
-            sts = m_FileReader.LoadNextFrame(pSurf);
+            if (m_mfxEncParams.mfx.EncodedOrder)
+                sts = m_FileReader.ChangePositionInFile(pSurf->Info, m_QPFileReader.GetCurrentDisplayOrder());
+            if (MFX_ERR_NONE == sts)
+                sts = m_FileReader.LoadNextFrame(pSurf);
```
pgribov: now calling looks like this

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,6 +174,23 @@ void CSmplYUVReader::Reset()
     }
 }
 
+mfxStatus CSmplYUVReader::ChangePositionInFile(const mfxFrameInfo& info, mfxU32 displayOrder)
+{
+    // If used encoded order (in qpfile mode),
+    // change file position for read to "frameLength * displayOrder".
+    mfxU16 w = info.CropW ? info.CropW : info.Width;
+    mfxU16 h = info.CropH ? info.CropH : info.Height;
+    mfxU32 frameLength;
```
mgonchar: newline

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,6 +174,23 @@ void CSmplYUVReader::Reset()
...
+    if (MFX_ERR_NONE != GetFrameLength(w, h, m_ColorFormat, frameLength))
+    {
+        msdk_printf(MSDK_STRING("Input color format %s is unsupported in qpfile mode\n"), ColorFormatToStr(m_ColorFormat));
+        return MFX_ERR_UNSUPPORTED;
+    }
```
mgonchar: newline

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,6 +174,23 @@ void CSmplYUVReader::Reset()
...
+    if (0 != fseek(m_files[info.FrameId.ViewId], frameLength * displayOrder, SEEK_SET))
+        return MFX_ERR_MORE_DATA;
```
mgonchar: newline

samples/sample_encode/src/pipeline_encode.cpp
mgonchar: yes, much better as I think

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2388,14 +2416,20 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
...
 
             sts1 = m_pMFXAllocator->Unlock(m_pMFXAllocator->pthis, pSurf->Data.MemId, &(pSurf->Data));
             MSDK_CHECK_STATUS(sts1, "m_pMFXAllocator->Unlock failed");
         }
         else
         {
-            sts = m_FileReader.LoadNextFrame(pSurf);
+            if (m_mfxEncParams.mfx.EncodedOrder)
+                sts = m_FileReader.ChangePositionInFile(pSurf->Info, m_QPFileReader.GetCurrentDisplayOrder());
+            if (MFX_ERR_NONE == sts)
```
mgonchar: too complicated.

```
if (m_mfxEncParams.mfx.EncodedOrder)
{
    sts = m_FileReader.ChangePositionInFile(pSurf->Info, m_QPFileReader.GetCurrentDisplayOrder());
    MSDK_CHECK_STATUS(sts, "m_FileReader.ChangePositionInFile failed");
}

sts = m_FileReader.LoadNextFrame(pSurf);
```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2408,12 +2442,24 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
...
+        pSurf->Data.FrameOrder = m_mfxEncParams.mfx.EncodedOrder ? m_QPFileReader.GetCurrentDisplayOrder() : m_nFramesRead;
     m_nFramesRead++;
 
     return sts;
 }
 
+mfxStatus CEncodingPipeline::LoadNextControl(mfxEncodeCtrlWrap*& pCtrl, mfxU32 encSurfIdx)
+{
+    pCtrl = &m_EncCtrls[encSurfIdx];
+    if (!pCtrl) return MFX_ERR_NULL_PTR;
```
mgonchar: it is impossible situation, you are taking address of an object, it can't be ```NULL```

samples/sample_common/src/sample_utils.cpp
```diff
@@ -152,6 +174,23 @@ void CSmplYUVReader::Reset()
     }
 }
 
+mfxStatus CSmplYUVReader::ChangePositionInFile(const mfxFrameInfo& info, mfxU32 displayOrder)
```
mgonchar: here is still issue with responsibilities. Reader class here doesn't care about surface at all. So this function should accept width and height as parameters

mgonchar: Also let's name it more precisely, something like:
```
SkipNframesFromBeginning
```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2408,12 +2454,23 @@ mfxStatus CEncodingPipeline::LoadNextFrame(mfxFrameSurface1* pSurf)
...
     // frameorder required for reflist, dbp, and decrefpicmarking operations
-    if (pSurf) pSurf->Data.FrameOrder = m_nFramesRead;
+    if (pSurf)
+        pSurf->Data.FrameOrder = m_mfxEncParams.mfx.EncodedOrder ? m_QPFileReader.GetCurrentDisplayOrder() : m_nFramesRead;
     m_nFramesRead++;
 
     return sts;
 }
 
+mfxStatus CEncodingPipeline::LoadNextControl(mfxEncodeCtrlWrap*& pCtrl, mfxU32 encSurfIdx)
```
mgonchar: should return ```void```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -463,8 +463,13 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
 
     m_mfxEncParams.mfx.NumSlice = pInParams->nNumSlice;
     ConvertFrameRate(pInParams->dFrameRate, &m_mfxEncParams.mfx.FrameInfo.FrameRateExtN, &m_mfxEncParams.mfx.FrameInfo.FrameRateExtD);
-    m_mfxEncParams.mfx.EncodedOrder            = 0; // binary flag, 0 signals encoder to take frames in display order
 
+    m_mfxEncParams.mfx.EncodedOrder = pInParams->QPFileMode ? 1 : 0; // binary flag, 0 signals encoder to take frames in display order
+    if (m_mfxEncParams.mfx.EncodedOrder)
```
mgonchar: here and everywhere further you use ```m_mfxEncParams.mfx.EncodedOrder``` to indicate that you should read QPs from file. But that's actually bad design, it is possible to use encoded order without manual QP settings... I suggest to have another class variable ```CEncodingPipeline:: bool QPFileMode;``` and use it instead of ```m_mfxEncParams.mfx.EncodedOrder```

pgribov: changed to m_bQPFileMode

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -463,8 +463,15 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
 
     m_mfxEncParams.mfx.NumSlice = pInParams->nNumSlice;
     ConvertFrameRate(pInParams->dFrameRate, &m_mfxEncParams.mfx.FrameInfo.FrameRateExtN, &m_mfxEncParams.mfx.FrameInfo.FrameRateExtD);
-    m_mfxEncParams.mfx.EncodedOrder            = 0; // binary flag, 0 signals encoder to take frames in display order
 
+    m_bQPFileMode = pInParams->QPFileMode ? 1 : 0;
```
mgonchar: these are bools, so assign directly:
```
m_bQPFileMode = pInParams->QPFileMode;
```

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -463,8 +463,15 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
...
+    m_mfxEncParams.mfx.EncodedOrder = m_bQPFileMode; // binary flag, 0 signals encoder to take frames in display order
```
mgonchar: and this is ```bool``` to ```short unsigned int``` implicit conversion, it will work fine, but you may convert to integer first as you did on line 467

pgribov: changed

samples/sample_encode/src/pipeline_encode.cpp
pgribov: changed

pgribov:
@dmitryermilov

pgribov:
@onabiull

Issue #1736, opened 2019-10-18
sample decode crashes if X server can't be initialized


dvrogozh:
I was following https://github.com/Intel-Media-SDK/MediaSDK/wiki/Intel-media-stack-on-Ubuntu article in conjunction with https://github.com/Intel-Media-SDK/MediaSDK/wiki/Running-on-GPU-under-docker and stepped into this issue.

Basically, I forgot to pass correct DISPLAY to the docker and sample decode crashed with:

```
/usr/share/mfx/samples/sample_decode h264 -i AUD_MW_E.264 -r -rgb4 -f 30

Thread 1 "sample_decode" received signal SIGSEGV, Segmentation fault.
0x00007ffff26002fd in XCloseDisplay () from /lib/x86_64-linux-gnu/libX11.so.6
```

I guess we attempted to execute XCloseDisplay while XOpenDisplay has actually failed:
https://github.com/Intel-Media-SDK/MediaSDK/blob/0a9caaa2d5487e5b2706249e6aa56600cfd8849f/samples/sample_common/src/vaapi_utils_x11.cpp#L105

dvrogozh:
Can be reproduced w/o docker if DISPLAY is wrong or unset.

PR #1738, opened 2019-10-18
samples/decode: fix crash if X server can't be initialized

dvrogozh:
Fixes: #1736


samples/sample_common/src/vaapi_utils_x11.cpp
```diff
@@ -32,104 +32,90 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
 {
-    //release context
+    Close();
+}
+
+void X11LibVA::Close()
+{
     if (m_contextID != VA_INVALID_ID)
     {
         m_libva.vaDestroyContext(m_va_dpy, m_contextID);
```
mgonchar: I understand that this is almost destructor code. But those functions returns statuses. Can we add at least some logging here to print errors, if they occur?

dvrogozh: why not? done. Thought I don't think this will be of much use.

PR #1742, opened 2019-10-23
Fix UV output height of P010 when CropH is odd.

wangyan42164:

### DIFF
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -858,6 +858,56 @@ void CSmplYUVWriter::Close()
     m_bInited = false;
 }
 
+mfxStatus GetChromaSize(const mfxFrameInfo & pInfo, mfxU32 & ChromaW, mfxU32 & ChromaH)
+{
+    switch (pInfo.FourCC)
+    {
+    case MFX_FOURCC_YV12:
+    {
+        ChromaW = (pInfo.CropW + 1) / 2;
+        ChromaH = (pInfo.CropH + 1) / 2;
+        break;
+    }
+    case MFX_FOURCC_NV12:
+    {
+        ChromaW = pInfo.CropW;
+        ChromaH = (pInfo.CropH + 1) / 2;
+        break;
+    }
+    case MFX_FOURCC_P010:
+    case MFX_FOURCC_P210:
+    {
+        ChromaW = pInfo.CropW;
+        ChromaH = pInfo.FourCC == MFX_FOURCC_P210 ? (mfxU32)pInfo.CropH : (mfxU32)(pInfo.CropH + 1) / 2;
+        break;
+    }
+
+    case MFX_FOURCC_RGB4:
+    case 100: //DXGI_FORMAT_AYUV
+    case MFX_FOURCC_AYUV:
+    case MFX_FOURCC_YUY2:
+    case MFX_FOURCC_A2RGB10:
+    {
+        if (pInfo.CropH > 0 && pInfo.CropW > 0)
+        {
+            ChromaW = pInfo.FourCC == MFX_FOURCC_YUY2 ? (pInfo.CropW + 1) / 2 : pInfo.CropW;
+            ChromaH = pInfo.CropH;
+        }
+        else
+        {
+            ChromaW = pInfo.FourCC == MFX_FOURCC_YUY2 ? (pInfo.Width + 1) / 2 : pInfo.Width;
+            ChromaH = pInfo.Height;
+        }
+        break;
+    }
+
+    default:
+        return MFX_ERR_UNSUPPORTED;
+    }
+
+    return MFX_ERR_NONE;
+}
+
 mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
 {
     MSDK_CHECK_ERROR(m_bInited, false, MFX_ERR_NOT_INITIALIZED);
@@ -866,7 +916,7 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
     mfxFrameInfo &pInfo = pSurface->Info;
     mfxFrameData &pData = pSurface->Data;
 
-    mfxU32 i, h, w;
+    mfxU32 i;
     mfxU32 vid = pInfo.FrameId.ViewId;
 
     mfxU32 shiftSizeLuma   = 16 - pInfo.BitDepthLuma;
@@ -885,6 +935,10 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
     }
 
     FILE* dstFile = m_bIsMultiView ? m_fDestMVC[vid] : m_fDest;
+
+    mfxU32 ChromaW, ChromaH;
+    if (MFX_ERR_NONE != GetChromaSize(pInfo, ChromaW, ChromaH))
+        return MFX_ERR_UNSUPPORTED;
 
     switch (pInfo.FourCC)
     {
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/sample_utils.cpp
```diff
@@ -1019,7 +1019,7 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
     case MFX_FOURCC_P010:
     case MFX_FOURCC_P210:
     {
-        mfxU32 height = pInfo.FourCC == MFX_FOURCC_P210 ? (mfxU32)pInfo.CropH : (mfxU32)pInfo.CropH / 2;
+        mfxU32 height = pInfo.FourCC == MFX_FOURCC_P210 ? (mfxU32)pInfo.CropH : (mfxU32)(pInfo.CropH + 1) / 2;
```
dvrogozh: +1? that's something not obvious...

dvrogozh: discussed offline. That's ok since we have CropH here, not Height which should be multiple of 2.

wangyan-intel: Yes. I find some VP9 P010 dynamic resoluion video clip has odd cropping height. So UV height will be wrong. Thanks.

dvrogozh: I think we likely need similar fix for other color formats where we have /2. Or we need to reconsider the change and require crops to be aligned by 2. Anyhow, this is a good observation, but I feel we need more work here. For this reason, -1 to avoid merging in partial fix.

dvrogozh:
@mgonchar : can you, please, advice on how to refactor code to reflect correct crops, width, height rounding on /2 operations? We likely need to use some std::ceil. And I think we might need to introduce some common temporary variables to cache rounded values.

samples/sample_common/src/sample_utils.cpp
wangyan42164: Sure. I will check the other color format and refine this PR. Thanks.

wangyan42164:
> @mgonchar : can you, please, advice on how to refactor code to reflect correct crops, width, height rounding on /2 operations? We likely need to use some std::ceil. And I think we might need to introduce some common temporary variables to cache rounded values.

@dvrogozh @mgonchar  I have used std:ceil() to refine all color formats for YUV output. Could you please review it? Thanks.

mgonchar:
Hi @wangyan42164 I'd better prefer some helper function which returns you chroma size as tuple for example, taking fourcc and luma plane size. Going from integer arithmetic to floating and then back to integer is not good as for me. Plus I don't like this spreading of similar functionality across the code (actually thats why I vote for helper function).

Also, why do we talk only about height? Width of chroma plane might suffer from same issue

wangyan42164:
> Hi @wangyan42164 I'd better prefer some helper function which returns you chroma size as tuple for example, taking fourcc and luma plane size. Going from integer arithmetic to floating and then back to integer is not good as for me. Plus I don't like this spreading of similar functionality across the code (actually thats why I vote for helper function).
> 
> Also, why do we talk only about height? Width of chroma plane might suffer from same issue

@mgonchar I have refined this PR based on your comments. I am not sure if it is aligned with your idea? I can continue to refine it.
1. Add GetChromaSize() function to calculate the chroma width and height.
2. Use the above help function in WriteNextFrame() and WriteNextFrameI420() for both chroma width and height.
3. Based on your comments, I give up std:ceil() and fall back (x + 1) / 2. If any better idea, I can refine it.
Thanks.

samples/sample_common/src/sample_utils.cpp
```diff
@@ -1030,20 +1084,20 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
                 // Bits will be shifted to the lower position
                 tmp.resize(pData.Pitch);
 
-                for (int idx = 0; idx < pInfo.CropW; idx++)
+                for (int idx = 0; idx < (int)ChromaW; idx++)
```
mgonchar: I'd better changed counter type to ```mfxU32```

samples/sample_common/src/sample_utils.cpp
```diff
@@ -858,6 +858,58 @@ void CSmplYUVWriter::Close()
     m_bInited = false;
 }
 
+mfxStatus GetChromaSize(const mfxFrameInfo & pInfo, mfxU32 & ChromaW, mfxU32 & ChromaH)
+{
+    mfxStatus sts = MFX_ERR_NONE;
```
mgonchar: you don't need this variable, since you just return ```ERR_NONE``` at the end

samples/sample_common/src/sample_utils.cpp
```diff
@@ -858,6 +858,58 @@ void CSmplYUVWriter::Close()
...
+    case MFX_FOURCC_P010:
+    case MFX_FOURCC_P210:
+    {
+        ChromaW = pInfo.CropW;
+        ChromaH = pInfo.FourCC == MFX_FOURCC_P210 ? (mfxU32)pInfo.CropH : (mfxU32)(pInfo.CropH + 1) / 2;
+        break;
+    }
+
+    case MFX_FOURCC_RGB4:
+    case 100: //DXGI_FORMAT_AYUV
```
mgonchar: why dxgi enum can be here?

wangyan42164:
> @wangyan42164 it is exactly what I meant, thanks for the patch! The only thing that is surprising is why there is a place for DXGI enum? I thought we accept only MFX enums
In fact, I don't know why either. I just refer to the implementation WriteNextFrame(). If necessary, I can remove it. Thanks.

samples/sample_common/src/sample_utils.cpp
wangyan42164: Sure. I will change it. Thanks.

samples/sample_common/src/sample_utils.cpp
wangyan42164: Sure. I will change it. Thanks.

samples/sample_common/src/sample_utils.cpp
wangyan42164: In fact, I don't know why either. I just refer to the implementation WriteNextFrame(). If necessary, I can remove it. Thanks.

wangyan42164:
@mgonchar I have updated this PR based on your comments. For dxgi enum, I just refer to the implementation of  WriteNextFrame(). If necessary, I can remove it in GetChromaSize() and WriteNextFrame(). Could you please review it again? Thanks.

samples/sample_common/src/sample_utils.cpp
```diff
@@ -992,36 +1046,34 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
     {
     case MFX_FOURCC_YV12:
     {
-        for (i = 0; i < (mfxU32)pInfo.CropH / 2; i++)
+        for (i = 0; i < ChromaH; i++)
         {
             MSDK_CHECK_NOT_EQUAL(
-                fwrite(pData.V + (pInfo.CropY * pData.Pitch / 2 + pInfo.CropX / 2) + i * pData.Pitch, 1, pInfo.CropW, dstFile),
+                fwrite(pData.V + (pInfo.CropY * pData.Pitch / 2 + pInfo.CropX / 2) + i * pData.Pitch, 1, ChromaW, dstFile),
```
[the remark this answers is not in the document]
wangyan42164: For YV12:
CropX and CropY should be left/top aligned. So if they are odd, CropX/2 and CropY/2 are right themselves. But in this case, the first row/column of U/V may be only corresponding with the first row/column of Y. And the 2nd row/column of U/V should be corresponding with the 2nd/3rd row/column Y. It may not match YV12 rule? Otherwise, we may have to do interpolation.
For NV12:
CropX should be OK because of U/V interleaved. But CropY should have the same issue.
For P010/P210:
They are similar with 8bit cases.

So I am not sure if we can assume CropX/CropY must not be odd?
In fact, my current case is the width/height of the clip is odd. So both CropX and CropY are zero.
Thanks.

wangyan42164:
@dmitryermilov Could you please review this PR? Thanks.

PR #1749, opened 2019-10-31
Add dispatcher for UWP apps

npletnev:


npletnev:
@dmitryermilov Most of the VS projects are included in AllBuild.sln and have separate solutions, so I made the same here. Or should we remove all the separate solutions?

dmitryermilov:
> @dmitryermilov Most of the VS projects are included in AllBuild.sln and have separate solutions, so I made the same here. Or should we remove all the separate solutions?

Only samples have them. Tools and tutorials don't have. Please remove.

dvrogozh:
Anything we put into repository should be supported (optimally at least:)). In our CI we validate only top level AlllBuild.sln. So, any other solution will be off the radar at some point. I really considering that we should drop individual samples solutions as well for this reason - I am just lazy enough to do that.

Or we will finally switch to cmake across board at some point and drop everything:).

onabiull:
Cmake is closer, than you may think :). So from my perspective number of sln files doesn't matter that much. They going to be phased out soon.

dvrogozh:
I'll be on skeptical side. So: any new solutions not covered by CI should not be allowed in. Will be easier to clean up later...

dmitryermilov:
@npletnev , please address concerns or remove the PR.

PR #1751, opened 2019-11-01
sample_fei: Reset DataFlag during input bitstream reset

mgonchar:


samples/sample_fei/src/auxiliary_interfaces.cpp
```diff
@@ -503,6 +503,7 @@ mfxStatus MFX_DecodeInterface::ResetState()
     // prepare bit stream
     m_mfxBS.DataOffset = 0;
     m_mfxBS.DataLength = 0;
+    m_mfxBS.DataFlag   = 0;
```
dmitryermilov: What about other fields? Do we need to reset?

mgonchar: well, we can reset everything except some fields:
```
        struct {
            mfxEncryptedData* EncryptedData;
            mfxExtBuffer **ExtParam;
            mfxU16  NumExtParam;
        };
         mfxU32  reserved[6];
     };
    mfxU8*  Data;
    mfxU32  DataOffset;
    mfxU32  DataLength;
```

But I don't really the necessity of this.

Issue #1737, opened 2019-10-18
chroma_siting feature doesn't work for sample_vpp in Linux


liumark:
I knew this feature was fixed on Windows and I can confirm with following command:
sample_vpp.exe -sw 960 -sh 540 -scc rgb4 -dw 960 -dh 540 -dcc nv12 -chroma_siting vcen hcen -i test.rgb -o output.nv12

But the same command line doesn't work on Linux.
Mark

NaveenKumar-Intel:
Hi @liumark,
I tried following command on CL & Linux and it works fine for me. 

**/usr/share/mfx/samples/sample_vpp -sw 1280 -sh 720 -scc rgb4 -dw 1280 -dh 720 -dcc nv12 -chroma_siting vcen hcen -i out_sintel_1280x720.rgb -o output_sintel.nv12 -n 100**

Please find following output file generated with above command:
[output_sintel.zip](https://github.com/Intel-Media-SDK/MediaSDK/files/3751631/output_sintel.zip)

ias@clr-d3c8ee47e70e45ce8e7c0b93146f7b97 /mnt/Video_Content/RGB $ /usr/share/mfx/samples/sample_vpp -sw 1280 -sh 720 -scc rgb4 -dw 1280 -dh 720 -dcc nv12 -chroma_siting vcen hcen -i out_sintel_1280x720.rgb -o output_sintel.nv12 -n 100

VPP_WRN: some filter(s) skipped
Input format    RGB4
Resolution      1280x720
Crop X,Y,W,H    0,0,1280,720
Frame rate      30.00
PicStruct       progressive
Output format   NV12
Resolution      1280x720
Crop X,Y,W,H    0,0,1280,720
Frame rate      30.00
PicStruct       progressive

Video Enhancement Algorithms
Deinterlace     OFF
Signal info     OFF
Scaling         OFF
CromaSiting             ON
Denoise         OFF
MCTF            OFF
ProcAmp         OFF
DetailEnh       OFF
GamutMapping    OFF
ColorSaturation OFF
ContrastEnh     OFF
SkinToneEnh     OFF
MVC mode        OFF
ImgStab         OFF

IOpattern type                  sys_to_sys
Number of asynchronious tasks   1
Time stamps checking            OFF
ROI checking                    OFF

MediaSDK impl   hw
HW accelaration is enabled
MediaSDK ver    1.28
VPP started
Frame number: 100
VPP finished

Total frames 100
[...]

PR #1756, opened 2019-11-08
Chroma Siting on Linux

fzhar:


_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1237,6 +1243,41 @@ mfxStatus VAAPIVideoProcessing::Execute(mfxExecuteParams *pParams)
         break;
     }
 
+#if (MFX_VERSION >= 1025)
+        uint8_t& chromaSitingMode = m_pipelineParam[0].input_color_properties.chroma_sample_location;
+        chromaSitingMode = VA_CHROMA_SITING_UNKNOWN;
```
dmitryermilov: should be in switch, I suppose

dmitryermilov: should be in switch, I suppose

dmitryermilov: Also you should add a reference to https://github.com/Intel-Media-SDK/MediaSDK/issues/1737 via:
Fixes: # 1737

fzhar: It can be left here - this is just local reference, it will be destroyed after exiting function scope.

fzhar: "Fixes" issue is addressed

PR #1762, opened 2019-11-15
Samples: Fix behavior when specified non-intel device

AntonGrishin:


AntonGrishin:
> Add:
> `Fixes: 4e3c14a ("samples: add support for render node specifying")`
> 
> To commit message. Question: did you verify that samples work with -device parameter?

Done.
Yes, i double-checked this, but in first commit I missed this behavior due to big log...

PR #1765, opened 2019-11-20
[sample_common] Fix wrong m_maxWinBitsLim

egorovdanil:


samples/sample_common/include/brc_routines.h
```diff
@@ -245,12 +245,13 @@ class AVGBitrate
         {
             if (bPanic || bSH)
             {
-                m_maxWinBitsLim = mfx::clamp((GetLastFrameBits(windowSize, false) + m_maxWinBits)/2, GetMaxWinBitsLim(), m_maxWinBits);
+                m_maxWinBitsLim = std::max(std::min((GetLastFrameBits(windowSize, false) + m_maxWinBits)/2, m_maxWinBits), GetMaxWinBitsLim());
             }
```
egorovdanil: Clamp returns `std::min(m_maxWinBits, std::max(GetLastFrameBits(windowSize, false) + m_maxWinBits)/2, GetMaxWinBitsLim())`, but before we used to get `IPP_MAX(IPP_MIN(GetLastFrameBits(windowSize, false) + m_maxWinBits)/2, m_maxWinBits), GetMaxWinBitsLim())`
[0b43390](https://github.com/Intel-Media-SDK/MediaSDK/pull/1647/commits/0b43390c8d268c58790d7f257fceed0b4dc38565)

samples/sample_common/include/brc_routines.h
```diff
@@ -245,12 +245,13 @@ class AVGBitrate
...
             else
             {
-                if (recode)
-                    m_maxWinBitsLim = mfx::clamp((GetLastFrameBits(windowSize, false) + GetStep())/2, m_maxWinBitsLim, m_maxWinBits);
+                if (recode) {
+                    m_maxWinBitsLim = std::min(std::max(GetLastFrameBits(windowSize,false) + GetStep()/2, m_maxWinBitsLim), m_maxWinBits);
```
mgonchar: this one is exactly clamp, so no need to change these lines

samples/sample_common/include/brc_routines.h
mgonchar: Hi @egorovdanil . I consulted with @takulako before this change. What exactly is the issue? I understand that code change here is not so straightforward, but anyway, min + max means clamp. So if you think that clamp is incorrect, you can rearrange parameters of the clamp function instead

egorovdanil: Hello, @mgonchar. Yes, the clamp is correct, the error was in the wrong brackets in the clamp function call below :)

mgonchar: yep, I rechecked the original code. You are right :)

PR #1766, opened 2019-11-21
Added access index check

filippenkos:
Added access index check to prevent possible access violation

### DIFF
--- a/_studio/shared/umc/codec/jpeg_enc/src/jpegenc.cpp
+++ b/_studio/shared/umc/codec/jpeg_enc/src/jpegenc.cpp
@@ -694,6 +694,8 @@ JERRCODE CJPEGEncoder::SetParams(
   }
   else
   {
+    if (m_jpeg_ncomp > MAX_HUFF_TABLES)
+      return JPEG_ERR_PARAMS;
     for(c = 0; c < m_jpeg_ncomp; c++)
     {
       jerr = m_dctbl[c].Create();


_studio/shared/umc/codec/jpeg_enc/src/jpegenc.cpp
```diff
@@ -696,6 +696,9 @@ JERRCODE CJPEGEncoder::SetParams(
   {
     for(c = 0; c < m_jpeg_ncomp; c++)
     {
+      if (c >= MAX_HUFF_TABLES)
```
mgonchar: definitely no need to check inside the loop. Also, is it possible that ```m_jpeg_ncomp >= MAX_HUFF_TABLES```?

filippenkos: Agree, there is definitely no need to check inside the loop. I changed commit to
```suggestion
      if (c >= MAX_HUFF_TABLES)
        return JPEG_ERR_PARAMS;
      for(c = 0; c < m_jpeg_ncomp; c++)
      {
```
It's possible that `m_jpeg_ncomp >= MAX_HUFF_TABLES ` because it is limited by another constant `MAX_COMPS_PER_SCAN`

mgonchar: ok

PR #1767, opened 2019-11-21
Replaced GetExtBuffer calls with GetExtBufferRef calls

filippenkos:
Replaced GetExtBufferCalls in InheritDefaultValues to exclude
pointer dereferencing

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -5124,16 +5124,16 @@ void MfxHwH264Encode::InheritDefaultValues(
     mfxVideoParam const * parResetIn)
 {
     mfxExtCodingOption const &  extOptInit      = GetExtBufferRef(parInit);
-    mfxExtCodingOption *        extOptReset     = GetExtBuffer(parReset);
+    mfxExtCodingOption &        extOptReset     = GetExtBufferRef(parReset);
     mfxExtCodingOption3 const & extOpt3Init     = GetExtBufferRef(parInit);
-    mfxExtCodingOption3 *       extOpt3Reset    = GetExtBuffer(parReset);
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -5132,8 +5132,14 @@ void MfxHwH264Encode::InheritDefaultValues(
 
     mfxU32 TCBRCTargetFrameSize = 0;
 
-    InheritOption(extOptInit.NalHrdConformance,         extOptReset->NalHrdConformance);
-    InheritOption(extOpt3Init.LowDelayBRC,              extOpt3Reset->LowDelayBRC);
+    if (extOptReset)
```
mgonchar: I believe that this pointer can never be ```null```.

I think that correct fix should be following:
```
mfxExtCodingOption &        extOptReset     = GetExtBufferRef(parReset);
...
```

filippenkos: This pointer can possibly be `null` because `GetExtBuf ` can return 0 in some cases
https://github.com/Intel-Media-SDK/MediaSDK/blob/045057e446a0e5c62ac6c25f7f2bedb3d16979e9/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L7000-L7015

mgonchar: look carefully on ```MfxVideoParam``` (capital M matters) constructor. Those buffers are always(!) exists in class, so you can take reference on them without any worry

filippenkos: Right, i misread type name which lead to wrong conclusion. I changed commit to `GetExtBufferRef` calls.

PR #1770, opened 2019-11-22
Add GeminiLake (GLK) device ids

dvrogozh:


_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2282,6 +2282,7 @@ mfxStatus  VideoVPPHW::Init(
             case MFX_HW_APL:
             case MFX_HW_KBL:
             case MFX_HW_CFL:
+            case MFX_HW_GLK:
```
felixbuenemann: I think you should list GLK before CFL for consistency.

dvrogozh: done

PR #1774, opened 2019-11-25
Add 12-bit format support

onabiull:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw.cpp
```diff
@@ -138,13 +149,41 @@ bool GetRecInfo(const MfxVideoParam& par, mfxFrameInfo& rec)
     }
     else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV420) && CO3.TargetBitDepthLuma == 10)
     {
-        rec.FourCC = MFX_FOURCC_NV12;
-        rec.Width  = mfx::align2_value(rec.Width, 32) * 2;
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+        if (par.m_platform >= MFX_HW_TGL_LP)
```
dmitry-gurulev: 1. why we change old behavior (was nv12 w/ doubled width), now it is p016?
2. why we use nv12 for tgl+ instead of p016, like it is for y216 above?

saosipov: everything is done according to the latest SAS to support MMC

PR #1775, opened 2019-11-26
samples: Fix android build

AntonGrishin:
Fixes: 4e3c14a ("samples: add support for render node specifying")


AntonGrishin:
> Build still fails. Is it infra issue? Need it to be addressed and test rerun to let this change in.

Currently this is infra, because build failed on "clear" stage

PR #1777, opened 2019-11-26
[SW BRC][HEVCe] Disable unnecessary checking in HDRConformance case

pgribov:
if NalHrdConformance is on and VuiNalHrdParameters is off (MFX_BRC_HRD_WEAK)
then there is no need to check some hrd parameters.
This checking was disabled for this case.

### DIFF
--- a/_studio/mfx_lib/shared/include/mfx_brc_common.h
+++ b/_studio/mfx_lib/shared/include/mfx_brc_common.h
@@ -160,6 +160,21 @@ protected:
 #define MAX_RACA 361.0
 #define RACA_SCALE 128.0
 
+/*
+NalHrdConformance | VuiNalHrdParameters   |  Result
+--------------------------------------------------------------
+    off                  any                => MFX_BRC_NO_HRD
+    default              off                => MFX_BRC_NO_HRD
+    on                   off                => MFX_BRC_HRD_WEAK
+    on (or default)      on (or default)    => MFX_BRC_HRD_STRONG
+--------------------------------------------------------------
+*/
+enum : mfxU16
+{
+    MFX_BRC_NO_HRD = 0,
+    MFX_BRC_HRD_WEAK,  // IF HRD CALCULATION IS REQUIRED, BUT NOT WRITTEN TO THE STREAM
+    MFX_BRC_HRD_STRONG
+};
 
 namespace MfxHwH265EncodeBRC
 {
@@ -168,7 +183,7 @@ class cBRCParams
 public:
     mfxU16 rateControlMethod; // CBR or VBR
 
-    mfxU16 bHRDConformance;  // is HRD compliance  needed
+    mfxU16 HRDConformance;   // is HRD compliance  needed
     mfxU16 bRec;             // is Recoding possible
     mfxU16 bPanic;           // is Panic mode possible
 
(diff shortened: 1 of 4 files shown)


_studio/mfx_lib/shared/include/mfx_brc_common.h
```diff
@@ -228,7 +243,7 @@ class cBRCParams
 public:
     cBRCParams() :
         rateControlMethod(0),
-        bHRDConformance(0),
+        HRDConformance(0),
```
mgonchar: use ```MFX_BRC_NO_HRD``` here

_studio/mfx_lib/shared/include/mfx_brc_common.h
```diff
@@ -160,6 +160,21 @@ class AVGBitrate
...
+--------------------------------------------------------------
+    off                  any                => MFX_BRC_NO_HRD
+    default              off                => MFX_BRC_NO_HRD
+    on                   off                => MFX_BRC_HRD_WEAK
+    on (or default)      on (or default)    => MFX_BRC_HRD_STRONG
+--------------------------------------------------------------
+*/
+enum : mfxU16
+{
+    MFX_BRC_NO_HRD,
```
mgonchar: You are highly relying on default values start from zero, but not specifying it explicitly.  Better to add initialization

```MFX_BRC_NO_HRD = 0,```

File: _studio/mfx_lib/shared/src/mfx_brc_common.cpp [hunk 3]
```diff
@@ -153,14 +153,21 @@ mfxStatus cBRCParams::Init(mfxVideoParam* par, bool fieldMode)
...
+    if (pExtCO)
+    {
+        if (!IsOff(pExtCO->NalHrdConformance) && !IsOff(pExtCO->VuiNalHrdParameters))
+            HRDConformance = MFX_BRC_HRD_STRONG;
+        else if (IsOn(pExtCO->NalHrdConformance) && IsOff(pExtCO->VuiNalHrdParameters))
+            HRDConformance = MFX_BRC_HRD_WEAK;
+    }
 
-    if (bHRDConformance)
+    if (HRDConformance)
```
mgonchar: Better to use explicit comparison with NO_HRD enum and not rely on defaults. Same applicable to many places below
```
if (HRDConformance != MFX_BRC_NO_HRD)
```

samples/sample_common/include/brc_routines.h
```diff
@@ -43,12 +43,28 @@ or https://software.intel.com/en-us/media-client-solutions-support.
...
+--------------------------------------------------------------
+    off                  any                => MFX_BRC_NO_HRD
+    default              off                => MFX_BRC_NO_HRD
+    on                   off                => MFX_BRC_HRD_WEAK
+    on (or default)      on (or default)    => MFX_BRC_HRD_STRONG
+--------------------------------------------------------------
+*/
+enum : mfxU16
+{
+    MFX_BRC_NO_HRD,
```
mgonchar: same applicable for this file as well

_studio/mfx_lib/shared/include/mfx_brc_common.h
pgribov: changed

_studio/mfx_lib/shared/include/mfx_brc_common.h
pgribov: changed

_studio/mfx_lib/shared/src/mfx_brc_common.cpp
pgribov: changed

samples/sample_common/include/brc_routines.h
pgribov: changed

PR #1783, opened 2019-11-29
Fix crash in dispatcher

npletnev:
When msdk dll is loaded, table of pointers to function implemented in this dll is filled.
The crash happens when: dll which implements MFXInitEx is loaded ->
pointer to MFXInitEx saved to the table -> dll is unloaded ->
table isn't cleaned -> dll which doesn't implement MFXInitEx is loaded ->
MFXInitEx called by the pointer that points to previous dll -> crash


gkrivor:
Thanks for investigation of algorithm in this part. Original change was incorrect.

dmitryermilov:
@npletnev , please rebase and trigger a new build. Seems the last one hanged

PR #1785, opened 2019-12-02
[h264e] Fix LA support on TGL

AntonGrishin:
LA TGL kernel already enabled, but query wasnt update.

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -244,6 +244,11 @@ namespace
             ;
     }
 
+    bool hasSupportVME(eMFXHWType platform)
+    {
+        return platform <= MFX_HW_TGL_LP;
+    }
+
     inline mfxU16 GetMaxSupportedLevel()
     {
         return MFX_LEVEL_AVC_52;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -2482,7 +2482,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
     }
 
     bool laEnabled = true;
-    if (platform > MFX_HW_ICL_LP)
+    if (platform > MFX_HW_ICL_LP && platform != MFX_HW_TGL_LP)
```
dmitryermilov: Please add a comment why we need such condition

dmitryermilov: Okay, it's VME dependency. Please name this accordingly. Please also check HEVC part. The shall the same condition there.

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -2482,7 +2487,7 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
     }
 
     bool laEnabled = true;
-    if (platform > MFX_HW_ICL_LP)
+    if (!isSupportVME(platform))
```
dmitryermilov: Two questions: 
# What about TGL?
# what about HEVC?

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -244,6 +244,11 @@ namespace
             ;
     }
 
+    bool isSupportVME(eMFXHWType platform)
```
dmitryermilov: I'd prefer HasVMESupport()

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
AntonGrishin: 1. <= TGL, so supports.
2. Didnt find any similar logic

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
AntonGrishin: changed

PR #1787, opened 2019-12-02
[sample] Pitch calculating refactoring

IlyaTikhanov:
When it is 16K resolution and 444 Chroma subsampling
we need value more then 16 bit variable allows
So need to use PitchLow and PitchHigh instead of just Pitch

Issue: MDP-58640

### DIFF
--- a/samples/sample_common/src/sysmem_allocator.cpp
+++ b/samples/sample_common/src/sysmem_allocator.cpp
@@ -107,46 +107,54 @@ mfxStatus SysMemFrameAllocator::LockFrame(mfxMemId mid, mfxFrameData *ptr)
     case MFX_FOURCC_NV12:
         ptr->U = ptr->Y + Width2 * Height2;
         ptr->V = ptr->U + 1;
-        ptr->Pitch = Width2;
+        ptr->PitchHigh = 0;
+        ptr->PitchLow = (mfxU16)MSDK_ALIGN32(fs->info.Width);
         break;
     case MFX_FOURCC_NV16:
         ptr->U = ptr->Y + Width2 * Height2;
         ptr->V = ptr->U + 1;
-        ptr->Pitch = Width2;
+        ptr->PitchHigh = 0;
+        ptr->PitchLow = (mfxU16)MSDK_ALIGN32(fs->info.Width);
         break;
     case MFX_FOURCC_YV12:
         ptr->V = ptr->Y + Width2 * Height2;
         ptr->U = ptr->V + (Width2 >> 1) * (Height2 >> 1);
-        ptr->Pitch = Width2;
+        ptr->PitchHigh = 0;
+        ptr->PitchLow = (mfxU16)MSDK_ALIGN32(fs->info.Width);
         break;
     case MFX_FOURCC_UYVY:
         ptr->U = ptr->Y;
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/sysmem_allocator.cpp
```diff
@@ -107,87 +107,102 @@ mfxStatus SysMemFrameAllocator::LockFrame(mfxMemId mid, mfxFrameData *ptr)
...
-        ptr->Pitch = 2 * Width2;
+        ptr->PitchHigh = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) / (1 << 16));
+        ptr->PitchLow = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) % (1 << 16));
         break;
 #if (MFX_VERSION >= 1028)
     case MFX_FOURCC_RGB565:
         ptr->G = ptr->B;
         ptr->R = ptr->B;
-        ptr->Pitch = 2 * Width2;
+        ptr->PitchHigh = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) >> 16);
```
mgonchar: please, use one of the approaches, either
```
ptr->PitchHigh = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) / (1 << 16));
ptr->PitchLow = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) % (1 << 16));
```
or
```
ptr->PitchHigh = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) >> 16);
ptr->PitchLow = (mfxU16)((2 * MSDK_ALIGN32(fs->info.Width)) & 0xffff);
```

samples/sample_common/src/sysmem_allocator.cpp
```diff
@@ -107,87 +107,102 @@ mfxStatus SysMemFrameAllocator::LockFrame(mfxMemId mid, mfxFrameData *ptr)
...
-        ptr->Pitch = 3 * Width2;
+        ptr->PitchHigh = (mfxU16)((3 * MSDK_ALIGN32(fs->info.Width)) / (1 << 16));
+        ptr->PitchLow = (mfxU16)((3 * MSDK_ALIGN32(fs->info.Width)) % (1 << 16));
         break;
 #if !(defined(_WIN32) || defined(_WIN64))
     case MFX_FOURCC_RGBP:
         ptr->G = ptr->B + Width2 * Height2;
         ptr->R = ptr->B + Width2 * Height2 * 2;
-        ptr->Pitch = Width2;
+        ptr->PitchHigh = (mfxU16)((3 * MSDK_ALIGN32(fs->info.Width)) / (1 << 16));
```
mgonchar: why ```3 * MSDK_ALIGN32```?

samples/sample_common/src/sysmem_allocator.cpp
IlyaTikhanov: Now only the 1st approach is used

samples/sample_common/src/sysmem_allocator.cpp
IlyaTikhanov: Changed. We don't need to multiply by 3.

PR #1788, opened 2019-12-02
Add missed lines in 12-bit formats support.

onabiull:
Fixes: 444ecb48 ("Add 12-bit format support")

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
@@ -1375,6 +1375,9 @@ mfxU16 GetMaxChroma(MfxVideoParam const & par)
     {
     case MFX_FOURCC_NV12:
     case MFX_FOURCC_P010:
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+    case MFX_FOURCC_P016:
+#endif
         c = std::min<mfxU16>(c, MFX_CHROMAFORMAT_YUV420);
         break;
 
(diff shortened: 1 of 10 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -1650,6 +1678,16 @@ mfxStatus CheckVideoParam(MfxVideoParam& par, MFX_ENCODE_CAPS_HEVC const & caps,
 
     switch (CO3.TargetBitDepthLuma)
     {
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+    case 12:
+        invalid += CheckOption(CO3.TargetBitDepthChroma, 12/*, 10, 8*/, 0);
+        invalid += CheckOption(par.mfx.CodecProfile
+//            , (mfxU16)MFX_PROFILE_HEVC_MAIN10
```
dvrogozh: what this commented line is about?

_studio/shared/src/libmfx_core.cpp
```diff
@@ -1381,7 +1381,45 @@ mfxStatus CoreDoSWFastCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc, int c
 
         break;
 
+#if (MFX_VERSION >= 1027)
```
dvrogozh: this one is not terminated

_studio/shared/src/libmfx_core.cpp
```diff
@@ -1381,7 +1381,45 @@ mfxStatus CoreDoSWFastCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc, int c
...
+    case MFX_FOURCC_Y210:
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
```
dvrogozh: And the whole block presents already below, line ~1470: https://github.com/Intel-Media-SDK/MediaSDK/pull/1788/commits/4ac1843e24f90d84a1e25559efafa3bf42784f84#diff-fae1222e835b7a0eea6af66261c4d4d0R1470

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
saosipov: For 12b encoding both Luma and Chroma TargetBitdepth should be 12 (at least other behavior is not confirmed yet) while for 10b encoding it is allowed to have LumaTargetBitdepth=10 and ChromaTargetBitdepth=8
Commented lines can be easily removed.

Issue #1655, opened 2019-09-20
sample_decode failed using drm render with -async 1 (However it works fine if async value is other than 1 (>1) )


NaveenKumar-Intel:
Reproducible command (Verified on CL platform): 

_root@clr-31ce3df6fc9142bba096c1aa5890c202 /tmp # **/usr/share/mfx/samples/sample_decode h264 -i 1080p_30fps.h264 -hw -vaapi -calc_latency -async 1 -rgb4 -rdrm**
pretending that aspect ratio is 1:1
libva info: VA-API version 1.5.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'iHD'
libva info: Trying to open /usr/lib64/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_4
libva info: va_openDriver() returns 0
drmrender: trying connection: HDMIA
drmrender: selected crtc already attached to connector
drmrender: succeeded...
drmrender: connected via HDMIA to 1920x1080@60 capable display
Decoding Sample Version 8.4.27.

Input video     AVC
Output format   RGB4 (using vpp)
Input:
  Resolution    1920x1088
  Crop X,Y,W,H  0,0,1920,1080
Output:
  Resolution    1920x1080
Frame rate      30.00
Memory type             vaapi
MediaSDK impl           hw
MediaSDK version        1.29

Decoding started
drmrender: warning: rendering via OVERLAY plane
**fatal: failed to find output surface, that's a bug!**
Frame    1, latency=91.74500 ms, fread_fps: 0.000, fwrite_fps: 154.107

Latency summary:

AVG=91.74500 ms, MAX=91.74500 ms, MIN=91.74500 ms
[ERROR], sts=MFX_ERR_NOT_FOUND(-9), RunDecoding, Unexpected error!! at /home/naveen/workspace/bkup_GP-2.0/MediaSDK/samples/sample_decode/src/pipeline_decode.cpp:1900

[ERROR], sts=MFX_ERR_NOT_FOUND(-9), main, Pipeline.RunDecoding failed at /home/naveen/workspace/bkup_GP-2.0/MediaSDK/samples/sample_decode/src/sample_decode.cpp:740
drmrender: failed to restore original mode
drmrender: warning: failure to restore original mode may lead to application segfault!
root@clr-31ce3df6fc9142bba096c1aa5890c202 /tmp #_

**Please refer to this #1622 referenced issue for more details & HOWTO reproduce.**

PR #1798, opened 2019-12-04
sample_decode: Add one more surface in rendering mode

AntonGrishin:
Need to add one more surface in rendering mode for correct "async 1"
behavior

Fixes: #1655


AntonGrishin:
> What if renderer starts supporting NV12. Then we won't allocate "vpp" frames, right?
 
I am not sure about that it. Also when I`m tried enable NV12 renderer, I got few errors from vaapi
https://github.com/Intel-Media-SDK/MediaSDK/blob/89b7dcf675b7a5a7c10e69c8ed3c943aa78bcbdc/samples/sample_common/src/vaapi_allocator.cpp#L399
https://github.com/Intel-Media-SDK/MediaSDK/blob/89b7dcf675b7a5a7c10e69c8ed3c943aa78bcbdc/samples/sample_common/src/vaapi_allocator.cpp#L400
https://github.com/Intel-Media-SDK/MediaSDK/blob/89b7dcf675b7a5a7c10e69c8ed3c943aa78bcbdc/samples/sample_common/src/vaapi_allocator.cpp#L401

PR #1799, opened 2019-12-04
API 1.31

onabiull:


samples/sample_multi_transcode/src/transcode_utils.cpp
```diff
@@ -1187,6 +1187,16 @@ mfxStatus ParseAdditionalParams(msdk_char *argv[], mfxU32 argc, mfxU32& i, Trans
     {
         InputParams.nVuiNalHrdParameters = MFX_CODINGOPTION_OFF;
     }
+#if (defined(_WIN32) || defined(_WIN64)) && (MFX_VERSION >= 1031)
```
dvrogozh: why did you move the block? you distant windows/linux adapter options from each other...

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -1290,7 +1290,7 @@ mfxStatus CEncodingPipeline::InitFileWriters(sInputParams *pParams)
     return sts;
 }
 
-#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= MFX_VERSION_NEXT)
+#if (defined(_WIN64) || defined(_WIN32)) && (MFX_VERSION >= 1031)
```
dvrogozh: why version check is here?! Just few lines below in a function you have #if (MFX_VERSION >= 1027) - it does not make sense!

samples/sample_multi_transcode/src/transcode_utils.cpp
onabiull: On windows we hit compiler limit here. Too long if/else chain lead to compiler error. This function specifically designed to workaround this limitation.

samples/sample_encode/src/pipeline_encode.cpp
onabiull: To split 12-bit enabling from discrete adapters.

PR #1815, opened 2019-12-09
[code] Sync'ed device_id table with the table from Linux kernel

dmitryermilov:

### DIFF
--- a/_studio/shared/src/libmfx_core_vaapi.cpp
+++ b/_studio/shared/src/libmfx_core_vaapi.cpp
@@ -62,117 +62,129 @@ typedef struct {
     eMFXGTConfig config;
 } mfx_device_item;
 
-// list of legal dev ID for Intel's graphics
- const mfx_device_item listLegalDevIDs[] = {
+// The list below should be sync'ed MANUALLY with include/drm/i915_pciids.h of Linux kernel.
+// PLEASE KEEP THE ORDER AS SIMILAR AS POSSIBLE TO SIMPLIFY MERGE.
+const mfx_device_item listLegalDevIDs[] = {
     /*IVB*/
-    { 0x0156, MFX_HW_IVB, MFX_GT1 },   /* GT1 mobile */
-    { 0x0166, MFX_HW_IVB, MFX_GT2 },   /* GT2 mobile */
-    { 0x0152, MFX_HW_IVB, MFX_GT1 },   /* GT1 desktop */
-    { 0x0162, MFX_HW_IVB, MFX_GT2 },   /* GT2 desktop */
-    { 0x015a, MFX_HW_IVB, MFX_GT1 },   /* GT1 server */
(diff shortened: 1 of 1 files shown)


_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -63,276 +63,329 @@ typedef struct {
...
     { 0x8A50, MFX_HW_ICL_LP, MFX_GT2 },
-    { 0x8A51, MFX_HW_ICL_LP, MFX_GT2 },
-    { 0x8A52, MFX_HW_ICL_LP, MFX_GT2 },
-    { 0x8A53, MFX_HW_ICL_LP, MFX_GT2 },
-    { 0x8A56, MFX_HW_ICL_LP, MFX_GT1 },
-    { 0x8A57, MFX_HW_ICL_LP, MFX_GT1 },
-    { 0x8A58, MFX_HW_ICL_LP, MFX_GT1 },
+    { 0x8A5C, MFX_HW_ICL_LP, MFX_GT1 },
     { 0x8A59, MFX_HW_ICL_LP, MFX_GT1 },
+    { 0x8A58, MFX_HW_ICL_LP, MFX_GT1 },
```
dvrogozh: what's the sense to reorder them? To me it is much easier when they are ordered

dmitryermilov: > what's the sense to reorder them?

To be able to compare with i915_pciids.h in bcompare

> To me it is much easier when they are ordered

Much easier to do what?

dvrogozh: > Much easier to do what?

To understand what you have added! or dropped.

I had in mind a little different approach to this change. I hoped to just copy kernel header as is and reuse.

dmitryermilov: If you want to review, you'll have to spend some time on it..

dvrogozh: I am spending it already. However, you are heavily changing the code and you don't explain anywhere what you did and how others are supposed to support the code. Again, you reorder device ID list. How someone is supposed to support that? Why someone won't submit PR to reorder them back. You order is unnatural! At least provide README.md and explain what you did and how someone needs to support that/compare/etc.

dmitryermilov: I explained it. Please see above: "To be able to compare with i915_pciids.h in bcompare"
There is a comment in code that "The list should be sync'ed manually with include/drm/i915_pciids.h". You're saying like current order is natural. No, it is not. New one is at least is synced with some reference point.

dvrogozh: > I explained it. Please see above: 
This should be in code base, not in a github comments.

dmitryermilov: NP. Added. Please review,

PR #1824, opened 2019-12-10
[avc] Changed ImplementationAvc Init and Reset behavior to match manual

filippenkos:
Changed ImplementationAvc Init and Reset behavior in case of
mfxExtCodingOption2 member IntRefCycleSize equal to 1 which is invalid
value by manual. In this case methods will change intra refresh cycle size
to 1s and return MFX_WRN_VIDEO_PARAM_CHANGED.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
@@ -944,6 +944,7 @@ mfxStatus ImplementationAvc::Init(mfxVideoParam * par)
         }
     }
     mfxExtCodingOption3 & extOpt3 = GetExtBufferRef(m_video);
+
     bool bPanicModeSupport = (m_video.mfx.RateControlMethod == MFX_RATECONTROL_LA) || (m_video.mfx.RateControlMethod == MFX_RATECONTROL_LA_HRD) || IsOn(extOpt2.ExtBRC);
     if (m_raw.NumFrameActual == 0 && bPanicModeSupport )
     {
@@ -1165,6 +1166,22 @@ mfxStatus ImplementationAvc::Init(mfxVideoParam * par)
     // initialization of parameters for Intra refresh
     if (extOpt2.IntRefType)
     {
+        if (extOpt2.IntRefCycleSize == 1)
+        {
+            if (extOpt2.IntRefType == MFX_REFRESH_SLICE)
+            {
+                extOpt2.IntRefCycleSize = extOpt3.NumSliceP;
+            }
+            else
+            {
+                // set intra refresh cycle to 1 sec
+                extOpt2.IntRefCycleSize =
+                    (mfxU16)((m_video.mfx.FrameInfo.FrameRateExtN + m_video.mfx.FrameInfo.FrameRateExtD - 1) /
+                             m_video.mfx.FrameInfo.FrameRateExtD);
+            }
+            checkStatus = MFX_WRN_VIDEO_PARAM_CHANGED;
+        }
+
         if (extOpt2.IntRefType == MFX_REFRESH_SLICE)
         {
             m_intraStripeWidthInMBs = 0;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -944,6 +944,23 @@ mfxStatus ImplementationAvc::Init(mfxVideoParam * par)
         }
     }
     mfxExtCodingOption3 & extOpt3 = GetExtBufferRef(m_video);
+
+    if (extOpt2.IntRefCycleSize == 1)
```
gkrivor: Could you clarify why do you check IntRefCycleSize out of verification of IntRefType?

I suggest to put this verification as part of IntRefType conidition https://github.com/Intel-Media-SDK/MediaSDK/blob/4e98b9c1355d0f81fbefc75f12542aeed8c63be5/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp#L1166

According documentation 0 is also wrong value (actually, in case of MFX_REFRESH_SLICE it will be ignored). And it should be explicitly verified because otherwise we have division by zero at https://github.com/Intel-Media-SDK/MediaSDK/blob/4e98b9c1355d0f81fbefc75f12542aeed8c63be5/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp#L1175

Also I'm wondering why I cannot find line where we check FrameRateExtD!=0 (for filling default value in case of IntRefType!=MFX_REFRESH_SLICE)...

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
```diff
@@ -1405,6 +1422,26 @@ mfxStatus ImplementationAvc::Reset(mfxVideoParam *par)
...
+            // set intra refresh cycle to 1 sec
+            extOpt2New.IntRefCycleSize =
+                (mfxU16)((newPar.mfx.FrameInfo.FrameRateExtN + newPar.mfx.FrameInfo.FrameRateExtD - 1) /
+                         newPar.mfx.FrameInfo.FrameRateExtD);
+        }
+        mfxExtCodingOption2 & extOpt2Par = GetExtBufferRef(*par);
+        extOpt2Par.IntRefCycleSize = extOpt2New.IntRefCycleSize;
+        checkStatus = MFX_WRN_VIDEO_PARAM_CHANGED;
+    }
+
```
gkrivor: Same question as above

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
filippenkos: I moved verification inside `IntRefType `condition as you suggested. 
0 value already being processed inside `SetDefaults `here
https://github.com/Intel-Media-SDK/MediaSDK/blob/7e3ef73f9b4503f68721d9b329536bf3d119dddb/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L5925
It called from `CheckVideoParam` here
https://github.com/Intel-Media-SDK/MediaSDK/blob/7e3ef73f9b4503f68721d9b329536bf3d119dddb/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L2067
And  `CheckVideoParam` is called before verification.
As for `FrameRateExtD!=0` it verified inside `CheckVideoParam` here 
https://github.com/Intel-Media-SDK/MediaSDK/blob/7e3ef73f9b4503f68721d9b329536bf3d119dddb/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L1964-L1966

_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp
filippenkos: For `ImplementationAvc::Reset` verification can't be moved inside `IntRefType` condition, because it is inside another condition which is true only when starting new sequence, but invalid value need to be processed regardless of that, so i added another check to condition.
The answer to another parts of your question is pretty much the same as above except that `CheckVideoParam` is called inside `ProcessAndCheckNewParameters` here 
https://github.com/Intel-Media-SDK/MediaSDK/blob/7e3ef73f9b4503f68721d9b329536bf3d119dddb/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw.cpp#L1264
And `ProcessAndCheckNewParameters` called at the beginning of `ImplementationAvc::Reset`

PR #1825, opened 2019-12-11
[hevce] fix ROI priority mode support

vilichev:
ROI priority mode is supported on platforms with corresponding caps
Otherwise it should be converted to delta QP mode


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -958,6 +958,7 @@ mfxStatus CheckAndFixRect(RectData *rect, MfxVideoParam const & par, ENCODE_CAPS
 #ifdef MFX_ENABLE_HEVCE_ROI
 mfxStatus CheckAndFixRoi(MfxVideoParam  const & par, ENCODE_CAPS_HEVC const & caps, mfxExtEncoderROI *ROI, bool &bROIViaMBQP)
 {
+    MFX_CHECK(ROI, MFX_ERR_INVALID_VIDEO_PARAM);
```
dmitryermilov: MFX_ERR_NULL_PTR ?

mgonchar: AFAIK, here we wanted ERR_INVALID to be returned. @vilichev could you comment?

vilichev: yes, I would keep MFX_ERR_INVALID_VIDEO_PARAM

PR #1829, opened 2019-12-12
[HEVCe]: Change alignment of surface

alexelizarov:
Revert https://github.com/Intel-Media-SDK/MediaSDK/pull/1706

Surface should be aligned by 16 according to manual
Stream's resolution determined by mfxExtHEVCParam::PicWidthInLumaSamples
and mfxExtHEVCParam::PicHeightInLumaSamples


mgonchar:
@eistarov could you comment here

eistarov:
> @eistarov could you comment here

width/height - is for input frame surface, for some codecs (like AVC) also used as a coded frame size
for HEVC coded frame size may be different and specified via mfxExtHEVCParam::PicWidthInLumaSamples/PicHeightInLumaSamples

see mediasdk-man:

> `Width`, `Height` | Width and height of the video frame in pixels; `Width` must be a multiple of 16. `Height` must be a multiple of 16 for progressive frame sequence and a multiple of 32 otherwise.

>## <a id='mfxExtHEVCParam'>mfxExtHEVCParam</a>
>...
>Attached to the [mfxVideoParam](#mfxVideoParam) structure extends it with HEVC-specific parameters. Used by both decoder and encoder.
>...
>`PicWidthInLumaSamples`  | Specifies the width of each coded picture in units of luma samples.
`PicHeightInLumaSamples` | Specifies the height of each coded picture in units of luma samples.

PR #1846, opened 2019-12-18
Unbreak -DMFX_ENABLE_SW_FALLBACK=OFF

jbeich:
Fixes https://github.com/Intel-Media-SDK/MediaSDK/issues/1845


dvrogozh:
Per my understanding `_studio/mfx_lib/encode` contains SW only implementation. I think we need to exclude it entirely from the build with the `-DMFX_ENABLE_SW_FALLBACK=OFF`. I created #1855 to address that.

PR #1854, opened 2019-12-18
[hevce][refactored] Enabled HEVCe refactored code path for Linux SKL

alariono:
Signed-off-by: Andrey Larionov <andrey.larionov@intel.com>


_studio/shared/include/mfx_config.h
```diff
@@ -123,6 +123,15 @@
     #endif
 #endif
 
+#if !defined(ANDROID)
+#define MFX_ENABLE_HEVCEHW_REFACTORING
```
dvrogozh: Why do we diverge for Android?

_studio/shared/include/mfx_config.h
```diff
@@ -123,6 +123,15 @@
...
+#endif //!defined(ANDROID)
+
+#if defined(MFX_ENABLE_HEVCEHW_REFACTORING)
+#define MFX_ENABLE_HEVCEHW_REFACTORING_WIN_SKL
```
dvrogozh: This will enable SKL only or it will enable gen9+ support?

mgonchar: Only gen9 graphics, not only SKL

dvrogozh: In this case this macro name is wrong and misleading. Please, rename to MFX_ENABLE_HEVCEHW_REFACTORING_WIN_GEN9 and MFX_ENABLE_HEVCEHW_REFACTORING_LIN_GEN9 first.

Secondly, why we switch new code for GEN9 only?

mgonchar: it is agreement with HEVCe scrum team to make switch platform by platform. now SKL has acceptable passrates and all perf, stress, quality data so it is ready for switch

_studio/shared/include/mfx_config.h
mgonchar: it is related to some build issues. @eistarov could you comment here?

eistarov: Need to update Android.mk and fix couple of warnings - see PR draft #1858

alariono:
ES done it in PR https://github.com/Intel-Media-SDK/MediaSDK/pull/1858

dvrogozh:
what is ES?

#1858 looks better. Thank you.

_studio/shared/include/mfx_config.h
dvrogozh: ok

PR #1856, opened 2019-12-19
allow ROI rectangle top/left/bottom/right to be 0 for query

guoyejun:
It is nature to fill the ROI rect with zero when query the max
supported ROI numbers. Without this change, the query does not
work for such case.

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_roi.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_roi.cpp
@@ -107,6 +107,10 @@ mfxStatus ROI::CheckAndFixROI(
 
     auto IsInvalidRect = [](const ROI::RectData& rect)
     {
+        if (rect.Top == 0 && rect.Left == 0 && rect.Right == 0 && rect.Bottom == 0)
+        {
+            return false;
+        }
         return ((rect.Left >= rect.Right) || (rect.Top >= rect.Bottom));
     };
     mfxU16 numValidROI = mfxU16(std::remove_if(roi.ROI, roi.ROI + roi.NumROI, IsInvalidRect) - roi.ROI);


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
```diff
@@ -1011,12 +1011,14 @@ mfxStatus CheckAndFixRoi(MfxVideoParam  const & par, ENCODE_CAPS_HEVC const & ca
         // Elimination of invalid rects
         // Driver requirement
         if (ROI->ROI[i].Left >= ROI->ROI[i].Right || ROI->ROI[i].Top >= ROI->ROI[i].Bottom) {
-            ROI->NumROI--;
-            for (mfxU16 j = i; j < ROI->NumROI; j++)
-                ROI->ROI[j] = ROI->ROI[j + 1];
-            changed++;
-            i--;
-            continue;
+            if (!(ROI->ROI[i].Top == 0 && ROI->ROI[i].Left == 0 && ROI->ROI[i].Right == 0 && ROI->ROI[i].Bottom == 0)) {
```
dmitryermilov: @guoyejun , HEVC encode got a new design.  Please extend the change for it. Seems this is the right place for it: https://github.com/Intel-Media-SDK/MediaSDK/blob/f65d9f2091baa584c4855405084255c1858ad809/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_roi.cpp#L108 .

mgonchar: @guoyejun Dmitry meant to introduce your change for both codebases. Also @eistarov could you look at this one?

eistarov: Need to separate ROI check behavior in Query and [run-time](https://github.com/Intel-Media-SDK/MediaSDK/blob/710d7ad7a2f9b4d025b72d5276fcc16be1a01f2e/_studio/mfx_lib/encode_hw/hevc/linux/g9/hevcehw_g9_roi_lin.cpp#L114-L120) - not sure about driver reaction on all-zero ROI.

guoyejun: @eistarov  good point, i'm afraid i have no time for such change, just pass the ball to msdk team, thanks.

dmitryermilov: @lakulako , @saosipov , can you please help to finalize the patch?

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_roi.cpp
```diff
@@ -107,6 +107,10 @@ mfxStatus ROI::CheckAndFixROI(
 
     auto IsInvalidRect = [](const ROI::RectData& rect)
     {
+        if (rect.Top == 0 && rect.Left == 0 && rect.Right == 0 && rect.Bottom == 0)
+        {
+            return false;
+        }
         return ((rect.Left >= rect.Right) || (rect.Top >= rect.Bottom));
```
saosipov: Why are we using ">=" instead of ">" here?
Left==Right and Top==Buttom is a valid case, isn't it?

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_roi.cpp
```diff
@@ -107,6 +107,10 @@ mfxStatus ROI::CheckAndFixROI(
 
     auto IsInvalidRect = [](const ROI::RectData& rect)
     {
+        if (rect.Top == 0 && rect.Left == 0 && rect.Right == 0 && rect.Bottom == 0)
```
saosipov: Why only zeros?
Any numbers should be OK.

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_roi.cpp
```diff
@@ -107,6 +107,10 @@ mfxStatus ROI::CheckAndFixROI(
...
+        {
+            return false;
+        }
         return ((rect.Left >= rect.Right) || (rect.Top >= rect.Bottom));
     };
```
saosipov: After CheckAndFixRect above we have here top/left aligned down and right/bottom aligned up for the cases when they are not aligned to BlockSize and right/bottom > top/left.
If top/left == right/bottom or top/left > right/bottom inside the same Block we would have top/left==right/bottom here and if top/left > right/bottom and they don't belong to the same Block we would have top/left > right/bottom.
Let me know if you don't agree with it?
So we have to consider a Rect as invalid only the way it was done before i.e. ((rect.Left >= rect.Right) || (rect.Top >= rect.Bottom))

guoyejun:
as mentioned above that "@eistarov good point, i'm afraid i have no time for such change, just pass the ball to msdk team, thanks."

PR #1859, opened 2019-12-19
[vp9e] Check for tiles group height exceeds picture height fix

IlyaSergeyev:
Using the wrong constant (64 instead 128) led to overstated
acceptable number of tiles (>4)


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
```diff
@@ -1505,8 +1505,8 @@ mfxStatus CheckParameters(VP9MfxVideoParam &par, ENCODE_CAPS_VP9 const &caps)
 
     if (rows && height)
     {
-        mfxU16 heightInSBs = static_cast<mfxU16>(CeilDiv(height, SB_SIZE));
```
eshiryae: Please delete definition of SB_SIZE as it's used only here

PR #1873, opened 2019-12-20
Adjust includes in some places

jbeich:
Upstreaming some fixes from [FreeBSD package](https://www.freshports.org/multimedia/intel-media-sdk):
- libc++ (unlike libstdc++) doesn't bootleg `<string>` via `<map>`
- `sysinfo` code was converted to `sysconf`, making it portable to other Unices

### DIFF
--- a/_studio/shared/umc/core/vm/src/vm_sys_info_linux32.c
+++ b/_studio/shared/umc/core/vm/src/vm_sys_info_linux32.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Intel Corporation
+// Copyright (c) 2017-2019 Intel Corporation
 // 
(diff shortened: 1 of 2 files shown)


_studio/shared/umc/core/vm/src/vm_sys_info_linux32.c
```diff
@@ -21,12 +21,8 @@
 #if defined(LINUX32)
 
 #include "vm_sys_info.h"
```
dvrogozh: Could you, please, modify:
`Copyright (c) 2017 Intel Corporation`
to
`Copyright (c) 2017-2019 Intel Corporation`
on the very top to unlock our CI?

PR #1877, opened 2019-12-23
hevce: avoid possible usage of garbage padding bytes in hash calculation

mgonchar:
Compiler may insert uninitialized padding bytes inside POD structs, so it is unsafe to calculate hash across fields boundary


dmitryermilov:
Seems it's a common, not HEVC related

mgonchar:
@dmitryermilov well, originally this code was introduced and used for hevce only. But of course change potentially have general impact, so I renamed PR

PR #1878, opened 2019-12-23
[HEVCe] Clapping of input QP

vilichev:
If QP supplied is out of range it is clapped to the supported range now
instead of dropping to the default value

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_par.cpp
@@ -2180,9 +2180,12 @@ mfxStatus CheckVideoParam(MfxVideoParam& par, MFX_ENCODE_CAPS_HEVC const & caps,
 
     if (par.mfx.RateControlMethod == MFX_RATECONTROL_CQP)
     {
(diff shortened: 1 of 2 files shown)


dmitryermilov:
@vilichev , what about new HEVCe design? Is there change already there?

vilichev:
> @vilichev , what about new HEVCe design? Is there change already there? Please confirm.

@dmitryermilov yes, it is implemented in Legacy::CheckBRC()

dmitryermilov:
I don't see similar non zero checks here https://github.com/Intel-Media-SDK/MediaSDK/blob/858dc0214619d4d6055650a46b33ed70a3836fcd/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_legacy.cpp#L4161

Is it in another place?

vilichev:
> I don't see similar non zero checks here
> 
> https://github.com/Intel-Media-SDK/MediaSDK/blob/858dc0214619d4d6055650a46b33ed70a3836fcd/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_legacy.cpp#L4161
> 
> Is it in another place?

It seems you are right, check for zero values is missing in new design. @eistarov what do you think?

PR #1881, opened 2019-12-24
[HEVCe] Update IR stripes width calculation

vilichev:
Implemented balancing of IR stripes width (they can have now different
width based on caps.IntraRefreshBlockUnitSize) to meet target CycleSize
more accurately. Stripes remainder is in the tail frames.


vilichev:
> @vilichev , what about new HEVCe design? Is there change already there? Please confirm and post a link to code.

@dmitryermilov corresponding changes are already here. https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_legacy.cpp#L2009

PR #1882, opened 2019-12-24
Extended UWP Samples are added

fzhar:


dmitryermilov:
Where I can find documentation?

npletnev:
@dmitryermilov , I'm going to upload documentation in separate change.

dmitryermilov:
> I'm going to upload documentation in separate change.

Alright.

dvrogozh:
>> I'm going to upload documentation in separate change.
> Alright.

No, this is not ok. Change can't be reviewed thoroughly without related documentation which sets expectations for the new code. Please, prepare documentation (it can be separate commit, but MUST be same PR) and return back once you will be ready.

Also, please rebase your change and remove merge commit. I see that there are some CI failures in major commit and would like to be sure we are really good to go.

dvrogozh:
This PR has another PR inside (#1906). Please decouple or we need to wait for #1906 to go in first.

fzhar:
Documentation is added. AllBuild.sln is modified

dmitryermilov:
@dvrogozh , @onabiull , please review

doc/samples/readme-extended-UWP-samples.md
```diff
@@ -0,0 +1,74 @@
+# Extended UWP Samples
+
+## Overview
+
+Extended UWP Samples solution (InteropSamples.sln) contains 3 projects:
```
dvrogozh: You recommend to use something (InteropSamples.sln) which is not under CI testing. We do not actually verify that this solution is actually functional. Why we can't direct everyone to use AllBuild.sln as we already did for tutorials?

samples/ExtendedUWPSamples/MSDKInterop/StreamWriter.cpp
```diff
@@ -0,0 +1,43 @@
+/******************************************************************************\
```
dvrogozh: That's a duplicate from sample_common. Why do we need a duplicate? If that was trivial 10 lines of code that would be ok, but it is not.... That's hundreds of lines, multiple files. From now on we will need to support same code in 2 places. Were there issues you've fixed? Refactoring done? Why you can't update, improve and reuse sample_common?

doc/samples/readme-extended-UWP-samples.md
fzhar: I think we should add it to CI than. This sln is essential for providing UWP samples as standalone package.

samples/ExtendedUWPSamples/MSDKInterop/StreamWriter.cpp
fzhar: These samples were initially designed as standalone samples package so we could provide the code with minimal dependencies.

PR #1884, opened 2019-12-25
[HEVCe] add check for qp parameters

vilichev:
If qpi is not set default qps are used


vilichev:
Corresponding changes for new HEVCe design is already here: 
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_legacy.cpp#L4170

PR #1885, opened 2019-12-25
Update codeowners

eistarov:


dmitryermilov:
@eistarov , can you please remain an hevce code reviewer while you're in a transition to a new hevce design?

eistarov:
> @eistarov , can you please remain an hevce code reviewer while you're in a transition to a new hevce design?

I can, but only for the refactored part - not the legacy one.

dmitryermilov:
Sure. Thank you!

PR #1886, opened 2019-12-25
[HEVCe] Fixed bug for LDB ref lists

vilichev:


vilichev:
Corresponding changes for new HEVCe design is already here: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_legacy_defaults.cpp#L1147

PR #1887, opened 2019-12-25
[HEVCe] disable RAB if no B slices

alexelizarov:
Set SliceIPonly for VDEnc depending on platform and TU


alexelizarov:
Corresponding changes for new HEVCe design are already here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_iddi_packer.h#L65
and
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/encode_hw/hevc/agnostic/g12/hevcehw_g12_caps.cpp#L83

PR #1889, opened 2019-12-26
[HEVCe] Remove hardcode of resolution's caps

alexelizarov:
Remove MaxPicWidth and MaxPicHeight from HardcodeCaps()
The first step in avoiding of hardcoding caps.

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/linux/g12/hevcehw_g12_caps_lin.h
+++ b/_studio/mfx_lib/encode_hw/hevc/linux/g12/hevcehw_g12_caps_lin.h
@@ -51,8 +51,6 @@ protected:
         caps.NoInterlacedField          = 1;
         caps.ParallelBRC                = 1;
         caps.MaxEncodedBitDepth         = 2;
-        caps.MaxPicWidth                = 16384;
-        caps.MaxPicHeight               = 16384;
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/encode_hw/hevc/linux/g12/hevcehw_g12_caps_lin.h
```diff
@@ -51,8 +51,8 @@ class Caps
         caps.NoInterlacedField          = 1;
```
dmitryermilov: for me this struct with hardcoded caps is a surprise. We have an interface for caps querying. @eistarov , @mgonchar , @dvrogozh , why do we need to hardcode caps?

eistarov: By our agreement with HEVC scrum, we aligned refactored code logic with legacy one by the moment of ww41, all changes made after (if any) supposed to be replicated in refactored code by HEVC scrum. I don't track status of TGL caps readiness in driver.

mgonchar: I heard that there were some issues with caps reported by driver. @saosipov may say more here, but he is on the vacation now

PR #1891, opened 2019-12-26
Enable HEVC SCC decoding.

wangyan42164:

### DIFF
--- a/api/include/mfxstructures.h
+++ b/api/include/mfxstructures.h
@@ -470,6 +470,9 @@ enum {
     MFX_PROFILE_HEVC_MAIN10           =2,
     MFX_PROFILE_HEVC_MAINSP           =3,
     MFX_PROFILE_HEVC_REXT             =4,
+#if (MFX_VERSION >= MFX_VERSION_NEXT)
+    MFX_PROFILE_HEVC_SCC              =9,
+#endif
 
     MFX_LEVEL_HEVC_1   = 10,
     MFX_LEVEL_HEVC_2   = 20,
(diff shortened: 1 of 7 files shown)


wangyan42164:
This is a draft still. I can continue to refine it. Welcome review comments. Thanks.

api/include/mfxstructures.h
```diff
@@ -470,6 +470,9 @@ enum {
     MFX_PROFILE_HEVC_MAIN10           =2,
     MFX_PROFILE_HEVC_MAINSP           =3,
     MFX_PROFILE_HEVC_REXT             =4,
+#if (MFX_VERSION >= 1031)
```
onabiull: Please replace 1031 here and in in other parts of this PR to MFX_VERSION_NEXT. There are CI configuration named **buildbot/build-mediasdk-api-next** which enable these features. API 1.31 scope already finalized.

onabiull: Please add respective changes in [mediasdk-man.md](./doc/mediasdk-man.md) to this PR.

wangyan-intel: > Please replace 1031 here and in in other parts of this PR to MFX_VERSION_NEXT. There are CI configuration named **buildbot/build-mediasdk-api-next** which enable these features. API 1.31 scope already finalized.

In fact, I try to use MFX_VERSION_NEXT at first. But it seems not to work. The macro section isn't enabled. May I set it in CMake command line? I can double check it.

onabiull: You can pass `-DAPI:STRING=latest` to CMake. API version would become **1.31+** instead of just 1.31. Respective line in product config: https://github.com/Intel-Media-SDK/product-configs/blob/master/conf_linux_public.py#L149

wangyan42164: I tried it. It works fine. I will use MFX_VERSION_NEXT instead of hard code. Thanks.

wangyan42164:
@onabiull @alariono I have refined this PR based on your comments. Could you please review it again? Thanks.

PR #1814, opened 2019-12-09
[hevce] Get MaxSliceSize/SliceStructure caps from driver

dmitryermilov:
Previosly they were hardcoded

Signed-off-by: Dmitry Ermilov <dmitry.ermilov@intel.com>

PR #1909, opened 2020-01-14
[MPEG2] Get some caps from driver

abykovsk:
Add request to driver for MPEG2 encoder configuration


_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -380,18 +387,74 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
+        (attrs[idx_map[VAConfigAttribEncMaxRefFrames]].value != 0))
+        caps.MaxNum_Reference  = attrs[idx_map[VAConfigAttribEncMaxRefFrames]].value;
+    else
+        caps.MaxNum_Reference = 1;
+
+    if ((attrs[idx_map[VAConfigAttribEncSliceStructure]].value != VA_ATTRIB_NOT_SUPPORTED) &&
+        (attrs[idx_map[VAConfigAttribEncSliceStructure]].value != 0))
+        caps.SliceStructure  = attrs[idx_map[VAConfigAttribEncSliceStructure]].value;
+    else
+        caps.SliceStructure = 2;
```
dmitryermilov: it was '1' previously. why did you change to '2'?

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -380,18 +387,74 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
+        VAConfigAttribEncMaxRefFrames,
+        VAConfigAttribEncSliceStructure
+    };
+
+    std::vector<VAConfigAttrib> attrs;
+    attrs.reserve(sizeof(attr_types) / sizeof(attr_types[0]));
+
+    for (size_t i=0; i < sizeof(attr_types)/sizeof(attr_types[0]); ++i)
+    {
+        attrs.push_back( createVAConfigAttr(attr_types[i], 0) );
```
dvrogozh: stupid question: why direct `{attr_types[i], 0}` bypassing `createVAConfigAttr` won't work? why do you need `createVAConfigAttr`?

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
dvrogozh: And what 2 actually means? Why we don't have some enum list with self speaking names?!

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -380,18 +387,74 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
+
+    if ((attrs[idx_map[VAConfigAttribEncSliceStructure]].value != VA_ATTRIB_NOT_SUPPORTED) &&
+        (attrs[idx_map[VAConfigAttribEncSliceStructure]].value != 0))
+        caps.SliceStructure  = attrs[idx_map[VAConfigAttribEncSliceStructure]].value;
+    else
+        caps.SliceStructure = 2;
 
     return MFX_ERR_NONE;
-} // mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
+} // mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps, mfxU8 codecProfileType)
```
dvrogozh: just drop this comment entirely. It is useless.

abykovsk: @dvrogozh all functions below have such a comment

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
abykovsk: @dmitryermilov  yes, it was. But driver returns SliceStructure = 2. What value should we use in this case?
@dvrogozh  SliceStructure = 2 means SliceDividerHsw, but I can't find any enum list for this parameter

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -380,18 +387,74 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
+        (attrs[idx_map[VAConfigAttribMaxPictureWidth]].value != 0))
+        caps.MaxPicWidth  = attrs[idx_map[VAConfigAttribMaxPictureWidth]].value;
+    else
+        caps.MaxPicWidth = 1920;
+
+    if ((attrs[idx_map[VAConfigAttribMaxPictureHeight]].value != VA_ATTRIB_NOT_SUPPORTED) &&
+        (attrs[idx_map[VAConfigAttribMaxPictureHeight]].value != 0))
+        caps.MaxPicHeight = attrs[idx_map[VAConfigAttribMaxPictureHeight]].value;
+    else
+        caps.MaxPicHeight = 1920;
```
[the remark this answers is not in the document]
abykovsk: @dmitryermilov ,  yes, it was. But driver returns MaxPicHeight = 1920. What value should we use in this case?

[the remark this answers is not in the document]
dmitryermilov: 'else' block for the case when driver doesn't return caps. I don't mind changing default value but it's a subject for separate commit

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
dmitryermilov: > @dmitryermilov yes, it was. But driver returns SliceStructure = 2. What value should we use in this case?

'else' block for the case when driver doesn't return caps. I don't mind changing default value but it's a subject for separate commit

> @dmitryermilov yes, it was. But driver returns SliceStructure = 2. What value should we use in this case?
see https://github.com/Intel-Media-SDK/MediaSDK/pull/1814/files#diff-21db52ca9c332ae4833c901c5be556ea

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
dmitryermilov: It's pretty useless comment..

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
mgonchar: I also don't like this function) I believe it won't work without proper casting. If you cast ```{attr_types[i], 0}``` to ```std::initializer_list<VAConfigAttrib>``` it should work:

```
attrs.emplace_back(std::initializer_list<VAConfigAttrib>(attr_types[i], 0));
```

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -345,6 +345,7 @@ VAAPIEncoder::VAAPIEncoder(VideoCORE* core)
     , m_initFrameWidth(0)
     , m_initFrameHeight(0)
     , m_layout()
+    , m_initCodecProfile(MFX_PROFILE_MPEG2_MAIN)
```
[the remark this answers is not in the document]
mgonchar: +1

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -380,18 +387,74 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
-    caps.SkipFrame         = 1;
 
     caps.SliceIPBOnly      = 1;
-    caps.MaxNum_Reference  = 1;
-    caps.MaxPicWidth       = 1920;
-    caps.MaxPicHeight      = 1088;
     caps.NoInterlacedField = 0; // Enable interlaced encoding
-    caps.SliceStructure    = 1; // 1 - SliceDividerSnb; 2 - SliceDividerHsw; 3 - SliceDividerBluRay; the other - SliceDividerOneSlice
 
+    std::map<VAConfigAttribType, size_t> idx_map;
```
mgonchar: I see we use this scheme of vector + map to handle caps in several places... I think we need a dedicated class which can handle this and we can put it in mfx_utils.h

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -389,9 +390,60 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
+        (attrs[idx_map[VAConfigAttribEncMaxRefFrames]].value != 0))
+        caps.MaxNum_Reference  = attrs[idx_map[VAConfigAttribEncMaxRefFrames]].value;
+
+    if ((attrs[idx_map[VAConfigAttribEncSliceStructure]].value != VA_ATTRIB_NOT_SUPPORTED) &&
+        (attrs[idx_map[VAConfigAttribEncSliceStructure]].value != 0))
+        caps.SliceStructure  = attrs[idx_map[VAConfigAttribEncSliceStructure]].value;
 
     return MFX_ERR_NONE;
-} // mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
+}
```
dvrogozh: Placing this comment in the very end of the function:
```} // mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)```
is a bad practice I am trying to fight with. It has 2 big problems:
1. You need to support this comment on each function prototype update. Without IDE with such integrated feature (and you can't demand the whole community to have the one) sooner or later comment becomes wrong
2. This comment gives a match on greps and actually obscures grep results. I prefer to match only those lines which actually are sensitive to my research.

File: _studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp (hunk 6 above)
Comment (dvrogozh) [reply _studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp]: Out of scope of this PR. If you @mgonchar have a BW to implement or (even better) to work with Alina to guide her to implement - please, do, but in separate PR which will cover other places in the code.

File: _studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp [hunk 8]
```diff
@@ -389,9 +390,60 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps)
...
+    for (size_t i=0; i < sizeof(attr_types)/sizeof(attr_types[0]); ++i)
+    {
+        attrs.emplace_back(VAConfigAttrib{attr_types[i], 0});
+        idx_map[ attr_types[i] ] = i;
+    }
+
+    VAEntrypoint entrypoint = VAEntrypointEncSlice;
+
+    if(codecProfileType == MFX_PROFILE_UNKNOWN)
+        codecProfileType = m_initCodecProfile;
```
DenWolf: I have a question about "m_initCodecProfile" - if I right understand your change, you introduce this member in the class, set to default value in constructor, and than use it only for setting this value for other variable.
If my understanding is correct - I think we can set default value here directly without introducing member "m_initCodecProfile"

Issue #1800, opened 2019-12-04
fake error in sample_decode 


dvrogozh:
Decoding succeeds, but error is reported:

```
# wget https://fate-suite.libav.org/h264-conformance/AUD_MW_E.264

# ./share/mfx/samples/sample_decode h264 -i AUD_MW_E.264 -hw -o aud.yuv
pretending that stream is 30fps one
pretending that aspect ratio is 1:1
libva info: VA-API version 1.6.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'iHD'
libva info: Trying to open /usr/lib64/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_6
libva info: va_openDriver() returns 0

[ERROR], sts=MFX_ERR_UNSUPPORTED(-3), AllocImpl, sts!=MFX_ERR_NONE error at /.../samples/sample_common/src/general_allocator.cpp:154
Decoding Sample Version 8.4.27.


Input video     AVC
Output format   NV12
Input:
  Resolution    176x144
  Crop X,Y,W,H  0,0,176,144
Output:
  Resolution    176x144
Frame rate      30.00
Memory type             system
MediaSDK impl           hw
MediaSDK version        1.30

Decoding started
Frame number:  100, fps: 4632.203, fread_fps: 0.000, fwrite_fps: 0.000
Decoding finished
```

dmitryermilov:
@AntonGrishin , could you take a look?

AntonGrishin:
It`s regression started from 40dbc2f98cc2f4a0a8b3e09edafb093015854988

We try to alloc video memory here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/006b768609930b938c3d32cc54b4e8d5b726ccb9/_studio/shared/src/libmfx_core_vaapi.cpp#L617

But when we pass to AllocImpl, we have no m_D3DAllocator and go to sys allocator, 
and here we get ``MFX_ERR_UNSUPPORTED`` due CheckRequestType failed
https://github.com/Intel-Media-SDK/MediaSDK/blob/006b768609930b938c3d32cc54b4e8d5b726ccb9/samples/sample_common/src/sysmem_allocator.cpp#L253

https://github.com/Intel-Media-SDK/MediaSDK/blob/006b768609930b938c3d32cc54b4e8d5b726ccb9/samples/sample_common/src/general_allocator.cpp#L154


And then we use default allocator
```
[...]

PR #1912, opened 2020-01-15
core: Use default allocator for allocate internal resources

AntonGrishin:
We use default allocator, if external allocator fails (also may leads
fake errs). So better to use default allocator right away for internal
recources.

fixes: #1800


_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -612,7 +612,9 @@ VAAPIVideoCORE::AllocFrames(
         } else
         {
             // external allocator
-            if (m_bSetExtFrameAlloc && request->Info.FourCC != MFX_FOURCC_P8)
+            if (m_bSetExtFrameAlloc &&
+                request->Info.FourCC != MFX_FOURCC_P8 &&
+                !(request->Type & MFX_MEMTYPE_INTERNAL_FRAME))
```
dmitryermilov: I didn't get it, Anton. at line 617/619, which allocator we call to? seems to be external? If we change a behavior and now allocates internal surfaces via ext allocator, then why do we need this call?

AntonGrishin: @dmitryermilov this code path still uses for alloc external resources.

dmitryermilov: From my understanding, this code block is used only for one case: when decoder at Init needs to access all memId's which had been allocated by app. I'd suggest to add a comment here:
`
// make 'fake' Alloc call to retrieve memId's of surfaces already allocated by app.
sts = (*m_FrameAllocator.frameAllocator.Alloc)(m_FrameAllocator.frameAllocator.pthis, &temp_request, response);
`

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -612,7 +612,9 @@ VAAPIVideoCORE::AllocFrames(
...
             {
                 sts = (*m_FrameAllocator.frameAllocator.Alloc)(m_FrameAllocator.frameAllocator.pthis, &temp_request, response);
```
dmitryermilov: From my understanding, this code block is used only for one case: when decoder at Init needs to access all memId's which had been allocated by app. I'd suggest to add a comment here:

```
// make 'fake' Alloc call to retrieve memId's of surfaces already allocated by app.
sts = (*m_FrameAllocator.frameAllocator.Alloc)(m_FrameAllocator.frameAllocator.pthis, &temp_request, response);

```

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -612,7 +612,9 @@ VAAPIVideoCORE::AllocFrames(
         } else
         {
             // external allocator
-            if (m_bSetExtFrameAlloc && request->Info.FourCC != MFX_FOURCC_P8)
+            if (m_bSetExtFrameAlloc &&
+                request->Info.FourCC != MFX_FOURCC_P8 &&
```
dmitryermilov: let's do "(request->Type & MFX_MEMTYPE_EXTERNAL_FRAME" instead of "!(request->Type & MFX_MEMTYPE_INTERNAL_FRAME))"

_studio/shared/src/libmfx_core_vaapi.cpp
dmitryermilov: when please see a comment below

_studio/shared/src/libmfx_core_vaapi.cpp
AntonGrishin: External block refactored.

_studio/shared/src/libmfx_core_vaapi.cpp
AntonGrishin: done

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -612,54 +612,36 @@ VAAPIVideoCORE::AllocFrames(
...
-                    {
-                        (*m_FrameAllocator.frameAllocator.Free)(m_FrameAllocator.frameAllocator.pthis, response);
-                        return MFX_ERR_MEMORY_ALLOC;
-                    }
-
-                    m_bUseExtAllocForHWFrames = true;
-                    sts = ProcessRenderTargets(request, response, &m_FrameAllocator);
-                    MFX_CHECK_STS(sts);
+                // Checking for unsupported mode - external allocator exist but Device handle doesn't set
+                if (!m_Display)
```
mgonchar: Let's use ```MFX_CHECK``` macro

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -612,54 +612,36 @@ VAAPIVideoCORE::AllocFrames(
...
-                    return TraceFrames(request, response, sts);
-                }
-                // error situation
-                else
+                if (response->NumFrameActual < request->NumFrameMin)
                 {
-                    m_bUseExtAllocForHWFrames = false;
-                    return sts;
+                    (*m_FrameAllocator.frameAllocator.Free)(m_FrameAllocator.frameAllocator.pthis, response);
+                    return MFX_ERR_MEMORY_ALLOC;
```
mgonchar: ```MFX_RETURN```

_studio/shared/src/libmfx_core_vaapi.cpp
AntonGrishin: done

_studio/shared/src/libmfx_core_vaapi.cpp
AntonGrishin: done

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -612,54 +612,35 @@ VAAPIVideoCORE::AllocFrames(
...
-                // if external allocator cannot allocate d3d frames - use default memory allocator
-                if (MFX_ERR_UNSUPPORTED == sts || MFX_ERR_MEMORY_ALLOC == sts)
-                {
-                    // Default Allocator is used for internal memory allocation only
-                    if (request->Type & MFX_MEMTYPE_EXTERNAL_FRAME)
-                        return sts;
-                    m_bUseExtAllocForHWFrames = false;
-                    sts = DefaultAllocFrames(request, response);
-                    MFX_CHECK_STS(sts);
+                m_bUseExtAllocForHWFrames = false;
```
dmitryermilov: Is it needed?

AntonGrishin: I think yes, because earlier, when sts! = mfx_err_none we set this variable to false /

dmitryermilov:
@dvrogozh , @mgonchar , can you please review?

Issue #1880, opened 2019-12-23
Clarify encoding mode (VME vs. VDENC) in sample encode and smt


dvrogozh:
Starting from 1.31 API implementation we started to change default encoding mode:
https://github.com/Intel-Media-SDK/MediaSDK/blob/073bf0f47ebe8dcf4221a7f1cbfa51d80aa4c161/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L1430

Here is code snippet:
```
#if (MFX_VERSION >= 1031)
    if ((platfrom == MFX_HW_JSL || platfrom == MFX_HW_EHL)
        && par.mfx.LowPower == MFX_CODINGOPTION_UNKNOWN)
    {
        par.mfx.LowPower = MFX_CODINGOPTION_ON;
        return sts;
    }
#endif
```

Can you, please, clarify:
1. Which is the default encoding mode for sample encode and smt for each platform?
2. If default mode for JSL/EHL is VDENC, how user can change it to VME? (I think there is only `-qsv-ff` which enables VDENC, but not disables it)

I recommend to print out some info clarifying the work mode of samples. Plus, I would consider changing command line option to something more sensible, like `--lowpower=on|off`

dmitryermilov:
> Which is the default encoding mode for sample encode and smt for each platform?

In which way would you like to have it? Just a comment in current issue or ..?

> If default mode for JSL/EHL is VDENC, how user can change it to VME? (I think there is only -qsv-ff which enables VDENC, but not disables it)
I recommend to print out some info clarifying the work mode of samples. Plus, I would consider changing command line option to something more sensible, like --lowpower=on|off

@AntonGrishin , could you please follow up?

AntonGrishin:
> > Which is the default encoding mode for sample encode and smt for each platform?
> 
> @dvrogozh , In which way would you like to have it? Just a comment in current issue or ..?

@dvrogozh any suggestions?

PR #1914, opened 2020-01-16
Add more sensible option for enable lowpower mode in se/smt

AntonGrishin:
fixes: #1880

### DIFF
--- a/samples/sample_encode/src/sample_encode.cpp
+++ b/samples/sample_encode/src/sample_encode.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************\
-Copyright (c) 2005-2019, Intel Corporation
+Copyright (c) 2005-2020, Intel Corporation
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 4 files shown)


samples/sample_encode/src/sample_encode.cpp
```diff
@@ -122,7 +122,7 @@ void PrintHelp(msdk_char *strAppName, const msdk_char *strErrorMessage, ...)
 #if (MFX_VERSION >= 1027)
     msdk_printf(MSDK_STRING("   [-round_offset_in file]  - use this file to set per frame inter/intra rounding offset(for AVC only)\n"));
 #endif
-    msdk_printf(MSDK_STRING("   [-qsv-ff]                - Enable QuickSync Fixed Function (low-power HW) encoding mode\n"));
+    msdk_printf(MSDK_STRING("   [-lowpower:<on,off>]     - Turn this option ON to enable QuickSync Fixed Function (low-power HW) encoding mode\n"));
```
dmitryermilov: I'm afraid it's not easy to "deprecate" the option. It's using in test system and some customers are used to it. I'd suggest to keep old one.

AntonGrishin: @dmitryermilov i removed them only from "help", but left it handler.

dmitryermilov: Alight

PR #1917, opened 2020-01-17
[build] Drop mfx_enc_common.cpp from mfx_common

Alexandr-Konovalov:
mfx_enc_common.cpp is added to both mfx_common_hw and mfx_common,
but later one missed MFX_VA flag, so mfx_enc_common.cpp is build
without MFX_VA_LINUX, that is incorrect.

Issue: MDP-51482


dvrogozh:
change PR name, please, to reflect what you have finally changed...

PR #1919, opened 2020-01-18
[VPP][MCTF] Fast path feature enabled for ICL

spossos:
MCTF hangs when is used in ICL with Opensource
driver, this is because Fast Path for enqueue
was not implemented. Now Fast path is available
in opensrouce driver but it needs to be called
implicitly for ICL platform. This fix determines
based on the platform what type of enqueue function
to use, EnqueueFast for Gen11 and Enqueue for all
others.

Issue: https://hsdes.intel.com/resource/2207632605
Test: manual on SKL and ICL


spossos:
I push more files than needed, need to restart the commit.

PR #1923, opened 2020-01-21
metrics_monitor: Let user specify a path to gfx device

dmitryermilov:

### DIFF
--- a/samples/metrics_monitor/sample/cttmetrics_sample.cpp
+++ b/samples/metrics_monitor/sample/cttmetrics_sample.cpp
@@ -1,6 +1,6 @@
 /* ****************************************************************************** *\
 
-Copyright (C) 2014-2018 Intel Corporation.  All rights reserved.
+Copyright (C) 2014-2020 Intel Corporation.  All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
@@ -33,7 +33,16 @@ File Name: cttmetrics_sample.cpp
 #include <string.h>
 #include <signal.h>
 #include <errno.h>
+#include <unistd.h>
 #include "cttmetrics.h"
+
+#define MIN_PERIOD_MS  10
+#define MAX_PERIOD_MS  1000
+#define DEFAULT_PERIOD_MS  500
+
+#define MIN_NUMSAMPLES  1
+#define MAX_NUMSAMPLES  1000
+#define DEFAULT_NUMSAMPLES 100
 
 volatile sig_atomic_t run = 1;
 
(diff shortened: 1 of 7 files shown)


samples/metrics_monitor/sample/cttmetrics_sample.cpp
```diff
@@ -64,9 +65,14 @@ int main(int argc, char *argv[])
         period_ms = atoi(argv[2]);
     }
 
+    if (3 < argc && NULL != argv[3])
```
dvrogozh: I think that's time to support --help option and print out help. Use getopt() standard function, please. I would like to see specification for device option format.

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -94,32 +94,91 @@ perf_event_open(struct perf_event_attr *attr,
...
+
+    /* strip off the relative path */
+    s = strrchr(path, '/');
+    if (s)
+        memmove(path, s + 1, len - (s - path) + 1);
+
+    return path;
+}
+
+const char *i915_perf_device(int i915, char *buf, long unsigned int buflen)
```
dvrogozh: static?

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -94,32 +94,91 @@ perf_event_open(struct perf_event_attr *attr,
...
+    dir = open(path, O_RDONLY);
+    if (dir != -1) {
+        len = readlinkat(dir, "device", path, pathlen - 1);
+        close(dir);
     }
-    if (n < 0)
+    if (len < 0)
+        return NULL;
+
+    path[len] = '\0';
```
dvrogozh: should dir be closed here?

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
dmitryermilov: ok

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
dmitryermilov: we close at line 114

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -94,32 +95,91 @@ perf_event_open(struct perf_event_attr *attr,
     return syscall(__NR_perf_event_open, attr, pid, cpu, group_fd, flags);
 }
 
-static uint64_t i915_type_id(void)
+static char *bus_address(int i915, char *path, int pathlen)
 {
-    char buf[1024];
-    int fd, n;
+    struct stat st;
```
dvrogozh: that's .cpp file. Maybe we will move definitions closer to where we actually use them?

samples/metrics_monitor/src/cttmetrics_i915_pmu.cpp
```diff
@@ -94,32 +95,91 @@ perf_event_open(struct perf_event_attr *attr,
...
+        if (*s == ':')
+            *s = '_';
+
+    return buf;
+}
+
+static uint64_t i915_type_id(const char *device)
+{
+    char buf[120];
+    ssize_t ret;
```
dvrogozh: same comment about definitions.

samples/metrics_monitor/sample/cttmetrics_sample.cpp
```diff
@@ -45,28 +54,67 @@ void signal_handler(int signo)
...
+{
+    printf("metrics_monitor - Monitors GPU usage per engine\n"
+        "\n"
+        "Usage: %s [OPTION]\n"
+        "\n"
+        "\tThe following parameters are optional:\n\n"
+        "\t[-h]          Show this help text.\n"
+        "\t[-s <num>]    Number of metric samples to collect during sampling period(valid range %u..%u, default %u).\n"
+        "\t[-p <ms>]     Sampling period in milliseconds(valid range %u..%u, default %u).\n"
+        "\t[-d <path>]   Path to gfx device (like /dev/dri/card* or /dev/dri/renderD*).\n"
```
dvrogozh: What's is program default behavior for device? I suggest to have some text describing app behavior in the very end after options.

dmitryermilov:
> You need to update https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/samples/readme-metrics_monitor_linux.md

@dvrogozh , Do you think we need to support versioning there?

dvrogozh:
That's something which bothers me. metrics library API/ABI might be something people have encountered for... So, I would at least bump major API version to make sure we signal we break API/ABI here. I hope this will be ok since that's still sample...

samples/metrics_monitor/sample/cttmetrics_sample.cpp
```diff
@@ -45,28 +54,68 @@ void signal_handler(int signo)
...
+        "\t              If device is not set, the tool uses render node device with smaller number."
```
dvrogozh: Will the tool search for the first Intel adapter or just try renderD128?

dmitryermilov: The tool does what's written in help output,  it uses render node device with smallest number. "smaller" is a misprint and will be fixed.

dvrogozh: Actually no. You have this line to initialize device:
```
g_ctx.gem_fd = device ? open(device, O_RDWR) : drmOpenWithType("i915", NULL, DRM_NODE_RENDER);
```

This means that by default you search for the first Intel GPU render node. You need to clarify what's the used algo and what means "first".

dmitryermilov: I DID NOT use word 'first' intentionally. Where did you find it?

dvrogozh: you do not mention that it will look for the Intel GPU either. I read this that it can found other vendor render node and fail to work.

dmitryermilov: added i915

PR #1924, opened 2020-01-21
HEVCe: changed behavior for input zero QP

### LINKED ISSUE #1894: [regression] hevc encoding
Recently MSDK broke the pre-check for gst-msdk (https://github.com/intel-media-ci/gst-plugins-bad/pull/52), the result of git bisect shows 9120f46035bca1899ea586856bca7b3a97741bb9 is the first bad commit. 

You may check  https://media-ci.ostc.intel.com/build/365163/overview for the error case

saosipov:
More details at
fixes #1894


doc/mediasdk-man.md
```diff
@@ -5734,7 +5734,7 @@ This structure specifies configurations for decoding, encoding and transcoding p
...
-`QPI, QPP, QPB`         | Quantization Parameters (`QP)  `for `I`, `P` and `B` frames, respectively, for the constant `QP` (CQP) mode.
+`QPI, QPP, QPB`         | Quantization Parameters (`QP`) for `I`, `P` and `B` frames for constant `QP` mode (CQP). All input non-zero QPs are clipped to QP range supported on hardware being used. Zero QP is not valid and means that some default value will be assigned by MediaSDK library. This value depends on the codec and can be changed in future MediaSDK releases. E.g. HEVC Encoder calculates default values based on the valid ones to follow the formula QPP = QPI + 2, QPB = QPP + 2. If QPI is not set the average value of QP range is used.
```
dmitryermilov: I'd propose this :
Quantization Parameters (`QP`) for `I`, `P` and `B` frames for constant `QP` mode (CQP). Zero QP is not valid and means that a default value will be assigned by SDK. Non-zero QPs might be clipped to supported QP range. 
**Note**: Default QP values for QPI/QPP/QPB are implementation dependent and subject to change without an additional notice in this document.

dmitryermilov: @stasefremov , @aleonenk , @eistarov , @dmitry-gurulev , can you please review this part?

stasefremov: I would add detailed information about ranges for different codecs and bit depths. Maybe a table.

dmitryermilov: >I would add detailed information about ranges for different codecs and bit depths. Maybe a table.

 @stasefremov , per codec/platform?

stasefremov: Not sure if platform limitations are subject for MSDK manual.. I'm talking about fundamental ranges which depend on codec/bitdepth.

saosipov: I'm afraid that in this case we'll have to update this data pretty often. E.g. for HEVC driver minQP is 10 for VDENC and 0 for VME, starting from ICL negative QP is supported on VME and so on. Some restrictions msdk can obtain through caps. So this table can be very hard to support.
I basically like Dima's formulation

stasefremov: Frequent changes would be required for platform limitations. I think fundamental ranges (e.g. 0-51 for HEVC 8-bit, 0-255 for VP9 8-bit, etc.) shouldn't be changed at all.

dmitryermilov: @stasefremov , we don't support '0'

stasefremov: Yes. This limitation (unsupported value 0) is essential part of parameter semantics. It should be excluded from "fundamental" range.

dvrogozh: Can we comment how AVC will behave in this case? If you will be able to cover all mediasdk supported encoders this will be really cool.

saosipov: Hey guys, are we really going to describe here all QP requirements for all codecs msdk supports? In this case we'll never have this code checked in.
Can we just say that QP range follows corresponding specification but still can be different on different platforms and to one has to use Query to get real numbers?

dmitryermilov: > Can we comment how AVC will behave in this case? If you will be able to cover all mediasdk supported encoders this will be really cool.

It is not an approach we can maintain. I'm against mentioning implementation details in the manual.

dvrogozh: In this case I suggest to drop this sentence entirely: "E.g. HEVC Encoder calculates default values based on the valid ones to follow the formula QPP = QPI + 2, QPB = QPP + 2. If QPI is not set the average value of QP range is used." Without it text seems to be ok. With it you need to add info for other codecs. At least for AVC.

dmitryermilov: We're on the same page.

stasefremov: > Can we just say that QP range follows corresponding specification

AFAIK it's not the case for HEVC high bit depth.
 
> but still can be different on different platforms and to one has to use Query to get real numbers?

Unfortunately Query can't be used to get an idea about proper range with adequate efforts. You can check particular value if you already understand what full range for the parameter is.

I agree that we don't need to mention anything related to particular implementation/platform (HW/SW restrictions for particular platform, rules for deduction of unknown QP type from known QP type, etc.). But I think we need to give an idea about full/fundamental range for QP parameter for particular codec/bitdepth.

saosipov:
Moved to #1930

dmitryermilov:
Sergey, why did we move to the pool request?

saosipov:
@dmitryermilov It was incorrectly done from external branch. Oleg asked me to remove it.
In this PR I only updated .md the way you advised earlier.

dmitryermilov:
I see. Thanks

PR #1925, opened 2020-01-22
[hevce] Enable Weighted Prediction and Fade Detection features

otsirule:


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -2377,7 +2377,11 @@ mfxStatus MfxVideoParam::GetSliceHeader(Task const & task, Task const & prevTask
                     }
 
                     if (task.m_ldb && Equal(task.m_refPicList[0], task.m_refPicList[1]))
-                        Copy(s.pwt[1], s.pwt[0]);
+                    {
```
dvrogozh: do we still need old Copy() or definition can be dropped?

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_weighted_prediction.cpp
```diff
@@ -44,7 +44,7 @@ void WeightPred::SetSupported(ParamSupport& blocks)
 void WeightPred::Query1WithCaps(const FeatureBlocks& /*blocks*/, TPushQ1 Push)
 {
     Push(BLK_CheckAndFix
-        , [this](const mfxVideoParam& /*in*/, mfxVideoParam& par, StorageW& global) -> mfxStatus
+        , [](const mfxVideoParam& /*in*/, mfxVideoParam& par, StorageW& global) -> mfxStatus
```
dvrogozh: hm. Wasn't that a fix for compiler issue to include 'this'? Not sure which compiler. I think @eistarov worked on it. Can you, please, comment?

_studio/shared/include/mfx_config.h
```diff
@@ -126,10 +126,10 @@
 #define CMAPIUPDATE
 
 #define MFX_ENABLE_HEVCEHW_REFACTORING
- 
+
 #if defined(MFX_ENABLE_HEVCEHW_REFACTORING)
     #define MFX_ENABLE_HEVCEHW_REFACTORING_LIN_GEN9
     #define MFX_ENABLE_HEVCEHW_REFACTORING_LIN_GEN11
-#endif //defined(MFX_ENABLE_HEVCEHW_REFACTORING) 
+#endif //defined(MFX_ENABLE_HEVCEHW_REFACTORING)
```
dvrogozh: can you, please, drop the comment if you will push any changes to this PR? It's clearly seen to which if belongs endif...

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_weighted_prediction.cpp
otsirule: @dvrogozh, I removed it because clang CI failed due to unused "this" in lambda. Before this code wasn't compiled because of disabled define in mfx_config.h

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
otsirule: @dvrogozh, It's probably dropped because compilation failed with old implementation. That's why I've changed the code. Before this code wasn't compiled due to disabled define in mfx_config.h

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_weighted_prediction.cpp
dvrogozh: ok. @eistarov likely fixed similar issue in the active code  that's why I remember

PR #1927, opened 2020-01-23
core: Use external allocator for find FEI ENC / PAK reconstructs

AntonGrishin:
FEI ENC/PAK uses external allocator to retrieve memId's of surfaces
which already allocated by app

Test: few different scenarios via sample_encode/sample_decode/smt/sample_fei 

fixes: 7eb0c0b (core: Use default allocator for allocate internal resources")

### DIFF
--- a/_studio/shared/src/libmfx_core_vaapi.cpp
+++ b/_studio/shared/src/libmfx_core_vaapi.cpp
@@ -614,7 +614,7 @@ VAAPIVideoCORE::AllocFrames(
             // external allocator
             if (m_bSetExtFrameAlloc &&
                 request->Info.FourCC != MFX_FOURCC_P8 &&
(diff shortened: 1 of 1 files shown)


dmitryermilov:
@mgonchar , can you please review as well? It's an important place.

mgonchar:
@dmitryermilov agree, it is very sensible place. not sure if it will work, need to check all layers. @AntonGrishin did you run full FEI validation and can you point places in code where mismatch happens and which is potential issue?

dmitryermilov:
> ... can you point places in code where mismatch happens and which is potential issue?

@mgonchar , this change is a fix for FEI regression caused by 7eb0c0b . We want to allocate all internal surfaces by internal allocator. The mismatch is for AVC FEI reconstituted surfaces **are external resources**

PR #1928, opened 2020-01-24
[hevce]: some caps are common for all platforms

alexelizarov:
Some caps should be hardcoded regardless platform


alexelizarov:
Need to eliminate HardcodeCaps function at all.

PR #1931, opened 2020-01-28
[hevce] Fix Linux MBQP

lakulako:
Now MBQP block size used by HW in cuqpMap can vary according to caps.
Added cuqpMap filling for unaligned LCU, because HW averages LCU blocks to compute per LCU QP.
 (61287).
Manually tested.
Backmerged fixes for windows.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019 Intel Corporation
+// Copyright (c) 2017-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -429,10 +429,12 @@ mfxStatus FillCUQPDataDDI(Task& task, MfxVideoParam &par, VideoCORE& core, mfxFr
         FrameLocker lock(&core, task.m_midCUQp);
         MFX_CHECK(lock.Y, MFX_ERR_LOCK_MEMORY);
 
-        for (mfxU32 i = 0; i < CUQPFrameInfo.Height; i++)
+        for (mfxU32 i = 0; i < CUQPFrameInfo.Height; i++) {
             for (mfxU32 j = 0; j < CUQPFrameInfo.Width; j++)
-                    lock.Y[i * lock.Pitch + j] = mbqp->QP[i*drBlkH/inBlkSize * pitch_MBQP + j*drBlkW/inBlkSize];
-
+                lock.Y[i * lock.Pitch + j] = mbqp->QP[i*drBlkH / inBlkSize * pitch_MBQP + j * drBlkW / inBlkSize];
+            for (mfxU32 j = CUQPFrameInfo.Width; j < lock.Pitch; j++) // hevc averages to LCU size
+                lock.Y[i * lock.Pitch + j] = lock.Y[i * lock.Pitch + CUQPFrameInfo.Width - 1];
+        }
     }
 #ifdef MFX_ENABLE_HEVCE_ROI
     else if (roi)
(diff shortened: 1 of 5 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
```diff
@@ -429,10 +429,12 @@ mfxStatus FillCUQPDataDDI(Task& task, MfxVideoParam &par, VideoCORE& core, mfxFr
         FrameLocker lock(&core, task.m_midCUQp);
         MFX_CHECK(lock.Y, MFX_ERR_LOCK_MEMORY);
 
-        for (mfxU32 i = 0; i < CUQPFrameInfo.Height; i++)
+        for (mfxU32 i = 0; i < CUQPFrameInfo.Height; i++) {
             for (mfxU32 j = 0; j < CUQPFrameInfo.Width; j++)
-                    lock.Y[i * lock.Pitch + j] = mbqp->QP[i*drBlkH/inBlkSize * pitch_MBQP + j*drBlkW/inBlkSize];
-
+                lock.Y[i * lock.Pitch + j] = mbqp->QP[i*drBlkH / inBlkSize * pitch_MBQP + j * drBlkW / inBlkSize];
+            for (mfxU32 j = CUQPFrameInfo.Width; j < lock.Pitch; j++) // hevc averages to LCU size
```
dmitryermilov: "hevc averages to LCU size" comment doesn't seem to be right. AFAIU, this new 'for' block populates last QP in a row to position [width, pitch]. So there is no averaging here.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1248,17 +1248,17 @@ bool operator!=(const ENCODE_ENC_CTRL_CAPS& l, const ENCODE_ENC_CTRL_CAPS& r)
     return !(l == r);
 }
 
-
-void CUQPMap::Init (mfxU32 picWidthInLumaSamples, mfxU32 picHeightInLumaSamples)
+// block width/height = 8 << blockSize
+void CUQPMap::Init (mfxU32 picWidthInLumaSamples, mfxU32 picHeightInLumaSamples, mfxU32 blockSize)
```
dmitryermilov: Do we begin supporting any block size? AFAIR, HEVC MBQP could work only with 16x16 blocks. It's captured in the manual. Do we need to update manual ?

dmitryermilov:
@lakulako , can you please describe in commit message what exact scenarios were fixed? E.g. did unfilled padding(because of pitch > width) could cause some non b2b?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
lakulako: The comment answers to "what for" not "what is it".
When frame is unaligned to LCU, qp for all LCU blocks are averaged in hw, so these values have to be set. Averaging is only horizontal here, btw.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
lakulako: Now only 16 and 32 are used by hw. From msdk code point of view any size 8<<N can work.

dmitryermilov: I was wrong, 16x16 is not captured in the manual:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#mfxExtMBQP
But I think it's worth mentioning in the manual that encoder can work with different block sizes and describe what API variable can set it. Leonid, can you please do it?

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
dmitryermilov: When I suggest to emphasize in comment that averaging is done in HW

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
lakulako: Oh no. This commit doesn't change anything related to mfxExtMBQP. It aligns msdk with driver/hw.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_ddi.cpp
lakulako: Added to commit comment.

dmitryermilov: I mean code comment "// hevc averages to LCU size"

lakulako: Please check

dmitryermilov:
LGTM. @saosipov , @vilichev , could you review?

PR #1932, opened 2020-01-29
[MPEG2] Refactor QueryEncodeCaps()

abykovsk:


_studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp
```diff
@@ -62,7 +62,14 @@ mfxStatus MfxHwMpeg2Encode::QueryHwCaps(VideoCORE* pCore, ENCODE_CAPS & hwCaps,
     ddi.reset( CreatePlatformMpeg2Encoder(pCore) );
     if(ddi.get() == NULL)
         return MFX_ERR_NULL_PTR;
-    mfxStatus sts = ddi->QueryEncodeCaps(hwCaps, codecProfileType);
+
+    mfxU16 codecProfile = par->mfx.CodecProfile;
```
stasefremov: Please either pass "par" by const& or check pointer for 0 before dereference. Better to pass using const& because it's not expected that "par" can be zero.

_studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp
```diff
@@ -62,7 +62,14 @@ mfxStatus MfxHwMpeg2Encode::QueryHwCaps(VideoCORE* pCore, ENCODE_CAPS & hwCaps,
...
+    if (codecProfile == MFX_PROFILE_UNKNOWN)
```
stasefremov: I would properly define profile at higher levels of the stack (e.g. where QueryHWCaps is called).

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -423,27 +418,34 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps, mfxU8 codecProfileTy
 
     if ((attrs[idx_map[VAConfigAttribMaxPictureWidth]].value != VA_ATTRIB_NOT_SUPPORTED) &&
```
stasefremov: Same check is done multiple times. Maybe better to make it inline function?

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -370,24 +371,21 @@ mfxStatus VAAPIEncoder::QueryEncodeCaps(ENCODE_CAPS & caps, mfxU8 codecProfileTy
...
-    caps.MaxPicWidth       = 1920;
-    caps.MaxPicHeight      = 1088;
-    caps.NoInterlacedField = 0; // Enable interlaced encoding
-    caps.SliceStructure    = 1; // 1 - SliceDividerSnb; 2 - SliceDividerHsw; 3 - SliceDividerBluRay; the other - SliceDividerOneSlice
+    m_caps.SliceIPBOnly      = 1;
+    m_caps.MaxNum_Reference  = 1;
+    m_caps.MaxPicWidth       = 1920;
+    m_caps.MaxPicHeight      = 1088;
+    m_caps.NoInterlacedField = 0; // Enable interlaced encoding
+    m_caps.SliceStructure    = 1; // 1 - SliceDividerSnb; 2 - SliceDividerHsw; 3 - SliceDividerBluRay; the other - SliceDividerOneSlice
```
dvrogozh: Why we don't have enum for different slice structures. @dmitryermilov : I think we touched something similar to that recently... Maybe that's time to fix that?

dmitryermilov: https://github.com/Intel-Media-SDK/MediaSDK/pull/1814

_studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp
```diff
@@ -46,7 +46,7 @@ typedef struct tagENCODE_QUERY_STATUS_DATA_tmp
 
 using namespace MfxHwMpeg2Encode;
 
-mfxStatus MfxHwMpeg2Encode::QueryHwCaps(VideoCORE* pCore, ENCODE_CAPS & hwCaps, mfxU8 codecProfileType)
+mfxStatus MfxHwMpeg2Encode::QueryHwCaps(VideoCORE* pCore, ENCODE_CAPS & hwCaps, mfxVideoParam const * par)
 {
     EncodeHWCaps* pEncodeCaps = QueryCoreInterface<EncodeHWCaps>(pCore);
```
mgonchar: beginning of this function is messy...

At least we don't need to use ```else``` after ```return``` and we should also switch to usage of ```MFX_CHECK``` macro. But that should be different commit

```
EncodeHWCaps* pEncodeCaps = QueryCoreInterface<EncodeHWCaps>(pCore);
MFX_CHECK(pEncodeCaps , MFX_ERR_UNDEFINED_BEHAVIOR);

if (pEncodeCaps->GetHWCaps<ENCODE_CAPS>(DXVA2_Intel_Encode_MPEG2, &hwCaps) == MFX_ERR_NONE)
            return MFX_ERR_NONE;

std::unique_ptr<DriverEncoder> ddi(CreatePlatformMpeg2Encoder(pCore));
MFX_CHECK_NULL_PTR1(ddi);
```

_studio/mfx_lib/shared/include/mfx_mpeg2_enc_common_hw.h
```diff
@@ -100,7 +100,7 @@ namespace MfxHwMpeg2Encode
 
     mfxStatus QueryHwCaps(VideoCORE* core,
         ENCODE_CAPS & hwCaps,
-        mfxU8 codecProfileType);
+        mfxVideoParam const * par);
```
mgonchar: Better to pass it as const ref

_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -46,9 +46,7 @@ namespace MPEG2EncoderHW
     {
         ENCODE_CAPS hwCaps {};
 
-        mfxU8 CodecProfileType = par->mfx.CodecProfile;
-
-        mfxStatus sts = MfxHwMpeg2Encode::QueryHwCaps(core, hwCaps, CodecProfileType);
+        mfxStatus sts = MfxHwMpeg2Encode::QueryHwCaps(core, hwCaps, par);
         if (sts != MFX_ERR_NONE)
```
mgonchar: this should be handled by ```MFX_CHECK```

_studio/mfx_lib/shared/include/mfx_mpeg2_encode_interface.h
```diff
@@ -149,7 +149,7 @@ namespace MfxHwMpeg2Encode
     public:
         virtual ~DriverEncoder(){}
 
-        virtual mfxStatus QueryEncodeCaps(ENCODE_CAPS & caps, mfxU8 codecProfileType) = 0;
+        virtual mfxStatus QueryEncodeCaps(ENCODE_CAPS & caps) = 0;
```
mgonchar: can we drop return status here? From implementation it seems that ```void``` is more suitable

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
abykovsk: commit wasn't merged

_studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp
abykovsk: Not need if we define profile at higher levels of the stack

_studio/mfx_lib/shared/src/mfx_mpeg2_enc_common_hw.cpp
abykovsk: Fixed

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
abykovsk: Done

dmitryermilov:
@mgonchar , @stasefremov , are your concerns addressed ?

PR #1943, opened 2020-02-03
Extension of manual  

RaguzovaTatyana:
Added a description of changes in function GetHandle()

### DIFF
--- a/doc/mediasdk-man.md
+++ b/doc/mediasdk-man.md
@@ -2073,11 +2073,11 @@ This function obtains system handles previously set by the [MFXVideoCORE_SetHand
 --- | ---
 `MFX_ERR_NONE` | The function completed successfully.
 `MFX_ERR_NOT_FOUND` | Specified handle type not found.
+`MFX_ERR_INVALID_HANDLE` | Specified handle type is invalid.
 
 **Change History**
 
-This function is available since SDK API 1.0.
-
+This function is available since SDK API 1.0. SDK API 1.30 added new return status, `MFX_ERR_INVALID_HANDLE` 
 ### <a id='MFXVideoCORE_SetBufferAllocator'>MFXVideoCORE_SetBufferAllocator</a>
 
 **Syntax**
--- a/doc/mediasdk-man.md
+++ b/doc/mediasdk-man.md
@@ -2077,7 +2077,7 @@ This function obtains system handles previously set by the [MFXVideoCORE_SetHand
 
 **Change History**
 
-This function is available since SDK API 1.0. SDK API 1.30 added new return status, `MFX_ERR_INVALID_HANDLE` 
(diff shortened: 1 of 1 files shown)


doc/mediasdk-man.md
```diff
@@ -2073,11 +2073,11 @@ This function obtains system handles previously set by the [MFXVideoCORE_SetHand
...
 `MFX_ERR_NONE` | The function completed successfully.
 `MFX_ERR_NOT_FOUND` | Specified handle type not found.
+`MFX_ERR_INVALID_HANDLE` | Specified handle type is invalid.
 
 **Change History**
 
-This function is available since SDK API 1.0.
-
+This function is available since SDK API 1.0. SDK API 1.30 added new return status, `MFX_ERR_INVALID_HANDLE`. 
 ### <a id='MFXVideoCORE_SetBufferAllocator'>MFXVideoCORE_SetBufferAllocator</a>
```
dmitryermilov: What's the difference in conditions (input arguments or something else) for returning MFX_ERR_NOT_FOUND vs MFX_ERR_INVALID_HANDLE? If we introduce MFX_ERR_INVALID_HANDLE probably, should we remove a possibility of MFX_ERR_NOT_FOUND?

mgonchar: In my understanding, ```MFX_ERR_INVALID_HANDLE``` indicates invalid ```mfxSession``` as it does in other functions

RaguzovaTatyana: MFX_ERR_INVALID_HANDLE is returned when we pass MFX_HANDLE_D3D11_DEVICE type into D3D9VideoCORE::GetHandle and similarly in case with d3d9. GetHandle for common core is checking also other handle types, and in this cases the function can return MFX_ERR_NOT_FOUND. 
Yes, I missed that MFX_ERR_INVALID_HANDLE is used in other functions as  mfxSession fail. Is it better to replace MFX_ERR_INVALID_HANDLE with MFX_ERR_NOT_FOUND?

AntonGrishin:
You need to squash commits, in the second you just added a dot

doc/mediasdk-man.md
dmitryermilov: My thoughts:
* MFX_ERR_INVALID_HANDLE is generally used for invalid MFXSession pointer (keep in mind C MSDK API). So it doesn't seem we can use this status for another purpose.
* MFX_ERR_NOT_FOUND for GetHandle is used for situations when the implementation can't find a handle_value for given HANDLE_TYPE. From my understanding, @RaguzovaTatyana , it's the case which you're describing above.
* You can't modify 1.30 because it's already finalized. 

 @mgonchar , @onabiull , @dvrogozh what do you think?

PR #1956, opened 2020-02-07
[sample_encode] Add BitrateLimit to cmd line options

pgribov:
-Added -BitrateLimit:<on,off> option to sample_encode command line options.
 BitrateLimit equals MFX_CODINGOPTION_UNKNOWN(0) by default,
 for encoder this means enabling (same behavior as MFX_CODINGOPTION_ON).
-Added warning in case when BitrateLimit is ON and bitrate changed by Query.
-Updated sample_encode and mediasdk docs.


doc/samples/readme-encode_linux.md
```diff
@@ -107,6 +107,7 @@ The following command-line switches are optional:
   | [-num_slice] | number of slices in each video frame. 0 by default. If num_slice equals zero, the encoder may choose any slice partitioning allowed by the codec standard.|
    |[-mss] |maximum slice size in bytes. Supported only with -hw and h264 codec. This option is not compatible with -num_slice option.|
   | [-mfs]|maximum frame size in bytes. Supported only with h264 and hevc codec for VBR mode.|
+  | [-BitrateLimit:<on,off>] | Turn this flag OFF to remove bitrate limitations imposed by the SDK encoder. Setting this flag OFF may lead to violation of HRD conformance and severe visual artifacts. The default value is ON, i.e. bitrate is limited.|
```
dvrogozh: These won't be artifacts, but the result of the requested bitrate, i.e. quality might be low.

Actually I disagree with the selected default value for this parameter. Following bitrate limit is needed for the small fraction of use cases, for example, encoding for DVD. Majority of the people don't need this. Moreover, they don't get it for libx264 which gives only a warning, but allows to encode with the selected bitrate. As a result when we have bitrate limit defaulted to ON we face the situation that people can't compare our encoder with someone else encoder apple to apple at least out of the box.

Summarizing:
1. I would like to see this feature be OFF by default at least on samples level. This implies both to sample encode and SMT.
2. I think we should not write about artifacts, but warn that there could be low quality if bitrate will be selected lower than required by profile/level/etc. combination. In which case it is recommended to enable this feature and see the difference.

spossos:
Please set the default behavior to OFF, limiting artificially the bitrate will cause wrong behavior for developers taking into account that not everybody (and most users are included here) are not trying to reach compliance. It should be given a warning about it if level and profile don't match the requested bitrate but should not be a limitation for the user.

pgribov:
@dmitryermilov, @stasefremov 
What do you think about changing of the default value for BitrateLimit to off.

dmitryermilov:
>What do you think about changing of the default value for BitrateLimit to off.

I'm fine about it.

GlaUgrub:
Small chance that any customer intentionally rely on current default behavior (BitrateLimit == on). For other usages we can properly communicate changed MSDK behavior (i.e. argue that customers app behavior is buggy and requires changes). So I can agree with making BitrateLimit = off by default in MSDK. It should SOLVE more problems than CAUSE (i.e. it's not a purely good change, but effect can be measured as positive).

stasefremov:
GlaUgrub == stasefremov :)

dmitryermilov:
Just to be on the same page: @stasefremov confirmed he would like to see  BitrateLimit==OFF **inside** MSDK. @dvrogozh , @spossos , what did you mean, samples level or MSDK internal behavior?

dvrogozh:
I actually meant samples level at the first place. That's where things are easier: I think we are more or less free to change defaults here especially when we first introduce the option.

I also would prefer to have this changed in mediasdk by default. Yes, some people might step into the changed behavior. However, if default does not make sense or is a bad default, I think we should be more flexible and be able to change it. The fact is that a lot of people (especially new to the mediasdk) tend to use default settings trusting they are good and when they are finding they are not good... well, this is not good:)...

So, I suggest 2 steps:
1. Let's have bitratelimit=OFF in samples, both in SE and SMT - that's this PR
2. Let's create another PR and suggest a change to the mediasdk default and discuss more there.

@GlaUgrub, @dmitryermilov : are you ok with such plan?

dmitryermilov:
>@GlaUgrub, @dmitryermilov : are you ok with such plan?

Yes

doc/mediasdk-man.md
```diff
@@ -5733,7 +5733,7 @@ This structure specifies configurations for decoding, encoding and transcoding p
...
...re for the constant bitrate (CBR), variable bitrate control (VBR) and [CQP HRD](#Appendix_F) algorithms.<br><br>The SDK encoders follow the Hypothetical Reference Decoding (HRD) model. The HRD model assumes that data flows into a buffer of the fixed size `BufferSizeInKB` with a constant bitrate `TargetKbps`. (Estimate the targeted frame size by dividing the framerate by the bitrate.)<br><br>The decoder starts decoding after the buffer reaches the initial size `InitialDelayInKB`, which is equivalent to reaching an initial delay of `InitialDelayInKB*8000/TargetKbps`ms. Note: In this context, KB is 1000 bytes and Kbps is 1000 bps.<br><br>If `InitialDelayInKB` or `BufferSizeInKB` is equal to zero, the value is calculated using bitrate, frame rate, profile, level, and so on.<br><br>`TargetKbps` must be specified for encoding initialization. In CBR/VBR modes, if BitrateLimit is on,`TargetKbps` may be changed on Query.<br><br>For variable bitrate control, the `MaxKbps` parameter specifies the maximum bitrate at which the encoded data enters the Video Buffering Verifier (VBV) buffer. If `MaxKbps` is equal to zero, the value is calculated from bitrate, frame rate, profile, level, and so on.
```
dmitryermilov: "In CBR/VBR modes, if BitrateLimit is on" - please put BitrateLimit under ''

"if BitrateLimit is on,`TargetKbps` may be changed on Query" - only on Query ? What about Init() without Query()?

samples/sample_encode/include/pipeline_encode.h
```diff
@@ -184,6 +184,8 @@ struct sInputParams
     mfxU16 AdaptiveB;
     mfxU32 nMaxFrameSize;
 
+    mfxU16 BitrateLimit;
```
dmitryermilov: Perhaps it's a time to introduce "Tri-state" as a type on samples level? Not in this commit, of course.
@dvrogozh ?

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -876,10 +877,18 @@ mfxStatus CEncodingPipeline::AllocFrames()
...
 
+    mfxU16 tmpTargetKbps = m_mfxEncParams.mfx.TargetKbps;
+
     // Querying encoder
     sts = GetFirstEncoder()->Query(&m_mfxEncParams, &m_mfxEncParams);
     MSDK_CHECK_STATUS(sts, "Query (for encoder) failed");
 
+    if (MFX_WRN_INCOMPATIBLE_VIDEO_PARAM == sts && tmpTargetKbps != m_mfxEncParams.mfx.TargetKbps)
+    {
+        msdk_printf(MSDK_STRING("WARNING: Target bitrate was changed from %d kbps to %d kbps due to default limitations.\n"), tmpTargetKbps, m_mfxEncParams.mfx.TargetKbps);
```
dmitryermilov: "due to default limitations" - IMHO this is unclear. Better to re-phrase or even remove.

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -876,10 +877,18 @@ mfxStatus CEncodingPipeline::AllocFrames()
...
+        msdk_printf(MSDK_STRING("         To prevent this, use -BitrateLimit:off flag (see docs for additional details).\n\n"));
```
dmitryermilov: Is BitrateLimit can be the only reason for TargetKbps adjustment? @GlaUgrub , can you please comment?

stasefremov: 1) AFAIK BitrateLimit is the only reason which can change bitrate value.
2) Value 'of "TargetKbps" can be changed by MSDK if pair ("TargetKbps", "BRCParamMultipier") was provided. Sometimes MSDK changes the multiplier (e.g. pair (10000, 20) can be changed to (40000, 5)). But bitrate value given by the pair remains unchanged.

samples/sample_encode/include/pipeline_encode.h
dvrogozh: I am not sure I understand what you mean. PR, please:).

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -876,10 +877,18 @@ mfxStatus CEncodingPipeline::AllocFrames()
...
+    mfxU16 initialTargetKbps = m_mfxEncParams.mfx.TargetKbps;
+    auto co2 = m_mfxEncParams.GetExtBuffer<mfxExtCodingOption2>();
+
     // Querying encoder
     sts = GetFirstEncoder()->Query(&m_mfxEncParams, &m_mfxEncParams);
     MSDK_CHECK_STATUS(sts, "Query (for encoder) failed");
 
+    if (co2->BitrateLimit != MFX_CODINGOPTION_OFF && initialTargetKbps != m_mfxEncParams.mfx.TargetKbps)
+    {
+        msdk_printf(MSDK_STRING("WARNING: Target bitrate was changed from %d kbps to %d kbps because -BitrateLimit:on option specified.\n"), initialTargetKbps, m_mfxEncParams.mfx.TargetKbps);
```
dvrogozh: I would put just that:
"WARNING: -BitrateLimit:on: target bitrate was changed from %d kbps to %d kbps"

doc/mediasdk-man.md
```diff
@@ -3980,7 +3980,7 @@ The application can attach this extended buffer to the [mfxVideoParam](#mfxVideo
...
-`BitrateLimit` | Turn off this flag to remove bitrate limitations imposed by the SDK encoder. This flag is intended for special usage models and usually the application should not set it. Setting this flag may lead to violation of HRD conformance and severe visual artifacts. See the [CodingOptionValue](#CodingOptionValue) enumerator for values of this option. The default value is ON, i.e. bitrate is limited. This parameter is valid only during initialization.
+`BitrateLimit` | Turn off this flag to remove bitrate limitations imposed by the SDK encoder. Setting this flag off may lead to violation of HRD conformance and there could be low quality if entered by the user bitrate will be lower than suggested by the SDK encoder. If BitrateLimit is on, for non CQP modes, if `TargetKbps` is lower than suggested by the SDK encoder, it will be changed to suggested. See the [CodingOptionValue](#CodingOptionValue) enumerator for values of this option. The default value is ON, i.e. bitrate is limited. This parameter is valid only during initialization.
```
dvrogozh: How about this text:

Modify bitrate to be in the range imposed by the SDK encoder. Setting this flag off may lead to violation of HRD conformance. Mind that specifying bitrate below the SDK encoder range might significantly affect quality. If on this option takes effect in non CQP modes: if `TargetKbps` is not in the range imposed by the SDK encoder, it will be changed to be in the range. See the [CodingOptionValue](#CodingOptionValue) enumerator for values of this option. The default value is ON, i.e. bitrate is limited. This parameter is valid only during initialization.

doc/samples/readme-encode_linux.md
```diff
@@ -107,6 +107,7 @@ The following command-line switches are optional:
   | [-num_slice] | number of slices in each video frame. 0 by default. If num_slice equals zero, the encoder may choose any slice partitioning allowed by the codec standard.|
    |[-mss] |maximum slice size in bytes. Supported only with -hw and h264 codec. This option is not compatible with -num_slice option.|
   | [-mfs]|maximum frame size in bytes. Supported only with h264 and hevc codec for VBR mode.|
+  | [-BitrateLimit:<on,off>] | Turn this flag ON to set bitrate limitations imposed by the SDK encoder: if target bitrate is lower than suggested by the SDK encoder, it will be changed to suggested. The default value is OFF, i.e. bitrate is not limited.|
```
dvrogozh: Modify bitrate to be in the range imposed by the SDK encoder. Setting this flag off may lead to violation of HRD conformance. The default value is OFF, i.e. bitrate is not limited.|

doc/samples/readme-multi-transcode_linux.md
```diff
@@ -102,6 +102,7 @@ ParFile is extension of what can be achieved by setting pipeline in the command
 |  -q <quality\>| Quality parameter for JPEG encoder; in range [1,100], 100 is the best quality|
 |-l <numSlices\>|  Number of slices for encoder; default value 0|
 |-mss <maxSliceSize\>|Maximum slice size in bytes. Supported only with -hw and h264 codec. This option is not compatible with -l option.|
+|-BitrateLimit:<on,off>| Turn this flag ON to set bitrate limitations imposed by the SDK encoder: if target bitrate is lower than suggested by the SDK encoder, it will be changed to suggested. The default value is OFF, i.e. bitrate is not limited.|
```
dvrogozh: Modify bitrate to be in the range imposed by the SDK encoder. Setting this flag off may lead to violation of HRD conformance. The default value is OFF, i.e. bitrate is not limited.|

samples/sample_encode/src/pipeline_encode.cpp
pgribov: okay, changed

doc/mediasdk-man.md
pgribov: changed

doc/samples/readme-encode_linux.md
pgribov: changed

doc/samples/readme-multi-transcode_linux.md
pgribov: changed

dmitryermilov:
@pgribov , please confirm you can do for SMT and then we'll merge this PR

pgribov:
@dvrogozh , @dmitryermilov 
in this PR smt changes have already been added

PR #1958, opened 2020-02-11
[hevce] Passing deblocking filter params for HEVC (backmerge)

YurSkur:


eistarov:
> Do not use `ExtraParametersCheck` for hevce it is stub for FEI only!
> 
> Additionally, I don't think we need this check at all. @eistarov could you comment here?

I don't see any sense to keep any special logic to report warning for run-time parameters from EncodeFrameAsync without any ability to report which parameter was "changed" (in fact it wasn't - we just assume invalid value will be properly processed in async routine), we even can't be sure user gets this warning because of ERR_MORE_DATA.

YurSkur:
> Do not use `ExtraParametersCheck` for hevce it is stub for FEI only!
> 
> Additionally, I don't think we need this check at all. @eistarov could you comment here?

done

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -1970,6 +1970,13 @@ void MfxVideoParam::SyncMfxToHeadersParam(mfxU32 numSlicesForSTRPSOpt)
     m_pps.deblocking_filter_control_present_flag  = 1;
     m_pps.deblocking_filter_disabled_flag = !!m_ext.CO2.DisableDeblockingIdc;
     m_pps.deblocking_filter_override_enabled_flag = 1; // to disable deblocking per frame
+#if MFX_VERSION >= MFX_VERSION_NEXT
+    if (!m_pps.deblocking_filter_disabled_flag)
+    {
+        m_pps.beta_offset_div2 = mfxI8(m_ext.CO3.DeblockingBetaOffset * 0.5);
```
mgonchar: no need to use float arithmetic, just do
```
m_pps.beta_offset_div2 = mfxI8(m_ext.CO3.DeblockingBetaOffset / 2);
```

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw_utils.cpp
```diff
@@ -2411,14 +2418,22 @@ mfxStatus MfxVideoParam::GetSliceHeader(Task const & task, Task const & prevTask
...
             s.beta_offset_div2 = 0;
             s.tc_offset_div2 = 0;
         }
      }
+#if MFX_VERSION >= MFX_VERSION_NEXT
+     mfxExtCodingOption3 *ext3 = ExtBuffer::Get(task.m_ctrl);
+     if (ext3 && (ext3->DeblockingAlphaTcOffset != m_ext.CO3.DeblockingAlphaTcOffset || ext3->DeblockingBetaOffset != m_ext.CO3.DeblockingBetaOffset)
+              && m_pps.deblocking_filter_override_enabled_flag && !s.deblocking_filter_disabled_flag)
+     {
+         s.beta_offset_div2 = mfxI8(ext3->DeblockingBetaOffset * 0.5);
```
mgonchar: same

dmitryermilov:
@YurSkur , please squash commits

PR #1969, opened 2020-02-14
HEVCe: Fix min num refs for P-pyramid

eistarov:


_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_legacy_defaults.cpp
```diff
@@ -247,7 +247,7 @@ class GetDefault
         mfxU16 NumRefActiveP[8];
         par.base.GetNumRefActive(par, &NumRefActiveP, nullptr, nullptr);
         mfxU16 RefActiveP = *std::max_element(NumRefActiveP, NumRefActiveP + Size(NumRefActiveP));
```
mgonchar: ```*std::max_element(std::begin(NumRefActiveP), std::end(NumRefActiveP));```

PR #1973, opened 2020-02-18
Modify for fixing static code scan issues in CIV on WW_08 2020

Kexiaox:


dmitryermilov:
And fix build please

dmitryermilov:
And fix code alignment

_studio/mfx_lib/encode_hw/vp9/include/*Backtrace*
```diff
@@ -0,0 +1,5 @@
+Debugger entered--Lisp error:(void-variable wq)
```
dmitryermilov: What is it?

_studio/mfx_lib/encode_hw/vp9/include/*Backtrace*~
```diff
@@ -0,0 +1,5 @@
+Debugger entered--Lisp error: (void-variable wq)
```
dmitryermilov: same

Kexiaox:
> I still would like to stack which calls static analyzer complains

line 255 in "mfx_enc_common.cpp" file

dmitryermilov:
> line 255 in "mfx_enc_common.cpp" file

this line contains another function
https://github.com/Intel-Media-SDK/MediaSDK/blob/c3b4182a028945369e4a45befaa510a6cd7b5819/_studio/mfx_lib/shared/src/mfx_enc_common.cpp#L255

Please give good example

LukeZheZhu:
> I still would like to stack which calls static analyzer complains

Hi @dmitryermilov,
This issue is caused by assert(p). The assert(p) works only when (#ifdef NDEBUG). In other situation, it will not work.

dmitryermilov:
> > I still would like to stack which calls static analyzer complains
> 
> Hi @dmitryermilov,
> This issue is caused by assert(p). The assert(p) works only when (#ifdef NDEBUG). In other situation, it will not work.

Still, please share example of call stack

dmitryermilov:
This is a false positive issue. mfxExtBufferRefProxy::operator T&() can be used only for cases when the buffer we're looking is available m_extParam. For cases when we're not sure , another getter function should be used: https://github.com/Intel-Media-SDK/MediaSDK/blob/97de8773fa0f66e7a2d0f9fd9de878fb698a5fbd/_studio/mfx_lib/shared/src/mfx_enc_common.cpp#L255

So right now I don't see any violations in the code of such rules.

PR #1974, opened 2020-02-18
[HEVCe R] Fix va context input width and height

YurSkur:


dmitryermilov:
@YurSkur , could you describe in **commit message** what par.mfx.FrameInfo.Width/par.mfx.FrameInfo.Hight was wrong. It's not clear to me.

dmitryermilov:
What does "External parameters weren't taken into account" mean? Both PicWidthInLumaSamples and par.mfx.FrameInfo.Width are external.

PR #1982, opened 2020-02-25
12b support on sample level

dmitryermilov:


xhaihao:
I may get a P016 YUV file by running the command below after applying this PR. 

$> sample_decode h265 -i input_12bit_420.h265 -o output.yuv 

However I saw a green image in a YUV viewer.

xhaihao:
Does it make sense to use P016 for a 12bit output? Why not to add a new FOURCC code (e.g. P012) ?

dmitryermilov:
MSDK is cross platform library. P016 was used because of MSFT: https://docs.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format
There is no P012 there.

dmitryermilov:
>>However I saw a green image in a YUV viewer.
Thanks. Need to check

sunxinpeng:
I also get green output image by running 12bit decode command after applying this patch.

dmitryermilov:
I checked on TGL and didn't find the issue you're talking about. I guess you used P016 as FOURCC format when opened YUV in some YUV viewer. Please use P012 instead
Screenshot from 2020-02-27 04-55-02
Screenshot from 2020-02-27 04-55-08

dmitryermilov:
Please also note that "sample_decode h265 -i input_12bit_420.h265 -o output.yuv -vaapi" won't work with current version of sample_decode. It's because at decoder Init the application (sample_decode in our case) has to set Shift=1 (see [mfxFrameInfo](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#mfxFrameInfo)::Shift). 
Inside DecodeHeader the decoder initializes mfxVideoParams with right Shift value:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp#L127

but it can work only if at this time application already passed vaDisplayId to MSDK session via SetHandle(). Right now sample_decode doesn't do it. It's a bug.

xhaihao:
It seems the P012 format used in your tool is different from other P012 formats, e.g. DRM_FORMAT_P012 (https://github.com/torvalds/linux/blob/bfdc6d91a25f4545bcd1b12e3219af4838142ef1/include/uapi/drm/drm_fourcc.h#L254-L259, where the 12 MSB in 2-bytes contains color value)

dmitryermilov:
Okay. Is it an issue? 10/12 bit formats can be LSB or MSB.

xhaihao:
Yes 10/12bit formats can be LSB or MSB, but if HW uses MSB, I think it would be better to use MSB in MediaSDK too. AFAIK HW uses MSB for 10bit decoding, so I presume it is MSB for 12bit decoding too.

dmitryermilov:
Did you have a change to check mfxFrameInfo::Shift ? As I said it depends on the Shift. If you need output in system memory with MSB, use Shift=1 in decoder mfxVideoParams.

dmitryermilov:
Did you have a change to check [mfxFrameInfo](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#mfxFrameInfo)::Shift ? As I said it depends on the Shift. If you need output in system memory with MSB, use Shift=1 in decoder mfxVideoParams.

xhaihao:
it seems mfxFrameAllocRequest::Shift=1 for internal surfaces but mfxFrameAllocRequest::Shift=0 for external surface.

According to MSDK's output, the format is P016, how does user know the 12 LSB contains the color values in the output file? 

```
Input video     HEVC
Output format   P016
```

dmitryermilov:
>how does user know the 12 LSB contains the color values in the output file?

By BitDepth:
https://github.com/Intel-Media-SDK/MediaSDK/blob/9890dff6064a7f8fe738899fc0c39b33a2d181b5/api/include/mfxstructures.h#L56

PR #1983, opened 2020-02-25
[sample_encode,smt][doc] Fix BitrateLimit option in samples

pgribov:
Flag works with MFX_CODEC_AVC only, it is ignored with other codecs.
A corresponding condition was corrected.


samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -607,9 +607,10 @@ mfxStatus CEncodingPipeline::InitMfxEncParams(sInputParams *pInParams)
 
     // configure the depth of the look ahead BRC if specified in command line
     if (pInParams->nLADepth || pInParams->nMaxSliceSize || pInParams->nMaxFrameSize || pInParams->nBRefType ||
+        (pInParams->BitrateLimit && pInParams->CodecId == MFX_CODEC_AVC) ||
```
dvrogozh: Why we need to check for AVC in sources if for other codecs it is just ignored? Does this leads to any failures or warnings or what?

pgribov: Only AVC part of the code processes this flag, other don't (same as ignoring).

On sample level we set BitrateLimit to MFX_CODINGOPTION_OFF (32) by default, so BitrateLimit is always not 0. Thus, codingOption2 is always created, but it may not be supported with other codecs, and it leads to an error in the next step of checking ext buffers.

pgribov:
@dvrogozh , @dmitryermilov , @onabiull

pgribov:
@dvrogozh , @dmitryermilov , @onabiull
Please, review

PR #1986, opened 2020-02-26
[hevce][refactored] Enabled HEVCe refactored code path for Linux

vilichev:


_studio/shared/include/mfx_config.h
```diff
@@ -130,6 +130,7 @@
 #if defined(MFX_ENABLE_HEVCEHW_REFACTORING)
     #define MFX_ENABLE_HEVCEHW_REFACTORING_LIN_GEN9
     #define MFX_ENABLE_HEVCEHW_REFACTORING_LIN_GEN11
+    #define MFX_ENABLE_HEVCEHW_REFACTORING_LIN_TGL
```
dvrogozh: GEN9, GEN11 and suddenly TGL. Why not GEN12? Can you, please, be consistent.

vilichev: Done. Removed usage of MFX_ENABLE_HEVCEHW_REFACTORING_LIN_* since currently all Linux platforms are moved to refactored path.

Issue #1874, opened 2019-12-23
V4L2 in MSDK 


salimerko:
1.) How to properly build sample_encode via cmake to support V4L2? All I found was about build.pl which is not in MediaSDK repository, but in samples repository which is deprecated.So I did checkout master and manually defined "ENABLE_V4L2_SUPPORT" in whole sample folder and build the code.
Also I had to uncomment pParams in sample_encode/src/pipeline_encode.cpp to build it

2.) FFplay show input from /dev/video0, but when I run encode sample:
./sample_encode h264 -i::v4l2 -o out -w 1920 -h 1080 -d /dev/video0 -UYVY

it report error in sample_common/src/v4l2_util.cpp:277) : STREAMON failed: Device or resource busy
Aborted (core dumped)

Thanks for any help

dmitryermilov:
From @onabiull :

> Build system look for for ENABLE_V4L2 variable. To enable V4L2 pass  `-DENABLE_V4L2=ON`
> to cmake.
> Looks like commit https://github.com/Intel-Media-SDK/MediaSDK/commit/0ff71e66fc6e6c8500ede212c2dcf3d997a5c784 erroneously comment out `pParams` in function signature.

salimerko:
Thank you, that eliminated first problem, now it builded with v4l2, however, the second problem still persists. Here is the output

~/msdk-19.3.1/build/__bin/release$ ./sample_encode h264 -i::v4l2 -o output -w 1920 -h 1080 -d /dev/video0 -UYVY
libva info: VA-API version 1.6.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'iHD'
libva info: Trying to open /opt/intel/mediasdk/lib64/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_6
libva info: va_openDriver() returns 0
Driver Caps:
  Driver: "uvcvideo"
  Card: "USB Capture HDMI+: USB Capture "
  Bus: "usb-0000:00:14.0-2"
  Version: 1.0
  Capabilities: 84a00001
G_FMT(start): width = 1280, height = 720, 4cc = YUYV, BPP = 2560 sizeimage = 1843200 field = 1
G_FMT(pre): width = 1920, height = 1080, 4cc = UYVY, BPP = 0 sizeimage = 0 field = 0
G_FMT(final): width = 1920, height = 1080, 4cc = YUYV, BPP = 3840
Encoding Sample Version 19.3.1

[...]

PR #1987, opened 2020-02-27
sample_encode: Fix V4L2

AntonGrishin:
Camera capture linear, but driver expected tiled memory with yuy2,uyvy
formats. When passed VA_SURFACE_ATTRIB_MEM_TYPE_VA with custom buffer,
driver sets memory to non-tiled.

Fixes:#1874

Signed-off-by: Anton Grishin <anton.grishin@intel.com>

### DIFF
--- a/samples/sample_common/src/vaapi_allocator.cpp
+++ b/samples/sample_common/src/vaapi_allocator.cpp
@@ -164,6 +164,7 @@ static mfxStatus GetVAFourcc(mfxU32 fourcc, unsigned int &va_fourcc)
         (VA_FOURCC_P208 != va_fourcc) &&
         (VA_FOURCC_P010 != va_fourcc) &&
         (VA_FOURCC_YUY2 != va_fourcc) &&
+        (VA_FOURCC_UYVY != va_fourcc) &&
 #if (MFX_VERSION >= 1027)
         (VA_FOURCC_Y210 != va_fourcc) &&
         (VA_FOURCC_Y410 != va_fourcc) &&
@@ -293,7 +294,7 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
         if( VA_FOURCC_P208 != va_fourcc )
         {
             unsigned int format;
-            VASurfaceAttrib attrib[2];
+            VASurfaceAttrib attrib[3];
             int attrCnt = 0;
 
             attrib[attrCnt].type          = VASurfaceAttribPixelFormat;
@@ -332,7 +333,25 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
             {
                 format = VA_RT_FORMAT_RGBP;
             }
+#ifdef ENABLE_V4L2_SUPPORT
+            if ((m_export_mode & vaapiAllocatorParams::PRIME) && (format == VA_RT_FORMAT_YUV422))
+            {
+                VASurfaceAttribExternalBuffers buffer_desc{};
+                buffer_desc.pixel_format        = va_fourcc;
+                buffer_desc.width               = request->Info.Width;
+                buffer_desc.height              = request->Info.Height;
 
+                attrib[attrCnt].type            = VASurfaceAttribMemoryType;
+                attrib[attrCnt].flags           = VA_SURFACE_ATTRIB_SETTABLE;
+                attrib[attrCnt].value.type      = VAGenericValueTypeInteger;
+                attrib[attrCnt++].value.value.i = VA_SURFACE_ATTRIB_MEM_TYPE_VA;
+
+                attrib[attrCnt].type            = (VASurfaceAttribType)VASurfaceAttribExternalBufferDescriptor;
+                attrib[attrCnt].flags           = VA_SURFACE_ATTRIB_SETTABLE;
+                attrib[attrCnt].value.type      = VAGenericValueTypePointer;
+                attrib[attrCnt++].value.value.p = &buffer_desc;
+            }
+#endif
             va_res = m_libva->vaCreateSurfaces(m_dpy,
                                     format,
                                     request->Info.Width, request->Info.Height,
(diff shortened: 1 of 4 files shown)


samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -332,7 +333,26 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
             {
                 format = VA_RT_FORMAT_RGBP;
             }
+#ifdef ENABLE_V4L2_SUPPORT
+            if ((m_export_mode & vaapiAllocatorParams::PRIME) && (format == VA_RT_FORMAT_YUV422))
+            {
+                VASurfaceAttribExternalBuffers buffer_desc;
```
dmitryermilov: VASurfaceAttribExternalBuffers buffer_desc{};

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -332,7 +333,26 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
...
+                buffer_desc.pixel_format        = va_fourcc;
+                buffer_desc.width               = request->Info.Width;
+                buffer_desc.height              = request->Info.Height;
+                buffer_desc.flags               = 0;
```
dmitryermilov: drop

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -332,7 +333,26 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
             {
                 format = VA_RT_FORMAT_RGBP;
             }
+#ifdef ENABLE_V4L2_SUPPORT
+            if ((m_export_mode & vaapiAllocatorParams::PRIME) && (format == VA_RT_FORMAT_YUV422))
```
dmitryermilov: why VA_RT_FORMAT_YUV422 only?

samples/sample_common/src/vaapi_allocator.cpp
```diff
@@ -332,7 +333,26 @@ mfxStatus vaapiFrameAllocator::AllocImpl(mfxFrameAllocRequest *request, mfxFrame
...
+                attrib[attrCnt].flags           = VA_SURFACE_ATTRIB_SETTABLE;
+                attrib[attrCnt].value.type      = VAGenericValueTypeInteger;
+                attrib[attrCnt++].value.value.i = VA_SURFACE_ATTRIB_MEM_TYPE_VA;
 
+                attrib[attrCnt].type            = (VASurfaceAttribType)VASurfaceAttribExternalBufferDescriptor;
+                attrib[attrCnt].flags           = VA_SURFACE_ATTRIB_SETTABLE;
+                attrib[attrCnt].value.type      = VAGenericValueTypePointer;
+                attrib[attrCnt++].value.value.p = &buffer_desc;
+            }
+#endif
```
dmitryermilov: just a question: does it force linear memory type?

samples/sample_encode/include/pipeline_encode.h
```diff
@@ -291,11 +291,11 @@ class CEncodingPipeline
 
     void SetNumView(mfxU32 numViews) { m_nNumView = numViews; }
     virtual void  PrintInfo();
-
+#if defined (ENABLE_V4L2_SUPPORT)
```
dmitryermilov: Silly question: do we need this #if...#endif

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -1206,9 +1206,9 @@ CEncodingPipeline::CEncodingPipeline()
 
     MSDK_ZERO_MEMORY(m_EncResponse);
     MSDK_ZERO_MEMORY(m_VppResponse);
-
+#ifdef ENABLE_V4L2_SUPPORT
```
dmitryermilov: same

samples/sample_common/src/vaapi_allocator.cpp
AntonGrishin: done

samples/sample_common/src/vaapi_allocator.cpp
AntonGrishin: done

samples/sample_common/src/vaapi_allocator.cpp
AntonGrishin: for V4L2 we used only YUY2/UYVY, so no need another formats

samples/sample_common/src/vaapi_allocator.cpp
AntonGrishin: Yep, we passed part of description with VA_SURFACE_ATTRIB_MEM_TYPE_VA  for force user description and disable tiling:

https://github.com/intel/media-driver/blob/7857d6bea214b755d610c1a478d4e1e23da31cdb/media_driver/linux/common/ddi/media_libva.cpp#L2307

https://github.com/intel/media-driver/blob/7857d6bea214b755d610c1a478d4e1e23da31cdb/media_driver/linux/common/ddi/media_libva_util.cpp#L329

samples/sample_encode/include/pipeline_encode.h
AntonGrishin: I think so, also it`s way to fix build w/ and w/o V4L2

samples/sample_common/src/vaapi_allocator.cpp
dmitryermilov: This is an allocator. It's transparent to a fixed range FOURCC particular cameras can output.

samples/sample_common/src/vaapi_allocator.cpp
dvrogozh: I am looking into that and ask myself: how I should get this from the VAAPI specification that specifying this attribute will force Linear type?!

IMHO: we need add this to VAAPI specification. What you are doing here is just a workaround matching (accidentally) current driver behavior. Start creating VAAPI spec. proposal!

PR #1991, opened 2020-03-02
Revert "[AVCe] Changed ImplementationAvc Init and Reset behavior ..."

pgribov:
This reverts commit 1f8456f10bdb204b0ea3067df7f5a3e8a1de407f.


dmitryermilov:
@pgribov , what is the motivation to revert? Could you add this info in commit message.

dmitryermilov:
> @pgribov , what is the motivation to revert? Could you add this info in commit message.

I see you did it. Thanks.

PR #1993, opened 2020-03-03
[sample_decode] Pass device to session right after session is created

### LINKED ISSUE #1988: sample_decode doesn't pass vaDisplayId to MSDK session before DecodeHeader
Please see https://github.com/Intel-Media-SDK/MediaSDK/pull/1982#issuecomment-591931538 . It's a bug and has to be addressed ASAP. SetHandle has to be called before any DecodeHeader, Query calls. 
See manual: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#working-with-va-api-applications :

> Because the SDK does not create internal acceleration device on Linux, the application must always share it with the SDK. This sharing should be done before any actual usage of the SDK, including capability query and component initialization.

Please check other samples, tutorials.
Once it's done, please remove this code:
https://github.com/Intel-Media-SDK/MediaSDK/blob/757ea72a2065c03942571a16c0cbe9cff51ef245/samples/sample_decode/src/pipeline_decode.cpp#L856

egorovdanil:
Fixes #1988

### DIFF
--- a/samples/sample_decode/src/pipeline_decode.cpp
+++ b/samples/sample_decode/src/pipeline_decode.cpp
@@ -389,6 +389,20 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
         m_bVppFullColorRange = pParams->bUseFullColorRange;
     }
 
+    m_bVppIsUsed = IsVppRequired(pParams);
+    if (m_bVppIsUsed)
+        m_bDecOutSysmem = pParams->bUseHWLib ? false : true;
+    else
+        m_bDecOutSysmem = pParams->memType == SYSTEM_MEMORY;
+
+    m_eWorkMode = pParams->mode;
+
+    m_monitorType = pParams->monitorType;
+    // create device and allocator
+#if defined(LIBVA_SUPPORT)
+    m_libvaBackend = pParams->libvaBackend;
+#endif // defined(MFX_LIBVA_SUPPORT)
+
     sts = GetImpl(*pParams, initPar.Implementation);
     MSDK_CHECK_STATUS(sts, "GetImpl failed");
 
@@ -401,6 +415,30 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
 
     sts = m_mfxSession.QueryIMPL(&m_impl); // get actual library implementation
     MSDK_CHECK_STATUS(sts, "m_mfxSession.QueryIMPL failed");
+
+    bool isDeviceRequired = false;
+    mfxHandleType hdl_t;
+#if D3D_SURFACES_SUPPORT
+    isDeviceRequired = m_memType != SYSTEM_MEMORY || !m_bDecOutSysmem;
+    hdl_t =
+#if MFX_D3D11_SUPPORT
+        D3D11_MEMORY == m_memType ? MFX_HANDLE_D3D11_DEVICE :
+#endif // #if MFX_D3D11_SUPPORT
+        MFX_HANDLE_D3D9_DEVICE_MANAGER;
+#elif LIBVA_SUPPORT
+    isDeviceRequired = true; // on Linux MediaSDK doesn't create device internally
+    hdl_t = MFX_HANDLE_VA_DISPLAY;
+#endif
+    if (isDeviceRequired)
+    {
+        sts = CreateHWDevice();
+        MSDK_CHECK_STATUS(sts, "CreateHWDevice failed");
+        mfxHDL hdl = NULL;
+        sts = m_hwdev->GetHandle(hdl_t, &hdl);
+        MSDK_CHECK_STATUS(sts, "m_hwdev->GetHandle failed");
+        sts = m_mfxSession.SetHandle(hdl_t, hdl);
+        MSDK_CHECK_STATUS(sts, "m_mfxSession.SetHandle failed");
+    }
 
     if (pParams->bIsMVC && !CheckVersion(&version, MSDK_FEATURE_MVC)) {
         msdk_printf(MSDK_STRING("error: MVC is not supported in the %d.%d API version\n"),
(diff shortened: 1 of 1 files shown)


egorovdanil:
> This is not what I meant. Can you please simply move SetHandle call from CreateAllocator() to place right after MSDK sessions is being created?

I moved the SetHandle call from CreateAllocator, but I also had to move the parameters settings for the Allocator.

egorovdanil:
@dmitryermilov, done for Linux.

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -402,6 +402,39 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
...
+    else
+        m_bDecOutSysmem = pParams->memType == SYSTEM_MEMORY;
+
+    m_eWorkMode = pParams->mode;
+
+    m_monitorType = pParams->monitorType;
+    // create device and allocator
+#if defined(LIBVA_SUPPORT)
+    m_libvaBackend = pParams->libvaBackend;
+#endif // defined(MFX_LIBVA_SUPPORT)
```
dmitryermilov: Why do we need this here? I think it should remain where it was. Or I'm missing somethig?

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -402,6 +402,39 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
...
+
+#ifdef LIBVA_SUPPORT
+    if(m_memType != SYSTEM_MEMORY || !m_bDecOutSysmem
```
dmitryermilov: For VAAPI path, we don't need this "if ()"

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -487,22 +520,15 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
         MSDK_CHECK_STATUS(sts, "Plugin load failed");
     }
 
-    // Populate parameters. Involves DecodeHeader call
```
dmitryermilov: Please keep

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1352,17 +1368,9 @@ mfxStatus CDecodingPipeline::CreateAllocator()
...
-
-        // provide device manager to MediaSDK
         VADisplay va_dpy = NULL;
         sts = m_hwdev->GetHandle(MFX_HANDLE_VA_DISPLAY, (mfxHDL *)&va_dpy);
         MSDK_CHECK_STATUS(sts, "m_hwdev->GetHandle failed");
-        sts = m_mfxSession.SetHandle(MFX_HANDLE_VA_DISPLAY, va_dpy);
-        MSDK_CHECK_STATUS(sts, "m_mfxSession.SetHandle failed");
 
         vaapiAllocatorParams *p_vaapiAllocParams = new vaapiAllocatorParams;
         MSDK_CHECK_POINTER(p_vaapiAllocParams, MFX_ERR_MEMORY_ALLOC);
```
dmitryermilov: This is right

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1397,22 +1405,6 @@ mfxStatus CDecodingPipeline::CreateAllocator()
...
-            MSDK_CHECK_STATUS(sts, "CreateHWDevice failed");
-
-            // provide device manager to MediaSDK
-            VADisplay va_dpy = NULL;
-            sts = m_hwdev->GetHandle(MFX_HANDLE_VA_DISPLAY, (mfxHDL *)&va_dpy);
-            MSDK_CHECK_STATUS(sts, "m_hwdev->GetHandle failed");
-            sts = m_mfxSession.SetHandle(MFX_HANDLE_VA_DISPLAY, va_dpy);
-            MSDK_CHECK_STATUS(sts, "m_mfxSession.SetHandle failed");
-        }
-#endif
```
dmitryermilov: and this

samples/sample_decode/src/pipeline_decode.cpp
egorovdanil: @dmitryermilov, we use [m_eWorkMode](https://github.com/Intel-Media-SDK/MediaSDK/blob/406334ff98465e8c5dd8be42b0934c8aac7c772e/samples/sample_decode/src/pipeline_decode.cpp#L1034), [m_monitorType ](https://github.com/Intel-Media-SDK/MediaSDK/blob/406334ff98465e8c5dd8be42b0934c8aac7c772e/samples/sample_decode/src/pipeline_decode.cpp#L1066), [m_libvaBackend](https://github.com/Intel-Media-SDK/MediaSDK/blob/406334ff98465e8c5dd8be42b0934c8aac7c772e/samples/sample_decode/src/pipeline_decode.cpp#L1070)  in CreateHWDevice(). And m_bDecOutSysmem will be needed for this condition `if(m_memType != SYSTEM_MEMORY || !m_bDecOutSysmem)` on windows.

samples/sample_decode/src/pipeline_decode.cpp
egorovdanil: Done

egorovdanil:
@dmitryermilov, done for Windows.

samples/sample_decode/src/pipeline_decode.cpp
egorovdanil: Done

samples/sample_decode/src/pipeline_decode.cpp
dmitryermilov: Okay.

You know, even code seems to be correct in terms of logic/language mistakes, I somehow don't like how it looks. There are "low level" details which hide initialization steps logic. It's generally a global problem of our samples. Two proposals for current PR:
- please move code at lines 405-413  before GetImpl()
- then do something like :
```
bool isDeviceRequired = false;
mfxHandleType hdl_t = unkown;
#if Windows
  isDeviceRequired = m_memType != SYSTEM_MEMORY || !m_bDecOutSysmem;
  hdl_t = D3D9 or D3D11;
#elseif Linux
  isDeviceRequired = true;
  hdl_t = VA_DISPLAY;
#endif
if (isDeviceRequired) 
{
   CreateHWDevice();
   mfxHDL hdl = NULL;
   m_hwdev->GetHandle(hdl_t, &hdl);
   m_mfxSession.SetHandle(hdl_t, hdl);
}
```

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1308,9 +1340,6 @@ mfxStatus CDecodingPipeline::CreateAllocator()
     if (m_memType != SYSTEM_MEMORY || !m_bDecOutSysmem)
     {
 #if D3D_SURFACES_SUPPORT
-        sts = CreateHWDevice();
-        MSDK_CHECK_STATUS(sts, "CreateHWDevice failed");
-
         // provide device manager to MediaSDK
```
dmitryermilov: remove!

dmitryermilov:
Also, please rename commit message to highlight nature of the change. Something like "Pass device to session right after session is created"

samples/sample_decode/src/pipeline_decode.cpp
egorovdanil: @dmitryermilov, done, but the `hdl_t` is not initialized as `unknown` when declared, because it is not provided in `mfxHandleType`.

Issue #1817, opened 2019-12-09
[h264] GPU hang with my ROI encoding code on skylake+ubuntu16.04


guoyejun:
Hi,

I'm trying to enable ROI encoding for ffmpeg->msdk(qsv) path, but met GPU hang issue, hope to get help, thanks.

my system is skylake + ubuntu 16.04
my msdk version: e6caad2d1e00380f8cab045de62567a0a4a53a53
my iHD version: 644fc6d2bb9e6271a2b91578b1bc2f63275d184a

Firstly, I tried with msdk's sample and verified that it works.
./sample_multi_transcode -i::h265 str352x288.h265 -hw -n 1 -o::h264 ./str.roi.h264 -roi_file roi

Then, i add roi encoding code based on msdk's sample (sample_encode), and it works too. See command blow (out352x288.yuv just contain 1 frame).
$ ./sample_encode h264 -i out352x288.yuv -hw -o out.h264 -w 352 -h 288 
I also uploaded my change at https://github.com/guoyejun/MediaSDK/commit/20644f25e5204ae9b9760219bf12f101e20ac52d in case you are interested.

Then, I did a similar change in ffmpeg to enable roi encoding.  See my change at https://github.com/guoyejun/ffmpeg/commit/4edb1516096e9b2880144d55afde2e1b2ad36f1d, (they are hard code for easy debug).

the h265 encoder works correctly with command:
./ffmpeg -s 352x288 -i out352x288.yuv -c:v hevc_qsv -y qsv.roi.h265

but, there is GPU hang with h264 encoder with command:
./ffmpeg -s 352x288 -i out352x288.yuv -c:v h264_qsv -y qsv.roi.h264 

Since msdk's roi encoding API does not tie with encoders, and my code for h265 works, I would say my code is basically right.   And there is GPU hang with ffmpeg + my code, while no issue with sample_encode + my code, I would guess there is a limitation (or special requirements) from msdk for the roi encoding, but looks that I did not find hints in msdk's document.

I've tried to enable LIBVA_TRACE for the two cases (ffmpeg + my code  VS sample_encode + my code), and no weird found between the logs. I even did some tricky code above libva to make the log of LIBVA_TRACE exactly the same (except the time records), still same issue.

i also attached GPU hang's log from /sys/kernel/debug/dri/0/i915_error_state, see
[i915_error_state.txt](https://github.com/Intel-Media-SDK/MediaSDK/files/3939516/i915_error_state.txt)

[...]

PR #1996, opened 2020-03-04
[h264e] Temporary disable MFE in MFE+ROI non CQP case

pgribov:
MFE+ROI encoding in non CQP mode produces gpu hang.
This commit temporary disabled this case.


dmitryermilov:
"зп

dmitryermilov:
@pgribov , is it a fix for https://github.com/Intel-Media-SDK/MediaSDK/issues/1817? 
If so please add "Fixes: #1817 " in commit message

dmitryermilov:
@pgribov , please address CI issue

PR #1998, opened 2020-03-05
[hevce] Update checking logic for valid dirty rectangle

JeremyShang:
Take dirty rectangle(0, 0, 0, 0) as a valid dirty rectangle and means
that frame is not changed.


JeremyShang:
PR already been pulled (see for PR#1997)

Issue #1995, opened 2020-03-04
sample_vpp cannot convert to Y210 and Y410 formats


VUNNYSODHI:
When using sample_vpp for conversion from yuv2 to Y210 or Y410 format there is unsupported error coming.
**E.g**
$ sample_vpp -vaapi -scc yuy2 -sw 352 -sh 288 -dcc y210 -i test_YUY2_352x288.yuv -o Y210-output.yuv
libva info: VA-API version 1.7.0
libva info: User environment variable requested driver 'iHD'
libva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_7
libva info: va_openDriver() returns 0

[ERROR], sts=MFX_ERR_UNSUPPORTED(-3), InitMemoryAllocator, pProcessor->pmfxVPP->Query failed at /home/test/github/msdk/samples/sample_vpp/src/sample_vpp_utils.cpp:759

[ERROR], sts=MFX_ERR_UNSUPPORTED(-3), InitResources, InitMemoryAllocator failed at /home/test/github/msdk/samples/sample_vpp/src/sample_vpp_utils.cpp:806

[ERROR], sts=MFX_ERR_UNSUPPORTED(-3), main, InitResources failed at /home/test/github/msdk/samples/sample_vpp/src/sample_vpp.cpp:523

However, conversion to NV12 is working fine.
sample_vpp -vaapi -scc yuy2 -sw 352 -sh 288 -dcc nv12 -i Prog_Packed_YUY2_352x288.yuv -o NV12output.yuv

**Environment details:**
Libva: Commit 48f446b32a4fe76d542c56c87f9427e8d14e073d
MSDK: Commit a92b7e0ef074a189f5057107b90aa9237620ab7a
Media-driver: 1b48bb5df6e91e224b66c9bf1e9f7ab686b75e95
OS: Ubuntu 18.04

Its should be working as per: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/samples/readme-vpp_linux.md

VUNNYSODHI:
Also found there is no support in "mfxU32 Str2FourCC" function for Y210 and Y410 inside "samples/sample_vpp/src/sample_vpp_parser.cpp" file

dmitryermilov:
[...]

PR #1999, opened 2020-03-05
[sample_vpp] Add conversion support for Y210 and Y410 formats

egorovdanil:
Issue #1995

Signed-off-by: egorovdanil <danil.egorov@intel.com>

### DIFF
--- a/samples/sample_vpp/src/sample_vpp_parser.cpp
+++ b/samples/sample_vpp/src/sample_vpp_parser.cpp
@@ -414,6 +414,16 @@ mfxU32 Str2FourCC( msdk_char* strInput )
     {
         fourcc = MFX_FOURCC_UYVY;
     }
+#if (MFX_VERSION >= 1027)
+    else if ( 0 == msdk_stricmp(strInput, MSDK_STRING("y210")) )
+    {
+        fourcc = MFX_FOURCC_Y210;
+    }
+    else if ( 0 == msdk_stricmp(strInput, MSDK_STRING("y410")) )
+    {
+        fourcc = MFX_FOURCC_Y410;
+    }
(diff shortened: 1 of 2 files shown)


samples/sample_vpp/src/sample_vpp_parser.cpp
```diff
@@ -414,6 +414,16 @@ mfxU32 Str2FourCC( msdk_char* strInput )
...
+#if (MFX_VERSION >= 1027)
+    else if ( 0 == msdk_stricmp(strInput, MSDK_STRING("y210")) )
+    {
+        fourcc = MFX_FOURCC_Y210;
+    }
+    else if ( 0 == msdk_stricmp(strInput, MSDK_STRING("y410")) )
+    {
+        fourcc = MFX_FOURCC_Y410;
+    }
+#endif
```
dmitryermilov: could you add else block and throw an exception if we can't find a corresponding if block

egorovdanil: @dmitryermilov, I did this via call vppPrintHelp() and return MFX_ERR_UNSUPPORTED status.

dmitryermilov:
Thanks. LGTM. One last question: did you try it on TGL? I mean does NV12-> Y210 start working?

egorovdanil:
> Thanks. LGTM. One last question: did you try it on TGL? I mean does NV12-> Y210 start working?

@dmitryermilov, I checked that yuy2 -> Y210 and yuy2 -> Y410 works on TGL.

VUNNYSODHI:
Thanks @egorovdanil and @dmitryermilov for quick fix for this issue.

PR #2019, opened 2020-03-18
[hevce] Fix Gen12 caps

vilichev:
FeatureBlocks queue is reordered to set Gen12 caps after Gen9


vilichev:
> Year check failed in CI. Please adjust to 2020.

Done

PR #2033, opened 2020-03-23
[AVCe] VA may return multiple values in VAConfigAttribEncSliceStructure

Alexandr-Konovalov:
Previously combined values were mapped to 0.

Issue: MDP-61999

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -90,21 +90,18 @@ VAProfile ConvertProfileTypeMFX2VAAPI(mfxU32 type)
 
 } // VAProfile ConvertProfileTypeMFX2VAAPI(mfxU8 type)
 
-mfxU8 ConvertSliceStructureVAAPIToMFX(mfxU8 structure)
+static mfxU8 ConvertSliceStructureVAAPIToMFX(mfxU8 structure)
 {
-    switch (structure)
-    {
-        case VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS:
-            return 1;
-        case VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS:
-            return 2;
-        case VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS:
-            return 3;
-        case VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS:
-            return 4;
-        default:
-            return 0;
-    }
+    if (structure & VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS)
+        return 1;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS)
+        return 2;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS)
+        return 3;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS)
+        return 4;
+    else
+        return 0;
 }
 
 mfxStatus SetHRD(
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -886,7 +886,6 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
         VAConfigAttribMaxPictureWidth,
         VAConfigAttribEncParallelRateControl,
         VAConfigAttribEncMaxRefFrames,
-        VAConfigAttribEncSliceStructure,
         VAConfigAttribEncROI,
         VAConfigAttribEncTileSupport
     };
--- a/_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
@@ -398,7 +398,6 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(mfxU16 codecProfile)
         VAConfigAttribMaxPictureWidth,
         VAConfigAttribEncSkipFrame,
         VAConfigAttribEncMaxRefFrames,
-        VAConfigAttribEncSliceStructure
     };
 
     std::vector<VAConfigAttrib> attrs;
@@ -432,9 +431,6 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(mfxU16 codecProfile)
 
     if (CheckAttribValue(attrs[idx_map[VAConfigAttribEncMaxRefFrames]].value))
         m_caps.MaxNum_Reference  = attrs[idx_map[VAConfigAttribEncMaxRefFrames]].value;
-
-    if (CheckAttribValue(attrs[idx_map[VAConfigAttribEncSliceStructure]].value))
-        m_caps.SliceStructure  = attrs[idx_map[VAConfigAttribEncSliceStructure]].value;
 
     return MFX_ERR_NONE;
 }


Alexandr-Konovalov:
Multiple values are set by the driver since [media_libva_caps: add VAConfigAttribEncSliceStructure caps for VAEntrypointEncSliceLP](https://github.com/intel/media-driver/pull/837/commits/edf3cc8a6ca788a1069b638a39dbef1dfb37a8e0).

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -92,19 +92,16 @@ VAProfile ConvertProfileTypeMFX2VAAPI(mfxU32 type)
...
+    if (structure & VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS)
+        return 1;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS)
+        return 2;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS)
+        return 3;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS)
+        return 4;
+    else
+        return 0;
```
dmitryermilov: If "VA may return multiple values in VAConfigAttribEncSliceStructure " so, we need changes for other encoders?
Plus, there should be a reference to something (va.h or whatever) which says that "VA may return multiple values in VAConfigAttribEncSliceStructure"

Alexandr-Konovalov: > Plus, there should be a reference to something (va.h or whatever) which says that "VA may return multiple values in VAConfigAttribEncSliceStructure"

@dmitryermilov , in va.h I see

[This attribute determines slice structures supported by the driver for encoding.](https://github.com/intel/libva/blob/64520e9ec90ed30e016d7c633d746b3bf538c702/va/va.h#L567)

Is having plurals in _structures_ enough?

dvrogozh: This is directly stated in the VAAPI as a generic thing for any config attributes: https://github.com/intel/libva/blob/64520e9ec90ed30e016d7c633d746b3bf538c702/va/va.h#L795
```
/**
 * Configuration attributes
 * If there is more than one value for an attribute, a default
 * value will be assigned to the attribute if the client does not
 * specify the attribute when creating a configuration
 */
typedef struct _VAConfigAttrib {
    VAConfigAttribType type;
    uint32_t value; /* OR'd flags (bits) for this attribute */
} VAConfigAttrib;

```

dvrogozh: And yes, highly likely we need same change for all other encoders.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -92,19 +92,16 @@ VAProfile ConvertProfileTypeMFX2VAAPI(mfxU32 type)
 
 mfxU8 ConvertSliceStructureVAAPIToMFX(mfxU8 structure)
```
dvrogozh: static

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
mgonchar: Not really familiar with this caps enums, so I'm a little bit confused. Are we sure that we have correct order here?

For example if ```structure == VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS | A_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE | VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS``` as in code snippet from link, should we return ```2``` or ```3```?

dvrogozh: ordering you say.... hm... so, if encoder supports few which one msdk should select? Less restrictive or more restrictive or what? Do we select something here or check caps?

Alexandr-Konovalov: @dmitryermilov , @mgonchar , @dvrogozh , 

Regarding other encoders, I see that only MPEG2e and HEVCe read `VAConfigAttribEncSliceStructure` attribute, and both don't use it. So quite natural is not read it.

Regarding ordering. CAPS is just read, the driver declares it support several `VA_ENC_SLICE_STRUCTURE_`, so it's for MediaSDK to decide. 

For `structure == VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS | VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE | VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS` the patch returns 2 to keep legacy (i.e., pre-[edf3cc8](https://github.com/intel/media-driver/pull/837/commits/edf3cc8a6ca788a1069b638a39dbef1dfb37a8e0)) behavior. That might be sub-optimal.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
Alexandr-Konovalov: Done.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
mgonchar: @Alexandr-Konovalov  yes, I understand that it is for MSDK to decide, so let's decide how we want to handle these caps if we assume that they might overlap

Alexandr-Konovalov: @mgonchar So, are you for reordering checks in `ConvertSliceStructureVAAPIToMFX()` to get maximal supported values? In [MakeSliceDivider()](https://github.com/Intel-Media-SDK/MediaSDK/blob/19d11bf1cfd4c7699d1d146d7c3e43c39ffd6038/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L7405l) `VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS` is falling back to `SliceDividerHsw()`, but that's not a problem.

dvrogozh: Well, unfortunately I don't know the background here. So, below is my gut feeling. It seems that these  slice structures are ordered in the more strict variant on top and more relaxed on the bottom. It seems that what you have on the bottom might be ultimately better since it gives more choices to encoder. So, I would revert the order of the list.

You need to talk with our teammates who better understand in encoders. @aleonenk, @splotnikv, @mgonchar

dmitryermilov:
@Alexandr-Konovalov , can you please check what slice_structure value UMD previously returned in caps?

Alexandr-Konovalov:
> can you please check what slice_structure value UMD previously returned in caps?

@dmitryermilov , Do you mean VAConfigAttribEncSliceStructure value? It was  `VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS | VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE`, and `VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE` is masked out my MediaSDK. See [media_libva_caps: add VAConfigAttribEncSliceStructure caps for VAEntrypointEncSliceLP](https://github.com/intel/media-driver/pull/837/commits/edf3cc8a6ca788a1069b638a39dbef1dfb37a8e0).

dmitryermilov:
> and VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE is masked out my MediaSDK

Here https://github.com/Intel-Media-SDK/MediaSDK/blob/ce37d6c41f763b9a913ef3d83528c3e9cd774b51/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp#L1573

Alexandr-Konovalov:
Close, as the issue was resolved via [Revert "media_libva_caps: add VAConfigAttribEncSliceStructure caps for](https://github.com/intel/media-driver/commit/e90d17e).

PR #2034, opened 2020-03-23
sample encode: Add support for LA plugin with HEVC

AntonGrishin:


dmitryermilov:
LGTM
@ogolikov , would you like to review MSDK LA part?
@AntonGrishin , please provide info about tests (manual and/or automation) you did to make sure this change didn't cause regressions.

_studio/mfx_lib/fei/h264_la/mfx_h264_la.cpp
```diff
@@ -483,6 +481,33 @@ mfxStatus VideoENC_LA::Init(mfxVideoParam *par)
     mfxFrameAllocRequest request = { };
     request.Info = m_video.mfx.FrameInfo;
 
+    if (m_video.IOPattern == MFX_IOPATTERN_IN_SYSTEM_MEMORY)
```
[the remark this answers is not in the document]
AntonGrishin: m_raw used for copy surfaces to video memory in case of system input.
If DS factor > 1, then we alloc pool with wrong size, because we already
changed request.

samples/sample_encode/include/pipeline_encode.h
```diff
@@ -401,7 +428,12 @@ class CEncodingPipeline
     virtual mfxStatus AllocateSufficientBuffer(mfxBitstreamWrapper& bs);
     virtual mfxStatus FillBuffers();
     virtual mfxStatus LoadNextFrame(mfxFrameSurface1* pSurf);
-    virtual void LoadNextControl(mfxEncodeCtrlWrap*& pCtrl, mfxU32 encSurfIdx);
+    virtual void LoadNextControl(mfxEncodeCtrl*& pCtrl, mfxU32 encSurfIdx);
```
alariono, line 431: Added new methods... Where is implementation?

AntonGrishin: @alariono please see pipeline_encode.cpp:2696
It`s just signature change.

Issue #2048, opened 2020-03-26
asc: avx2 code path is not actually enabled


dvrogozh:
AVX2 optimized path in ASC implementation:
https://github.com/Intel-Media-SDK/MediaSDK/blob/19d11bf1cfd4c7699d1d146d7c3e43c39ffd6038/_studio/shared/asc/src/asc_avx2_impl.cpp#L22

isn't actually enabled because we build mediasdk for sse4.2:
https://github.com/Intel-Media-SDK/MediaSDK/blob/19d11bf1cfd4c7699d1d146d7c3e43c39ffd6038/builder/ConfTargets.cmake#L23

PR #2049, opened 2020-03-26
asc: enable avx2 code path

dvrogozh:
Fixes: #2048

### DIFF
--- a/_studio/shared/asc/CMakeLists.txt
+++ b/_studio/shared/asc/CMakeLists.txt
@@ -28,14 +28,22 @@ set( defs "" )
 set( sources "" )
 set( sources.plus "" )
 
+add_library(asc_avx2 OBJECT ${CMAKE_CURRENT_SOURCE_DIR}/src/asc_avx2_impl.cpp)
+target_compile_options(asc_avx2 PRIVATE -mavx2)
+configure_build_variant(asc_avx2 none)
+
+add_library(asc_sse4 OBJECT ${CMAKE_CURRENT_SOURCE_DIR}/src/asc_sse4_impl.cpp)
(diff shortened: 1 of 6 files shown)


_studio/shared/asc/CMakeLists.txt
```diff
@@ -28,14 +28,22 @@ set( defs "" )
...
+
 list( APPEND sources
     ${CMAKE_CURRENT_SOURCE_DIR}/src/asc.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/src/asc_avx2_impl.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/asc_c_impl.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/src/asc_sse4_impl.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/iofunctions.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/motion_estimation_engine.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/tree.cpp
+    $<TARGET_OBJECTS:asc_avx2>
```
onabiull, line 45: Object libraries should go to the LIBS or LIBS_NOVARIANT variables. Please move.

dvrogozh: See here: https://cmake.org/cmake/help/v3.5/command/add_library.html#object-libraries

dvrogozh: @onabiull : do you consider this resolved or you still think we need some change?

onabiull: My bad :)

PR #2052, opened 2020-03-27
[h264e] Enable MaxSliceSize feature for Linux LowPower encoder

dmitrybrazhkin:

### DIFF
--- a/_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
+++ b/_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -229,6 +229,7 @@ namespace MfxHwH264Encode
         VABufferID m_rirId;                     // VAEncMiscParameterRIR
         VABufferID m_qualityParamsId;           // VAEncMiscParameterEncQuality
         VABufferID m_miscParameterSkipBufferId; // VAEncMiscParameterSkipFrame
+        VABufferID m_maxSliceSizeId;            // VAEncMiscParameterMaxSliceSize
 #if defined (MFX_ENABLE_H264_ROUNDING_OFFSET)
         VABufferID m_roundingOffsetId;          // VAEncMiscParameterCustomRoundingControl
 #endif
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -317,6 +317,41 @@ mfxStatus SetFrameRate(
     return MFX_ERR_NONE;
 }
 
+static mfxStatus SetMaxSliceSize(
+    const UINT   userMaxSliceSize,
+    VADisplay    vaDisplay,
+    VAContextID  vaContextEncode,
+    VABufferID & maxSliceSizeBuf_id)
+{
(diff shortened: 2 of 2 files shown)


_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
```diff
@@ -229,6 +229,7 @@ namespace MfxHwH264Encode
         VABufferID m_rirId;                     // VAEncMiscParameterRIR
         VABufferID m_qualityParamsId;           // VAEncMiscParameterEncQuality
         VABufferID m_miscParameterSkipBufferId; // VAEncMiscParameterSkipFrame
+        VABufferID m_maxSliceSizeId;            // VAEncMiscParameterMaxSliceSize
```
dmitryermilov: need to init at constructor

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1536,6 +1572,7 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
         const auto sliceCapabilities = AV(VAConfigAttribEncSliceStructure);
         const auto sliceStructure = sliceCapabilities & ~VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE;
         m_caps.ddi_caps.SliceStructure = ConvertSliceStructureVAAPIToMFX(sliceStructure);
+        m_caps.ddi_caps.SliceLevelRateCtrl = sliceCapabilities & VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE ? 1 : 0;
```
dmitryermilov: Please avoid hardcode (like 1/0). Give name them

_studio/mfx_lib/shared/include/mfx_h264_encode_vaapi.h
dmitrybrazhkin: @dmitryermilov , please explain what is wrong with mfx_h264_encode_vaapi.cpp, line 1329:

>     , m_maxSliceSizeId(VA_INVALID_ID)

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -2789,6 +2837,9 @@ mfxStatus VAAPIEncoder::Execute(
     configBuffers.push_back(m_frameRateId);
     configBuffers.push_back(m_qualityLevelId);
 
+    if (VA_INVALID_ID != m_maxSliceSizeId)
```
dmitryermilov: m_maxSliceSizeId might be VA_INVALID_ID  because of some mistake.
```
if (extOpt2->MaxSliceSize != 0)
    configBuffers.push_back(m_maxSliceSizeId);
```
is better and it keeps the same style.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
dmitrybrazhkin: SliceLevelRateCtrl is binary flag. So using 1/0 is reasonable for this case. Even if we try to give meaningful name, it would be "enabled/disabled", "true/false", "supported/unsupported" and so on, the same meaning to 1/0

dmitryermilov: Ahh, it's just bool:
https://github.com/Intel-Media-SDK/MediaSDK/blob/bcaea9ec4aca9ac8c08a524d40403c609c143a1a/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h#L241

I thought it's enum. Then okay.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
dmitrybrazhkin: done

PR #2059, opened 2020-03-31
[HEVCe r] Fix MBQP passing to driver

alexelizarov:


dmitryermilov:
@alexelizarov  , is this final one? :)

alexelizarov:
Yes, seems to be final

PR #2062, opened 2020-04-03
Enlarge execution latency threshold for decoding one frame

aobolensk:


aobolensk:
> This is not good. In sample_defs.h there is a hardcoded macro. Instead of explicit modification of this macro you modify place where it's used. Why?

Fixed

samples/sample_common/include/sample_defs.h
```diff
@@ -116,7 +116,7 @@ enum LibVABackend
 #define MSDK_ENC_WAIT_INTERVAL 300000
 #define MSDK_VPP_WAIT_INTERVAL 300000
 #define MSDK_SURFACE_WAIT_INTERVAL 300000
-#define MSDK_DEVICE_FREE_WAIT_INTERVAL 30000
+#define MSDK_DEVICE_FREE_WAIT_INTERVAL 300000
```
dmitryermilov, line 119: And now the question, why do we really need separate MSDK_DEVICE_FREE_WAIT_INTERVAL while generally MFX_WNR_DEVICE_BUSY can last the same time as synchronization of some internal MSDK task? In other words, it's a time of SyncOperation execution. And for SyncOperation we use MSDK_WAIT_INTERVAL.

dmitryermilov:
@aobolensk , please check what's wrong with CI

PR #2064, opened 2020-04-06
Fast_copy: add C implementation

egorovdanil:


_studio/shared/src/fast_copy.cpp
```diff
@@ -0,0 +1,65 @@
...
+mfxI32 CpuFeature_AVX2() {
+    return((__builtin_cpu_supports("avx2")));
+}
+
+void copyVideoToSys(const mfxU8* src, mfxU8* dst, int width)
+{
+    static const int m_SSE4_available = CpuFeature_SSE41();
+    static const int m_AVX2_available = CpuFeature_AVX2();
+
+    static const t_copyVideoToSys copyVideoToSys = FAFT_COPY_CPU_DISP_INIT_AVX2_SSE4_C(copyVideoToSys);
```
dvrogozh, line 42: please, rename variable to be different from the function name. here and below.

egorovdanil: done

PR #2068, opened 2020-04-10
[sample_decode] Fix wrong m_bVppIsUsed

egorovdanil:
m_bVppIsUsed should be defined after the InitMfxParams call

### DIFF
--- a/samples/sample_decode/src/pipeline_decode.cpp
+++ b/samples/sample_decode/src/pipeline_decode.cpp
@@ -391,8 +391,9 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
         m_bVppFullColorRange = pParams->bUseFullColorRange;
     }
 
-    m_bVppIsUsed = IsVppRequired(pParams);
-    if (m_bVppIsUsed)
+    bool bResolutionSpecified = pParams->Width || pParams->Height; // potentially VPP can be inserted
+
+    if (bResolutionSpecified)
         m_bDecOutSysmem = pParams->bUseHWLib ? false : true;
     else
         m_bDecOutSysmem = pParams->memType == SYSTEM_MEMORY;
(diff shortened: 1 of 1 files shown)


egorovdanil:
This bug caused by 83cbfa commit, where I moved the call IsVppRequired() before call MFXInitParams() that needed to define m_bDecOutSysmem for CreateHWDevice(). And I think is not correct, because m_bVppIsUsed most likely will always 'true'.
```
    m_bVppIsUsed = IsVppRequired(pParams);
    if (m_bVppIsUsed)
        m_bDecOutSysmem = pParams->bUseHWLib ? false : true;
    else
        m_bDecOutSysmem = pParams->memType == SYSTEM_MEMORY;
...
#if D3D_SURFACES_SUPPORT
    isDeviceRequired = m_memType != SYSTEM_MEMORY || !m_bDecOutSysmem;
...
    if (isDeviceRequired)
    {
        sts = CreateHWDevice();
```
@dmitryermilov, what do you think?

dmitryermilov:
In other words at CreateHWDevice() stage we don't have decoder mfxVideoParams yet so we can't say if VPP is really needed or not, right? And in order to get decoder's mfxVideoParams we have to call DecodeHeader. But SetHandle(device) must occur before DecodeHeader. 
:)

Ideal solution here is if user specifies -w -h cmd options, VPP has to be inserted. I'm not sure if we need to change this logic right now. 

Let me think ...

egorovdanil:
@dmitryermilov, done

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -389,8 +389,9 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
         m_bVppFullColorRange = pParams->bUseFullColorRange;
     }
 
-    m_bVppIsUsed = IsVppRequired(pParams);
-    if (m_bVppIsUsed)
+    bool bResolutionSpecified = pParams->Width || pParams->Height; // potentially VPP can be inserted
+
+    if (bResolutionSpecified)
         m_bDecOutSysmem = pParams->bUseHWLib ? false : true;
```
dmitryermilov, line 395: This is a bit different from what I proposed. If we can't rely on IsVppRequired here, I'm not sure we can set m_bDecOutSysmem here.  I don't say it's wrong, just not sure.

PR #2069, opened 2020-04-13
[jpegd][jpege] Check that SW fallback is supported before using it

Alexandr-Konovalov:
jpeg software fallback is supported starting SSE4.2.

### DIFF
--- a/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -28,6 +28,7 @@
 #include "mfx_common_decode_int.h"
 
 #ifdef MFX_ENABLE_SW_FALLBACK
+#include <ippcore.h> // for mfxInit()
 #include "mfx_mjpeg_task.h"
 #include "umc_mjpeg_mfx_decode.h"
 #include "mfx_thread_task.h"
@@ -2019,6 +2020,9 @@ VideoDECODEMJPEGBase_SW::VideoDECODEMJPEGBase_SW()
 
 mfxStatus VideoDECODEMJPEGBase_SW::Init(mfxVideoParam *decPar, mfxFrameAllocRequest *request, mfxFrameAllocResponse *response, mfxFrameAllocRequest *, bool isUseExternalFrames, VideoCORE *core)
 {
+    auto ippSt = mfxInit();
+    MFX_CHECK(ippSt == ippStsNoErr, MFX_ERR_UNSUPPORTED);
+
     UMC::Status umcSts = m_FrameAllocator->InitMfx(0, core, decPar, request, response, isUseExternalFrames, true);
     MFX_CHECK(umcSts == UMC::UMC_OK, MFX_ERR_MEMORY_ALLOC);
 
--- a/_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
+++ b/_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -23,6 +23,8 @@
 #if defined (MFX_ENABLE_MJPEG_VIDEO_ENCODE)
 
 #include <thread> // for thread::hardware_concurrency()
+
+#include <ippcore.h> // for mfxInit()
 
 #include "mfx_common.h"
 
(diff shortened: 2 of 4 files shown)


_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
```diff
@@ -1102,13 +1104,12 @@ mfxStatus MFXVideoENCODEMJPEG::Init(mfxVideoParam *par_in)
...
     mfxVideoParam* par = par_in;
 
-    if(m_isInitialized)
-        return MFX_ERR_UNDEFINED_BEHAVIOR;
+    MFX_CHECK(m_isInitialized, MFX_ERR_UNDEFINED_BEHAVIOR);
+    MFX_CHECK(par == NULL, MFX_ERR_NULL_PTR);
 
-    if(par == NULL)
-        return MFX_ERR_NULL_PTR;
+    MFX_CHECK(mfxInit() == ippStsNoErr, MFX_ERR_UNSUPPORTED);
```
dvrogozh, line 1110: Don't do that, please. MFX_CHECK macro might (actually will) expand in a way that it will use expressions few times. In this case it will call mfxInit() few times which is not what we want. When you use MFX_CHECK like macro defs, always use trivial expressions, not functions. Should be something like:
```
auto ippst = mfxInit();
MFX_CHECK(ippst == ippStsNoErr, MFX_ERR_UNSUPPORTED);
```

_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
```diff
@@ -1102,13 +1104,12 @@ mfxStatus MFXVideoENCODEMJPEG::Init(mfxVideoParam *par_in)
...
 
-    MFX_CHECK(CheckExtBuffers(par->ExtParam, par->NumExtParam)== MFX_ERR_NONE, MFX_ERR_INVALID_VIDEO_PARAM);
+    MFX_CHECK(CheckExtBuffers(par->ExtParam, par->NumExtParam) == MFX_ERR_NONE, MFX_ERR_INVALID_VIDEO_PARAM);
```
dvrogozh, line 1112: same as above. And that's a slipover from old times since we had this problem before.

contrib/ipp/src/ippinit.c
```diff
@@ -0,0 +1,28 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include "ippcore.h"
+
+IppStatus mfxInit(void)
+{
+    return __builtin_cpu_supports("sse4.1") > 0? ippStsNoErr : ippStsNotSupportedCpu;
```
dvrogozh, line 25: I thought we use sse4.2 actually

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -2019,6 +2020,8 @@ VideoDECODEMJPEGBase_SW::VideoDECODEMJPEGBase_SW()
 
 mfxStatus VideoDECODEMJPEGBase_SW::Init(mfxVideoParam *decPar, mfxFrameAllocRequest *request, mfxFrameAllocResponse *response, mfxFrameAllocRequest *, bool isUseExternalFrames, VideoCORE *core)
 {
+    MFX_CHECK(mfxInit() == ippStsNoErr, MFX_ERR_UNSUPPORTED);
```
dvrogozh, line 2023: ditto below.

_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
Alexandr-Konovalov: @dvrogozh , macro with multiple evaluation of some argument is time bomb. But in [the code](https://github.com/Intel-Media-SDK/MediaSDK/blob/c92b652bb0f9bb8504c54076a3a1db9fed3fa50e/_studio/shared/include/mfx_utils.h#L58) I see
`#define MFX_CHECK(EXPR, ERR)    { if (!(EXPR)) MFX_RETURN(ERR); }
`
So, it looks like EXPR is used only once.

It's matter of taste, and I agree that a code with potential side effects looks suspiciously in macro arguments.

_studio/mfx_lib/encode/mjpeg/src/mfx_mjpeg_encode.cpp
Alexandr-Konovalov: Done.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
Alexandr-Konovalov: Done.

contrib/ipp/src/ippinit.c
Alexandr-Konovalov: Done.

contrib/ipp/include/ippcore.h
```diff
@@ -80,6 +80,8 @@ IPPAPI( void*, mfxMalloc,  (int length) )
 IPPAPI( void, mfxFree, (void* ptr) )
 
 
+IppStatus mfxInit(void);
```
dvrogozh, line 83: Should that be declared IPPAPI()?

Alexandr-Konovalov: For current build mode, this is no-op. But yes, for uniformity let's add IPPFUN/IPPAPI.

PR #2070, opened 2020-04-13
[HEVCe][MPEG2e] Remove unused VAConfigAttribEncSliceStructure attribute

Alexandr-Konovalov:
For HEVCe, value of VAConfigAttribEncSliceStructure has not read.
For MPEG2e, one read but not used.


Alexandr-Konovalov:
> Why did you remove it?
> Who has requested that?
> There's an activity in driver and msdk teams now to fully support capabilities and align them with Windows

@saosipov , while working on a fix related to using of VAConfigAttribEncSliceStructure in AVC, I was asked if some issue exists for other usage of VAConfigAttribEncSliceStructure. So I found that HEVC and MPEG2 just read it value that is not used later. 

Is it better to add a comment that VAConfigAttribEncSliceStructure is actually not used by HEVC?

PR #2071, opened 2020-04-13
1) fix small buffer for 16k wpp stream;  2)  clear vector when re-allocate buffer

weizhu-intel:
There two changes are all cherry-picked from internal branch.
1) For 16k wpp stream,  it has up to 300 bytes' prevention, so increase 300 bytes here. Meanwhile, 300 bytes is the  biggest size we met till now.

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_va_supplier.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_va_supplier.cpp
@@ -174,7 +174,7 @@ UMC::Status VATaskSupplier::AllocateFrameData(H265DecoderFrame * pFrame, mfxSize
 H265Slice * VATaskSupplier::DecodeSliceHeader(UMC::MediaDataEx *nalUnit)
 {
     size_t dataSize = nalUnit->GetDataSize();
-    nalUnit->SetDataSize(std::min<size_t>(2560, dataSize));
+    nalUnit->SetDataSize(std::min<size_t>(SliceHeaderSize, dataSize));
 
     H265Slice * slice = TaskSupplier_H265::DecodeSliceHeader(nalUnit);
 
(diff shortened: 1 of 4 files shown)


dmitryermilov:
@weizhu-intel , please give a better name to this PR

_studio/shared/umc/codec/h265_dec/src/umc_h265_va_supplier.cpp
```diff
@@ -174,7 +174,7 @@ UMC::Status VATaskSupplier::AllocateFrameData(H265DecoderFrame * pFrame, mfxSize
 H265Slice * VATaskSupplier::DecodeSliceHeader(UMC::MediaDataEx *nalUnit)
 {
     size_t dataSize = nalUnit->GetDataSize();
-    nalUnit->SetDataSize(std::min<size_t>(2560, dataSize));
+    nalUnit->SetDataSize(std::min<size_t>(2860, dataSize));
```
dvrogozh, line 177: This is duplicated change with #2092. Please, left a single PR for review.

eponomar:
Currently I don't see any issues in CI tests after last force-push.

_studio/shared/umc/codec/h265_dec/src/umc_h265_va_supplier.cpp
dvrogozh, line 177: why this is 2860, not 2861 or 2863? Can you, please, create a macro def or constant with the self-speaking name explaining the value. If 2860 is composite value - explain how it was constructed.

weizhu-intel: Thanks Dmitry V. Pls review agian.

BTW,  will this comment be public to open source? Or just internal comment?
If public, is there any format/rule to comply will?

dvrogozh: This is open source project and this should be a public comment. Basically, not explaining what this magic value actually means and/or where it comes from undermines open source status of the project.

PR #2091, opened 2020-04-22
Fix HEVCd EntryPoint/disp_win_offset/PredWeightTable parsing issues

guowillyy:
Fix HEVCd EntryPoint/disp_win_offset/PredWeightTable parsing issues.

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_bitstream_headers.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_bitstream_headers.cpp
@@ -1340,7 +1340,15 @@ void H265HeadersBitstream::xParsePredWeightTable(const H265SeqParamSet *sps, H26
             wp[1].log2_weight_denom = sliceHdr->chroma_log2_weight_denom;
             wp[2].log2_weight_denom = sliceHdr->chroma_log2_weight_denom;
 
-            wp[0].present_flag = Get1Bit(); // luma_weight_lX_flag
+            if (sliceHdr->m_RefPOCList[eRefPicList][iRefIdx] == sliceHdr->m_poc)
+            {
+                wp[0].present_flag = 0;         // luma_weight_lX_flag
+            }
+            else
+            {
+                wp[0].present_flag = Get1Bit(); // luma_weight_lX_flag
+            }
+
             uTotalSignalledWeightFlags += wp[0].present_flag;
         }
 
@@ -1349,7 +1357,16 @@ void H265HeadersBitstream::xParsePredWeightTable(const H265SeqParamSet *sps, H26
             for (int32_t iRefIdx=0 ; iRefIdx < sliceHdr->m_numRefIdx[eRefPicList] ; iRefIdx++ )
             {
                 wp = sliceHdr->pred_weight_table[eRefPicList][iRefIdx];
-                wp[1].present_flag = wp[2].present_flag = Get1Bit(); // chroma_weight_lX_flag
+                
+                if (sliceHdr->m_RefPOCList[eRefPicList][iRefIdx] == sliceHdr->m_poc)
+                {
+                    wp[1].present_flag = wp[2].present_flag = 0;         // chroma_weight_lX_flag
+                }
+                else
+                {
+                    wp[1].present_flag = wp[2].present_flag = Get1Bit(); // chroma_weight_lX_flag
+                }
+
                 uTotalSignalledWeightFlags += 2*wp[1].present_flag;
             }
         }
@@ -1447,7 +1464,7 @@ UMC::Status H265HeadersBitstream::GetSliceHeaderPart1(H265SliceHeader * sliceHdr
 }
 
 // Parse remaining of slice header after GetSliceHeaderPart1
-void H265HeadersBitstream::decodeSlice(H265Slice *pSlice, const H265SeqParamSet *sps, const H265PicParamSet *pps)
+void H265HeadersBitstream::decodeSlice(H265Slice *pSlice, const H265SeqParamSet *sps, const H265PicParamSet *pps, PocDecoding *pocDecoding)
 {
     if (!pSlice || !pps || !sps)
         throw h265_exception(UMC::UMC_ERR_INVALID_STREAM);
@@ -1524,6 +1541,7 @@ void H265HeadersBitstream::decodeSlice(H265Slice *pSlice, const H265SeqParamSet 
         if (sliceHdr->IdrPicFlag)
         {
             sliceHdr->slice_pic_order_cnt_lsb = 0;
+            sliceHdr->m_poc = 0;
             ReferencePictureSet* rps = pSlice->getRPS();
             rps->num_negative_pics = 0;
             rps->num_positive_pics = 0;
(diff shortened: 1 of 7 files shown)


dvrogozh:
@guowillyy : please, create PRs with _reasonable_ names reflecting what actually has changed.

_studio/shared/umc/codec/h265_dec/src/umc_h265_bitstream_headers.cpp
```diff
@@ -1340,7 +1341,16 @@ void H265HeadersBitstream::xParsePredWeightTable(const H265SeqParamSet *sps, H26
...
+            if (sliceHdr->m_RefPOCList[eRefPicList][iRefIdx] == sliceHdr->m_poc)
+            {
+                iPresent = 0;
+            }
+            else
+            {
+                iPresent = Get1Bit();
+            }
+
+            wp[0].present_flag = (iPresent == 1); // luma_weight_lX_flag
```
mgonchar, line 1353: why do you need to compare with one, if you acquiring just one bit?

I believe following code should work
```
            if (sliceHdr->m_RefPOCList[eRefPicList][iRefIdx] == sliceHdr->m_poc)
            {
                wp[0].present_flag = 0; // luma_weight_lX_flag
            }
            else
            {
                wp[0].present_flag = Get1Bit(); // luma_weight_lX_flag
            }
```

guowillyy: Good suggestion! Code is updated.

dmitryermilov: @mgonchar , +1 from you?

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
```diff
@@ -317,6 +386,122 @@ int H265Slice::getNumRpsCurrTempList() const
...
+    m_SliceHeader.m_numPicLtCurr = NumPicLtCurr;
+
+    H265PicParamSet const* pps = GetPicParam();
+    VM_ASSERT(pps);
+
+    int32_t numPicTotalCurr = NumPicStCurr0 + NumPicStCurr1 + NumPicLtCurr;
+    if (pps->pps_curr_pic_ref_enabled_flag)
+        numPicTotalCurr++;
+
+    int32_t rpsPOCCurrList1[MAX_NUM_REF_PICS + 1];
```
dvrogozh, line 437: if you use in-the-middle declarations of variables, do that closer to real usage. This particular one needs to be moved below closer to lie 455 where you first start using it.

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
```diff
@@ -317,6 +386,122 @@ int H265Slice::getNumRpsCurrTempList() const
...
+    if (m_SliceHeader.slice_type == P_SLICE)
+    {
+        m_SliceHeader.m_numRefIdx[1] = 0;
+    }
+    else
+    {
+        for (rIdx = 0; rIdx <= m_SliceHeader.m_numRefIdx[1] - 1; rIdx++)
+        {
+            cIdx = refPicListModification.ref_pic_list_modification_flag_l1 ? refPicListModification.list_entry_l1[rIdx] : rIdx % numPicTotalCurr;
+            m_SliceHeader.m_RefPOCList[1][rIdx] = rpsPOCCurrList1[cIdx];
```
dvrogozh, line 500: I start to fix myself as static analyzer:). But it seems that you can use `rpsPOCCurrList1` uninitialized here. You actually fill out this array above if and only if `m_SliceHeader.slice_type == B_SLICE`, but you will enter this block of code both for I_SLICE and B_SLICE. For I_SLICE it will be unitialized. Should you update condition and avoid this code execution for the I_SLICE? or maybe you should add the whole function code protection to avoid execution for I_SLICE?

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
```diff
@@ -411,64 +608,58 @@ UMC::Status H265Slice::UpdateReferenceList(H265DBPList *pDecoderFrameList, H265D
...
-            {
-                /* Reporting about missed reference */
-                m_pCurrentFrame->SetErrorFlagged(UMC::ERROR_FRAME_REFERENCE_FRAME);
-                /* And because frame can not be decoded properly set flag "ERROR_FRAME_MAJOR" too*/
-                m_pCurrentFrame->SetErrorFlagged(UMC::ERROR_FRAME_MAJOR);
-            }
-            // pcRefPic->setCheckLTMSBPresent(false);
+        if (pFrm)
+            pFrm->SetisLongTermRef(false);
+        RefPicSetStCurr1[NumPicStCurr1] = pFrm;
```
dvrogozh, line 647: just to double check: if pFram==null, you still need to assign it, i.e. nul to reference list and increase reference number?

_studio/shared/umc/codec/h265_dec/src/umc_h265_bitstream_headers.cpp
mgonchar: @dmitryermilov looks fine to me, but I'm not familiar with this code

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
guowillyy: if (sliceHdr->slice_type != I_SLICE)
{
            pSlice->setRefPOCListSliceHeader();
}
Because the code above, I_Slice will not enter setRefPOCListSliceHeader function.

dvrogozh: Ok, I see. However, I don't think that current solution is the best one. I think you need to do one of 2 things:
1. Either move `if (sliceHdr->slice_type != I_SLICE)` check inside the function in a form like:
```
void H265Slice::setRefPOCListSliceHeader()
{
    if (m_SliceHeader.slice_type == I_SLICE)
         return;
    ...
}
```

2. or add assert to check that setRefPOCListSliceHeader really works with the correct expected slice type.

I strongly recommend to follow 1st solution...

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
dvrogozh: And I believe you better to initialize it with 0es to satisfy static analyzers in advance...

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
guowillyy: IMHO, do not think it is best to move if (sliceHdr->slice_type != I_SLICE) inside function, because in its parent function decodeSlice, many operations have dependency w/ slice type, all of them check slice type in decodeSlice, it is more readable. I can add one more check inside function, but has code redundancy concern.

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
guowillyy: This is legacy logic, not changed by this fix, I can check further, but should not block this fix's code review and check-in.

_studio/shared/umc/codec/h265_dec/src/umc_h265_slice_decoding.cpp
dvrogozh: And I hove code maintenance concern. Someone will need to add yet another call to this function in some other place and will forget to add a check for I_SLICE to prevent the function call. If you want to minimize impact - add assert. It is specifically designed for that purpose: notify that something impossible had happenned:
```
void H265Slice::setRefPOCListSliceHeader()
{
    assert(m_SliceHeader.slice_type != I_SLICE);
    ...
}

```

dvrogozh: @guowillyy : I see your change in the beginning of the function. It's ok with me. Thank you for the fix.

PR #2092, opened 2020-04-23
Wayne 4 23

weizhu-intel:
this two changes are cherry-picked from internal commit


dmitryermilov:
@weizhu-intel , could you give this PR a better name? "Wayne 4 23" doesn't make sense.

PR #2095, opened 2020-04-24
[jpege] Improve caps usage

Alexandr-Konovalov:
1) Cache jpege caps, not ask them at every EncodeFrameAsync.
2) Request some encoding limitation from driver, not use hardcoded values.


_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
```diff
@@ -195,47 +238,8 @@ mfxStatus VAAPIEncoder::QueryBitstreamBufferInfo(mfxFrameAllocRequest& request)
 mfxStatus VAAPIEncoder::QueryEncodeCaps(JpegEncCaps & caps)
 {
     MFX_CHECK_NULL_PTR1(m_core);
```
mgonchar, line 240: Is there any possibility to call this function when ```m_caps``` is not initialized yet? or ```m_core``` check guaranties that it won't happen?

PR #2096, opened 2020-04-24
Add filtering for VTune mfx_trace messages

aobolensk:


_studio/shared/mfx_trace/src/mfx_trace_vtune_ftrace.cpp
```diff
@@ -92,12 +92,11 @@ mfxTraceU32 MFXTraceFtrace_Init()
 
     const char* file_name = get_tracing_file("tracing/trace_marker");
     trace_handle = open(file_name , O_WRONLY);
-    if (trace_handle == -1){
-        // printf("@@@@ Can't open trace file!\n");
+    if (trace_handle == -1)
+    {
+        fprintf(stderr, "Can't open trace file!\n");
```
dvrogozh, line 97: don't we have this trace module on by default? This will flood messages to stderr... I am not sure I like that.

aobolensk: Removed this message.

PR #2097, opened 2020-04-24
Break ipp build into 2 parts

egorovdanil:
ippinit to be build without -msee4.2
other source files to be build for -msee4.2

Signed-off-by: egorovdanil <egorovdanil2905@gmail.com>

### DIFF
--- a/contrib/ipp/CMakeLists.txt
+++ b/contrib/ipp/CMakeLists.txt
@@ -37,15 +37,79 @@ endif()
 set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ipp_defs}" )
 set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ipp_defs}" )
 
+### ipp_sse4
 set( sources "" )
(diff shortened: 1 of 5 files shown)


contrib/ipp/CMakeLists.txt
```diff
@@ -37,15 +37,77 @@ endif()
 set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ipp_defs}" )
 set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ipp_defs}" )
 
+### ippinit
+set( sources "" )
```
dvrogozh, line 41: Can we, please, use the same approach as we had here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/CMakeLists.txt#L131

Please, create OBJECT library from all ipp sources except ippinit.c and link single library, called ipp at the very end.

egorovdanil: done

Issue #2047, opened 2020-03-26
Sigill on CPUs without sse4 extension like AMD


sndirsch:
Applications are crashing with SIGILL on CPUs without sse4 extension like AMD, which offer only up to "sse2". Library can't be build with "-msse2" either. See some more details on

https://bugzilla.suse.com/show_bug.cgi?id=1167346

dmitryermilov:
@sndirsch , what MSDK API level function call causes the crash ? mfxInit() or something else ?

sndirsch:
@dmitryermilov Honestly I don't know. I don't know more than you can find in the referenced bug above. Did you have a look?

dmitryermilov:
Yes, I looked but didn't find that info there. It's okay.

dvrogozh:
As correctly pointed in the suse bug we force sse4.2 here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/19d11bf1cfd4c7699d1d146d7c3e43c39ffd6038/builder/ConfTargets.cmake#L23

And we have at least the following places to take care of we will try to remove this restriction:
1. Fast copy which is a mix of SSE2 and SSE4.1: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/include/fast_copy.h

2. ASC has instruction set runtime dispatcher which will require adjustments to compile correctly + it has unprotected SSE4.2(?) ME path: https://github.com/Intel-Media-SDK/MediaSDK/tree/master/_studio/shared/asc

3. IPP which uses SSE4.2: https://github.com/Intel-Media-SDK/MediaSDK/tree/master/contrib/ipp

Fast Copy is a primary thing to take care of I think

dvrogozh:
We finished code refactoring and now we are ready to remove enforced `-m<instruction-set>` from msdk build system. #2099 should take care of that, it's under testing and review now.

sndirsch:
> We finished code refactoring and now we are ready to remove enforced `-m<instruction-set>` from msdk build system. #2099 should take care of that, it's under testing and review now.

Build doesn't work for me with that patch.

[   57s] /usr/lib64/gcc/x86_64-suse-linux/9/include/tmmintrin.h:136:1: error: inlining failed in call to always_inline '_mm_shuffle_epi8': target specific option mismatch
[   57s]   136 | _mm_shuffle_epi8 (__m128i __X, __m128i __Y)

dvrogozh:
[...]

PR #2099, opened 2020-04-27
build: do not enforce specific instruction set

dvrogozh:
Fixes: #2047

We should not enforce specific instruction set target (-m<inst-set>
option) since end-users might experience crashes with sigill at
least for the following reasons:
1. msdk lib might be probed at runtime even on non-intel CPUs if it
is included into some plugin based framework (like gstreamer)
2. unit tests included into build system might be run on the system
where build had happenned which might be non-intel

So, this change remove msdk global instruction set flags. We still
have few places in the library tight to the particular instruction
set. These are optimization specific algorithms such as ipp, asc
(scene change detection) and memory copy operations. In these cases
we use runtime checks to detect underlying instruction sets. If we
don't support particular one - we are supposed to gracefully fail.

### DIFF
--- a/builder/ConfTargets.cmake
+++ b/builder/ConfTargets.cmake
@@ -20,23 +20,7 @@
 
 message( STATUS "Global Configuration of Targets" )
 
-set( T_ARCH "sse4.2" )
-message( STATUS "Target Architecture to compile: ${T_ARCH}" )
-
 append("-std=c++11" CMAKE_CXX_FLAGS)
-
-# SW HEVC decoder & encoder require SSE4.2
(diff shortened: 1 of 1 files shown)


sndirsch:
Build doesn't work for me with that patch:

[   57s] /usr/lib64/gcc/x86_64-suse-linux/9/include/tmmintrin.h:136:1: error: inlining failed in call to always_inline '_mm_shuffle_epi8': target specific option mismatch
[   57s]   136 | _mm_shuffle_epi8 (__m128i __X, __m128i __Y)

PR #2103, opened 2020-04-30
[encoders] Mark DriverEncoder-s functions as overriden

Alexandr-Konovalov:
Add override to functions in descendants of Mfx*Encode::DriverEncoder.

Issue: MDP-61382


_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_vaapi.h
```diff
@@ -65,45 +65,39 @@ namespace MfxHwVP9Encode
...
-            VP9MfxVideoParam const & par);
+            VP9MfxVideoParam const & par) override;
 
         virtual
         mfxStatus Reset(
-            VP9MfxVideoParam const & par);
-
-        // empty  for Lin
-        virtual
-        mfxStatus Register(
```
mgonchar, line 80: maybe we should return error here?

Alexandr-Konovalov: @mgonchar This function (I mean, with that signature) is not present in base class and never called. Why should we keep it?

_studio/mfx_lib/encode_hw/mjpeg/include/mfx_mjpeg_encode_vaapi.h
```diff
@@ -59,48 +59,48 @@ namespace MfxHwMJpegEncode
...
 
     private:
-        VAAPIEncoder(VAAPIEncoder const &);              // no implementation
-        VAAPIEncoder & operator =(VAAPIEncoder const &); // no implementation
         mfxStatus DestroyBuffers();
 
         VideoCORE       * m_core;
         mfxU32            m_width;
         mfxU32            m_height;
-        JpegEncCaps       m_caps;
```
Alexandr-Konovalov, line 103: Right solution is in [PR 2095](https://github.com/Intel-Media-SDK/MediaSDK/pull/2095), here I just make clang happy.

_studio/mfx_lib/encode_hw/vp9/include/mfx_vp9_encode_hw_vaapi.h
mgonchar: ok

Issue #2104, opened 2020-04-30
asc: remove/hide instruction-set specific code


dvrogozh:
https://github.com/Intel-Media-SDK/MediaSDK/blob/99712bd1c86ea8201dcf533bfb79c694dd6a4e3c/_studio/shared/asc/include/asc_common_impl.h#L56

We need to check whether this code is actually used and if it is used need to hide it inside instuction-set specific library.

spossos:
This particular function is used on both AVX2 and SSE4.1 optimized code, I will work on it

dvrogozh:
I wish we had all intrinsic functions grouped in .cpp files rather than .h files. Since this completely eliminates a risk that the header file will be included in the generic code path (which is not built with required instruction set) and function accidentally used. If that's not possible, I wish we will refactor the source code placement in the folders making sure that headers with intrinsics are grouped in a dedicated folder, This would allow easier catching of problems.

dvrogozh:
I believe we fixed that in #2107.

PR #2105, opened 2020-04-30
msdk-20.1.x: do not enforce specific instruction-set building msdk

### LINKED ISSUE #2047: Sigill on CPUs without sse4 extension like AMD

dvrogozh:
This PR intends to fix #2047 in latest release branch.


dvrogozh:
I just spotted this issue: #2104. It eventually did not impact our CI which makes me think we have some intrinsics code in the headers which we actually don't use. Anyhow, I would like to investigate more on #2104 before deciding whether we need to apply #2105 permanently for the release branch.

@dmitryermilov, @onabiull : can you, please, add -1 to this PR to avoid accidental merge?

dmitryermilov:
@dvrogozh , is it ready?

dvrogozh:
@dmitryermilov : we were waiting for one more fix for #2104. Now it is ready for review.

PR #2107, opened 2020-05-01
[VPP][ASC] remove/hide asc instruction-set specific code

spossos:
Moved optimized functions located in asc_common_impl.h
to their respective optimization file. These functions
deal with incomplete lines for full AVX2 and SSE4.1
registers.

Related to: #2104
Test: vpp_sc_di


_studio/shared/asc/include/asc_common_impl.h
```diff
@@ -43,138 +43,7 @@ ASC_ALIGN_DECL(16) static const mfxU16 tab_killmask[8][8] = {
...
-    }
-    if (len & 1) {
-        xlo = _mm_insert_epi8(xlo, *pSrc, 0);
-    }
-    return _mm256_inserti128_si256(_mm256_castsi128_si256(xlo), xhi, 1);
-}
-#endif //defined(__AVX2__)
-
-static inline void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
+void calc_RACA_4x4_C(mfxU8 *pSrc, mfxI32 pitch, mfxI32 *RS, mfxI32 *CS) {
```
dvrogozh, line 46: This seems to be a duplicate: you have same function declared in asc_common_impl.cpp file.

If you will decide to left implementation in header, you will need to have it declared static inline....

_studio/shared/asc/include/asc_avx2_impl.h
```diff
@@ -22,6 +22,28 @@
 #define _ASC_AVX2_IMPL_H_
 
 #include "asc_common_impl.h"
+// Load 0..7 floats to YMM register from memory
+// NOTE: elements of YMM are permuted [ 4 2 - 1 ]
+static inline __m256 LoadPartialYmm(
```
dvrogozh, line 27: These are just declarations. I think 'static inline' will be just ignored. Do we really need them here?

_studio/shared/asc/include/asc_common_impl.h
spossos: Corrrect, I am removing the duplicate code.

_studio/shared/asc/include/asc_avx2_impl.h
spossos: We do not need them, I am cleaning that out.

PR #2114, opened 2020-05-07
[DISP]: Change the API version verification

Gluhoedov:
Change the API version verification condition in MFXJoinSession
function from the desired version of the library implementation
to loaded SDK implementation version.


dmitryermilov:
@Gluhoedov , please fix CI

PR #2116, opened 2020-05-08
[h264][sample_encode][smt] Add VCM mode

pgribov:
Added VCM mode support.
Added -vcm key to sample_encode and smt.


pgribov:
docs have been updated

PR #2118, opened 2020-05-13
HEVCe: rename Gen9 -> Base

eistarov:


dmitryermilov:
@eponomar , please check what's wrong with opencl build

PR #2119, opened 2020-05-13
[sample_encode] Fixed Shift setting

eshiryae:
Shift should be set only for P010/P210/Y210 formats.


dmitryermilov:
@eponomar , please check what's wrong with opencl build

eshiryae:
@dmitryermilov opencl build is finished successfully now

PR #2120, opened 2020-05-13
HEVCe: P016 recon format instead of NV12 for input P010

alexelizarov:


dmitryermilov:
@alexelizarov please rebase

PR #2124, opened 2020-05-18
[avce] Check VAConfigAttribQPBlockSize to get MBQP support status

Alexandr-Konovalov:
Before MBQP was treated as always supported.

Issue: VSMGWL-26248

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -1492,7 +1492,8 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
         VAConfigAttribEncSliceStructure,
         VAConfigAttribEncROI,
         VAConfigAttribEncSkipFrame,
-        VAConfigAttribCustomRoundingControl
+        VAConfigAttribCustomRoundingControl,
+        VAConfigAttribQPBlockSize
     };
 
     std::vector<VAConfigAttrib> attrs;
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1549,7 +1550,9 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     m_caps.ddi_caps.UserMaxFrameSizeSupport = 1;
     m_caps.ddi_caps.MBBRCSupport            = 1;
-    m_caps.ddi_caps.MbQpDataSupport         = 1;
+    // non-zero VAConfigAttribQPBlockSize means MBQP supported,
+    // VA_ATTRIB_NOT_SUPPORTED returned by legacy drivers, treat it as supported
```
dvrogozh, line 1554: > // VA_ATTRIB_NOT_SUPPORTED returned by legacy drivers, treat it as supported

What does this mean? I don't see  you handle this here...

Alexandr-Konovalov: `VA_ATTRIB_NOT_SUPPORTED` is non-zero, so this is supported automatically. Ok, let's do this explicit.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1549,7 +1550,10 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
 
     m_caps.ddi_caps.UserMaxFrameSizeSupport = 1;
     m_caps.ddi_caps.MBBRCSupport            = 1;
-    m_caps.ddi_caps.MbQpDataSupport         = 1;
+    // Zero VAConfigAttribQPBlockSize means MBQP is not supported.
+    // VA_ATTRIB_NOT_SUPPORTED is returned by legacy drivers, treat this as supported.
+    m_caps.ddi_caps.MbQpDataSupport         =
```
dvrogozh, line 1555: emm. So, driver explicitly returns "unsupported" and we treat this as supported. Does not mean any sense to me:). What are "legacy" drivers behaving in this way?

Alexandr-Konovalov: @dvrogozh This was added in [[Encode] Set VAConfigAttribQPBlockSize VA attribute to report MBQP AVC support](https://github.com/intel/media-driver/commit/f403839f357c89897b309475b1252adc5cc38fbd).

I treat `VA_ATTRIB_NOT_SUPPORTED`  as "the attribute is not supported", in this case we are treating it optimistically and hope that functionality is supported (that is true for AVC VME). If the attribute is supported, we are reading a value and treat zero as "the functionality is not supported".

Sure, this may be strange, but disabling AVC VME MBQP for "new" MediaSDK working with "old" driver may be even worse.

Should we just forget about "new MediaSDK + old driver" combination? The support is cheap.

PR #2129, opened 2020-05-20
[sample_decode] Fixed Shift param for 10bit VP9 and AV1 decoding (backmerge)

mpushki:


mpushki:
> You're adding back internationally removed code. See #2057

@dmitryermilov AV1 does not work without this changes. Can I remove only VP9 or need to find another place for fix?

dmitryermilov:
> > You're adding back internationally removed code. See #2057
> 
> @dmitryermilov AV1 does not work without this changes. 

It's too high-level. What exactly doesn't work? AV1 MediaSDK decoder? Why? Does it return some error? Why? Where? Do you have confidence that it's not AV1 issue?

mpushki:
> > > You're adding back intentionally removed code. See #2057
> > 
> > 
> > @dmitryermilov AV1 does not work without this changes.
> 
> It's too high-level. What exactly doesn't work? AV1 MediaSDK decoder? Why? Does it return some error? Why? Where? Do you have confidence that it's not AV1 issue?

I've investigated this problem again and found new solution, thanks that you noticed adding of removed code.

PR #2131, opened 2020-05-25
[hevce] remove usage of extddi buffer for LCUSize

vilichev:
LCUSize is removed from mfxExtCodingOptionDDI buffer since
it was moved to mfxExtHEVCParam (official API)


vilichev:
This changes affects HEVCe legacy code path only (no changes are required for current HEVCe path)

PR #2132, opened 2020-05-25
[sample_decode] Set m_bVppIsUsed before using

egorovdanil:
In the commits 113614a and 83cbfa6, I repeatedly moved the IsVppRequired() call, and I missed the fact that it's needed in InitMfxParams() to set m_bVppIsUsed.


samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -739,6 +738,11 @@ mfxStatus CDecodingPipeline::InitMfxParams(sInputParams *pParams)
             sts = m_pmfxDEC->DecodeHeader(&m_mfxBS, &m_mfxVideoParams);
         }
 
+        if(!sts)
```
dmitryermilov, line 741: if (MFX_ERR_NONE == sts)

egorovdanil: Done

PR #2133, opened 2020-05-26
vpp: Fix list of platforms which supports MCTF

AntonGrishin:


_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -2597,10 +2597,8 @@ mfxStatus VideoVPPHW::QueryCaps(VideoCORE* core, MfxHwVideoProcessing::mfxVppCap
 
 #ifdef MFX_ENABLE_MCTF
     eMFXHWType  hwType = core->GetHWType();
-    if (hwType != MFX_HW_BDW &&
-        hwType != MFX_HW_SCL &&
-        hwType != MFX_HW_KBL &&
-        hwType != MFX_HW_CNL)
+    if (hwType < MFX_HW_BDW ||
```
dvrogozh, line 2600: I strongly believe `hwType > MFX_HW_TGL_LP` condition is incorrect and actually dangerous. As of now we don't actually support any platform beyond TGL_LP and it is better to deny MCTF at the moment when incompatible platform will first appear.

@spossos : FYI

AntonGrishin: @dvrogozh, changed to mirror condition.

dmitryermilov:
@dvrogozh , was you concern addressed?

dvrogozh:
> @dvrogozh , was you concern addressed?

yes.

PR #2136, opened 2020-05-27
[h264e] VA may return multiple values in VAConfigAttribEncSliceStructure

Alexandr-Konovalov:
1) Previously combined values leads to creation of single slice.
2) enum is used instead of magic numbers.

Issue: MDP-61999

### DIFF
--- a/_studio/mfx_lib/shared/include/mfx_h264_enc_common_hw.h
+++ b/_studio/mfx_lib/shared/include/mfx_h264_enc_common_hw.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -942,6 +942,15 @@ namespace MfxHwH264Encode
 
     ENCODE_FRAME_SIZE_TOLERANCE ConvertLowDelayBRCMfx2Ddi(mfxU16 type, bool bTCBRC);
 
+    enum class SliceDividerType
+    {
+        ONESLICE            = 0, // Once slice for the whole frame
+        ROW2ROW             = 1, // Slices are power of 2 number of rows, all slices the same
+        ROWSLICE            = 2, // Slices are any number of rows, all slices the same
+        ARBITRARY_ROW_SLICE = 3, // Slices are any number of rows, slices can be different
+        ARBITRARY_MB_SLICE  = 4, // Slices are any number of MBs, slices can be different
+    };
+
     struct MfxMemId
     {
     public:
@@ -1169,9 +1178,9 @@ namespace MfxHwH264Encode
     };
 
 
-    struct SliceDividerBluRay : SliceDivider
+    struct SliceDividerArbitraryRowSlice : SliceDivider
     {
-        SliceDividerBluRay(
+        SliceDividerArbitraryRowSlice(
             mfxU32 numSlice,
             mfxU32 widthInMbs,
             mfxU32 heightInMbs);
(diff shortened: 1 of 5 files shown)


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -91,21 +91,19 @@ VAProfile ConvertProfileTypeMFX2VAAPI(mfxU32 type)
...
-        case VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS:
-            return 3;
-        case VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS:
-            return 4;
-        default:
-            return 0;
-    }
+    if (structure & VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS)
+        return SliceDividerArbitrarySliceSizeT;
+    else if (structure & VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS)
```
mgonchar, line 98: no need in else after return. Without multiple ```else``` code will be clearer

_studio/mfx_lib/shared/include/mfx_h264_enc_common_hw.h
```diff
@@ -942,6 +942,15 @@ namespace MfxHwH264Encode
 
     ENCODE_FRAME_SIZE_TOLERANCE ConvertLowDelayBRCMfx2Ddi(mfxU16 type, bool bTCBRC);
 
+    enum SliceDividerType
```
mgonchar, line 945: well, if you are introducing new enum, let's make it scoped enum to avoid implicit conversions:
```
enum class SliceDividerType
{
...
};
```

Alexandr-Konovalov: Done.

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
Alexandr-Konovalov: Done.

_studio/mfx_lib/shared/include/mfx_h264_enc_common_hw.h
```diff
@@ -942,6 +942,15 @@ namespace MfxHwH264Encode
 
     ENCODE_FRAME_SIZE_TOLERANCE ConvertLowDelayBRCMfx2Ddi(mfxU16 type, bool bTCBRC);
 
+    enum class SliceDividerType
+    {
+        OneSlice = 0,
```
dvrogozh, line 947: Imho, that's a weird to see such a mix of platforms, use cases and slice types! It blows my mind when we are supposed to use what? It depends on platform, on usage or on what? @mgonchar : probably not in this commit, but I wonder, can this somehow be improved?

dmitryermilov:
@dvrogozh , I always respect your review but here you're inconsistent. Previously for the same change you put -1: https://github.com/Intel-Media-SDK/MediaSDK/pull/1814/files . Please explain why you changed you mind.

_studio/mfx_lib/shared/include/mfx_h264_enc_common_hw.h
```diff
@@ -942,6 +942,15 @@ namespace MfxHwH264Encode
...
+        Snb = 1,
```
dmitryermilov, line 948: This enum doesn't make sense to me. No one, reading Hsw BluRay etc, will understand what it exactly means. I propose to do it this way: https://github.com/Intel-Media-SDK/MediaSDK/pull/1814/commits/91584f6455de9d008b3b02d5c354fc8fd1576871#diff-21db52ca9c332ae4833c901c5be556eaR995

Alexandr-Konovalov: Done.

PR #2137, opened 2020-05-27
Add Android build support for Intel MSDK 20.1.1

LukeZheZhu:


_studio/shared/asc/Android.mk
```diff
@@ -1,22 +1,88 @@
 LOCAL_PATH:= $(call my-dir)
 
+# =============================================================================
+
+include $(CLEAR_VARS)
+include $(MFX_HOME)/android/mfx_defs.mk
+
+LOCAL_SRC_FILES := $(addprefix src/, $(notdir $(wildcard $(LOCAL_PATH)/src/asc_avx2_impl.cpp)))
```
dvrogozh, line 8: here and below...

I don't see the reason to use wildcards expression here when you can just explicitly specify full path to the file. I think this will be simpler.

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_ext_brc.cpp
```diff
@@ -494,7 +494,13 @@ void ExtBRC::QueryTask(const FeatureBlocks& /*blocks*/, TPushQT Push)
             break;
         case MFX_BRC_PANIC_BIG_FRAME:
             task.bSkip = true;
+#ifndef ANDROID
```
dvrogozh, line 497: There is nothing specific to Android in this code. I think it's better to remove ANDROID define here and fix clang warning for all OSes. For example, something like that:
```
        case MFX_BRC_PANIC_BIG_FRAME:
            task.bSkip = true;
#if __clang__
            [[clang::fallthrough]];
#endif 
        case MFX_BRC_BIG_FRAME:
#if __clang__
            [[clang::fallthrough]];
#endif
```

@mgonchar : advice whether we want more generic approach. For example, we can have something like
```
#if __cplusplus > 201703L
#define MFX_FALLTHROUGH [[fallthrough]]
#elif __clang__ 
#define MFX_FALLTHROUGH [[clang::fallthrough]]
#else 
#define MFX_FALLTHROUGH
#endif 
```

mgonchar: @dvrogozh yes, your last suggestion looks fine. Let's put it to utility header (or separate header which we will remove after starting support of C++17)

LukeZheZhu:
> Did you provide PR for master branch already?

Hi @dvrogozh 
As too many changes in latest master branch code, I think this patch only fit for version 20.1.1

dvrogozh:
@eponomar : is that possible to enable gcc-10 build for this branch?

_studio/mfx_lib/encode_hw/hevc/agnostic/g9/hevcehw_g9_ext_brc.cpp
```diff
@@ -494,7 +494,13 @@ void ExtBRC::QueryTask(const FeatureBlocks& /*blocks*/, TPushQT Push)
             break;
         case MFX_BRC_PANIC_BIG_FRAME:
             task.bSkip = true;
+#if __clang__
```
dvrogozh, line 497: Can you, please, create MFX_FALLTHROUGH macro in this file: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/include/mfx_utils.h and use it here? Implementation should be something similar to:
```
#if __cplusplus > 201703L
#define MFX_FALLTHROUGH [[fallthrough]]
#elif __clang__ 
#define MFX_FALLTHROUGH [[clang::fallthrough]]
#else 
#define MFX_FALLTHROUGH
#endif 
```

LukeZheZhu: I am going to do this

eponomar:
> @eponomar : is that possible to enable gcc-10 build for this branch?

Yes, I will do it tomorrow

LukeZheZhu:
> Gcc-10 enabled for this and all new release branches.
> Please rebase or push your changes again to re-trigger CI checks with new infrastructure

I have re-push my pr to re-trigger CI checks. Thanks

PR #2145, opened 2020-06-03
mfxFrameAllocResponse: allow NumFrameActual be zero.

xuguangxin:
The current implementation needs all decode surfaces be registered at MFXInit time. 
In some use cases like "encoder level dynamic change", "video analytics buffers batch", the downstream will dynamically adjust buffer count. So the request may not so reasonable.  
Since libva and our driver do not request surface register when we create va context anymore, It's no reason to enforce it at media sdk level. 
These behavior changes will make middleware like gstremer and FFmpeg got simpler code.


xuguangxin:
the example allocator function be 

```
mfxStatus  FrameAllocator::_Alloc(mfxHDL pthis, mfxFrameAllocRequest *request, mfxFrameAllocResponse *response) {
    response->mids = NULL;
    response->NumFrameActual = 0;
    return MFX_ERR_NONE;
}
```

xuguangxin:
@xhaihao , @XinfengZhang. please help review it.
thanks

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -630,7 +630,7 @@ VAAPIVideoCORE::AllocFrames(
                 // Checking for unsupported mode - external allocator exist but Device handle doesn't set
                 MFX_CHECK(m_Display, MFX_ERR_UNSUPPORTED)
 
-                if (response->NumFrameActual < request->NumFrameMin)
+                if (response->NumFrameActual && (!response->mids || (response->NumFrameActual < request->NumFrameMin)))
```
wangyan-intel, line 633: This PR seems to be only for vaapi buffer? May it need for system memory buffer?

xuguangxin: yes, I only tested vaapi, but for system memory buffer. It may not need this. since it will copy the buffer.

xuguangxin:
@mgonchar any suggestion on this?
thanks

mgonchar:
@xuguangxin well, I don't see any sense in such allocator, which actually does nothing. Why can't you avoid allocation call? Could you point me to place in code, which force you to make allocation when you don't want it?

xuguangxin:
hi @mgonchar ,
A simple example is zero-copy multi-channels transcode for live streaming. Imaging a decoder output NV1 to 50 encoders, and we may add more encoders latter if users joined. Each encoder will cache some buffers, and we can't foresee how many buffers we need in MFXVideoDECODE_Init time

_studio/mfx_lib/decode/vp9/src/mfx_vp9_dec_decode_hw.cpp
```diff
@@ -416,7 +416,7 @@ mfxStatus VideoDECODEVP9_HW::Init(mfxVideoParam *par)
 #endif
             reallocFrames;
 
-    if (isUseExternalFrames && !reallocFrames)
+    if (isUseExternalFrames)
```
mgonchar, line 419: need to understand if such correction doesn't harm some use case

xuguangxin: yes, thanks. 
@onabiull do you know who made the original changes? why reallocFrames not need m_FrameAllocator->SetExternalFramesResponse(&m_response);? 

thanks

mgonchar: @xuguangxin I've checked git history and this code was introduced intentionally to handle Dynamic Resolution Change scenarios

xuguangxin: @mgonchar , thanks for checking. It's for vp9 p frame resolution changing. It may need time to found a decent change for this. Please focus on other changes. I can drop the vp9 related and fix it later.
thanks

xuguangxin:
@dmitry-gurulev any suggestion for this patch?
thanks

xuguangxin:
@dmitry-gurulev , thanks for the review. 
You mean like https://github.com/Intel-Media-SDK/MediaSDK/blob/95e1735206bfd91cb5c6ae809ebb85c87f0db599/api/include/mfxstructures.h#L330
I can add "mfxU16  SkipFrameAllocFree" if you agreed.
thanks

dmitry-gurulev:
@xuguangxin,

> You mean like 
> mfxU16  EnableReallocRequest; 

Yes, exactly.

> I can add "mfxU16  SkipFrameAllocFree" if you agreed.

IMO, that's not a perfect name, maybe 'EnableEmptyRequest\EnableZeroNumFrameActual\EnableNoAllocatedFrames'?

xuguangxin:
> IMO, that's not a perfect name, maybe 'EnableEmptyRequest\EnableZeroNumFrameActual\EnableNoAllocatedFrames'?
use those names means mediasdk still need to call FrameAllocator's allocate/free functions, but we expect the FrameAllocator to give zero allocated frame to us. right? 
Is it possible to skip allocate and free functions? It does not give the caller/mediasdk any information. The mfxFrameAllocRequest can be provided by MFXVideoDECODE_QueryIOSurf

xuguangxin:
@dmitry-gurulev @mgonchar 
fixed, please help review it again.
thanks

xuguangxin:
@dmitry-gurulev @mgonchar , any suggestion on the update?
thanks

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -741,6 +741,14 @@ VAAPIVideoCORE::CreateVA(
     if (!(request->Type & MFX_MEMTYPE_FROM_DECODE) ||
         !(request->Type & MFX_MEMTYPE_DXVA2_DECODER_TARGET))
         return MFX_ERR_NONE;
+    m_bEnableZeroNumFrameActual = (param->mfx.EnableZeroNumFrameActual == MFX_CODINGOPTION_ON);
+    if (m_bEnableZeroNumFrameActual) {
+	if (response->NumFrameActual || response->mids)
```
mgonchar, line 746:
```
MFX_CHECK(response->NumFrameActual && response->mids, MFX_ERR_INVALID_VIDEO_PARAM);
```

_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -741,6 +741,14 @@ VAAPIVideoCORE::CreateVA(
...
+           return MFX_ERR_INVALID_VIDEO_PARAM;
+    } else {
+      if (!response->NumFrameActual || !response->mids)
```
mgonchar, line 749:
```
MFX_CHECK(!response->NumFrameActual && !response->mids, MFX_ERR_INVALID_VIDEO_PARAM);
```

api/include/mfxstructures.h
```diff
@@ -328,7 +328,8 @@ typedef struct {
             mfxU16  SliceGroupsPresent;
             mfxU16  MaxDecFrameBuffering;
             mfxU16  EnableReallocRequest;
-            mfxU16  reserved2[7];
+            mfxU16  EnableZeroNumFrameActual;
```
mgonchar, line 331: I'm afraid API change is not so simple, you need to
1) get approval from @varistar 
2) add static assert to [this](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/src/mfx_static_assert_structs.cpp) file
3) add description to doc with example on when we need to use this control

xuguangxin: @mgonchar thanks for the review, fixed. 
I will do the changes after aligning the direction with @varistar . 

Hi @varistar 
What's your suggestion on this?

thanks

mgonchar:
@xuguangxin check copyrights in files you've modified. Some CI checks failed because of incorrect copyright

xuguangxin:
@mgonchar thanks for the reminder, fixed.

_studio/shared/src/mfx_umc_alloc_wrapper.cpp
```diff
@@ -384,7 +387,9 @@ void mfx_UMC_FrameAllocator::SetExternalFramesResponse(mfxFrameAllocResponse *re
 {
     m_externalFramesResponse = 0;
 
-    if (!response || !response->NumFrameActual)
+    if (!response)
+        return;
+    if (!response->NumFrameActual && !m_lazyRegistration)
```
dmitry-gurulev, line 392: May it makes sense to add check like
`if (response->NumFrameActual && m_lazyRegistration)`

PS I'd rather add `assert` (or even `throw std::logic_error`) before `return`

xuguangxin: done.

xuguangxin:
Hi @varistar , 
This patch is important. It will greatly simplify the FFmpeg/gstreamer mediasdk plugins code.
And it's will open the possibility of complex use cases.  like https://github.com/Intel-Media-SDK/MediaSDK/pull/2145#issuecomment-639195981. 
Any suggestions on this? 
Thanks

api/include/mfxstructures.h
varistar: Hi, several questions:
1. Why we cannot remove AllocFrames from Decoders' initialization at all instead of making fake AllocFrames with zero output? The complexity of changes or minimize Linux specific or what? The current solution looks like a workaround to me. 
2. Who will care about close source Win changes? This mode should be prohibited for DX9, but can be enabled in (or not) DX11?

xuguangxin: Hi @varistar 
thanks for the reply.

>Why we cannot remove AllocFrames from Decoders' initialization at all instead of making fake AllocFrames with zero output? The complexity of changes or minimize Linux specific or what? The current solution looks like a workaround to me.

Yes, I'd love to remove AllocFrames. But it may introduce back capability issues for old MediaSDK users. So I just provide this workaround.

> Who will care about close source Win changes? This mode should be prohibited for DX9, but can be enabled in (or not) DX11?

It's a design flaw of DX9, DX11 removed this constrain. 
Win7 is the last windows uses DX9. I hope we can use this mode for both windows and Linux after Win7 retired.

varistar: You are right, we cannot just remove AllocFrames, but we can do the following - 
if (!mfxVideoParam->lazy_allocation)
AllocFrames()
....
else
// no call of AllocFrames.

Right now anybody who wants to support this mode needs to create very special allocator which allocates zero surfaces that looks very cont-intuitive. 
Ideally, I'd like to see a solution where applications don't need to set external allocator and MediaSDK doesn't call AllocFrames/Free when it's not needed

xuguangxin: > You are right, we cannot just remove AllocFrames, but we can do the following -
> if (!mfxVideoParam->lazy_allocation)
> AllocFrames()
> ....
> else
> // no call of AllocFrames.
> 

Yes. we are fully aligned. 
This is exactly I suggested in https://github.com/Intel-Media-SDK/MediaSDK/pull/2145#issuecomment-644501955
I will send a patch for this.
Before the patch, could you share your preferred name? SkipFrameAllocFree or LazyAllocation in mfxInfoMFX?

> Ideally, I'd like to see a solution where applications don't need to set external allocator and MediaSDK doesn't call AllocFrames/Free when it's not needed

This may need more works in MediaSDK. Because MeidaSDK needs mfxFrameAllocator.getHDL to get platform related surface handle.

xuguangxin: github is broken last week. @varistar made some valuable suggestions through mail. let me attach here

replays from @varistar 

>  > This may need more works in MediaSDK. Because MeidaSDK needs mfxFrameAllocator.getHDL to get platform related surface handle.
>  
>  Agree, but for such mode we can specify that mfxMemID should contain native handle making getHDL call redundant.
>  
> 
>  
> >  Before the patch, could you share your preferred name? SkipFrameAllocFree or LazyAllocation in mfxInfoMFX?
>  
>  I'd like to call it like DeferredFrameAllocation or DelayedFrameAllocation
>

xuguangxin: >Agree, but for such mode we can specify that mfxMemID should contain native handle making getHDL call redundant.

Done. But allocator is a session level param. It can't controlled by mfxVideoParam. So I use     `MFXVideoCORE_SetFrameAllocator(m_session, NULL);`
to indicate a delayed frame allocator.
refer to https://github.com/Intel-Media-SDK/MediaSDK/pull/2145/files#diff-fae1222e835b7a0eea6af66261c4d4d0R904 fro implementation details

The patch tested for following test cases:
**decode**: vp9 resolution change at P frame.
**transcode with vpp**: h264->h264, h264->h265

@mgonchar @dmitry-gurulev @varistar , please help  review again.

thanks

_studio/shared/src/libmfx_core_vaapi.cpp
xuguangxin: the code is simplified, now we do not need this line anymore.
@wangyan-intel please help review it again.
thanks

_studio/shared/src/libmfx_core.cpp
```diff
@@ -874,14 +897,16 @@ mfxBaseWideFrameAllocator* CommonCORE::GetAllocatorByReq(mfxU16 type) const
...
     UMC::AutomaticUMCMutex guard(m_guard);
-    if (!allocator)
-        return MFX_ERR_NONE;
 
     if (!m_bSetExtFrameAlloc)
     {
-        m_FrameAllocator.frameAllocator = *allocator;
         m_bSetExtFrameAlloc = true;
-        m_session->m_coreInt.FrameAllocator = *allocator;
+        if (allocator) {
```
xuguangxin, line 904: most of `if (m_bDelayedFrameAllocation)` in this PR  can be removed, if we implement and set an internal DealyedFrameAllocator here.

varistar: AFAIU you use SetFrameAllocator(allocator = NULL) as a trigger to enable this mode. I'd like to have something explicit  - separate mode DelayedFrameAllocation as we talked before.

xuguangxin: Hi @varistar , 
the allocator or getHDL is a session/core level function. It will shared by decoder/encoder/vpp/. We can't use a decoder param to signal it. If we do not use SetFrameAllocator(allocator = NULL), we need add other function like  MFXVideoCORE_SetFrameAllocatMode(mfxSession sessio, MODE mode).

xuguangxin: Hi @varistar,
Are you prefer add new MFXVideoCORE function  or you have other insights on this?
thanks

api/include/mfxstructures.h
```diff
@@ -361,7 +361,7 @@ MFX_PACK_BEGIN_STRUCT_W_PTR()
 typedef struct {
     mfxU32  AllocId;
     mfxU32  reserved[2];
-    mfxU16  reserved3;
+    mfxU16  DelayedFrameAllocation;
```
mgonchar, line 364: please add static asserts for offset in \_studio\shared\src\mfx_static_assert_structs.cpp

mgonchar, line 364: Also we need changes in doc with description

xuguangxin: sure.

xuguangxin: will change it if @varistar agree to current changes.

xuguangxin:
@varistar updated for h264dec, vpp and h264enc. could you help review it again.
I will change other codecs if current code is ok to you.

thanks

PR #2146, opened 2020-06-03
dispatcher: add runtime check for API version

AntonGrishin:
This resolve problem, when we try to use lib w/ old API (<1.19) and
samples, which build w/ new API (>=1.31).


api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
```diff
@@ -494,14 +494,27 @@ mfxStatus MFXQueryAdaptersDecode(mfxBitstream* bitstream, mfxU32 codec_id, mfxAd
...
+            sts = MFXVideoCORE_QueryPlatform(dummy_session.operator mfxSession(), &info.Platform);
+
+            if (sts != MFX_ERR_NONE)
+            {
+                continue;
+            }
+        }
+        else
+        {
+            info.Platform.DeviceId = static_cast<mfxU16>(DeviceID); // Device id sets in QueryPlatform
```
mgonchar, line 515: A little bit confusing comment:
```
// for API versions greater than 1.19 Device id is set inside QueryPlatform call
```

eponomar:
Please re-push commit to trigger CI checks.

api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
AntonGrishin: done

PR #2150, opened 2020-06-08
Copy MaxDecFrameBuffering when querying

### LINKED ISSUE #2140: [APL] H264: SPS header values set to zero, once iHD driver is loaded
There is issue in MediaSDK library for H264 stream.
If H264 stream has SPS header information with "bitstream_restriction_flag = 1" and  "max_dec_frame_buffering = 4", as H264 Spec it should use max_dec_frame_buffering, rather than using default dpzSize for frames calculation. Which is not happening currently.

**Command Used:**
 ./sample_decode h264 -hw -async 1 -vaapi -i ~/test_2880x960.h264 -n 10

**Test Environment:**
OS: Ubuntu 18.04
Hardware: APL
Latest MediaSDK Commit: 2e3dbfa7b30a1efa27e021f34dd6791f6e9447bc 
Latest MediaDriver Commit: d7061b608ad4d55fccab00f8d80b852f20eab1d7 (20.1.pre)
Latest Libva Commit: ccf3b081ff3d2b0784cd66daf04bbc317277cf8b (2.8.0.pre1)

wangyan42164:
Fixes #2140.

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_mfx_supplier.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_mfx_supplier.cpp
@@ -1055,6 +1055,8 @@ mfxStatus MFX_Utility::Query(VideoCORE *core, mfxVideoParam *in, mfxVideoParam *
 
     if (in)
     {
+        out->mfx.MaxDecFrameBuffering = in->mfx.MaxDecFrameBuffering;
+
         if (in->mfx.CodecId == MFX_CODEC_AVC)
             out->mfx.CodecId = in->mfx.CodecId;
 
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_utils.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_utils.cpp
@@ -692,6 +692,8 @@ mfxStatus Query_H265(VideoCORE *core, mfxVideoParam *in, mfxVideoParam *out, eMF
 
     if (in)
     {
+        out->mfx.MaxDecFrameBuffering = in->mfx.MaxDecFrameBuffering;
+
         if (in->mfx.CodecId == MFX_CODEC_HEVC)
             out->mfx.CodecId = in->mfx.CodecId;
 


wangyan-intel:
> Let’s make it this way:
> 
> 1. Always ignore max_dec_frame_buffering  == 0. Treat zero values is default (like "undefined"). It's needed to avoid potential issues with streams which invalidly set max_dec_frame_buffering.
> 2. Accept the rest values during header parsing but check they’re compliant (not bigger than maxDPB calculated from profile/level, not less than max_num_ref_frames, etc). If not, ignore max_dec_frame_buffering from bitstream and calculate max_dec_frame_buffering from profile/level.

Sure. I will change it.

dmitryermilov:
@wangyan-intel , can you please also check HEVC case?

wangyan42164:
> @wangyan-intel , can you please also check HEVC case? 

Sure.

wangyan42164:
> Let’s make it this way:
> 
> 1. Always ignore max_dec_frame_buffering  == 0. Treat zero values is default (like "undefined"). It's needed to avoid potential issues with streams which invalidly set max_dec_frame_buffering.
> 2. Accept the rest values during header parsing but check they’re compliant (not bigger than maxDPB calculated from profile/level, not less than max_num_ref_frames, etc). If not, ignore max_dec_frame_buffering from bitstream and calculate max_dec_frame_buffering from profile/level.

@dmitryermilov I debugged it and double checked it again. Your comments may have been implemented by the following:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp#L2460-L2471

So for h264 decoding, It may be enough after apply my current change?

wangyan42164:
> @wangyan-intel , can you please also check HEVC case?

@dmitryermilov For HEVC, it is similar. We need add one copying.  The variable should be max_dec_pic_buffering.
SPS header decoding has covered it:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/h265_dec/src/umc_h265_task_supplier.cpp#L1069-L1078

I will add another commit for HEVC into this PR.
Thanks.

dmitryermilov:
Thanks!

PR #2157, opened 2020-06-09
Update coding guide

mgonchar:
Some minor improvements to our coding guide

### DIFF
--- a/doc/Coding_guidelines.md
+++ b/doc/Coding_guidelines.md
@@ -613,7 +613,7 @@ Feel free to improve this document. Bring more examples and rules. But please tr
 
     - Always protect shared resource access if simultaneous read/write is possible with appropriate synchronization mechanism.
 
-    - Avoid time of check and time of usage mismatch. In multithread application value of shared variable may be changed at any moment by other thread.
+    - Avoid time of check and time of usage mismatch. In multithread application value of shared variable may be changed at any moment by other thread. For ```std::atomic``` variables this issue can be avoided by putting modifications of variable inside Compare And Swap (CAS) loop with ```std::compare_exchange_weak``` (prefer this to ```std::compare_exchange_strong```).
 
     - Limit usage of synchronizations in multi thread applications. The more synchronization primitives application uses, the less benefit it gets from threading.
 
--- a/doc/Coding_guidelines.md
+++ b/doc/Coding_guidelines.md
@@ -191,9 +191,11 @@ Feel free to improve this document. Bring more examples and rules. But please tr
 
 10. Use unsigned types whenever it possible. Unsigned overflow is well defined behavior while signed is not.
 
+11. Avoid usage of local ```static``` variables because it might bring unexpectable overhead. Since C++11 initialization of local ```static``` variables must be thread safe, so compiler generates additional code to provide it.
+
 # Memory management
 
-11. Avoid manual memory management
+12. Avoid manual memory management
 
     - Functions ```memcpy```, ```memset```, ```sizeof``` are dangerous, use them only if you really need them.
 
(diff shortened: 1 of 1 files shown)


doc/Coding_guidelines.md
```diff
@@ -191,9 +191,11 @@ Feel free to improve this document. Bring more examples and rules. But please tr
 
 10. Use unsigned types whenever it possible. Unsigned overflow is well defined behavior while signed is not.
 
+11. Avoid usage of ```static``` variables because it might bring unexpectable overhead. Since C++11 initialization of ```static``` variables must be thread safe, so compiler generates additional code to provide it.
```
dmitry-gurulev, line 194: I'd like to note that it is so only for [Static local variables](https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables).

doc/Coding_guidelines.md
```diff
@@ -613,7 +613,7 @@ Feel free to improve this document. Bring more examples and rules. But please tr
 
     - Always protect shared resource access if simultaneous read/write is possible with appropriate synchronization mechanism.
 
-    - Avoid time of check and time of usage mismatch. In multithread application value of shared variable may be changed at any moment by other thread.
+    - Avoid time of check and time of usage mismatch. In multithread application value of shared variable may be changed at any moment by other thread. If atomics are used, need to put their's modification inside Compare And Swap (CAS) loops with ```std::compare_exchange_weak``` (prefer this to ```std::compare_exchange_strong```).
```
dmitry-gurulev, line 616: Didn't catch. Wheter we need looping or not depends on code logic. We need a loop when implementing something like `mutex.lock` but when we have simply `if (check) do this; else do that;` we don't. And recommendation about `std::compare_exchange_weak` over `std::compare_exchange_strong` is true only when we use loops.

doc/Coding_guidelines.md
mgonchar: I will reflect this in description

doc/Coding_guidelines.md
mgonchar: Well I actually mentioned that I'm talking about modification, not just usage. I meant something like atomic counter. Em, didn't get your last point, I can't imagine usage of any of the ```std::compare_exchange_X``` without loop

doc/Coding_guidelines.md
```diff
@@ -191,9 +191,11 @@ Feel free to improve this document. Bring more examples and rules. But please tr
 
 10. Use unsigned types whenever it possible. Unsigned overflow is well defined behavior while signed is not.
 
+11. Avoid usage of local ```static``` variables because it might bring unexpectable overhead. Since C++11 initialization of local ```static``` variables must be thread safe, so compiler generates additional code to provide it.
```
dmitry-gurulev, line 194: For me, this is too strong statement, because in some cases local `static` is better than namespace scope's or even class member ones just because of more local scope. I'd rather re-phrase it something like
"Note that usage of local ```static``` variables because it might bring unexpectable overhead"

mgonchar: @dmitry-gurulev I agree that it is not 100% rule and violation with a good reason is allowed. But as for me usage of local static variables shouldn't be a wide practice. Moreover, your concern is covered by statement
```
These tips shouldn't be considered as strict rules, they are just good piece of advice on how to implement your solution in modern C++. But please be sure to know recommended way before implementing something different. Violations of these rules must be committed consciously, otherwise better solution is most likely just to follow guidelines.
```

dmitry-gurulev: OK, agreed w/ that interpretatin

PR #2161, opened 2020-06-11
smt: Do not force HW adapter type, when requested SW implementation.

AntonGrishin:
Fixed wrong usage of HW library with "-sw" option.


AntonGrishin:
Something went wrong with driver public CI @eponomar

mgonchar:
@romka0017  could you please look

AntonGrishin:
> I think it is better to do it [inside](https://github.com/Intel-Media-SDK/MediaSDK/blob/c0599c65b48f1e8be07134f86d67197cae3a269b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp#L725) `ForceImplForSession`

done

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -721,32 +721,35 @@ mfxStatus Launcher::QueryAdapters()
 
 void Launcher::ForceImplForSession(mfxU32 idxSession)
 {
-    //change only 8 bit of the implementation. Don't touch type of frames
-    mfxIMPL impl = m_InputParamsArray[idxSession].libType & mfxI32(~0xFF);
-
-    mfxU32 idx = GetPreferredAdapterNum(m_Adapters, m_InputParamsArray[idxSession]);
-    switch (m_Adapters.Adapters[idx].Number)
+    if (m_InputParamsArray[idxSession].libType != MFX_IMPL_SOFTWARE)
```
mgonchar, line 724: Well, I'd prefer other condition to simplify code and reduce delta:
```
if (m_InputParamsArray[idxSession].libType == MFX_IMPL_SOFTWARE)
    return;
```

AntonGrishin: Agreed. Done.

PR #2165, opened 2020-06-16
[disp] Add dispatcher project for vs8

npletnev:


dmitryermilov:
> imho, it is more reasonable to switch to dynamic dispatcher and distribute it along with runtime library. In this way there will be no need to provide multiple solutions for each studio version and limit ourselves in C/C++ standard usage tight to particular studio version.
> 
> Secondly, Microsoft ended support for VS 2005 ended in 4/12/2016 (https://support.microsoft.com/en-us/lifecycle/search?alpha=Visual%20Studio%202005) and VS 2008 in 4/10/2018 (https://support.microsoft.com/en-us/lifecycle/search?alpha=Visual%20Studio%202008). I can hardly see how you can motivate merging in this change in the mediasdk upstream project. If we will do that, who will be responsible for VS security updates? This is simply no go from my perspective.

@Dmi3shin , @mvmaximov , @npletnev , please reply to Dmitry's message.

npletnev:
@dvrogozh, we have always had binaries of dispatcher, built with VS2008, in MSDK Windows releases. I'm adding this project to include binaries built with it in the upcoming release to be aligned with our previous releases, because we have customers demand for that.

dvrogozh:
@npletnev : this is explanation, thank you for it. However, to me this is not enough justification. We need to consider life cycles of the projects we depend on and encourage customers to move forward and provide better structured releases on our side which allow to step out from the customer limitations to the outdated versions of some ingredients.

PR #2171, opened 2020-06-17
[hevce] Fix RefPicList control

lakulako:
Fixed resulting RPL size computation.

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
@@ -1350,8 +1350,8 @@ public:
             ++pRplEnd[lx];
         });
 
-        l0 = std::min<mfxU8>((mfxU8)maxL0, mfxU8(tmpRPL[0] - pRplEnd[0]));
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
```diff
@@ -1350,8 +1350,8 @@ class GetDefault
             ++pRplEnd[lx];
         });
 
-        l0 = std::min<mfxU8>((mfxU8)maxL0, mfxU8(tmpRPL[0] - pRplEnd[0]));
-        l1 = std::min<mfxU8>((mfxU8)maxL1, mfxU8(tmpRPL[1] - pRplEnd[1]));
+        l0 = std::min<mfxU8>((mfxU8)maxL0, mfxU8(pRplEnd[0] - tmpRPL[0]));
```
mgonchar, line 1353: why not just ```l0``` and ```l1```?
```
l0 = std::min(mfxU8(maxL0), l0);
```

lakulako: pRplEnd[0] - tmpRPL[0] != l0. tmpRPL can also include frames not listed in preferred list.
And it is only a bug fix that is easy to understand, btw.

PR #2172, opened 2020-06-17
[tests] Fixed unitialized values

filippenkos:
Added initialization of statuses to prevent uninitialized use

### DIFF
--- a/tests/unit/suites/tracer/linux/mfx_tracer_test_cases_libs.cpp
+++ b/tests/unit/suites/tracer/linux/mfx_tracer_test_cases_libs.cpp
@@ -41,7 +41,7 @@ auto func_case_list = ::testing::Values(
 TEST_P(TracerLibsTestParametrizedFunc, ShouldSucceedForSeeminglyGoodMockLibrary)
 {
     MockCallObj& mock = *g_call_obj_ptr;
-    mfxStatus sts;
+    mfxStatus sts = MFX_ERR_UNKNOWN;
 
     mfxI32 func = GetParam();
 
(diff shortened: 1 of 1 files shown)


tests/unit/suites/tracer/linux/mfx_tracer_test_cases_libs.cpp
```diff
@@ -41,7 +41,7 @@ auto func_case_list = ::testing::Values(
 TEST_P(TracerLibsTestParametrizedFunc, ShouldSucceedForSeeminglyGoodMockLibrary)
 {
     MockCallObj& mock = *g_call_obj_ptr;
-    mfxStatus sts;
+    mfxStatus sts = MFX_ERR_NONE;
```
mgonchar, line 44: so the issue would be if ```func``` is not ```MFX_INIT``` and not ```MFX_INIT_EX```. I think in this case we should still fail with error, so for this particular case we should use different initialization status, not ```MFX_ERR_NONE```

filippenkos: In this test `func` have only `MFX_INIT` and `MFX_INIT_EX` values
https://github.com/Intel-Media-SDK/MediaSDK/blob/40ab977d9b50865f2248fd3b628e07738c9633a4/tests/unit/suites/tracer/linux/mfx_tracer_test_cases_libs.cpp#L37-L39

mgonchar: then you don't need to fix it. Personally I believe that better to fix it, because calling conditions may change at some point.

Static analysis complains on this code because it sees possible control path where ```sts``` is not set. If you decide to fix this "issue" I suggest to assume that any value can come to this function at ```func```

mgonchar: Thanks

Issue #2160, opened 2020-06-10
bump required libva version to 2.8.0 or protect code with version check


dvrogozh:
https://github.com/Intel-Media-SDK/MediaSDK/blob/81470616597c06deee3e6e1b85d6bbf5e20e0e75/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp#L101

As pointed out here:
https://github.com/Intel-Media-SDK/MediaSDK/commit/fb631a16cf3ecb1f2b900f841934eb5601b2cd79#commitcomment-39813152
VA_ENC_SLICE_STRUCTURE_EQUAL_MULTI_ROWS first appeared in libva-2.8.0. We either need to protect the code with VAAPI version check or bump minimal required libva version here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/81470616597c06deee3e6e1b85d6bbf5e20e0e75/builder/FindPackages.cmake#L484

I vote to bump the version.

PR #2175, opened 2020-06-18
Enabled the GPU context priority setting.

hongbolv:
This PR depends on VAAPI PR https://github.com/intel/libva/pull/416.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
@@ -25,7 +25,7 @@
 
 #include <va/va.h>
 #include <va/va_enc_hevc.h>
-
+#include "mfx_session.h"
 #include "libmfx_core_vaapi.h"
 #include "mfx_common_int.h"
 #include "mfx_h265_encode_vaapi.h"
--- a/_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
@@ -18,6 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+#include "mfx_session.h"
 #include "mfx_common.h"
 #include "hevcehw_base_va_lin.h"
 
@@ -170,16 +171,18 @@ void DDI_VA::InitAlloc(const FeatureBlocks& /*blocks*/, TPushIA Push)
     Push(BLK_CreateService
         , [this](StorageRW& strg, StorageRW& local) -> mfxStatus
     {
-        const auto&             par     = Glob::VideoParam::Get(strg);
+        auto& core = Glob::VideoCore::Get(strg);
+        const auto& par = Glob::VideoParam::Get(strg);
         const mfxExtHEVCParam&  hevcPar = ExtBuffer::Get(par);
         mfxStatus sts;
 
         m_callVa = Glob::DDI_Execute::Get(strg);
 
-        std::vector<VAConfigAttrib> attrib(2);
+        std::vector<VAConfigAttrib> attrib(3);
 
         attrib[0].type = VAConfigAttribRTFormat;
         attrib[1].type = VAConfigAttribRateControl;
+        attrib[2].type = VAConfigAttribContextPriority;
 
         sts = MfxEncodeHW::DeviceVAAPI::QueryCaps(attrib.data(), attrib.size() * sizeof(VAConfigAttrib));
         MFX_CHECK_STS(sts);
@@ -191,6 +194,18 @@ void DDI_VA::InitAlloc(const FeatureBlocks& /*blocks*/, TPushIA Push)
         MFX_CHECK((attrib[1].value & vaRCType), MFX_ERR_DEVICE_FAILED);
 
         attrib[1].value = vaRCType;
+        uint32_t maxContextPriority = attrib[2].value;
+        mfxPriority priority = core.GetSession()->m_priority;
+        if (priority == MFX_PRIORITY_LOW)
+        {
+            attrib[2].value = 0;
+        } else if (priority == MFX_PRIORITY_HIGH)
+        {
+            attrib[2].value = maxContextPriority;
+        } else
+        {
+            attrib[2].value = maxContextPriority/2;
+        }
 
         sts = MfxEncodeHW::DeviceVAAPI::Init(
             hevcPar.PicWidthInLumaSamples
--- a/_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
@@ -22,6 +22,7 @@
 
 #if defined (MFX_ENABLE_MJPEG_VIDEO_ENCODE) && defined (MFX_VA_LINUX)
 
+#include "mfx_session.h"
 #include "mfx_mjpeg_encode_hw_utils.h"
 #include "libmfx_core_factory.h"
 #include "libmfx_core_interface.h"
@@ -187,25 +188,40 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(mfxVideoParam const & par)
     }
 
     // Configuration
-    VAConfigAttrib attrib;
+    std::vector<VAConfigAttrib> attrib(2);
 
-    attrib.type = VAConfigAttribRTFormat;
+    attrib[0].type = VAConfigAttribRTFormat;
+    attrib[1].type = VAConfigAttribContextPriority;
 
     vaSts = vaGetConfigAttributes(m_vaDisplay,
                           VAProfileJPEGBaseline,
                           VAEntrypointEncPicture,
-                          &attrib, 1);
+                          attrib.data(),
+                          attrib.size());
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
-    if (!(attrib.value & VA_RT_FORMAT_YUV420) || !(attrib.value & VA_RT_FORMAT_YUV422) ) //to be do
+    if (!(attrib[0].value & VA_RT_FORMAT_YUV420) || !(attrib[0].value & VA_RT_FORMAT_YUV422) ) //to be do
         return MFX_ERR_DEVICE_FAILED;
+
+    uint32_t maxContextPriority = attrib[1].value;
+    mfxPriority priority = m_core->GetSession()->m_priority;
+    if (priority == MFX_PRIORITY_LOW)
+    {
+        attrib[1].value = 0;
+    } else if (priority == MFX_PRIORITY_HIGH)
+    {
+        attrib[1].value = maxContextPriority;
+    } else
+    {
+        attrib[1].value = maxContextPriority/2;
+    }
 
     vaSts = vaCreateConfig(
         m_vaDisplay,
         VAProfileJPEGBaseline,
         VAEntrypointEncPicture,
-        NULL,
-        0,
+        attrib.data(),
+        attrib.size(),
         &m_vaConfig);
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
--- a/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
@@ -19,6 +19,7 @@
 // SOFTWARE.
 
 
+#include "mfx_session.h"
 #include "mfx_vp9_encode_hw_utils.h"
 #include "mfx_vp9_encode_hw_par.h"
 #include "mfx_vp9_encode_hw_vaapi.h"
@@ -815,14 +816,16 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(VP9MfxVideoParam const & par)
     }
 
     // Configuration
-    VAConfigAttrib attrib[2];
+    std::vector<VAConfigAttrib> attrib(3);
 
     attrib[0].type = VAConfigAttribRTFormat;
     attrib[1].type = VAConfigAttribRateControl;
+    attrib[2].type = VAConfigAttribContextPriority;
     vaSts = vaGetConfigAttributes(m_vaDisplay,
                           va_profile,
                           VAEntrypointEncSliceLP,
-                          &attrib[0], 2);
+                          attrib.data(),
+                          attrib.size());
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
     uint32_t vaRTFormat = ConvertRTFormatMFX2VAAPI(par.mfx.FrameInfo.ChromaFormat);
@@ -836,13 +839,25 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(VP9MfxVideoParam const & par)
 
     attrib[0].value = vaRTFormat;
     attrib[1].value = vaRCType;
+    uint32_t maxContextPriority = attrib[2].value;
+    mfxPriority priority = m_pmfxCore->GetSession()->m_priority;
+    if (priority == MFX_PRIORITY_LOW)
+    {
+        attrib[2].value = 0;
+    } else if (priority == MFX_PRIORITY_HIGH)
+    {
+        attrib[2].value = maxContextPriority;
+    } else
+    {
+        attrib[2].value = maxContextPriority/2;
+    }
 
     vaSts = vaCreateConfig(
         m_vaDisplay,
         va_profile,
         VAEntrypointEncSliceLP,
-        attrib,
-        2,
+        attrib.data(),
+        attrib.size(),
         &m_vaConfig);
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -25,6 +25,7 @@
 #include <va/va.h>
 #include <va/va_enc_h264.h>
 #include "mfxfei.h"
+#include "mfx_session.h"
 #include "libmfx_core_vaapi.h"
 #include "mfx_common_int.h"
 #include "mfx_h264_encode_vaapi.h"
@@ -1701,18 +1702,19 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
 #endif
 
     // Configuration
-    VAConfigAttrib attrib[3];
+    std::vector<VAConfigAttrib> attrib(4);
     mfxI32 numAttrib = 0;
     mfxU32 flag = VA_PROGRESSIVE;
 
     attrib[0].type = VAConfigAttribRTFormat;
     attrib[1].type = VAConfigAttribRateControl;
-    numAttrib += 2;
+    attrib[2].type = VAConfigAttribContextPriority;
+    numAttrib += 3;
 
 #if defined(MFX_ENABLE_H264_VIDEO_FEI_ENCPAK) || defined(MFX_ENABLE_H264_VIDEO_FEI_PREENC)
     if ( m_isENCPAK )
     {
-        attrib[2].type = VAConfigAttribFEIFunctionType;
+        attrib[numAttrib].type = VAConfigAttribFEIFunctionType;
         numAttrib++;
     }
 #endif
@@ -1720,7 +1722,7 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
     vaSts = vaGetConfigAttributes(m_vaDisplay,
                           ConvertProfileTypeMFX2VAAPI(par.mfx.CodecProfile),
                           entryPoint,
-                          &attrib[0], numAttrib);
+                          attrib.data(), numAttrib);
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
     if ((attrib[0].value & VA_RT_FORMAT_YUV420) == 0)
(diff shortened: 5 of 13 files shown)


_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
```diff
@@ -191,6 +195,17 @@ void DDI_VA::InitAlloc(const FeatureBlocks& /*blocks*/, TPushIA Push)
         MFX_CHECK((attrib[1].value & vaRCType), MFX_ERR_DEVICE_FAILED);
 
         attrib[1].value = vaRCType;
+        priority = core.GetSession()->m_priority;
+        if (priority == MFX_PRIORITY_LOW)
+        {
+            attrib[2].value = -512;
```
dmitryermilov, line 201: -512/512 should go via dymanic caps from UMD or some harcoded value from va.h

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
```diff
@@ -187,25 +188,39 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(mfxVideoParam const & par)
...
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
-    if (!(attrib.value & VA_RT_FORMAT_YUV420) || !(attrib.value & VA_RT_FORMAT_YUV422) ) //to be do
+    if (!(attrib[0].value & VA_RT_FORMAT_YUV420) || !(attrib[0].value & VA_RT_FORMAT_YUV422) ) //to be do
         return MFX_ERR_DEVICE_FAILED;
 
+    priority = m_core->GetSession()->m_priority;
+    if (priority == MFX_PRIORITY_LOW)
+    {
+        attrib[1].value = -512;
```
dmitryermilov, line 209: same

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -836,13 +839,24 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(VP9MfxVideoParam const & par)
 
     attrib[0].value = vaRTFormat;
     attrib[1].value = vaRCType;
+    priority = m_pmfxCore->GetSession()->m_priority;
+    if (priority == MFX_PRIORITY_LOW)
+    {
+        attrib[2].value = -512;
```
dmitryermilov, line 845: same

hongbolv:
> This change contradicts manual:
> _With joined sessions, the application can set the priority of session operations through the MFXSetPriority function. A lower priority session receives less CPU cycles. Session priority does not affect hardware accelerated processing._

Synced with @Aristarkhov, Vasily, the MFXSetPriority will be leverage to set the GPU priority. Then the document should be updated.

hongbolv:
> This change contradicts manual:
> _With joined sessions, the application can set the priority of session operations through the MFXSetPriority function. A lower priority session receives less CPU cycles. **Session priority does not affect hardware accelerated processing**._
> ...
> 
> MFX_PRIORITY_LOW	Low priority: the session operation halts when high priority tasks are executing and more than 75% of the CPU is being used for normal priority tasks.
> MFX_PRIORITY_NORMAL	Normal priority: the session operation is halted if there are high priority tasks.
> MFX_PRIORITY_HIGH	High priority: the session operation blocks other lower priority session operations.

The MFXSetPriority functionality will be changed for GPU priority setting. Manual should be updated.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1131,6 +1132,7 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
         return MFX_ERR_DEVICE_FAILED;
 
     attrib[1].value = vaRCType;
+    attrib[2].value = m_core->GetSession()->m_priority;
```
varistar, line 1135: Why we don't make mapping here?

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
varistar: Yep, no magic values

hongbolv: Depends on the VAAPI change (https://github.com/intel/libva/pull/416) final decision now.  Once the VAAPI change is merged, it will be updated then.

dmitryermilov:
> Synced with @Aristarkhov, Vasily, the MFXSetPriority will be leveraged to set the GPU priority. Then the document should be updated.

1. I don't understand why you're appealing to Vasily. Documentation shall be updated together with code. 
2. It's **non backward compatible behavior change in SDK**. It's not good even with doc update.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
hongbolv: Pending on the VAAPI definition. When VAAPI definition is finalized, it will be updated.

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
```diff
@@ -191,6 +195,17 @@ void DDI_VA::InitAlloc(const FeatureBlocks& /*blocks*/, TPushIA Push)
...
+        } else if (priority == MFX_PRIORITY_HIGH)
+        {
+            attrib[2].value = 512;
+        } else
```
varistar, line 205: I'd prefer to have explicit else if (MFX_PRIORITY_NORMAL  == priority) to avoid mysterious behavior in the future if we add new priority policy

hongbolv: Agree. Will update it when VA interface change is done.

hongbolv:
> > Synced with @Aristarkhov, Vasily, the MFXSetPriority will be leveraged to set the GPU priority. Then the document should be updated.
> 
> 1. I don't understand why you're appealing to Vasily. Documentation shall be updated together with code.

   It is typo:(. I means Aristarkhov, Vasily.
> 2. It's **non backward compatible behavior change in SDK**. It's not good even with doc update.
   Do you have a idea to cover the case? Implement the new interface such as MFXSetGpuPriority?

varistar:
Formally, it's not backward compatible from behavior point of view, but it's OK for me - because previously we just didn't have control to change HW sessions priority. 
So, for me, it looks like fixing of our issue/limitation. Need to update the doc

dmitryermilov:
> 1. Do you have a idea to cover the case? Implement the new interface such as MFXSetGpuPriority?

Currently:
```
typedef enum
{
    MFX_PRIORITY_LOW = 0,
    MFX_PRIORITY_NORMAL = 1,
    MFX_PRIORITY_HIGH = 2

} mfxPriority;
```
We can add more values which would indicate GPU priorities

varistar:
I don't want to introduce more abstractions for the same needs. We are working on oneVPL - what would be behavior of SetGPUPriority on CPU version? Another vendor? We have a limitation in our MediaSDK thay setting of higher priority doesn't lead to anything. Now, we are removing it. 
Let me provide analogy - imagine we don't have interlace support in the decoder and after we add it we introduce new codec id to support backward compatibility instead just working as execpected

varistar:
And one more consideration - Linux/Windows are not Real-time OS and MSDK is working in user-mode, so all our descriptions in the doc are not super accurate. Any system services may brake it any time. So, from this perspective, I'm fine with current approach

dmitryermilov:
>I don't want to introduce more abstractions for the same needs

For me this is why we don't understand each other here. GPU and CPU priority are different things. And we're trying to mix it under the same interface. What if a user doesn't want to touch CPU and only influence GPU?

>  We are working on oneVPL - what would be behavior of SetGPUPriority on CPU version?

It will return an error. 

> Another vendor?

Error

varistar:
"For me this is why we don't understand each other here. GPU and CPU priority are different things. And we're trying to mix it under the same interface. What if a user doesn't want to touch CPU and only influence GPU?"
I don't care about CPU or GPU or FF or anythings else - I care about how my media tasks are executed - I even don't want to know where they are exectuted. What about Hybrid encoders like AV-1 - how you will explain behavior there? No need to make things more complex then they are.
We have abstract interface which should keep all implementation-specific inside. If customers want to work with GPU context - they can use VA

dmitryermilov:
One more thing, I assume we need a doc (wiki page or internal doc) how gpu priorities work in i915. At least rough description of i915 logic. 
Does it work on both i915 scheduler level and GuC scheduler level?

varistar:
> One more thing, I assume we need a doc (wiki page or internal doc) how gpu priorities work in i915. At least rough description of i915 logic.
> Does it work on both i915 scheduler level and GuC scheduler level?
Yep, such explanation is required

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
```diff
@@ -187,25 +188,39 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(mfxVideoParam const & par)
...
+        attrib[1].value = 0;
+    }
+
     vaSts = vaCreateConfig(
         m_vaDisplay,
         VAProfileJPEGBaseline,
         VAEntrypointEncPicture,
-        NULL,
-        0,
+        &attrib[0],
```
mgonchar, line 222: I don't like such hardcoding. Could you use ```vector``` instead? then we will have much clear code:

```
vaSts = vaCreateConfig(
        m_vaDisplay,
        VAProfileJPEGBaseline,
        VAEntrypointEncPicture,
        attrib.data(),
        attrib.size(),
        &m_vaConfig);
```

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
```diff
@@ -187,25 +188,39 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(mfxVideoParam const & par)
...
 
-    attrib.type = VAConfigAttribRTFormat;
+    attrib[0].type = VAConfigAttribRTFormat;
+    attrib[1].type = VAConfigAttribContextPriority;
 
     vaSts = vaGetConfigAttributes(m_vaDisplay,
                           VAProfileJPEGBaseline,
                           VAEntrypointEncPicture,
-                          &attrib, 1);
+                          &attrib[0], 2);
```
mgonchar, line 200: see my comment about ```vector``` below

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
```diff
@@ -170,16 +171,19 @@ void DDI_VA::InitAlloc(const FeatureBlocks& /*blocks*/, TPushIA Push)
...
+        auto& core = Glob::VideoCore::Get(strg);
+        const auto& par = Glob::VideoParam::Get(strg);
         const mfxExtHEVCParam&  hevcPar = ExtBuffer::Get(par);
         mfxStatus sts;
 
         m_callVa = Glob::DDI_Execute::Get(strg);
 
-        std::vector<VAConfigAttrib> attrib(2);
+        std::vector<VAConfigAttrib> attrib(3);
+        mfxPriority priority;
```
mgonchar, line 182: minor: it is actually a codeline waste, better to declare variable where you actually need it, i.e. on line 198
```
mfxPriority priority = core.GetSession()->m_priority;
```

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
```diff
@@ -187,25 +188,39 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(mfxVideoParam const & par)
     }
 
     // Configuration
-    VAConfigAttrib attrib;
+    mfxPriority priority;
```
mgonchar, line 191: same about variables declaration

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -815,14 +816,16 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(VP9MfxVideoParam const & par)
     }
 
     // Configuration
-    VAConfigAttrib attrib[2];
+    mfxPriority priority;
+    VAConfigAttrib attrib[3];
```
mgonchar, line 820: same about ```vector```

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
```diff
@@ -544,12 +545,13 @@ mfxStatus VAAPIEncoder::Init(ENCODE_FUNC func, ExecuteBuffers* pExecuteBuffers)
...
     attrib[0].type = VAConfigAttribRTFormat;
     attrib[1].type = VAConfigAttribRateControl;
+    attrib[2].type = VAConfigAttribContextPriority;
     //attrib[2].type = VAConfigAttribEncSkipFrame;
 
     vaGetConfigAttributes(m_vaDisplay,
         ConvertProfileTypeMFX2VAAPI(pExecuteBuffers->m_sps.Profile),
         VAEntrypointEncSlice,
-        &attrib[0], 2);
+        &attrib[0], 3);
```
mgonchar, line 554: same here

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -224,6 +225,9 @@ mfxStatus VAAPIVideoProcessing::Init(_mfxPlatformAccelerationService* pVADisplay
 
         VAEntrypoint* va_entrypoints = NULL;
         VAStatus vaSts;
+        VAConfigAttrib attrib[1];
```
mgonchar, line 228: I'd still prefer ```vector``` since it is easy extendable

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1709,10 +1711,12 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
 #if defined(MFX_ENABLE_H264_VIDEO_FEI_ENCPAK) || defined(MFX_ENABLE_H264_VIDEO_FEI_PREENC)
     if ( m_isENCPAK )
     {
-        attrib[2].type = VAConfigAttribFEIFunctionType;
+        attrib[numAttrib].type = VAConfigAttribFEIFunctionType;
```
mgonchar, line 1714: I don't like this hack with ```numAttrib``` variable. Better to use ```push_back``` and don't have explicit index manipulation

_studio/shared/umc/io/umc_va/src/umc_va_linux.cpp
```diff
@@ -330,7 +330,7 @@ Status LinuxVideoAccelerator::Init(VideoAcceleratorParams* pInfo)
     MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_HOTSPOTS, "LinuxVideoAccelerator::Init");
     Status         umcRes = UMC_OK;
     VAStatus       va_res = VA_STATUS_SUCCESS;
-    VAConfigAttrib va_attributes[4];
+    VAConfigAttrib va_attributes[5];
```
mgonchar, line 333: same applies here

dmitryermilov:
> What about Hybrid encoders like AV-1 - how you will explain behavior there? No need to make things more complex then they are.

I don't propose to describe per codec in the manual. But ideally it should be documented in release notes or some whitepaper. 

Yes, MSDK manual is pretty abstract from implementation. From an SDK point of view it's fine. But MediaSDK is an introduction point to HW media stack for many users. VA describes even less than our. Documentation at MediaSDK GitHub project is not manual-focud only. If you're against manual update, I'm fine. But the description needs to posted at some other public resource (wiki, whitepapar).

varistar:
> If you're against manual update, I'm fine. But the description needs to posted at some other public resource (wiki, whitepapar).

Yes, I'm fully OK to specify it anywhere beyond manual :)

dmitryermilov:
> > If you're against manual update, I'm fine. But the description needs to posted at some other public resource (wiki, whitepapar).
> 
> Yes, I'm fully OK to specify it anywhere beyond manual :)

Deal:)

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
```diff
@@ -1097,12 +1097,13 @@ mfxStatus VAAPIEncoder::CreateAccelerationService(MfxVideoParam const & par)
...
 
     // Configuration
-    std::vector<VAConfigAttrib> attrib(2);
+    std::vector<VAConfigAttrib> attrib(3);
 
     mfxU32 flag = VA_PROGRESSIVE;
 
     attrib[0].type = VAConfigAttribRTFormat;
     attrib[1].type = VAConfigAttribRateControl;
+    attrib[2].type = VAConfigAttribContextPriority;
```
dvrogozh, line 1106: This resembles the following issue filed after the prev. introduction of new VAAPI feature into msdk code: #2160.  Effectively, if you introduce dependency from new VAAPI, you either need to wrap this code here in the VAAPI version check or update mediasdk libva version requirement here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/81470616597c06deee3e6e1b85d6bbf5e20e0e75/builder/FindPackages.cmake#L484

I vote for the second solution to bump the version.

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
dvrogozh: That's not actually a minor. This will remove the concern of variable not being initialized. Please, move declaration below.

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
dvrogozh: I agree. We need to get these values from somewhere, either having them declared in VAAPI or VAAPI returning a supported range sorted be priority in a API declared way. In this case we can do calculation of min, (min+max)/2, max and map that to LOW, NORMAL, HIGH.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
dvrogozh: yep. Please, move.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
dvrogozh: I am lazy to dig for @mgonchar comment, but suspect I will write the same...

imho, hardcoded 2 is avoidable. Use std::vector attrib(2), use attrib.size()

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
dvrogozh: +1

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
dvrogozh: +1

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
hongbolv: Updated. The max value is reported from media driver. LOW is map to 0, NORMAL is map to max_value/2; HIGH is map to max value.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
hongbolv: Updated

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
hongbolv: updated.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
hongbolv: updated.

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
hongbolv: The default value will be used for the normal priority and other priority policy.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
hongbolv: Just inherit from the previous implementation. Anyway, i change to use std::vector.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
hongbolv: Done

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_lin.cpp
hongbolv: Done.

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
hongbolv: Done

_studio/mfx_lib/shared/src/mfx_mpeg2_encode_vaapi.cpp
hongbolv: Done

_studio/shared/src/mfx_vpp_vaapi.cpp
hongbolv: done.

_studio/shared/umc/io/umc_va/src/umc_va_linux.cpp
hongbolv: Done.

_studio/mfx_lib/encode_hw/mjpeg/src/mfx_mjpeg_encode_vaapi.cpp
hongbolv: Done

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
hongbolv: Done.

hongbolv:
> > > If you're against manual update, I'm fine. But the description needs to posted at some other public resource (wiki, whitepapar).
> > 
> > 
> > Yes, I'm fully OK to specify it anywhere beyond manual :)
> 
> Deal:)
> @hongbolv , please reply if it's fine for you.

It is fine to me.

_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp
hongbolv: Implemented the second solution.

hongbolv:
Changed to dynamic session priority setting implementation.

PR #2181, opened 2020-06-24
Eliminate self-invented shared_ptr-like classes

### LINKED ISSUE #1739: Eliminate self-invented shared_ptr-like classes
We have some manual implementation of refcounter class [here](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/samples/sample_common/include/avc_structures.h#L223). I believe we can replace it with ```std::shared_ptr```

egorovdanil:
Fixes #1739

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/include/umc_h264_au_splitter.h
+++ b/_studio/shared/umc/codec/h264_dec/include/umc_h264_au_splitter.h
@@ -61,7 +61,7 @@ public:
     UMC_H264_DECODER::H264SEIPayLoad* FindPayload(SEI_TYPE type) const;
 
 protected:
-    typedef std::vector<UMC_H264_DECODER::H264SEIPayLoad*> PayloadArray;
+    typedef std::vector<std::shared_ptr<UMC_H264_DECODER::H264SEIPayLoad>> PayloadArray;
     PayloadArray m_payloads;
 
     int32_t FindPayloadPos(SEI_TYPE type) const;
@@ -99,7 +99,7 @@ public:
     SetOfSlices& operator=(const SetOfSlices& set);
 
 protected:
-    std::vector<H264Slice*> m_pSliceQueue;
+    std::vector<std::shared_ptr<H264Slice>> m_pSliceQueue;
 };
 
 /****************************************************************************************************/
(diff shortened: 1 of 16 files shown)


_studio/shared/umc/codec/h265_dec/include/umc_h265_frame.h
```diff
@@ -82,7 +82,9 @@ class H265DecoderRefPicList
 };
 
 // HEVC decoder frame class
-class H265DecoderFrame : public H265DecYUVBufferPadded, public RefCounter
+class H265DecoderFrame;
+typedef std::shared_ptr<std::shared_ptr<H265DecoderFrame>> pH265DecoderFrame;
```
dvrogozh, line 86: 1. `pH265DecoderFrame` looks ugly. I mix it with variable declaration. Probably `H265DecoderFrameP` would be better.
2. `std::shared_ptr<std::shared_ptr>` - do you really need such construction and why?

_studio/mfx_lib/decode/h265/include/mfx_h265_dec_decode.h
```diff
@@ -102,12 +102,12 @@ class VideoDECODEH265 : public VideoDECODE
     bool IsSameVideoParam(mfxVideoParam * newPar, mfxVideoParam * oldPar, eMFXHWType type);
 
     // Fill up frame parameters before returning it to application
-    void FillOutputSurface(mfxFrameSurface1 **surface_out, mfxFrameSurface1 *surface_work, UMC_HEVC_DECODER::H265DecoderFrame * pFrame);
+    void FillOutputSurface(mfxFrameSurface1 **surface_out, mfxFrameSurface1 *surface_work, std::shared_ptr<std::shared_ptr<UMC_HEVC_DECODER::H265DecoderFrame>> pFrame);
```
dvrogozh, line 105: why don't you use new `pH265DecoderFrame` here?

mgonchar:
Absolutely agree with @dmitry-gurulev and @dvrogozh

egorovdanil:
@mgonchar, @dvrogozh, @dgurulev
I completely redid this commit, no longer using the `std::shared_ptr<std::shared_ptr>` construction, and only use shared pointers where it affects the frame lifetime.

At the moment, I have changed only `H265DecoderFrame` class, and I want to know your opinion on the design, whether it is worth changing the rest of the classes this way.

PR #2186, opened 2020-06-25
Fix uninitialized variables and add checks for null pointers

aobolensk:


_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -664,7 +664,7 @@ mfxStatus VideoDECODEMJPEG::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     // It can be useful to run threads right after first frame receive
     MFX_SAFE_CALL(DecodeFrameCheck(bs, surface_work, surface_out));
 
-    UMC::FrameData *dst;
+    UMC::FrameData *dst = NULL;
```
mgonchar, line 667: ```nullptr```

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -3090,6 +3090,7 @@ mfxStatus VideoVPPHW::VppFrameCheck(
             }
         }
 
+        MFX_CHECK_NULL_PTR1(pWorkOutSurf);
```
mgonchar, line 3093: Why it is not inside ```AssignTask```?

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
aobolensk: Done

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
aobolensk: Moved this check inside `AssignTask` function

PR #2192, opened 2020-06-30
[hevce] Fix reported frame order

YurSkur:
m_frameorder is set to 0 after Reset.
User applications cannot match the encoded frame to the bitstream
during encode using mfxExtAVCEncodedFrameInfo.FrameOrder.

### DIFF
--- a/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_hw.cpp
```diff
@@ -848,7 +848,7 @@ mfxStatus   MFXVideoENCODEH265_HW::Reset(mfxVideoParam *par)
         return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
 
     // waiting for submitted in driver tasks
-    MFX_CHECK_STS(WaitingForAsyncTasks(isIdrRequired));
+    MFX_CHECK_STS(WaitingForAsyncTasks(isIdrRequired, false));
```
mgonchar, line 851: Please, pull it off the ```MFX_CHECK_STS``` macro, it will cause double call on unhappy path. I suggest to use ```MFX_SAFE_CALL``` instead

YurSkur: > Please, pull it off the `MFX_CHECK_STS` macro, it will cause double call on unhappy path. I suggest to use `MFX_SAFE_CALL` instead

Changed.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.h
```diff
@@ -129,10 +129,10 @@ namespace Base
         NotNull<Defaults*> m_pQNCDefaults;
         eMFXHWType m_hw = MFX_HW_UNKNOWN;
 
-        void ResetState()
+        void ResetState(bool bZeroFrameOrder = true)
         {
-            m_frameOrder     = mfxU32(-1);
-            m_lastIDR        = 0;
+            m_frameOrder     = bZeroFrameOrder ? mfxU32(-1) : m_frameOrder;
+            m_lastIDR        = bZeroFrameOrder ? 0 : m_frameOrder + 1;
```
mgonchar, line 135: doesn't match code in ```ZeroParams```, is it intentional?

mgonchar:
@eistarov could you look at this one?

PR #2193, opened 2020-06-30
[hevce] fix recon format for specific cases

vilichev:
Fix recons format for specific cases when input format differs
from the output. For example, for HEVCe VDENC
Y210 format is supported as input while the output is P010

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
@@ -2598,6 +2598,22 @@ bool Legacy::GetRecInfo(
         return false;
     }
 
+    if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV420) && CO3.TargetBitDepthLuma == 8)
+    {
+        rec.FourCC = MFX_FOURCC_NV12;
+    }
+    else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV444) && CO3.TargetBitDepthLuma == 8)
+    {
+        rec.FourCC = MFX_FOURCC_AYUV;
+    }
+    else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV420) && CO3.TargetBitDepthLuma == 10)
+    {
+        rec.FourCC = MFX_FOURCC_P010;
+    }
+    else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV444) && CO3.TargetBitDepthLuma == 10)
+    {
+        rec.FourCC = MFX_FOURCC_Y410;
+    }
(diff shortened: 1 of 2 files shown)


vilichev:
> How about RGB input format support

It is supported both for 8b & YUV420 and 8b & YUV444 as output

_studio/mfx_lib/encode_hw/hevc/agnostic/g12/hevcehw_g12_rext.cpp
```diff
@@ -59,7 +59,11 @@ void RExt::InitInternal(const FeatureBlocks& /*blocks*/, TPushII Push)
         auto& rec = pRI->Info;
 
         rec = par.mfx.FrameInfo;
-        
+
+        if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV420) && CO3.TargetBitDepthLuma == 10)
+        {
+            rec.FourCC = MFX_FOURCC_P010;
+        }
```
saosipov, line 66: Why we need it if we have the same in agnostic base?

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
```diff
@@ -2598,6 +2598,22 @@ bool Legacy::GetRecInfo(
...
+    {
+        rec.FourCC = MFX_FOURCC_NV12;
+    }
+    else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV444) && CO3.TargetBitDepthLuma == 8)
+    {
+        rec.FourCC = MFX_FOURCC_AYUV;
+    }
+    else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV420) && CO3.TargetBitDepthLuma == 10)
+    {
+        rec.FourCC = MFX_FOURCC_P010;
```
saosipov, line 2611: We need nv12 instead of P010 for VME, right?

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
```diff
@@ -2598,6 +2598,22 @@ bool Legacy::GetRecInfo(
...
+    }
+    else if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV444) && CO3.TargetBitDepthLuma == 10)
+    {
+        rec.FourCC = MFX_FOURCC_Y410;
+    }
```
saosipov, line 2616: why don't we have here branches for 422?

vilichev: This case is covered by 
if (CO3.TargetChromaFormatPlus1 == (1 + MFX_CHROMAFORMAT_YUV420) && CO3.TargetBitDepthLuma == 8)
{
    rec.FourCC = MFX_FOURCC_NV12;
}
since MSDK sets 420 target chroma format when YUY2 is input in case of VDENC

PR #2194, opened 2020-07-01
hevce: use Low Power mode for 444 encoding by default

xhaihao:
According to
https://github.com/intel/media-driver/#decodingencoding-features, 444
encoding supports Low Power mode only. So we should turn the LowPower
flag on if user doesn't specify the encoding mode.

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
@@ -1438,20 +1438,24 @@ public:
         , const mfxVideoParam& par
         , eMFXHWType hw)
     {
-        bool bValid =
-            par.mfx.LowPower
-            && !CheckTriState(par.mfx.LowPower);
+        if (!Check<mfxU16, MFX_CODINGOPTION_ON, MFX_CODINGOPTION_OFF>(par.mfx.LowPower))
+            return par.mfx.LowPower;
 
+        auto fcc = par.mfx.FrameInfo.FourCC;
         bool bOn =
-            (hw == MFX_HW_CNL
-                && par.mfx.TargetUsage >= 6
-                && par.mfx.GopRefDist < 2
-                && !bValid);
+            ((hw == MFX_HW_CNL
+              && par.mfx.TargetUsage >= 6
+              && par.mfx.GopRefDist < 2) ||
+             (hw >= MFX_HW_ICL &&
+              (fcc == MFX_FOURCC_AYUV
+#if (MFX_VERSION >= 1027)
+               || fcc == MFX_FOURCC_Y410
+#endif
+                  )));
 
         return mfxU16(
             bOn * MFX_CODINGOPTION_ON
-            + bValid * par.mfx.LowPower
-            + !(bOn || bValid) * MFX_CODINGOPTION_OFF);
+            + !bOn * MFX_CODINGOPTION_OFF);
     }
 
     static std::tuple<mfxU16, mfxU16> NumTiles(


saosipov:
So we should turn the LowPower
> flag on if user doesn't specify the encoding mode.
Why do we have to do it?
Historically default behavior is VME behavior?
Do you suggest to switch to VDENC if some feature is not supported on VME e.g. weighted prediction or Rolling-I or RGB input?

xhaihao:
> Why do we have to do it?
> Historically default behavior is VME behavior?

Yes, the current default behavior is VME even if 444 encoding doesn't work with VME.  Another thing SCC encoding doesn't work with VME too,  and MSDK is able to use VDEnc by default for SCC encoding.  I think MSDK should keep consistency in behavior. 

> Do you suggest to switch to VDENC if some feature is not supported on VME e.g. weighted prediction or Rolling-I or RGB input?

I think so because it is hard for user to know the details in the underlying driver.

xhaihao:
@lakulako @vilichev any comment for this PR? Thanks!

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
```diff
@@ -1438,15 +1438,23 @@ class GetDefault
...
-                && par.mfx.TargetUsage >= 6
-                && par.mfx.GopRefDist < 2
-                && !bValid);
+            (!bValid &&
+             ((hw == MFX_HW_CNL
+               && par.mfx.TargetUsage >= 6
+               && par.mfx.GopRefDist < 2) ||
+              (hw >= MFX_HW_ICL &&
+               (fcc == MFX_FOURCC_AYUV
+#if (MFX_VERSION >= 1027)
```
lakulako, line 1454: Looks like #if is not needed, MFX_FOURCC_Y410 is already used without it.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
```diff
@@ -1438,15 +1438,23 @@ class GetDefault
         , const mfxVideoParam& par
         , eMFXHWType hw)
     {
+        auto fcc = par.mfx.FrameInfo.FourCC;
```
lakulako, line 1441: move it close to the place where it is used

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
```diff
@@ -1438,15 +1438,23 @@ class GetDefault
...
+
         bool bValid =
```
lakulako, line 1443: We can make this fussy function simpler, by
if (!Check<mfxU16, MFX_CODINGOPTION_ON, MFX_CODINGOPTION_OFF>(par.mfx.LowPower))
  return par.mfx.LowPower;
the rest will assume !bValid.
It is up to committer.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
mgonchar: Disagree here, this is how it defined in header:
```
#if (MFX_VERSION >= 1027)
    MFX_FOURCC_Y210         = MFX_MAKEFOURCC('Y','2','1','0'),
    MFX_FOURCC_Y410         = MFX_MAKEFOURCC('Y','4','1','0'),
#endif
```

If you know some places where Y410 used unprotected, please open issue ticket.

lakulako: I insist. There are huge amount of use. MFX_VERSION is used to to control enabling new features, but not  every other name definition. It would make code unreadable. It is assumed to be true. At least removing it will not break anything and match other code.

mgonchar: @lakulako what do you mean by other name definition? ```MFX_VERSION``` used to give ability to compile with certain API version. I should be able to compile latest code with API 1.25 for example, and all unsupported in API 1.25 features should be turned off. If you pull ```MFX_FOURCC_Y410``` out of define here, code will not compile.

lakulako: @mgonchar, I meant every 2nd name in the code. There are already many names used without such #if. Including MFX_FOURCC_Y410 in the same module. The code will not compile with such old MFX versions neither with nor without this #if. Compilation with old mfx versions is not an option. In the development #if (MFX_VERSION >= MFX_VERSION_NEXT) is used. After VERSION_NEXT is advanced, obsolete #if are deleted from code.
If one needs older mfx, he can checkout older version from github.

mgonchar: >> After VERSION_NEXT is advanced, obsolete #if are deleted from code.

No, it is done in other way, those ```#if``` remains in code with correct version of API. And all places which violates such rule should be fixed.

@onabiull I believe we need your input here.

dvrogozh: The only user known to me which gets use of older API versions built from the same source code base is Android:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/android/mfx_defs.mk#L54

Just looking into their build scripts, we get:
* Android O: 1.26
* Android P: 1.30
* everything else: 1.32

No idea which build paths are still alive from support perspective though...

I can't say that I very much like approach we use to support API versions and library code corresponding to them, but that's what we are doing as of toda. And to start doing otherwise at this point requires considerations.

xhaihao: @mgonchar @dvrogozh @lakulako Thanks for your comments, so here Y410 should be used with protection, right?

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
xhaihao: Thanks for the comment, I will update the PR.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
xhaihao: I like this idea and will change the commit. Thanks!

saosipov: @mgonchar I'm afraid we don't support #ifdef preprocessor options to build the code for a specific MSDK API. We used to do this some time ago (a couple of years) when driver teams could use old msdk code (they didn't update it often). If we have to support some old MSDK API for some customer it's better to create branch for it.
Moreover some MSDK API version may require some specific driver version with corresponding libva (ddi) structures. Is it possible to do it in driver?
Anyway we've never tested msdk functionality with api versions other than the latest one.
@xhaihao I can't approve this PR because nobody can confirm that it won't break our validations. We need to check first if we use changing default behavior anywhere. Besides if we change default behavior for 444 then we need to do the same for 422 (vme by default), SCC (vdenc by default), 12b (vme by default), right? And this is only if we try to distinguish HW by profiles but we can continue the same exercise basing on feature scope.

xhaihao: @saosipov The current SDK has already used the right mode for 422 (vme), SCC (vdenc), 12bit (vme) by default. Only for 444, the SDK uses the wrong mode (vme) by default. User has to specify low-power mode for 444 encoding, which  which confuses me.

xhaihao: > I can't approve this PR because nobody can confirm that it won't break our validations.

I don't know how did you test this feature in your validations. Per my testing, -qsv-ff (or -lowpower:on) must be used for 444 encoding when using sample_encode without this path, otherwise user will get MFX errors.

```
$> sample_encode h265 -hw -ayuv -i input.ayuv.352x288.yuv -o output.h265 -w 352 -h 288

[ERROR], sts=MFX_ERR_ABORTED(-12), SynchronizeFirstTask, SyncOperation failed at /home/xhh/graphics/ex/gstreamer/MediaSDK/samples/sample_encode/src/pipeline_encode.cpp:168

[ERROR], sts=MFX_ERR_ABORTED(-12), GetFreeTask, m_TaskPool.SynchronizeFirstTask failed at /home/xhh/graphics/ex/gstreamer/MediaSDK/samples/sample_encode/src/pipeline_encode.cpp:2188

[ERROR], sts=MFX_ERR_ABORTED(-12), Run, m_pmfxENC->EncodeFrameAsync failed at /home/xhh/graphics/ex/gstreamer/MediaSDK/samples/sample_encode/src/pipeline_encode.cpp:2492

[ERROR], sts=MFX_ERR_ABORTED(-12), main, pPipeline->Run failed at /home/xhh/graphics/ex/gstreamer/MediaSDK/samples/sample_encode/src/sample_encode.cpp:1660
```

So I presumed the LowPower flag is set via -qsv-ff (or -lowpower:on) in your validation, and my change shouldn't have impact to your validations unless you have some negative cases in your validation to expect for these MFX errors.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
dmitryermilov: > @mgonchar @dvrogozh @lakulako Thanks for your comments, so here Y410 should be used with protection, right?

2 cents: 
IMHO the fact that we're putting feature definitions under `#if SOME_VERSION` is an extra burden (especially from implementation point of view: we have to put implementations under corresponding `#if SOME_VERSION` ) without real benefits. I don't know real cases where it's really (objectively!) required.  
BUT we _already_ have it, so in order to be consistent, we have either to support it or fully drop. 
So my answer is **yes**, we need the protection there.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
saosipov: @xhaihao Could you update PR according to what you promised to do and we will take it?
We discussed this change internally and confirmed there's no risk with it.
As for your example with sample_encoder we'll have  a look.
I think you should have got UNSUPPORTED error instead of ABORT

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
xhaihao: @dmitryermilov thanks for confirming this.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
xhaihao: @saosipov I updated the PR, please review it when you get a chance.

PR #2203, opened 2020-07-06
[Core] Check if planes are continuously allocated

asmirno1:
This check disables CM usage if surface planes
are not allocated as one memory block.
This is must-condition for CM usage.

### DIFF
--- a/_studio/shared/src/cm_mem_copy.cpp
+++ b/_studio/shared/src/cm_mem_copy.cpp
@@ -3164,6 +3164,48 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
 
 #define CM_RGB_SUPPORTED_COPY_SIZE(ROI) (ROI.width <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT && ROI.height <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT )
 
+bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(const mfxFrameSurface1 &surf)
+{
+    switch (surf.Info.FourCC)
+    {
+    // Packed formats like YUY2, UYVY, AYUV, Y416, Y210, Y410, A2RGB10, RGB565, RGB3 
+    // does not need to be handled here - they got only one plane with all the luma and chroma data interleaved
+
+    // Handling semi-planar formats
+    case MFX_FOURCC_P010:
+    case MFX_FOURCC_P016:
+    case MFX_FOURCC_P210:
+    case MFX_FOURCC_NV12:
+    case MFX_FOURCC_NV16:
+        {
+            mfxU32 stride_in_bytes = surf.Data.PitchLow + ((mfxU32)surf.Data.PitchHigh << 16);
+            size_t luma_size_in_bytes = stride_in_bytes * mfx::align2_value(surf.Info.Height);
+            return surf.Data.Y + luma_size_in_bytes == surf.Data.UV;
+            break;
+        }
+
+    // Handling planar formats
+    case MFX_FOURCC_YV12:
+        {
+            mfxU32 stride_in_bytes = surf.Data.PitchLow + ((mfxU32)surf.Data.PitchHigh << 16);
+            size_t luma_size_in_bytes = stride_in_bytes * mfx::align2_value(surf.Info.Height);
+            size_t chroma_size_in_bytes = luma_size_in_bytes / 4; //stride of the V plane is half the stride of the Y plane; and the V plane contains half as many lines as the Y plane
+            return surf.Data.Y + luma_size_in_bytes == surf.Data.V && surf.Data.V + chroma_size_in_bytes == surf.Data.U;
+            break;
+        }
+
+    // Handling RGB-like formats
+    case MFX_FOURCC_RGBP:
+        {
+            mfxU32 stride_in_bytes = surf.Data.PitchLow + ((mfxU32)surf.Data.PitchHigh << 16);
+            size_t channel_size_in_bytes = stride_in_bytes * mfx::align2_value(surf.Info.Height);
+            return surf.Data.B + channel_size_in_bytes == surf.Data.G && surf.Data.G + channel_size_in_bytes == surf.Data.R;
+            break;
+        }
+    }
+    return true;
+}
+
 bool CmCopyWrapper::CanUseCmCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc)
 {
     mfxSize roi = {std::min(pSrc->Info.Width, pDst->Info.Width), std::min(pSrc->Info.Height, pDst->Info.Height)};
@@ -3187,7 +3229,7 @@ bool CmCopyWrapper::CanUseCmCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc)
         verticalPitch = (verticalPitch % pDst->Data.Pitch)? 0 : verticalPitch / pDst->Data.Pitch;
         if (isNV12LikeFormat(pDst->Info.FourCC) && isNV12LikeFormat(pSrc->Info.FourCC) && CM_ALIGNED(pDst->Data.Y) && CM_ALIGNED(pDst->Data.UV) && CM_SUPPORTED_COPY_SIZE(roi) && verticalPitch >= pDst->Info.Height && verticalPitch <= 16384)
         {
-            return true;
+            return CheckSurfaceContinuouslyAllocated(*pDst);
         }
         else if(isSinglePlainFormat(pDst->Info.FourCC) && isSinglePlainFormat(pSrc->Info.FourCC) && pSrc->Info.Shift == pDst->Info.Shift && CM_ALIGNED(dstPtr) && CM_SUPPORTED_COPY_SIZE(roi))
         {
--- a/_studio/shared/include/cm_mem_copy.h
+++ b/_studio/shared/include/cm_mem_copy.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019 Intel Corporation
+// Copyright (c) 2017-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -137,6 +137,7 @@ public:
     mfxStatus IsCmCopySupported(mfxFrameSurface1 *pSurface, mfxSize roi);
 
     static bool CanUseCmCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc);
+    static bool CheckSurfaceContinuouslyAllocated(const mfxFrameSurface1 &surf);
     static bool isSinglePlainFormat(mfxU32 format);
     static bool isNeedSwapping(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc);
     static bool isNeedShift(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc);


_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -26,6 +26,7 @@
 
 typedef const mfxU8 mfxUC8;
 
+#define ALIGN32(X) (((mfxU32)((X)+31)) & (~ (mfxU32)31))
```
mgonchar, line 29: please use ```mfx::align2_value``` instead

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
 
 #define CM_RGB_SUPPORTED_COPY_SIZE(ROI) (ROI.width <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT && ROI.height <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT )
 
+bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(mfxFrameSurface1 *surf)
```
mgonchar, line 3168: accept by reference, not pointer, or you have to check it against ```NULL```

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+{
+    mfxFrameData* ptrFrameData = &surf->Data;
+
+    switch (surf->Info.FourCC)
+    {
+    case MFX_FOURCC_P010:
+    case MFX_FOURCC_P016:
+    case MFX_FOURCC_P210:
+    case MFX_FOURCC_NV16:
+    case MFX_FOURCC_NV12:
```
mgonchar, line 3178: nv12 and nv16 are 8 bpp!

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+            {
+                return false;
+            }
+
+            break;
+        }
+
+    case MFX_FOURCC_YV12:
+        {
+            mfxU32 Y_stride_in_bytes = ptrFrameData->PitchLow + ((mfxU32)ptrFrameData->PitchHigh << 8);
```
lakulako, line 3195: It must be << 16! PitchLow is 16 lower bits of U32, PitchHigh is upper 16 bits. I never depends on CF.

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(mfxFrameSurface1 *surf)
+{
+    mfxFrameData* ptrFrameData = &surf->Data;
+
+    switch (surf->Info.FourCC)
+    {
+    case MFX_FOURCC_P010:
+    case MFX_FOURCC_P016:
+    case MFX_FOURCC_P210:
+    case MFX_FOURCC_NV16:
```
lakulako, line 3177: P210 and NV16 are 422 formats, unlike 420 their UV size is equal to Y. Don't divide height /2.

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+    case MFX_FOURCC_NV12:
+        {
+            mfxU32 stride_in_bytes = ptrFrameData->PitchLow + ((mfxU32)ptrFrameData->PitchHigh << 16);
+
+            size_t totalBufferUPSize = stride_in_bytes * ALIGN32(surf->Info.Height)
+                + stride_in_bytes * surf->Info.Height / 2;
+
+            if (ptrFrameData->UV + stride_in_bytes * surf->Info.Height / 2 > ptrFrameData->Y + totalBufferUPSize)
```
lakulako, line 3185: Just substitute totalBufferUPSize here and simplify. You will see that you only check UV > Y + stride_in_bytes * ALIGN32(surf->Info.Height).
So the code can be much simpler.
How is guaranteed that alignment is ALWAYS 32?
The function works incorrectly if UV < Y, i.e. allocated by separate allocations. Why not compare with ==?

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+
+            break;
+        }
+
+    case MFX_FOURCC_YV12:
+        {
+            mfxU32 Y_stride_in_bytes = ptrFrameData->PitchLow + ((mfxU32)ptrFrameData->PitchHigh << 8);
+            mfxU32 VU_stride_in_bytes = Y_stride_in_bytes / 2;
+
+            size_t totalBufferUPSize = Y_stride_in_bytes * ALIGN32(surf->Info.Height)
```
lakulako, line 3198: It is the same value as in previous case.

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+            if (ptrFrameData->U + VU_stride_in_bytes * surf->Info.Height / 2 > ptrFrameData->Y + totalBufferUPSize)
+                return false;
+
+            break;
+        }
+    case MFX_FOURCC_RGBP:
+        {
+            mfxU32 stride_in_bytes = ptrFrameData->PitchLow + ((mfxU32)ptrFrameData->PitchHigh << 8);
+
+            size_t totalBufferUPSize = stride_in_bytes * surf->Info.Height * 3;
```
lakulako, line 3210: Height is not aligned? Guaranteed?

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+
+            if (ptrFrameData->R + stride_in_bytes * surf->Info.Height > ptrFrameData->B + totalBufferUPSize)
```
lakulako, line 3212: BGR order is assumed, but GBR  or ??B will pass as well. Again, doesn't work if planes were allocated separately.

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3165,59 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+
+            if (ptrFrameData->UV + stride_in_bytes * surf->Info.Height / 2 > ptrFrameData->Y + totalBufferUPSize)
+            {
+                return false;
+            }
+
+            break;
+        }
+
+    case MFX_FOURCC_YV12:
```
mgonchar, line 3193: Absolutely agree with comments from @lakulako We already discovered most of them with @asmirno1 in offline discussion. I will just point one more time that for planar formats like YV12 we must check all three Y/U/V pointers

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3164,72 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+    // Packed formats like YUY2, UYVY, AYUV, Y416, Y210, Y410, A2RGB10, RGB565, RGB3 
+    // does not need to be handled here - they got only one plane with all the luma and chroma data interleaved
+
+    // Handling semi-planar formats
+    case MFX_FOURCC_NV12:
+    case MFX_FOURCC_NV16:
+        {
+            mfxU32 stride_in_bytes = ptrFrameData->PitchLow + ((mfxU32)ptrFrameData->PitchHigh << 16);
+            size_t Y_size = stride_in_bytes * mfx::align2_value(surf.Info.Height, 32);
+            if (ptrFrameData->Y + Y_size + 1 != ptrFrameData->UV)
```
lakulako, line 3182: What for +1 ?

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3164,72 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+            {
+                return false;
+            }
+            break;
```
lakulako, line 3186: can "return true;" instead

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3164,72 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
...
+        }
+
+    case MFX_FOURCC_P010:
+    case MFX_FOURCC_P016:
+    case MFX_FOURCC_P210:
+        {
+            mfxU32 stride_in_bytes = ptrFrameData->PitchLow + ((mfxU32)ptrFrameData->PitchHigh << 16);
+            size_t Y_size = stride_in_bytes * mfx::align2_value(surf.Info.Height, 32) * 2;
```
lakulako, line 3194: What for "* 2"? Stride is already in bytes.

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3164,72 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
 
 #define CM_RGB_SUPPORTED_COPY_SIZE(ROI) (ROI.width <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT && ROI.height <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT )
 
+bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(mfxFrameSurface1 &surf)
+{
+    mfxFrameData* ptrFrameData = &surf.Data;
```
lakulako, line 3169: What is the reason to change from reference to pointer?

_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3164,6 +3164,72 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
 
 #define CM_RGB_SUPPORTED_COPY_SIZE(ROI) (ROI.width <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT && ROI.height <= CM_RGB_MAX_GPUCOPY_SURFACE_HEIGHT )
 
+bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(mfxFrameSurface1 &surf)
```
lakulako, line 3167: const mfxFrameSurface1 &surf

_studio/shared/src/cm_mem_copy.cpp
mgonchar: just ```return ptrFrameData->Y + Y_size != ptrFrameData->UV```

no need in explicit ```true```/```false```

_studio/shared/src/cm_mem_copy.cpp
mgonchar: Yes, it is not needed, you've already skipped entire plane

PR #2212, opened 2020-07-13
[VP9e] Added check on max number of reference frames for TU7

eshiryae:
Only 1 reference frame is supported for TargetUsage 7.

### DIFF
--- a/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
+++ b/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp
```diff
@@ -1337,15 +1337,20 @@ mfxStatus CheckParameters(VP9MfxVideoParam &par, ENCODE_CAPS_VP9 const &caps)
             changed = true;
         }
 
-        // TargetUsage 4 and 7 don't support 3 reference frames
+        // For targetUsage 4: MaxNum_Reference is 2
+        // For targetUsage 7: MaxNum_Reference is 1
```
dmitryermilov, line 1341: Is it the same for all gen platforms?

eshiryae: yes

dmitryermilov: NumRefFrame defines DPB size. Why does TU influence DPB size? Aren't you mixing DPB size and number of active references?

eshiryae: This is restriction from driver side.
Also from DDI:
There is a relationship between the TargetUsage set and the number of L0/L1 references that will be used. MaxNum_Reference0 and MaxNum_Reference1 returned in ENCODE_CAPS are the maximum supported of all the TargetUsages.
Also change for caps (MaxNum_Reference0) for VP9e is on review.

dmitryermilov: Okay, you're mixing DPB size and number of active references. Please refer to manual:

_NumRefFrame: Max number of all available reference frames (for AVC/HEVC NumRefFrame defines DPB size); if NumRefFrame = 0, this parameter is not specified.See also mfxExtCodingOption3::NumRefActiveP, NumRefActiveBL0 and NumRefActiveBL1 which set a number of active references._

dmitryermilov:
Any review from @nastsokolova ?

PR #2216, opened 2020-07-14
[h264e] Return UNSUPPORTED, if sliding window is not supported by caps

Alexandr-Konovalov:
Issue: MDP-62514

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -3277,6 +3277,12 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         extOpt2->MBBRC = MFX_CODINGOPTION_OFF;
     }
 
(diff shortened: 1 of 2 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -3277,6 +3277,12 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
         extOpt2->MBBRC = MFX_CODINGOPTION_OFF;
     }
 
+    if (extOpt3->WinBRCSize && !hwCaps.ddi_caps.FrameSizeToleranceSupport)
+    {
+        unsupported = true;
```
dmitryermilov, line 3282: It looks like we need to modify this line: https://github.com/Intel-Media-SDK/MediaSDK/blob/089296a90c83ed54e95b20597437a6839406e10b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp#L234

Alexandr-Konovalov: Done.

PR #2220, opened 2020-07-16
[h264e] AdaptiveMaxFrameSize is unsupported under libva

Alexandr-Konovalov:
Issue: MDP-50286

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -788,6 +788,12 @@ namespace
 #if MFX_VERSION >= 1023
         if (!CheckTriStateOption(extOpt3.AdaptiveMaxFrameSize)) changed = true;
 
+        if(!hwCaps.AdaptiveMaxFrameSizeSupport && IsOn(extOpt3.AdaptiveMaxFrameSize))
+        {
+            extOpt3.AdaptiveMaxFrameSize = MFX_CODINGOPTION_UNKNOWN;
+            unsupported = true;
+        }
+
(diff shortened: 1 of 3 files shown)


_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -788,6 +788,13 @@ namespace
 #if MFX_VERSION >= 1023
         if (!CheckTriStateOption(extOpt3.AdaptiveMaxFrameSize)) changed = true;
 
+        // libva doesn't accept AdaptiveMaxFrameSize and MaxFrameSizeP, so return unsupported
+        if(MFX_HW_VAAPI == vaType && IsOn(extOpt3.AdaptiveMaxFrameSize))
```
dmitryermilov, line 792: @Alexandr-Konovalov , can we do it via "custom" caps? I mean by extension of [this ](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/shared/include/mfx_h264_encode_struct_vaapi.h#L328)

Alexandr-Konovalov: @dmitryermilov 
>  can we do it via "custom" caps?

Thank you,  static constraints are better than dynamic conditions.

> And in parallel, can you please submit an issue at https://github.com/intel/libva with a request to add missing caps reporting.

Sure, this is in our plans.

dmitryermilov: @Alexandr-Konovalov , when this commits back-merges into "windows" repo, please don't forget to update windows path..

PR #2227, opened 2020-07-18
[sample_encode] set default idr_interval

andreyor:
set default idr_interval for AVC=0 for HEVC=1


lakulako:
Why it is needed? If IdrInterval is not set, it mean that codec decides, not application.
Making every intra frame IDR by default will decrease default encoding quality.

PR #2233, opened 2020-07-23
Complete av1d output surface information.

### LINKED ISSUE #2231: AV1d doesn't fill required surface fields
timestamp, aspect ratio, FrameRateExt, FrameOrder, DataFlag, mfxExtDecodedFrameInfo, etc handling are missed in #2218 _studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp

wangyan42164:
Fixes #2231.

### DIFF
--- a/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
@@ -153,6 +153,7 @@ VideoDECODEAV1::VideoDECODEAV1(VideoCORE* core, mfxStatus* sts)
     , m_request()
     , m_response()
     , m_is_init(false)
+    , m_in_framerate(0)
 {
     if (sts)
     {
@@ -213,6 +214,8 @@ mfxStatus VideoDECODEAV1::Init(mfxVideoParam* par)
     }
 
     m_first_par = m_init_par;
+    m_in_framerate = (mfxF64) m_first_par.mfx.FrameInfo.FrameRateExtD / m_first_par.mfx.FrameInfo.FrameRateExtN;
+    m_decoder->SetInFrameRate(m_in_framerate);
 
     //mfxFrameAllocResponse response{};
     bool internal = ((m_platform == MFX_PLATFORM_SOFTWARE) ?
@@ -397,6 +400,15 @@ mfxStatus VideoDECODEAV1::Reset(mfxVideoParam* par)
     MFX_CHECK(m_allocator->Reset() == UMC::UMC_OK, MFX_ERR_MEMORY_ALLOC);
 
     m_first_par = *par;
+
+    if (0 == m_first_par.mfx.FrameInfo.FrameRateExtN || 0 == m_first_par.mfx.FrameInfo.FrameRateExtD)
+    {
+        m_first_par.mfx.FrameInfo.FrameRateExtD = 1;
+        m_first_par.mfx.FrameInfo.FrameRateExtN = 30;
+    }
+
+    m_in_framerate = (mfxF64) m_first_par.mfx.FrameInfo.FrameRateExtD / m_first_par.mfx.FrameInfo.FrameRateExtN;
+    m_decoder->SetInFrameRate(m_in_framerate);
 
     return MFX_ERR_NONE;
 }
@@ -959,7 +971,11 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
                     MFX_CHECK_STS(sts);
             }
 
-            UMC_AV1_DECODER::AV1DecoderFrame* frame = GetFrameToDisplay();
+            UMC_AV1_DECODER::AV1DecoderFrame* frame = m_decoder->GetCurrFrame();
+            if (frame && bs->TimeStamp != static_cast<mfxU64>(MFX_TIMESTAMP_UNKNOWN))
+                frame->SetFrameTime(GetUmcTimeStamp(bs->TimeStamp));
+
+            frame = GetFrameToDisplay();
             if (frame)
             {
                 sts = FillOutputSurface(surface_out, surface_work, frame);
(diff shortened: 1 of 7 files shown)


_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -1171,8 +1190,14 @@ mfxStatus VideoDECODEAV1::FillOutputSurface(mfxFrameSurface1** surf_out, mfxFram
     UMC::VideoDataInfo const* vi = fd->GetInfo();
     MFX_CHECK(vi, MFX_ERR_DEVICE_FAILED);
 
+    surface_out->Data.TimeStamp = bs->TimeStamp != static_cast<mfxU64>(MFX_TIMESTAMP_UNKNOWN) ? bs->TimeStamp : GetMfxTimeStamp(m_frame_order * m_in_framerate);
```
IlyaSergeyev, line 1193: Seems using bitstream here is not good idea. It is better to store timestamp in the pFrame struct (somewhere in the umc_av1_* code) and then just use GetMfxTimeStamp() function here.

dmitryermilov: Echo Ilya. It's better to set/get TS via pFrame . Two reasons:
1) Just to code style (we do this way for the rest decoders)
2) I assume AV1 has a re-ordering. If so, TS from bistream (in encoded order) can't be set to decoded surface (in display order).

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -1171,8 +1190,14 @@ mfxStatus VideoDECODEAV1::FillOutputSurface(mfxFrameSurface1** surf_out, mfxFram
...
+    surface_out->Data.FrameOrder = m_frame_order;
+    surface_out->Data.Corrupted = 0;
+    surface_out->Info.PicStruct = MFX_PICSTRUCT_PROGRESSIVE;
     surface_out->Info.CropW = static_cast<mfxU16>(pFrame->GetUpscaledWidth());
-    surface_out->Info.CropH = static_cast<mfxU16>(pFrame->GetHeight());
+    surface_out->Info.CropH = static_cast<mfxU16>(pFrame->GetFrameHeight());
+    surface_out->Info.AspectRatioW = 1;
+    surface_out->Info.AspectRatioH = 1;
```
dmitryermilov, line 1200: There is still no:
- FrameRate info.
- mfxExtDecodedFrameInfo
 - proper handling of  Corrupted

wangyan42164: FrameRateInfo: I have updated in the latest commit.
mfxExtDecodedFrameInfo: I think we have done it in SetFrameType() which is called by :
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L126-L146
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L1167
Corrupted: IMHO, AV1 should have not special handling like VP9. @IlyaSergeyev Do you think so?

dmitryermilov: Okay. Thanks. So only 'Corrupted' remained not covered here.

dmitryermilov: Ahh, I just saw it's already covered here: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L674

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -1149,10 +1162,16 @@ inline void CopyFilmGrainParam(mfxExtAV1FilmGrainParam &extBuf, UMC_AV1_DECODER:
...
+mfxStatus VideoDECODEAV1::FillOutputSurface(mfxBitstream *bs, mfxFrameSurface1** surf_out, mfxFrameSurface1* surface_work, UMC_AV1_DECODER::AV1DecoderFrame* pFrame)
 {
     MFX_CHECK_NULL_PTR3(surface_work, pFrame, surf_out);
 
+    if (!pFrame->GetShowFrame())
+    {
+        surf_out = 0;
+        return (mfxStatus)MFX_ERR_MORE_DATA_SUBMIT_TASK;
+    }
+
```
dmitryermilov, line 1174: It should be handled in GetFrameToDisplay(), not in the FillOutputSurface function which just adds some meta info to mfxFrameSurface1

wangyan42164: Yes. Its has been checked in the following:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp#L546
I have removed the section in the latest commit.
Thanks.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
dmitryermilov: @wangyan42164 , everything except this part is fine to me. Can you please re-factor the item which Ilya pointed at? I understand that you borrowed the logic from VP9 but it wan't written a good way there.

wangyan42164: Sure. I have changed it in the latest commit. @dmitryermilov @dmitry-gurulev @IlyaSergeyev Could you please review it again? Thanks.

wangyan42164:
@dmitryermilov I have refined this PR based on our offline discussion. Could you please review it again? Thanks.

PR #2235, opened 2020-07-24
Add AV1 codec level checking for decoding

wangyan42164:

### DIFF
--- a/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
@@ -426,6 +426,44 @@ mfxTaskThreadingPolicy VideoDECODEAV1::GetThreadingPolicy()
     return MFX_TASK_THREADING_SHARED;
 }
 
+inline
+mfxStatus CheckLevel(mfxVideoParam* in, mfxVideoParam* out)
+{
+    MFX_CHECK_NULL_PTR1(out);
+
+    mfxStatus sts = MFX_ERR_NONE;
+
+    if (in)
+    {
+        switch(in->mfx.CodecLevel)
+        {
+        case MFX_LEVEL_AV1_2:
+        case MFX_LEVEL_AV1_21:
+        case MFX_LEVEL_AV1_22:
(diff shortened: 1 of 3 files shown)


api/include/mfxstructures.h
```diff
@@ -507,6 +507,22 @@ enum {
     MFX_PROFILE_AV1_MAIN                    = 1,
     MFX_PROFILE_AV1_HIGH                    = 2,
     MFX_PROFILE_AV1_PRO                     = 3,
+
+    MFX_LEVEL_AV1_2                         = 0,
```
dmitry-gurulev, line 511: Value '0' is reserved for `MFX_LEVEL_UNKNOWN`. Please do mapping of public API's levels definition to specific `seq_level_idx` internally

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -450,6 +450,35 @@ mfxStatus VideoDECODEAV1::Query(VideoCORE* core, mfxVideoParam* in, mfxVideoPara
 #endif
     }
 
+    if (in)
+    {
+        switch(in->mfx.CodecLevel)
```
mgonchar, line 455: can we wrap checking to separate helper function? Just to not expand code of the Query

api/include/mfxstructures.h
mgonchar: How about doing it similarly to HEVC in this enum, i.e.:
```
MFX_LEVEL_AV1_2                         = 20,
MFX_LEVEL_AV1_21                       = 21,
...
```

wangyan42164: Sure. I have changed it in the latest commit.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
wangyan42164: Sure. I have changed it in the latest commit.

dmitryermilov:
@dmitry-gurulev , are there remaining concerns from you?

PR #2240, opened 2020-07-26
Coding guide improvements

mgonchar:
- Remove issues of doc -> md conversion
- Clarify auto usage

### DIFF
--- a/doc/Coding_guidelines.md
+++ b/doc/Coding_guidelines.md
@@ -461,7 +461,7 @@ Feel free to improve this document. Bring more examples and rules. But please tr
       {
       ...
       
-      size_t size() **const** { return size; }
+      size_t size() const { return size; }
       
       private:
           size_t size;
@@ -524,10 +524,10 @@ Feel free to improve this document. Bring more examples and rules. But please tr
         
       ```
       // This is always true statement, default second argument of DetectFrameNumGap masked typo
-      if (view.GetPOCDecoder(0)->DetectFrameNumGap(slice**),** true)
+      if (view.GetPOCDecoder(0)->DetectFrameNumGap(slice), true)
       
       // Correct code is following
-      if (view.GetPOCDecoder(0)->DetectFrameNumGap(slice**,** true**)**)
+      if (view.GetPOCDecoder(0)->DetectFrameNumGap(slice, true))
       ```
     - Use ```const``` qualifiers for parameters passed by reference/pointer and for class member functions everywhere it is possible.
 
@@ -561,7 +561,7 @@ Feel free to improve this document. Bring more examples and rules. But please tr
       Example:
 
       ```
-      template <typename T> void foo(**T&&**)
+      template <typename T> void foo(T&&)
       ```
 
     - Don't forget about Perfect forwarding (```std::forward```).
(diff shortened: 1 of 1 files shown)


doc/Coding_guidelines.md
```diff
@@ -556,6 +556,10 @@ Feel free to improve this document. Bring more examples and rules. But please tr
 
     - ```auto``` is the only way to create lambda function object.
 
+    - Remember that usage of ```auto``` may reduce available context of code in non-IDE text editors (or during code review on GitHub), several suggestions on limiting this effect of ```auto``` usage are given below:
+        * do not use ```auto``` for buit-in types (such as ```int```, ```char```, etc.) and trivially castable user types (such as ```enums```);
+        * do not use ```auto``` for initialization with returned object from some user-defined function.
```
dmitry-gurulev, line 561: Don't understand why ```auto time_point = std::chrono::system_clock::now();``` is good example but ```auto obj = my::create_my_cool_obj();``` (returning object from UD function) is bad. For me, the only one thing matters - improving readability. And in case, when context is not clear, let's say ```auto obj = my::do_something();``` it's better to give better name to function and variables rather than restrict ```auto``` usage.

mgonchar: Well, it is complicated thing indeed. Here I tried to formulate some insights on how engineering judgement may work for distinguishing between cases where type deduction for ```auto``` is "trivial" and when it just hides important information. My thought was that standard library is a "well known" code and user-defined code is most likely not so widely known (especially if brought in some recent commit), so using explicit types may help users to show more context.

This clarification was formulated after long discussion with @gkrivor, who described some difficulties during review of code which uses ```auto```. It was very hard to formulate something a bit more precise than "use your judgement" or "clear enough" or other statement that will not be fully subjective.

In general I agree that in final the only things which matters are simplicity and clearness of code. Developers should not perform compiler or static analysis tool work during review or code reading, so usage of meaningful names is of course preferred than any formal restrictions that are placed in this guide. In my opinion if you think that code is good structured and easy for understanding, then it doesn't matter how your code uses ```auto```.

If you have suggestion how to rephrase or modify this part I'm fully opened for it.

gkrivor: @mgonchar thanks! Maybe enum's may need some clarification like "in case enum was defined without class keyword". What do you think? But your description is better for me than adding some exceptions...

@dmitry-gurulev good catch. Is it good to formulate rule as "use auto in case left and right expression might be easily understood"?

Good examples:
auto it = myVector.begin(); //regular iterator notation
auto pTypeNameObj = reinterpret_cast<typename*>(pObj); //obvious type definition on right side, good name on left side
auto pTypeNameObj = getTypeNameObj(...); //obvious type definition in both names

Bad examples:
auto v = getObject(); //which type of object we will get after function call?
auto e = parseString(); //worse example, which type of object we will get after function call?
auto Obj = reinterpret_cast<typename*>(pObj); //left side doesn't provide information about type, also non-ptr object might be expected
auto sts = S_OK; //type of status is unclear, might cause wrong comparison later

dmitry-gurulev: OK, looks what u said is fully aligned w/ my understanding. Let's wait for @gkrivor suggestions.

PR #2242, opened 2020-07-27
[av1d] fix potential memleak issue

Jexu:


_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
```diff
@@ -378,11 +378,19 @@ namespace UMC_AV1_DECODER
                 updated_refs = prevFrameDPB;
                 Repeat_show = 0;
             }
+            if(Curr_temp)
```
mgonchar, line 381: can we use ```std::unique_ptr``` instead ?

dvrogozh: +1

And I am not 100% sure that you've fully addressed a leak issue. Actually, you delete some memory in a GetFrame() function which, from the name, is a function called regularly on each frame. So, can this happen that in the very end of decoding, on decoder reset, etc. this memory will still be allocated and not destroyed? IMHO, std::unique_ptr would be much better way to handle that and guarantee no-leak state.

Jexu:
Hi @mgonchar @dvrogozh 

can you please help review again?

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
```diff
@@ -378,12 +378,12 @@ namespace UMC_AV1_DECODER
                 updated_refs = prevFrameDPB;
                 Repeat_show = 0;
             }
-            Curr_temp = Curr;
+            Curr_temp.reset(Curr.get());
```
dvrogozh, line 381: This does not make sense to me. Unique_ptr is specifically designed to have single owner of the object. By assigning result of get() you actually end up with 2 unique_ptr managing the same object. It seems you need to have :
1. Either assignment as you had before so Curr will lose ownership
2. Or (if you still need ownership for Curr) you need you declare these as std::shared_ptr objects to have multiple owners.

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
```diff
@@ -378,12 +378,12 @@ namespace UMC_AV1_DECODER
...
         }
         if (!pPrevFrame)
         {
             updated_refs = refs_temp;
-            Curr_temp = Curr;
+            Curr_temp.reset(Curr.get());
```
dvrogozh, line 386: same

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
```diff
@@ -647,23 +647,22 @@ namespace UMC_AV1_DECODER
...
-               Poutput = pPrevFrame;
+               Poutput.reset(pPrevFrame);
             }
             else
             {
                 if (Repeat_show || FH_OutTemp.show_frame)
                 {
                     Poutput ->DecrementReference();
-                    Poutput = Curr;
+                    Poutput.reset(Curr.get());
```
dvrogozh, line 657: same as above, but for Poutput

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
mgonchar: +1

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
wangyan-intel: IMHO, 
1. AV1DecoderFrame inherits from RefCounter. So we may need not use std::unique_ptr/share_ptr. Just need increase reference count when assignment and decrease reference count at the end of the scope of the pointer.
2. Curr/Curr_temp/Poutput should not be initialized by new operator when AV1Decoder constructor is called because  they are just used to pointer to some AV1DecoderFrame object in dpb array.
https://github.com/Intel-Media-SDK/MediaSDK/blob/5ff3e4c1a8ac459f51ff36e7a23cc1a1c9087b5c/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp#L40-L42
https://github.com/Intel-Media-SDK/MediaSDK/blob/5ff3e4c1a8ac459f51ff36e7a23cc1a1c9087b5c/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp#L627-L630
3. When AV1Decoder is destroyed, AV1DecoderFrame object in dpb array will be freed. No memory leak.

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
wangyan-intel: Same with my previous comments.

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
wangyan-intel: Same with my previous comments.

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
wangyan-intel: Same with my previous comments.

Jexu:
Hi @mgonchar  @dvrogozh 

as yan memsioned before, the ptr seems no need to init since it always points to the frame from dpb array. so can you help review again?

dvrogozh:
@Jexu : please, confirm that this patch is still actual?

Jexu:
@dvrogozh yeah it is, pls help review it

wangyan-intel:
@Jexu After I debug thread sync between lib and app, I find the "new" Poutput frame will be used to check UID in CompleteDecodeFrame(). Current logic isn't designed by me.  Although I don't think it is a good design, I suggest we pend this PR so far. They should be refined together with thread sync logic.

PR #2247, opened 2020-07-29
[ted] Add README and use only installed binaries

aobolensk:


tests/README.md
```diff
@@ -0,0 +1,24 @@
...
+#### Prerequisites:
+Before running tests ensure that required packages are installed on your system:
+```
+git
+git-lfs
+python >= 3.6
+```
+As for default we assume:
+
+MediaStack tarball installed using install_media.sh or Media SDK has been previously compiled and installed to `/opt/intel/mediasdk`
```
dvrogozh, line 14:
```
$ fgrep -rsn install_media.sh
$
```

You refer to non-existing file. Drop it.

> Media SDK has been previously compiled and installed to

Be specific and provide cmake configure instructions. Do you imply where libva should be stored as well?

> As for default we assume:

Any way to change the default?

tests/ted.json
```diff
@@ -1,6 +1,6 @@
 {
-  "libmfx_folder": "$MFX_HOME/__cmake/intel64.make.release/__bin/release",
-  "samples_folder": "$MFX_HOME/__cmake/intel64.make.release/__bin/release",
+  "libmfx_folder": "$MFX_HOME/build/__bin/release",
```
dvrogozh, line 2: using `build` is actually a convention that user will build mediasdk as follows:
```
mkdir build && cd build
cmake <...> ..
```

Also, why do we actually need a path to non-installed medisdk binaries. So, what do we actually test: installed mediasdk version or non-installed one?

tests/ted/test.py
```diff
@@ -75,7 +75,8 @@ def __init__(self, fn, base, cfg, gold):
         
         extended_path = {
             'PATH': str(base.resolve() / subdir / 'bin') + os.pathsep + os.environ['PATH'],
-            'LD_LIBRARY_PATH': str(base.resolve() / subdir / 'bin' / 'lib'),
+            'LD_LIBRARY_PATH': (str(base.resolve() / subdir / 'bin' / 'lib') +
```
dvrogozh, line 78: `(str(base.resolve() / subdir / 'bin' / 'lib')`

I am not python guy, can someone explain what's going on here?

aobolensk: It's path concatenation. It resolves to:
```
/home/<user>/.../MediaSDK/tests/results/bin/lib
````

Comment (dvrogozh) [reply tests/ted/test.py:78]: imho, `bin/lib` path does not make sense to me.

Comment (aobolensk) [reply tests/ted/test.py:78]: msdk samples binaries are located in `MediaSDK/tests/results/bin`, libmfxhw.so is located in `MediaSDK/tests/results/bin/lib`

File: tests/ted.json (hunk 2 above)
Comment (aobolensk) [reply tests/ted.json:2]: We test non-installed binaries (test system copies them from "libmfx_folder").

File: tests/README.md (hunk 1 above)
Comment (aobolensk) [reply tests/README.md]: Fixed

File: tests/ted/test.py (hunk 3 above)
Comment (dvrogozh) [reply tests/ted/test.py:78]: At this point I am officially lost!

> Media SDK should be compiled and installed to `/opt/intel/mediasdk`

> We test non-installed binaries (test system copies them from "libmfx_folder").

> msdk samples binaries are located in MediaSDK/tests/results/bin, libmfxhw64.so is located in MediaSDK/tests/results/bin/lib

There are 3 (three, T-H-R-E-E) :) locations which you noted here and which are supposed to hold binaries which we test. Which one do we actually use? If we use all 3 then how all that actually works?

I can say how I envision the layout. There could be 2 approaches or both of them implemented at once:
1. We test non-installed binaries. In this case we need to know the location where the build had happened - i.e. the one called CMAKE_BINARY_DIR in cmake
2. We test installed binaries. In this case we need to know GnuInstallDirs layout, i.e. PREFIX, LIBDIR, BINDIR (and maybe few others).

IF we will have 1) we potentially can integrate these tests into `make test`. IF we will have 2) we have ability to verify the installed msdk version. Hence - both approaches are valuable and I would suggest to implement both.

Comment (aobolensk) [reply tests/ted/test.py:78]: We copy binaries from "libmfx_folder" defined in JSON to `MediaSDK/tests/results/bin` and use them for testing.
The first version implied that we need to have whole media stack installed in `/opt/intel/mediasdk` to make it work. I think I need to replace
```
Media SDK should be compiled and installed to `/opt/intel/mediasdk`
```
with
```
Media stack should be installed to `/opt/intel/mediasdk`
```

Comment (dvrogozh) [reply tests/ted/test.py:78]: libmfx_folder in the JSON file does not point to /opt/intel/mediasdk. It points to non-installed build location. This gives a confusion.

> Media stack should

In this case you need to list ingredients of the media stack.

Comment (aobolensk) [reply tests/ted/test.py:78]: I see. That's why I said:
> We test non-installed binaries (test system copies them from "libmfx_folder").
Do I need to point out that we test binaries from "libmfx_folder" in README?

File: tests/README.md [hunk 4]
```diff
@@ -0,0 +1,22 @@
...
+git-lfs
+python >= 3.6
+```
+Media stack ingredients (libva, libmfx, libigdgmm, libigfxcmrt) should be installed to `/opt/intel/mediasdk`
+
+Media driver should be compiled and installed to `/opt/intel/msdk_driver`
+
+#### Run tests:
+```sh
+export LD_LIBRARY_PATH=/opt/intel/mediasdk/lib64
```
dvrogozh, line 18: Here is how I would imagine run of the `ted.py` in most generic case:
```
export PREFIX=/path/to/something
export BINDIR=$PREFIX/bin
export LIBDIR=$PREFIX/lib

cmake -DCMAKE_INSTALL_PREFIX=$PREFIX -DCMAKE_INSTALL_LIBDIR=$LIBDIR ..
make && make install

export PATH=$BINDIR:$PATH
export LD_LIBRARY_PATH=$LIBDIR

python3 ted.py --gold  # to collect reference results
python3 ted.py
```

I don't quite get why do we need to runs of ted with and without --gold, but ok for now.

tests/README.md
```diff
@@ -0,0 +1,40 @@
+# Intel® Media SDK Open Source TEst Driver
```
dvrogozh, line 1: How about to simplify the title? I think "Test Driver" without capital E will be much better...

aobolensk: Done

Issue #2250, opened 2020-07-30
use prefix variable thru pkg-config files


dvrogozh:
https://github.com/Intel-Media-SDK/MediaSDK/blob/12ed7aea41f5ff583b86cbdca1e61947888da07d/_studio/mfx_lib/pkg-config.pc.cmake#L5

prefix variables is declared, but we don't use it to simplify the .pc file.

PR #2251, opened 2020-07-30
Preserve {prefix} in generated .pc

onabiull:
Fixes: #2250

### DIFF
--- a/_studio/mfx_lib/pkg-config.pc.cmake
+++ b/_studio/mfx_lib/pkg-config.pc.cmake
@@ -3,7 +3,8 @@ Description: Intel(R) Media SDK
 Version: @mfx_version_major@.@mfx_version_minor@
 
 prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@CMAKE_INSTALL_FULL_LIBDIR@
-includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+
 Libs: -L${libdir} -l@mfxlibname@
 Cflags: -I${includedir} -I${includedir}/mfx
--- a/api/mfx_dispatch/linux/pkg-config-shared.pc.cmake
+++ b/api/mfx_dispatch/linux/pkg-config-shared.pc.cmake
@@ -3,8 +3,9 @@ Description: Intel(R) Media SDK Dispatcher
 Version: @MFX_VERSION_MAJOR@.@MFX_VERSION_MINOR@
 
 prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@CMAKE_INSTALL_FULL_LIBDIR@
-includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+
 Libs: -L${libdir} -ldispatch_shared -lstdc++ -ldl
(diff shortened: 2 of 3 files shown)


_studio/mfx_lib/pkg-config.pc.cmake
```diff
@@ -3,7 +3,8 @@ Description: Intel(R) Media SDK
 Version: @mfx_version_major@.@mfx_version_minor@
 
 prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@CMAKE_INSTALL_FULL_LIBDIR@
-includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```
dvrogozh, line 7: I usually configure as follows:

`cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib64...`

Will your change produce:

`includedir=/usr//usr/lib64`

which is eventually wrong?

dvrogozh: You might need to have something like the following in cmake:

`string(REPLACE ${CMAKE_INSTALL_PREFIX} \${prefix} OUT_VAR ${CMAKE_INSTALL_FULL_LIBDIR})`

I am not sure I have put it in correct cmake language... just giving an idea...

onabiull: Yep it will. But given defaults listed at https://cmake.org/cmake/help/latest/command/install.html#command:install you do it wrong way. Correct way to specify that would be: `cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=lib64`

onabiull: Looks GNUInstallDirs module does exactly this internally.

dvrogozh: > Looks GNUInstallDirs module does exactly this internally.

`cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib64 ..`

You mean that even this will actually work?

onabiull: Let me try to explain my point. Line: `cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib64 ..` is malformed. Previously it did work, but from GNUInstallDirs module perspective CMAKE_INSTALL_LIBDIR_FULL derived from CMAKE_INSTALL_PREFIX and CMAKE_INSTALL_LIBDIR or specified explicitly, but not vice-versa. Also there are few special values like "/" in prefix handling.

dvrogozh: It is not malformed. Read https://cmake.org/cmake/help/v3.18/module/GNUInstallDirs.html carefully:

> CMAKE_INSTALL_FULL_<dir>

> The absolute path generated from the corresponding CMAKE_INSTALL_dir value. If the value is not already an absolute path, an absolute path is constructed typically by prepending the value of the CMAKE_INSTALL_PREFIX variable. However, there are some special cases as documented below.

In prev. documentation it was a little more clear from my perspective:

https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

> Each CMAKE_INSTALL_<dir> value may be passed to the DESTINATION options of install() commands for the corresponding file type. If the includer does not define a value the above-shown default will be used and the value will appear in the cache for editing by the user. Each CMAKE_INSTALL_FULL_<dir> value contains an absolute path constructed from the corresponding destination by prepending (if necessary) the value of CMAKE_INSTALL_PREFIX.

PR #2261, opened 2020-08-04
[smt] Fix possible memory leak in raw input case

aobolensk:


samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -3200,7 +3200,9 @@ mfxStatus CTranscodingPipeline::AllocFrames(mfxFrameAllocRequest *pRequest, bool
             surface->Data.MemId = pResponse->mids[i];
         }
 
-        (isDecAlloc) ? m_pSurfaceDecPool.push_back(surface) : m_pSurfaceEncPool.push_back(surface);
+        auto surface_ptr = surface.release();
+
+        (isDecAlloc) ? m_pSurfaceDecPool.push_back(surface_ptr) : m_pSurfaceEncPool.push_back(surface_ptr);
```
mgonchar, line 3205: Sorry for a little bit late review, but this is not complete fix yet. if ```push_back``` will throw an exception memory will still leak. I propose following update for this fix:
```
(isDecAlloc) ? m_pSurfaceDecPool.push_back(surface.get()) : m_pSurfaceEncPool.push_back(surface.get());

std::ignore = srface.release();
```

aobolensk: Thank you for this remark. I am going to address it.

PR #2267, opened 2020-08-05
[CM][AVC LA] Fix PictureWidth value

asmirno1:
This change makes it possible to use 4k streams with AVC LA
Now work with PictureWidth value is implemented same as PictureHeightMinusOne
Previously, overflow happens when dividing 4096 / 16, now it is fixed
CM kernels for all platforms were re-compiled.


dmitryermilov:
@asmirno1 , please fix CI

asmirno1:
Seems we have some problems on CI side

onabiull:
@eponomar, @dlobanox JFYI. Looks this change can't cause iHD load error.

Issue #1654, opened 2019-09-20
Pass through mode for VIDEO->SYSTEM with MFX_FOURCC_A2RGB10 format


sunxinpeng:
When I support the feature "convert P010 to 10bit RGBA" in ffmpeg qsv, I find that if downloading the output surface from driver in MSDK, it will use MSDK VPP path(although I just want to get the surface without any vpp conversion). In this step, MFXVideoVPP_Init will fail because of the format of input context is MFX_FOURCC_A2RGB10(this format is supported for output only). So if MSDK can provide a method which can directly download the a2r10g10b10 surface without passing the vpp path?

dmitryermilov:
It was fixed by https://github.com/Intel-Media-SDK/MediaSDK/pull/1664. 
Could you confirm?

sunxinpeng:
> It was fixed by #1664.
> Could you confirm?
I have tested with the newest code. The path with IO pattern of "sys_to_sys"(sample_vpp) has passed. But hwdownload(video_to_sys) still failed because A2RGB10 is supported as output only which cause vpp init failed in msdk.

dmitryermilov:
@AntonGrishin , can you please take a look?

AntonGrishin:
Hi @sunxinpeng,
I tested with sample_vpp and  all iopattern`s works fine for me:
```
./sample_vpp -o outa2rgb10.yuv -i stream_352x288_24_p010.yuv  \
-sw 352 -sh 288 -iopattern d3d_to_sys -lib hw -sbitshift 1 -scc p010 \
-dcc a2rgb10 -n 10
```

AntonGrishin:
Hi @sunxinpeng,
Can you please provide any updates?

sunxinpeng:
Hi @AntonGrishin 
Sorry for the delay reply. Maybe you misunderstood my question. When I do vpp in ffmpeg qsv, the hwdownload stage need MSDK download the surface from the driver. I track the code and I find that it use MFXVideoVPP to do this task, which mean that the input format and the output format are both A2RGB10 for MFXVideoVPP_Init at the hwdownload stage. But MSDK limits the input format of 
MFXVideoVPP_Init to not be A2RGB10, which will cause failure when hwdownload. So if there is another API path to do this hwdownload of if MSDK can support download A2RGB10 surface using MFXVideoVPP?

AntonGrishin:
Hi @sunxinpeng,

[...]

PR #2268, opened 2020-08-06
Fix passthru copy when input X2RGB10

AntonGrishin:
FFmpeg uses x2rgb10 format, which matches to VA X2R10G10B10 and
MFX a2rgb10 formats

cmd to check:
./ffmpeg -hwaccel qsv -v verbose -c:v hevc_qsv -i \
  ../p010.h265 -vf scale_qsv=format=x2rgb10,hwdownload,format=x2rgb10 \
  -vframes 1 out.yuv -y

fixes: #1654

### DIFF
--- a/_studio/shared/src/libmfx_allocator_vaapi.cpp
+++ b/_studio/shared/src/libmfx_allocator_vaapi.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019 Intel Corporation
+// Copyright (c) 2017-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


_studio/shared/src/libmfx_core.cpp
```diff
@@ -1508,6 +1507,12 @@ mfxStatus CoreDoSWFastCopy(mfxFrameSurface1 *pDst, mfxFrameSurface1 *pSrc, int c
             MFX_CHECK_STS(sts);
             break;
         }
+case MFX_FOURCC_A2RGB10:
+            roi.width *= 4;
+
+            sts = FastCopy::Copy(pDst->Data.B, dstPitch, pSrc->Data.B, srcPitch, roi, copyFlag);
```
mgonchar, line 1513: why we will receive ```nullptr``` in case of a2rgb? shouldn't all 4 pointer (```Data.A```, ```Data.R```, ```Data.G```, ```Data.B```) pointing to exactly same location?

mgonchar: Saying this I'm referring to our doc, which says "The application has to specify pointers to all color channels even for packed formats. For example, for YUY2 format the application has to specify Y, U and V pointers. For RGB32 – R, G, B and A pointers." @dmitryermilov @onabiull your thoughts?

AntonGrishin: @mgonchar this is good point, thanks!
If so, then it will be enough to add support for x2rgb10.

PR #2269, opened 2020-08-06
[ted] Move patches from infrastructure repo, update instructions

aobolensk:
- print test log to stdout if test is failed
- add `mpeg2d_conf.json` test
- remove function for creating reports
- update README
    - change path to samples folder
    - add necessary environment variables: `LIBVA_DRIVER_NAME` and `LIBVA_DRIVERS_PATH`


aobolensk:
> Please, add it to commit message! Such information should be there.

Done

PR #2272, opened 2020-08-10
Refine the member name of film grain flags

wangyan42164:

### DIFF
--- a/api/include/mfxstructures.h
+++ b/api/include/mfxstructures.h
@@ -2394,7 +2394,7 @@ MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxExtBuffer Header;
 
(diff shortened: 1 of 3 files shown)


api/include/mfxstructures.h
```diff
@@ -2394,7 +2394,7 @@ MFX_PACK_BEGIN_USUAL_STRUCT()
 typedef struct {
     mfxExtBuffer Header;
 
-    mfxU16 Flags;           /* FilmGrainFlags */
+    mfxU16 FilmGrainFlags;  /* FilmGrainFlags */
```
mgonchar, line 2397: @onabiull do you see any issues on renaming fields of structs which are still under ```VERSION_NEXT```?

onabiull: No :) Fields under that define considered volatile. They may be even dropped without any warning.

PR #2274, opened 2020-08-10
[CM][Core] Fix stride alignment

asmirno1:
This change fixes incorrect behavior of
CheckSurfaceContinuouslyAllocated()
in case of 2160x3840 stream size.
Change was tested with streams with
different resolutions - works as expected.

### DIFF
--- a/_studio/shared/src/cm_mem_copy.cpp
+++ b/_studio/shared/src/cm_mem_copy.cpp
@@ -3164,6 +3164,8 @@ mfxStatus CmCopyWrapper::CopyMirrorVideoToVideoMemory(void *pDst, void *pSrc, mf
 
 bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(const mfxFrameSurface1 &surf)
 {
+    mfxU32 stride_in_bytes = surf.Data.PitchLow + ((mfxU32)surf.Data.PitchHigh << 16);
+
     switch (surf.Info.FourCC)
     {
     // Packed formats like YUY2, UYVY, AYUV, Y416, Y210, Y410, A2RGB10, RGB565, RGB3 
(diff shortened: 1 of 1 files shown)


_studio/shared/src/cm_mem_copy.cpp
```diff
@@ -3176,7 +3176,7 @@ bool CmCopyWrapper::CheckSurfaceContinuouslyAllocated(const mfxFrameSurface1 &su
     case MFX_FOURCC_NV16:
         {
             mfxU32 stride_in_bytes = surf.Data.PitchLow + ((mfxU32)surf.Data.PitchHigh << 16);
-            size_t luma_size_in_bytes = stride_in_bytes * mfx::align2_value(surf.Info.Height);
+            size_t luma_size_in_bytes = stride_in_bytes * mfx::align2_value(surf.Info.Height, 32);
```
mgonchar, line 3179: These two lines of code are completely identical in all cases. Please pull them out of ```switch```

dmitryermilov: @asmirno1 , will it work if application makes 16 alignment?

dmitryermilov, line 3179: Will it works if user application makes 16 alignment?

dmitryermilov: @asmirno1 ?

asmirno1: I need some time to check it

dmitryermilov: Any news?

asmirno1: Done

asmirno1: Sorry for delay.
I checked - it works if user app makes 16 alignment.

dmitryermilov (approved):

PR #2275, opened 2020-08-10
[mpeg2e] Fix KW issue

pgribov:
Check potential division by zero

### DIFF
--- a/_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
+++ b/_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
@@ -822,7 +822,8 @@ namespace MPEG2EncoderHW
                 out->mfx.CodecLevel = MFX_LEVEL_UNKNOWN;
                 bWarning = true;
             }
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/mpeg2/src/mfx_mpeg2_encode_utils_hw.cpp
```diff
@@ -822,7 +822,8 @@ namespace MPEG2EncoderHW
                 out->mfx.CodecLevel = MFX_LEVEL_UNKNOWN;
                 bWarning = true;
             }
-            if (CorrectProfileLevelMpeg2(out->mfx.CodecProfile, out->mfx.CodecLevel,
+            if (out->mfx.FrameInfo.FrameRateExtD != 0 &&
```
mgonchar, line 825: why you put this condition here? You don't do any divisions on this level. I guess you need to put it inside ```CalculateUMCFramerate```

pgribov: I hurried, checking is already done inside the CalculateUMCFramerate function, this PR is not needed

PR #2279, opened 2020-08-11
[Tools][Tracer] Fix possible memory leaks

egorovdanil:


tools/tracer/tracer/tracer_linux.cpp
```diff
@@ -83,6 +83,7 @@ mfxStatus MFXInit(mfxIMPL impl, mfxVersion *ver, mfxSession *session)
         if(!loader->dlhandle)
             loader->dlhandle = dlopen(g_mfxlib_in_dir, RTLD_NOW|RTLD_LOCAL|RTLD_DEEPBIND);
         if (!loader->dlhandle){
+            free(loader);
```
mgonchar, line 86: can we switch to ```std::unique_ptr``` instead?

egorovdanil: I think not, because from the `loader` we get the `mfxSession` below, and this function is a replacement for the dispatcher `MFXInit` function which pass the session as a raw pointer.
```
*session = (mfxSession)loader;
```

PR #2284, opened 2020-08-13
Enable RKL Support

aidan2020sh:


_studio/shared/src/libmfx_core_vaapi.cpp
```diff
@@ -381,6 +381,13 @@ typedef struct {
     { 0x4906, MFX_HW_DG1, MFX_GT2 },
     { 0x4908, MFX_HW_DG1, MFX_GT2 },
 
+    /* RKL */
+    { 0x4C80, MFX_HW_RKL, MFX_GT1 },
```
FurongZhang, line 385: RKL only has GT1, no GT2?

aidan2020sh:
Hi ,
       Yes. here is the spec link: https://gfxspecs.intel.com/Predator/Home/Index/44501

Best regards,
Dan

_studio/shared/src/libmfx_core_vaapi.cpp
FurongZhang: Thanks for t he information.

PR #2276, opened 2020-08-10
API 1.34 - Keem Bay platform

onabiull:
Incorporate changes from 1.34 published in [oneVPL spec](https://spec.oneapi.com/versions/latest/elements/oneVPL/source/index.html).

PR #2294, opened 2020-08-18
Support AVC Level 6/6.1/6.2

guowillyy:
Support Decode AVC Level 6/6.1/6.2

Issue: MDP-47088

Change-Id: I42aa5f67b7480a264f86822d5e5e3cd750d8d090


onabiull (changes requested):
Please reflect new fields in tracer and mfx_static_asserts.cpp like in #2276

onabiull (approved):

PR #2297, opened 2020-08-19
[ted] Add av1d test and add support for target platforms specification

aobolensk:


tests/ted/platform.py
```diff
@@ -0,0 +1,66 @@
...
+def check_platform(current_device_ids, supported_platforms):
+    current_platform = '<unknown>'
+    for platform, ids in DEVICE_IDS.items():
+        if any(device_id in ids for device_id in current_device_ids):
+            current_platform = platform
+    return current_platform in {plat for plat in DEVICE_IDS.keys() if plat in supported_platforms}
+
+
+def get_current_device_ids():
+    proc = subprocess.Popen("lspci -n -m | grep 0300", shell=True, stdout=subprocess.PIPE)
```
dvrogozh, line 57: We need a new feature for TED: device id parameter on which to run tests (default: /dev/dri/renderD128). And we should not parse lspci output but look into device ID corresponding to /dev/dri/renderD128.

```
# cat /sys/class/drm/renderD128/device/device
0x193b
```

aobolensk: Done. Now we can specify device in `/dev/dri/renderD128` style in json file and its device id is checked without using lspci.

tests/ted/test.py
```diff
@@ -144,6 +144,12 @@ def run(self):
         return (total, passed, details)
 
     def generate_cases(self):
+        self.platforms = self.config.get('platforms', None)
+        device_name = Path(self.config.get('device', '/dev/dri/renderD128')).name
```
dvrogozh, line 148: I think it will be better if this will be command line option. Also, break change into few: av1 and device path are definitely different things. Do not forget to document new command line option.

aobolensk: Done

tests/README.md
```diff
@@ -32,6 +32,8 @@ export PATH=$BINDIR:$PATH
 export LD_LIBRARY_PATH=$LIBDIR
 
 python3 ted.py
+# or
+python3 ted.py --device /dev/dri/renderD128  # to specify device directly
```
dvrogozh, line 36: add comment to default path:

```
python3 ted.py # will execute on default device (<...>)
# or
python3 ted.py --device /dev/dri/renderD128  # to specify device directly
```

Also, what "default device" means? Do we explicitly specify /dev/dri/renderD128 or rely on apps device selection - this also needs to be reflected in the above comment...

tests/ted.py
```diff
@@ -39,6 +39,10 @@
         nargs='*',
         help='tests to run, if not specified all tests will be executed'
     )
+    parser.add_argument(
+        '--device', action='store', default='/dev/dri/renderD128',
+        help='provide device on which to run tests'
```
dvrogozh, line 44: again: what is the default?

tests/ted/platform.py
```diff
@@ -0,0 +1,60 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import os
+
+from . import objects
+
+DEVICE_IDS = {
```
dvrogozh, line 27: This part w/ device id selection belongs to av1 commit where you first start to use it... Please, move.

tests/ted/test.py
```diff
@@ -164,6 +171,8 @@ def generate_cases(self):
 
         for vals in itertools.product(*values):
             case = collections.OrderedDict(zip(keys, vals))
+            case['device'] = self.device
```
dvrogozh, line 174: well, I don't understand how you pass selected device to the apps... I expect to see -device parameter added to sample command lines and I can't find that...

aobolensk: This happens in `def other_options` run.py:108
So `-device /dev/dri/renderD128` flag appears in result parameters for sample.

tests/README.md
aobolensk: If we don't provide `--device` flag script automatically use `/dev/dri/renderD128` as default value. If we provide device explicitly we use provided one.
```
python3 ted.py # uses /dev/dri/renderD128 (implicitly, we set this device as default without any specific checks)
python3 ted.py --device /dev/dri/renderD129 # uses /dev/dri/renderD129
```

tests/ted.py
aobolensk: Default argument is used if it's not provided by user

tests/ted/test.py
dvrogozh: ok

tests/ted.py
dvrogozh: does it appear in the ted.py --help print out?

aobolensk: Yes, it does

aobolensk: I meant this option appears, but default value doesn't appear. Do I need to add it?

dvrogozh: yes, it will be helpful. Optimally I would like to see something like the following to be printed out:

`--device  provide device on which to run tests (default: /dev/dri/renderD128)`

tests/ted/platform.py
aobolensk: Done

tests/ted.py
aobolensk: Done

dvrogozh (dismissed):

dvrogozh:
@aobolensk: does av1 test pass on TGL? I think you were saying you saw some issues?

dmitryermilov:
@aobolensk , please fix the conflict

aobolensk:
> @aobolensk , please fix the conflict

Done

dvrogozh (approved):

dvrogozh:
Driver build failure is not related to this PR: https://github.com/intel/media-driver/issues/1049.

PR #2301, opened 2020-08-20
av1d: assign a VA surface ID to ref_frame_map

xhaihao:
The type of ref_frame_map is VASurfaceID, not FrameMemID, so we should
assign a VA surface ID to ref_frame_map
    
In addition, return VA_INVALID_SURFACE from
LinuxVideoAccelerator::GetSurfaceID() once catching an exception


wangyan-intel:
LGTM.

PR #2308, opened 2020-08-24
[Tutorials]: Added VPP sample code for scaling which set scaling mode…

FurongZhang:
… and interpolation method.

Signed-off-by: FurongZhang <furong.zhang@intel.com>


dmitryermilov (changes requested):
@FurongZhang , please update AllBuild.sln. Only this solution is used in Windows CI.  VS solutions from particular folders are not used in CI.

dmitryermilov:
> @FurongZhang , please update AllBuild.sln. Only this solution is used in Windows CI. VS solutions from particular folders are not used in CI.

Furong, any news?

PR #2310, opened 2020-08-25
Platform code name for JSL/EHL/TGL is only defined after API 1.31.

taste1981:
Platform code name for JSL/EHL/TGL needs to be protected against 1.31 API.

### DIFF
--- a/api/include/mfxcommon.h
+++ b/api/include/mfxcommon.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


onabiull (commented):
Sorry for slow reaction. I was hoping to finish my PR and refer to it, but still had not found enough time to finish it. We going to strip all version-specific checks and spare  only MFX_VERSION >= MFX_VERSION_NEXT. Could you please share if there are rationale to add that particular check?

taste1981:
> Sorry for slow reaction. I was hoping to finish my PR and refer to it, but still had not found enough time to finish it. We going to strip all version-specific checks and spare only MFX_VERSION >= MFX_VERSION_NEXT. Could you please share if there are rationale to add that particular check?

Without this, when we're referring to the platform code when writing our application, you may forget to add MFX_VERSION >= 1031 when using the TGL platform code. (which will result in a compilation error when developers is linking to an old MSDK.

dmitryermilov (approved):

PR #2323, opened 2020-08-27
turned off YUV  writing

dmitryermilov:


samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -1990,6 +1990,7 @@ mfxStatus CTranscodingPipeline::Surface2BS(ExtendedSurface* pSurf,mfxBitstreamWr
         MSDK_CHECK_ERR_NONE_STATUS(sts, MFX_ERR_ABORTED, "SyncOperation failed");
         pSurf->Syncp=0;
 
+	/*
```
dvrogozh, line 1993: emmm. motivation? it is broken? this is duplicated operation?

dmitryermilov: debug purpose. won't be merged as it is

PR #2332, opened 2020-09-03
Fix Linux 32-bits encoded timestamp limitation - 2038K Bug

jeremy-compostella:
Historically time_t has been encoded on Linux as 32-bits signed
integer and with Linux 32 bits <5.6 it is still encoded as a 32-bits
signed integer. As a consequence beyond January 2038 gettimeofday()
will start returning a negative timestamp and this negative timestamp
will be supplied to pthread_cond_timedwait() by MediaSDK. The Linux
kernel will detects the invalid timestamp and return -EINVAL making
MediaSDK unusable.

This patch makes use of the 'clock_gettime(CLOCK_MONOTONIC, ...)' to
get current time since boot and configure the condition variable to
use the same clock.

Since this patch only impact the timestamped used for the
pthread_cond_timedwait() calls it has no impact in the other
components.

Signed-off-by: Jeremy Compostella <jeremy.compostella@intel.com>

### DIFF
--- a/_studio/shared/umc/core/vm/src/vm_event_linux32.c
+++ b/_studio/shared/umc/core/vm/src/vm_event_linux32.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Intel Corporation
+// Copyright (c) 2017-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -59,6 +59,7 @@ int32_t vm_event_is_valid(vm_event *event)
 /* Init an event. Event is created unset. return 1 if success */
 vm_status vm_event_init(vm_event *event, int32_t manual, int32_t state)
 {
+    pthread_condattr_t cond_attr;
     int res = 0;
 
     /* check error(s) */
@@ -67,19 +68,24 @@ vm_status vm_event_init(vm_event *event, int32_t manual, int32_t state)
 
     event->manual = manual;
     event->state = state ? 1 : 0;
-    res = pthread_cond_init(&event->cond, 0);
+    res = pthread_condattr_init(&cond_attr);
     if (!res)
     {
-        res = pthread_mutex_init(&event->mutex, 0);
-        if (res)
+        pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC);
+        res = pthread_cond_init(&event->cond, &cond_attr);
+        if (!res)
         {
-            int cres = pthread_cond_destroy(&event->cond);
-            assert(!cres); // we experienced undefined behavior
-            (void)cres;
-            vm_event_set_invalid_internal(event);
+            res = pthread_mutex_init(&event->mutex, 0);
+            if (res)
+            {
+                int cres = pthread_cond_destroy(&event->cond);
+                assert(!cres); // we experienced undefined behavior
+                (void)cres;
+                vm_event_set_invalid_internal(event);
+            }
         }
+        pthread_condattr_destroy(&cond_attr);
     }
-
     return (res)? VM_OPERATION_FAILED: VM_OK;
 
 } /* vm_status vm_event_init(vm_event *event, int32_t manual, int32_t state) */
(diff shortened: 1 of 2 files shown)


jeremy-compostella:
Any review for this patch ?

_studio/shared/umc/core/vm/src/vm_event_linux32.c
```diff
@@ -67,19 +68,24 @@ vm_status vm_event_init(vm_event *event, int32_t manual, int32_t state)
...
     event->manual = manual;
     event->state = state ? 1 : 0;
-    res = pthread_cond_init(&event->cond, 0);
+    res = pthread_condattr_init(&cond_attr);
     if (!res)
     {
-        res = pthread_mutex_init(&event->mutex, 0);
-        if (res)
+        pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC);
+        res = pthread_cond_init(&event->cond, 0);
```
onabiull, line 75: Didn't you miss `res = pthread_cond_init(&event->cond, &cond_attr);`  here?

_studio/shared/umc/core/vm/src/vm_semaphore_linux32.c
```diff
@@ -69,44 +70,55 @@ vm_status vm_semaphore_init(vm_semaphore *sem, int32_t init_count)
 
     sem->count = init_count;
     sem->max_count = 1;
-    res = pthread_cond_init(&sem->cond, 0);
+    res = pthread_condattr_init(&cond_attr);
     if (!res)
     {
-        res = pthread_mutex_init(&sem->mutex, 0);
-        if (res)
+        res = pthread_cond_init(&sem->cond, 0);
```
onabiull, line 76: And here?

onabiull (changes requested):
Thank you for  for contribution! I'd definitely like to see this code running in 2038. 

Please modify copyright year in the file header to 2020, to unblock CI checks.

dvrogozh (changes requested):
Instead of fixing all this old vm/umc code we just need to drop it and rewrite couple of places it still being used to c++ std::mutex and other standard primitives.

_studio/shared/umc/core/vm/src/vm_semaphore_linux32.c
jeremy-compostella: Done. I missed this when I ported from my test environment.

_studio/shared/umc/core/vm/src/vm_event_linux32.c
jeremy-compostella: Done. I missed this when I ported from my test environment.

jeremy-compostella:
> Instead of fixing all this old vm/umc code we just need to drop it and rewrite couple of places it still being used to c++ std::mutex and other standard primitives.

Indeed, using c++ std::condition_variable would be more appropriate.  Note that the current implementation (https://github.com/llvm/llvm-project/blob/master/libcxx/src/condition_variable.cpp#L49) has the same issue and will not run properly beyond 2038 if compiled in 32-bits on Linux <5.6 kernel.

_studio/shared/umc/core/vm/src/vm_event_linux32.c
jeremy-compostella: Yes indeed. I missed this when I ported from my test environment (Android).

_studio/shared/umc/core/vm/src/vm_semaphore_linux32.c
jeremy-compostella: Yes indeed. I missed this when I ported from my test environment (Android).

dvrogozh:
>> Instead of fixing all this old vm/umc code we just need to drop it and rewrite couple of places it still being used to c++ std::mutex and other standard primitives.

> Indeed, using c++ std::condition_variable would be more appropriate. Note that the current implementation (https://github.com/llvm/llvm-project/blob/master/libcxx/src/condition_variable.cpp#L49) has the same issue and will not run properly beyond 2038 if compiled in 32-bits on Linux <5.6 kernel. So I don't think it can be considered a good solution for the moment.

Today is 2020. Maintenance period even for enterprise OSes is maximum 10 years. By 2038 this will be different llvm, different kernel and different. I strongly believe llvm/kernel/os will be patched by that time.

jeremy-compostella:
> > > Instead of fixing all this old vm/umc code we just need to drop it and rewrite couple of places it still being used to c++ std::mutex and other standard primitives.
> 
> > Indeed, using c++ std::condition_variable would be more appropriate. Note that the current implementation (https://github.com/llvm/llvm-project/blob/master/libcxx/src/condition_variable.cpp#L49) has the same issue and will not run properly beyond 2038 if compiled in 32-bits on Linux <5.6 kernel. So I don't think it can be considered a good solution for the moment.
> 
> Today is 2020. Maintenance period even for enterprise OSes is maximum 10 years. By 2038 this will be different llvm, different kernel and different. 

Well I can tell you that  some manufacturers are currently shipping products with OSes life expectancy of around 20 to 25 years. They include MediaSDK. And these products will not receive any major update (except maybe security patches). Also, in my sense, the maintenance period is mostly about handling security problems which unforeseeable as of today.  This is very different this is an issue we see as of today by setting the system date beyond 2038 and observing MediaSDK stop operating properly.  

> I strongly believe llvm/kernel/os will be patched by that time.

As you may have deduced from my commit message and comment above, this is already fix with kernel change 5.6 but it obviously require to align the libc as this long due fix is changing `time_t` a from 32-bits to a 64-bits integer. 

I must say that I don't understand this push-back. A very simple with no impact patch is made available to the MediaSDK developer team and instead of performing a good faith patch review we get into this non-sense discussion. I understand that you may want to abandon this pthread interface and move to another interface and this is perfectly fine by me but to my know knowledge this is not done yet and it does not change the fact that there is currently a bug preventing to include the MediaSDK library in long life-expectency product with a current Linux based Operating Systems. This patch is not making any dangerous changes neither is a hack.  It just moving from the absolute clock to the boot relative clock to allow the code to smoothly work longer.

onabiull (approved):

onabiull:
I believe in current form this patch may be merged and would bring more goodness than harm :).  In a long term we'll proceed with a refactoring: #2344 .

PR #2334, opened 2020-09-04
Add visual traces

aobolensk:


_studio/mfx_lib/shared/src/libmfxsw_async.cpp
```diff
@@ -26,6 +26,8 @@
 mfxStatus MFXVideoCORE_SyncOperation(mfxSession session, mfxSyncPoint syncp, mfxU32 wait)
 {
     MFX_AUTO_LTRACE(MFX_TRACE_LEVEL_API, "MFX_SyncOperation");
+    MFX_TRACE_CHROME(SyncOperation, sync);
```
dvrogozh, line 29: I don't quite see how MFX_TRACE_CHROME differs from MFX_AUTO_LTRACE... Hence: why new macro?

_studio/shared/include/mfx_trace.h
```diff
@@ -404,6 +404,33 @@ mfxTraceU32 MFXTrace_EndTask(mfxTraceStaticHandle *static_handle,
...
+    std::vector <std::pair <std::string, std::string> > info;
+public:
+    void add_info(const std::string &key, const std::string &value);
+    void add_info(const std::string &key, void *value);
+    ChromeTraceEvent(const char *name, const char *tid);
+    ~ChromeTraceEvent();
+};
+
+#define MFX_TRACE_CHROME(name, type) \
+    ChromeTraceEvent chrome_trace_event_##name(#name, #type)
```
dvrogozh, line 425: well, all that concept is already implemented in mfx_trace, but instead of integrating into existing functionality being a normal trace module you implement yet-another-one tracing module with separate settings, support, etc. I am against that. Basically, if you don't like something in the existing trace module - this needs to be fixed per some plan rather than implementing something new.

dvrogozh (changes requested):

aobolensk:
Got rid of `MFX_TRACE_CHROME` macro, but `MFX_TRACE_CHROME_ADD_INFO` is still used

_studio/shared/include/mfx_trace.h
```diff
@@ -432,6 +473,17 @@ class MFXTraceTask
 // auto tracing of the functions
 
 #ifdef MFX_TRACE_ENABLE
+#if defined(MFX_TRACE_ENABLE_CHROME) && (defined(LINUX32))
+    // Chrome traces
+    #define _MFX_AUTO_LTRACE_(_level, _task_name, _bCreateID)                      \
```
dvrogozh, line 478: You still pretend to integrate into trace system, not really integrating...

_studio/mfx_lib/shared/src/libmfxsw_encode.cpp
```diff
@@ -844,6 +844,7 @@ mfxStatus MFXVideoENCODE_EncodeFrameAsync(mfxSession session, mfxEncodeCtrl *ctr
     mfxStatus mfxRes;
 
     MFX_AUTO_LTRACE_WITHID(MFX_TRACE_LEVEL_API, "MFX_EncodeFrameAsync");
+    MFX_TRACE_CHROME_ADD_INFO("session", session);
```
dvrogozh, line 847: It seems you are trying to associate some name with variable, in the trivial case same name as variable name. Question: why? I believe you are trying to draw some dependency arrows in chrome.... So, do you look for variable names or variable values to draw them?

What's the reason why you can't use existing MFX_LTRACE_P/I/etc.? That's because there are too many of them or there is some other reason?

dvrogozh (changes requested):

PR #2335, opened 2020-09-07
Remove UMC_VA_LINUX macro for av1d

wangyan42164:


wangyan42164:
@dmitryermilov @onabiull This patch should only for open repo. We should not back port it. Thanks.

onabiull (approved):

PR #2345, opened 2020-09-09
fix issue of writing output yuv for p010 format surface

Jexu:
Signed-off-by: Xu, Zhengguo <zhengguo.xu@intel.com>

### DIFF
--- a/samples/sample_common/src/sample_utils.cpp
+++ b/samples/sample_common/src/sample_utils.cpp
@@ -1093,6 +1093,14 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
     break;
 #endif
     case MFX_FOURCC_P010:
+        for (i = 0; i < pInfo.CropH; i++)
+        {
+            mfxU16* shortPtr = (mfxU16*)(pData.Y + (pInfo.CropY * pData.Pitch + pInfo.CropX) + i * pData.Pitch);
+            MSDK_CHECK_NOT_EQUAL(
+                fwrite(shortPtr, 1, (mfxU32)pInfo.CropW * 2, dstFile),
+                (mfxU32)pInfo.CropW * 2, MFX_ERR_UNDEFINED_BEHAVIOR);
+        }
+        break;
 #if (MFX_VERSION >= 1031)
     case MFX_FOURCC_P016:
 #endif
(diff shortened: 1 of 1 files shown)


samples/sample_common/src/sample_utils.cpp
```diff
@@ -1093,6 +1093,14 @@ mfxStatus CSmplYUVWriter::WriteNextFrame(mfxFrameSurface1 *pSurface)
     break;
 #endif
     case MFX_FOURCC_P010:
+        for (i = 0; i < pInfo.CropH; i++)
+        {
```
AntonGrishin, line 1097: Why do we need this? As I see it`s just a copy of the code below without the "shift" part.
But what if the stream with shifted FourCC? 

Can you please describe the issue more clearly?

Jexu: Hi @AntonGrishin 
i am not very sure of so called 'shifted FourCC', but i get tried several stream that runs into 'shift' part and only get green output yuv. In fact, the driver has helped calculate the offset for Y, U, V, then msdk may have no need to do such shift.

AntonGrishin: What kind of scenario did you use? Cmd?

AFAIR, all of our samples have special options to specify shift https://github.com/Intel-Media-SDK/MediaSDK/tree/master/doc/samples (-msb10 for sample encode/ -sbitshift 0|1 for sample_vpp)

Jexu: @AntonGrishin 
I am just run hevc decode like:
sample_decode -i hevc.bin  h265 -async 1 -n 1 -hw -o out.yuv. i assume it has no shift, but it will run into shift part when writing yuv.

dmitryermilov: > i am not very sure of so called 'shifted FourCC',

Please see https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#mfxframeinfo. 
Idea is that 10,12 bits formats are stored in two bytes. So you can located bits in lowers or highest bits.

AntonGrishin: Shift value gets from the stream (parsed from the header). Are you sure that you use correct shift values in your yuv-viewer application? Even with an incorrect shift, usually, you should see "edges" on the green frame.
Based on your description, I see no problem in code.

Jexu: for hevc codec, i see 'request.Info.Shift' and 'par->mfx.FrameInfo.Shift' in https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp are always be set, is it correct?

dmitryermilov: Yes, it's limitation of HW decoders. SW components (not available at GitHub) may behave differently.

PR #2362, opened 2020-09-16
[MCTF] Surface cleanup for no filter case.

spossos:
When filter switches from some strength to zero,
the target surface is kept and not returned.

Issue: -
Tests: Manual testing


spossos:
This pull request is already covered here: https://github.com/Intel-Media-SDK/MediaSDK/pull/2366

PR #2365, opened 2020-09-17
[VPP][MCTF] Not inEnc-pipeline zero strength fix

spossos:
When strength zero is selected for MCTF in VPP, the
denoising path was not being executed after the fourth
frame because settings from the inEnc-pipeline operation
which is not active were disabling it.

Issue: Regression detected
Tests: vpp_mctf_denoise (all test passed)
       Manual testing for in-encoding pipeline


dvrogozh (changes requested):
Cleanup history, please. Single change which fixes the issue.

spossos:
Sure, I will make a new commit.

PR #2372, opened 2020-09-21
[av1d] Add handling UMC_ERR_INVALID_PARAMS in SubmitFrame

aobolensk:


_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -958,6 +958,10 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
                 sts = MFX_ERR_MORE_DATA;
             }
 
+            if (umcRes == UMC::UMC_ERR_INVALID_PARAMS)
```
dvrogozh, line 961: I think we just need to add handling of unknown error here. And we need to figure out why we get unknown error from GetFrame call. Basically such unknown errors from driver should have been mapped to device failed error.

dvrogozh (changes requested):

dvrogozh:
Also, please, provide issue description you are working on. It's unclear to the audience why you are giving such PR.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Unhandled driver errors are mapped to MFX_ERR_DEVICE_FAILED. And rest of the errors are mapped to MFX_ERR_UNKNOWN.

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
```diff
@@ -544,7 +544,7 @@ namespace UMC_AV1_DECODER
 
         UMC::Status umcRes = SubmitTiles(*pCurrFrame, firstSubmission);
         if (umcRes != UMC::UMC_OK)
-            return umcRes;
+            return UMC::UMC_ERR_FAILED;
```
dvrogozh, line 547: This might break something. Can you go inside SubmitTiles and find where you get the error?

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -988,6 +988,16 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
                 return MFX_ERR_NONE;
             }
 
+            if (umcRes == UMC::UMC_ERR_FAILED)
```
dvrogozh, line 991: I would not allow code at line 980 an below execute on the error. I would move your code before line 972 and actually reused error handling at lines 974-977.

dvrogozh (changes requested):

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp
aobolensk: This call returns error from driver:
https://github.com/Intel-Media-SDK/MediaSDK/blob/49f1e2907683c76f1043b2672e223be09af5f178/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder_va.cpp#L91-L93
Added mapping to UMC_ERR_FAILED.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Done

aobolensk:
> Also, please, provide issue description you are working on. It's unclear to the audience why you are giving such PR.

Not all errors from driver are handled properly in this piece of code. If we ignore an error it can lead to memory corruption on msdk or sample side. This PR is intended to fix these kind of errors.

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder_va.cpp
```diff
@@ -89,6 +89,8 @@ namespace UMC_AV1_DECODER
             packer->EndFrame();
 
         sts = va->Execute();
+        if (sts != UMC::UMC_OK)
```
dvrogozh, line 92: aha. So, here it is where we missed an error from media driver. Ok, that's a gap. However, why do you change the status here instead of just 'return sts'? /which status we actually get from Execute?/ And  you could just return UMC::UMC_ERR_DEVICE_FAILED to avoid one more if in the above code...

dvrogozh (changes requested):

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder_va.cpp
aobolensk: We receive UMC::UMC_ERR_INVALID_PARAMS from driver

aobolensk: UMC::UMC_ERR_DEVICE_FAILED is located here:
https://github.com/Intel-Media-SDK/MediaSDK/blob/07d6b5f50a2839f570bfb856179083d7592283fd/_studio/shared/umc/core/umc/include/umc_va_base.h#L182-L187
And it is used in switch-case in mfx_av1_dec_decode.cpp under `#if defined(MFX_VA)`

dvrogozh: @onabiull : how about promoting these 3 guys to normal UMC status codes?

@aobolensk : the whole this file seems to be VA specific. Can we just use UMC::UMC_ERR_DEVICE_FAILED here?

aobolensk: Replaced UMC::UMC_ERR_FAILED with UMC::UMC_ERR_DEVICE_FAILED

dvrogozh (approved):

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -988,6 +988,11 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
                 return MFX_ERR_NONE;
             }
 
+            if (sts == MFX_ERR_NONE && umcRes != UMC::UMC_OK)
+            {
+                sts = MFX_ERR_UNKNOWN;
```
dmitry-gurulev, line 993: maybe i missed smth but here u just set error status and continue `for` loop, so i see no side effects

dmitry-gurulev (dismissed):

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Exit from loop depends on this variable:
https://github.com/Intel-Media-SDK/MediaSDK/blob/6b6cc790c08426b4111094fa311dd34e2e4541fb/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L897
It is equal to `umcRes` (or it can be set to `UMC::UMC_ERR_NOT_ENOUGH_BUFFER` during error handling).
Anyway, if it is not `UMC_OK` (so `sts` is not `MFX_ERR_NONE`) we are going to exit the loop. According to this:
https://github.com/Intel-Media-SDK/MediaSDK/blob/6b6cc790c08426b4111094fa311dd34e2e4541fb/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L996-L997

dvrogozh: @dmitry-gurulev : was your question answered?

guowillyy (approved):

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
dvrogozh: Well... I looked for the second time into the change... I think @dmitry-gurulev is actually right... This is incorrect fix.

@aobolensk : can you, please, rewrite this whole code block. We need a clear sequence like:

```
if (umcRes == UMC::UMC_OK) {do something}
if (umcRes == UMC::UMC_ERR_INVALID_STREAM) {do something}
else if (UMC::UMC_NTF_NEW_RESOLUTION) {do something}
else { do something}  // <<<<< your are actually should add this.
```

or you can do the same with switch if this will make more sense....

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -988,6 +988,11 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
                 return MFX_ERR_NONE;
             }
 
+            if (sts == MFX_ERR_NONE && umcRes != UMC::UMC_OK)
```
dvrogozh, line 991: Actually, it seems this can actually be easily satisfied with for example UMC::UMC_ERR_NOT_ENOUGH_DATA....

dvrogozh (changes requested):

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: In fact, no, in case of `UMC::UMC_ERR_NOT_ENOUGH_DATA` sts is set to `MFX_ERR_MORE_DATA` so this condition will not be satisfied.
https://github.com/Intel-Media-SDK/MediaSDK/blob/6b6cc790c08426b4111094fa311dd34e2e4541fb/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L954-L959

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Replaced it with switch. Errors and warnings that change umcRes itself are handled separately.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -924,38 +946,25 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
...
-            if (umcRes == UMC::UMC_ERR_NOT_ENOUGH_DATA || umcRes == UMC::UMC_ERR_SYNC)
-            {
-                if (!bs || bs->DataFlag == MFX_BITSTREAM_EOS)
-                    force = true;
-                sts = MFX_ERR_MORE_DATA;
+                default:
+                    if (sts < 0)
+                        sts = MFX_ERR_UNKNOWN;
+                    break;
             }
```
dvrogozh, line 970: github does not show lines below, but they are:
```
if (umcRes != UMC::UMC_NTF_NEW_RESOLUTION)
{...}
```

This condition does not make sense to me since it is always true, isn't it? (we have return checking this status in switch)

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -906,10 +906,32 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
                  m_video_par.AsyncDepth = static_cast<mfxU16>(vp.async_depth);
             }
 
-            switch (umcRes)
+            if (umcRes == UMC::UMC_WRN_REPOSITION_INPROGRESS)
```
dvrogozh, line 909: Ok.... So, we check umcRes here for UMC::UMC_WRN_REPOSITION_INPROGRESS while we entered condition on the same status right above... and we overrode the status... Question: do we expect that m_decoder->GetInfo will return same status? why? Honestly, I am not sure which status all the code below actually handles... which is not good...

aobolensk: Seems like this should never happen, according to GetInfo implementation for AV1.
https://github.com/Intel-Media-SDK/MediaSDK/blob/52c71f460abea29fce1dea69df2b67c8d869a80e/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder.cpp#L141-L151
Because it returns different statuses.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Fair enough. Removed unnecessary condition.

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -924,45 +933,29 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
...
 
-            if (umcRes == UMC::UMC_OK && m_allocator->FindFreeSurface() == -1)
-            {
-                sts = MFX_ERR_MORE_SURFACE;
-                umcFrameRes = UMC::UMC_ERR_NOT_ENOUGH_BUFFER;
-            }
+                case UMC::UMC_ERR_NOT_ENOUGH_BUFFER:
+                case UMC::UMC_WRN_INFO_NOT_READY:
+                case UMC::UMC_ERR_NEED_FORCE_OUTPUT:
+                    force = (umcRes == UMC::UMC_ERR_NEED_FORCE_OUTPUT);
```
dvrogozh, line 940: This force variable is not actually used anywhere.... delete...

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
```diff
@@ -924,45 +933,29 @@ mfxStatus VideoDECODEAV1::SubmitFrame(mfxBitstream* bs, mfxFrameSurface1* surfac
...
+                    break;
 
-            if (umcRes == UMC::UMC_ERR_NOT_ENOUGH_DATA || umcRes == UMC::UMC_ERR_SYNC)
-            {
-                if (!bs || bs->DataFlag == MFX_BITSTREAM_EOS)
-                    force = true;
-                sts = MFX_ERR_MORE_DATA;
+                default:
+                    if (sts < 0)
+                        sts = MFX_ERR_UNKNOWN;
```
dmitry-gurulev, line 953: MFX_ERR_UNDEFINED_BEHAVIOR

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Done

_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp
aobolensk: Done

dvrogozh (approved):

guowillyy (approved):

dvrogozh:
I believe we can merge in this change.

onabiull (approved):

PR #2375, opened 2020-09-23
[hevce] Fix build by ICC

Alexandr-Konovalov:
Workaround for an internal compiler error.

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_data.h
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_data.h
@@ -900,10 +900,10 @@ namespace Base
         PackedData VPS;
         PackedData SPS;
         PackedData PPS;
-        PackedData AUD[3];
         PackedData PrefixSEI;
         PackedData SuffixSEI;
         std::vector<PackedData> SSH;
+        PackedData AUD[3];
     };
 
     using MfxEncodeHW::DDIExecParam;


Alexandr-Konovalov:
I'm not sure is it valuable or not. Please disregard if you think if it is pointless.

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_data.h
```diff
@@ -900,10 +900,10 @@ namespace Base
         PackedData VPS;
         PackedData SPS;
         PackedData PPS;
-        PackedData AUD[3];
         PackedData PrefixSEI;
         PackedData SuffixSEI;
         std::vector<PackedData> SSH;
+        PackedData AUD[3];
```
dvrogozh, line 906: Please, provide more details in commit message:
1. Which ICC compiler version you've tried?
2. What was the warning/error you got from the compiler?

Also, I think you better post some comment here in the code like:
```
PackedData AUD[3]; // ICC: this declaration should be here because....
```

Finally, we need a bug against ICC compiler for that.

dvrogozh (changes requested):

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_data.h
dmitryermilov: @Alexandr-Konovalov ?

PR #2376, opened 2020-09-23
set the initial value of frame_order to 0

chuanli1:


daleksan (changes requested):
please change commit message to the "av1d: set the initial value for frame_order" or something like that. Because "update" is not clear.

daleksan (dismissed):

guowillyy (approved):

onabiull (approved):

PR #2379, opened 2020-09-24
increase decode picture buffer

weizhu-intel:
For current AV1's DPB implementation, after decoding a frame, Output
frame will be saved in a temporary variable for DPB update when next
 frame coming. Need to give one more buffer to save temporary frame
for DPB update.

For this stream, it has 8 reference frames in DPB, if we only
 allocate 8 + 1 + async_depth buffers, DPB is not update immediately
so no output, in app side, it can't get any more free buffer to
feed next frame, then deadlock.

Issue: MDP-63996
Tests: ./sample_decode -hw av1 -i ./test_stream_av1.ivf -o 0001.yuv
        -async 1  -vaapi

Change-Id: I10ab4b697cd392b76068e4e83935688fa1279ee8

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
@@ -472,7 +472,7 @@ namespace MFX_VPX_Utility
         // Increase minimum number by one
         // E.g., decoder unlocks references in sync part (NOT async), so in order to free some surface
         // application need an additional surface to call DecodeFrameAsync()
-        p_request->NumFrameMin += 1;
+        p_request->NumFrameMin += 2;
 
         p_request->NumFrameSuggested = p_request->NumFrameMin;
 
--- a/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder_va.cpp
+++ b/_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder_va.cpp
@@ -57,7 +57,7 @@ namespace UMC_AV1_DECODER
         packer.reset(Packer::CreatePacker(va));
 
         uint32_t const dpb_size =
-            params.async_depth + TOTAL_REFS;
+            params.async_depth + TOTAL_REFS + 2;
         SetDPBSize(dpb_size);
         SetRefSize(TOTAL_REFS);
         return UMC::UMC_OK;


daleksan (changes requested):
please remove one space

_studio/shared/umc/codec/av1_dec/src/umc_av1_decoder_va.cpp
```diff
@@ -57,7 +57,7 @@ namespace UMC_AV1_DECODER
         packer.reset(Packer::CreatePacker(va));
 
         uint32_t const dpb_size =
-            params.async_depth + TOTAL_REFS;
+            params.async_depth  + TOTAL_REFS + 2;
```
daleksan, line 60: please remove one space

daleksan (changes requested):

dmitry-gurulev (changes requested):
In the formula [8 + 1 + async_depth buffers] this '+1' is exactly additional frame we need to hold a the frame until the next frame is coming. I still don't understand why we need yet another one frame. Could u please describe DPB state (which are referenced, which are displayed, which are being decoded asynchronously etc.) in time when u think u need this additional frame?

weizhu-intel:
Thanks Dmitry.G.
It is DPB update implementation related. Pls check GetFrame in umc_av1_decoder.cpp,   It firstly check whether pPrevFrame hold frame,  DPBupdate if have,  then  decode current frame,  and save current frame in pPrevFrame if completed.   When next frame come, it will check pPrevFrame(current frame saved here)  and update DPB again.
 
 Since it did not update DPB immediately(updated when next frame come). Another one frame is allocated for pPrevFrame.    If we don't give another one frame,  when DPB is full,  decode current frame will not update DPB immediate, since no output, APP side will have no chance to get a free buffer to decode next frame.

_studio/mfx_lib/shared/src/mfx_vpx_dec_common.cpp
```diff
@@ -472,7 +472,7 @@ namespace MFX_VPX_Utility
         // Increase minimum number by one
         // E.g., decoder unlocks references in sync part (NOT async), so in order to free some surface
         // application need an additional surface to call DecodeFrameAsync()
-        p_request->NumFrameMin += 1;
+        p_request->NumFrameMin += 2;
```
dvrogozh, line 475: I have a concern for system memory path where we have p_request->NumFrameSuggested != p_request->NumFrameMin:

https://github.com/Intel-Media-SDK/MediaSDK/blob/52c71f460abea29fce1dea69df2b67c8d869a80e/_studio/mfx_lib/decode/av1/src/mfx_av1_dec_decode.cpp#L554

I suspect this is a bug and we need to have these 2 values equal to each other. Can you, please, review this place.

dvrogozh (changes requested):

weizhu-intel:
Dmitry.R,
With system memory,  following code is used to  allocate output buffer(system memory), not DPB buffer(video memory)- This is what I modified.  Output buffer and DPB buffer's number could be different since there was a memory copy from DPB buffer to system memory when decoder finished a frame.

request->NumFrameSuggested = request->NumFrameMin + (par->AsyncDepth ? par->AsyncDepth : MFX_AUTO_ASYNC_DEPTH_VALUE); 

Thanks
wayne zhu

dvrogozh (approved):
That's WA. Let's track dpb management improvement here: #2382

dmitry-gurulev:
@weizhu-intel,

> It is DPB update implementation related. Pls check GetFrame in umc_av1_decoder.cpp, It firstly check whether pPrevFrame hold frame, DPBupdate if have, then decode current frame, and save current frame in pPrevFrame if completed. When next frame come, it will check pPrevFrame(current frame saved here) and update DPB again.

All u mentioned is a implementation related things, actuall I don't see any reasons for this change either from AV1 specification requirements or public API (architectural) poin of view.

dmitry-gurulev (approved):
Conditionally approved as WA, futher work is tracked by https://github.com/Intel-Media-SDK/MediaSDK/issues/2382

daleksan (approved):

PR #2386, opened 2020-09-29
[HEVCe] Correct caps report values

JeremyShang:
Correct some caps report values with LowPower on

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
@@ -54,10 +54,11 @@ void VAPacker::Query1WithCaps(const FeatureBlocks& /*blocks*/, TPushQ1 Push)
         caps.MbQpDataSupport            = 1;
         caps.TUSupport                  = 73;
         caps.SliceStructure             = 4;
+        caps.ParallelBRC                = bLP ? 0 : 1;
 
         caps.MaxEncodedBitDepth |= (!caps.BitDepth8Only);
         caps.YUV444ReconSupport |= (!caps.Color420Only && IsOn(par.mfx.LowPower));
-        caps.YUV422ReconSupport |= (!caps.Color420Only && IsOff(par.mfx.LowPower));
+        caps.YUV422ReconSupport &= (!caps.Color420Only && !IsOn(par.mfx.LowPower));
 
         return MFX_ERR_NONE;
     });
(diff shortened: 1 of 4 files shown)


vilichev (approved):

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
```diff
@@ -53,10 +53,11 @@ void VAPacker::Query1WithCaps(const FeatureBlocks& /*blocks*/, TPushQ1 Push)
         caps.MbQpDataSupport            = 1;
         caps.TUSupport                  = 73;
         caps.SliceStructure             = 4;
+        caps.ParallelBRC                = bLP ? 0 : 1;
 
         caps.MaxEncodedBitDepth |= (!caps.BitDepth8Only);
         caps.YUV444ReconSupport |= (!caps.Color420Only && IsOn(par.mfx.LowPower));
-        caps.YUV422ReconSupport |= (!caps.Color420Only && IsOff(par.mfx.LowPower));
+        caps.YUV422ReconSupport &= (!caps.Color420Only && !IsOn(par.mfx.LowPower));
```
dmitryermilov, line 60: @JeremyShang , @vilichev , could you explain:
1) Why do we need  |= and &= at all here? I don't follow the logic.
2) what is the difference between IsOff(par.mfx.LowPower) and !IsOn(par.mfx.LowPower) ?
3) Why did you change '|=' to '&=' for YUV422 only?

dmitryermilov (changes requested):

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
JeremyShang: 1. Because we initially assign the value to caps.YUV422ReconSupport in hevcehw_base_va_lin.cpp(Line:378), and we need the logic here to finalize this value.
2. IsOff(par.mfx.LowPower) means this option is off, !IsOn(par.mfx.LowPower) means this option is off or unknown.
3. The thing is that we don't support 422 recon format with LowPower on,  so after this change, caps.YUV422ReconSupport reports as 1 with LowPower on (Previously was 0).
@vilichev  Please correct me if I miss anything.

dmitryermilov: Okay. Thank you.

dmitryermilov (approved):

PR #2388, opened 2020-09-29
unmark the DPB's frame with frame_num duplicating with the coming frame

hongbolv:


dmitryermilov (changes requested):
need validation results

PR #2391, opened 2020-09-30
[sample_decode] Enable CodecLevel cmd parameter

egorovdanil:


dmitryermilov (changes requested):
Why do we need this change? CodecLevel comes from bitstream. Why do we need to force it via cmd option?

egorovdanil:
such PR is already uploaded https://github.com/Intel-Media-SDK/MediaSDK/pull/2416

PR #2397, opened 2020-10-06
[debug] Trace errors in mfx_common_int

Alexandr-Konovalov:


dvrogozh (approved):
At least someone listens to me breaking changes into commits.

dmitryermilov (dismissed):

onabiull (approved):

_studio/mfx_lib/shared/src/mfx_common_int.cpp
```diff
@@ -116,37 +112,41 @@ mfxStatus CheckFrameInfoCommon(mfxFrameInfo  *info, mfxU32 /* codecId */)
...
-            && info->FourCC != MFX_FOURCC_P016 && info->FourCC != MFX_FOURCC_Y216
-            && info->FourCC != MFX_FOURCC_Y416
+    case MFX_FOURCC_P016:
+    case MFX_FOURCC_Y216:
+    case MFX_FOURCC_Y416:
 #endif
-            )
-            return MFX_ERR_INVALID_VIDEO_PARAM;
+        // HEVC HW support both shifts, check rest of cases
+        if (!(isHW && codecId == MFX_CODEC_HEVC))
```
dgurulev, line 135: As I understand, the prev. logic was - 'the `Shift` can be set only for the certain FOURCCs'. It doesn't restrict to have `Shift` both values (zero/one) and I see no reasons to put codec's specific logic (HEVC) right here. I'd rather put it either to component's code or to `CheckFrameInfoCodecs` at least, where we already have both `codecId` and `isHW`.

dgurulev (changes requested):

_studio/mfx_lib/shared/src/mfx_common_int.cpp
Alexandr-Konovalov: @dgurulev , it seems `CheckFrameInfoCodecs()` already has all useful checks. So I'll abandon separate 
[HEVCd] supports both kind of shifts

dgurulev (approved):

dmitryermilov (approved):

PR #2402, opened 2020-10-09
[AVCe] Fix caps for 422/444 yuv formats

egorovdanil:
Test: `sample_encode h264 -rgb4 -i stream.rgb -o out.h264 -w 352 -h 288`

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
@@ -1562,7 +1562,7 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     m_caps.ddi_caps.MaxNum_WeightedPredL0   = 4;
     m_caps.ddi_caps.MaxNum_WeightedPredL1   = 2;
 
-    m_caps.ddi_caps.Color420Only = !(AV(VAConfigAttribRTFormat) & (VA_RT_FORMAT_YUV422 | VA_RT_FORMAT_YUV444));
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
```diff
@@ -1562,7 +1562,7 @@ mfxStatus VAAPIEncoder::CreateAuxilliaryDevice(
     m_caps.ddi_caps.MaxNum_WeightedPredL0   = 4;
     m_caps.ddi_caps.MaxNum_WeightedPredL1   = 2;
 
-    m_caps.ddi_caps.Color420Only = !(AV(VAConfigAttribRTFormat) & (VA_RT_FORMAT_YUV422 | VA_RT_FORMAT_YUV444));
+    m_caps.ddi_caps.Color420Only = !(AV(VAConfigAttribRTFormat) | VA_RT_FORMAT_YUV422 | VA_RT_FORMAT_YUV444);
```
AntonGrishin, line 1565: Looks wrong.
In old realization, we check that VaConfigRtFormat is not set to YUV422 or 444.

In your patch, Color420Only will be always false, because VA_RT_FORMAT are not zero constants.

AntonGrishin (changes requested):

_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp
dmitryermilov: @egorovdanil . please reply

egorovdanil: Yes, Anton is right, if we want to enable this, we need to edit the caps in the driver, but now the decision to support RGB using VPP in AVC has not been made.

PR #2410, opened 2020-10-19
fix RGB32->RGB32 composition with tile issue

yuqinw:
set output_background_color to fix RGB32->RGB32 composition with tile
incorrect output issue

Signed-off-by: Yuqin Wei <yuqin.wei@intel.com>

### DIFF
--- a/_studio/shared/src/mfx_vpp_vaapi.cpp
+++ b/_studio/shared/src/mfx_vpp_vaapi.cpp
@@ -1689,7 +1689,7 @@ mfxStatus VAAPIVideoProcessing::Execute_Composition_TiledVideoWall(mfxExecutePar
 
         m_pipelineParam[i].filters      = 0;
         m_pipelineParam[i].num_filters  = 0;
-        m_pipelineParam[i].output_background_color = 0;
+        m_pipelineParam[i].output_background_color = 0xff000000;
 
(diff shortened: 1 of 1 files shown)


FurongZhang (approved):

onabiull (approved):

_studio/shared/src/mfx_vpp_vaapi.cpp
```diff
@@ -1689,7 +1689,7 @@ mfxStatus VAAPIVideoProcessing::Execute_Composition_TiledVideoWall(mfxExecutePar
 
         m_pipelineParam[i].filters      = 0;
         m_pipelineParam[i].num_filters  = 0;
-        m_pipelineParam[i].output_background_color = 0;
+        m_pipelineParam[i].output_background_color = 0xff000000;
```
dmitryermilov, line 1692: @yuqinw ,what does 0xff000000 mean?

FurongZhang:
If output_background_color is not 0, we will do color fill in VPP, hence fixed the corruption. If it is 0, no color fill is called.

dmitryermilov:
@FurongZhang , what does 0xff000000 magic value mean?

FurongZhang:
It means alpha 0xff, but rgb or yuv is all 0.

Issue #2412, opened 2020-10-22
HEVC extbrc:implicit is broken w/ DEVICE_FAILED on SMT


dvrogozh:
See https://github.com/intel/media-delivery/issues/43

This command lines fails:
```
$ /usr/share/mfx/samples/sample_multi_transcode -fps 60 -i::h265 /tmp/WAR.hevc \
  -hw -async 1 -u 4 -gop_size 256 -dist 8 -num_ref 5 -vbr -b 9000 \
  -NalHrdConformance:off -VuiNalHrdParameters:off -hrd 5000 -InitialDelayInKB 2500 \
  -extbrc::implicit -ExtBrcAdaptiveLTR:on -o::h264 /opt/data/artifacts/measure/perf/output/WAR_hevc_0.h264  \
  -device /dev/dri/renderD128

ERROR], sts=MFX_ERR_DEVICE_FAILED(-17), Transcode, <EncodeOneFrame|Surface2BS> failed at /opt/src/msdk/samples/sample_multi_transcode/src/pipeline_transcode.cpp:1870

[ERROR], sts=MFX_ERR_DEVICE_FAILED(-17), Run, CTranscodingPipeline::Run::Transcode() [0x5565f30fff30] failed at /opt/src/msdk/samples/sample_multi_transcode/src/pipeline_transcode.cpp:4538
```

For input stream see https://github.com/intel/media-delivery/issues/43

Notes:
* This is regression on 20.3.0, I did not check master
* AVC is not affected
* ffmpeg (AVC/HEVC) is not affected

dvrogozh:
Reproduced on latest master: 2a7cfaa4, and this commit from master is currently the last known which worked good: 0e585ddc (that's a `git merge-base origin/master intel-mediasdk-20.2.1`)

dvrogozh:
Bisect point to d978f4ece (@spossos : fyi):
```
commit d978f4ece7952c527b2398a5a78cca6efc0e0b01
Author: possosse <sebastian.possos@intel.com>
Date:   Wed Aug 26 11:07:42 2020 -0700

    [AVC][MCFT] Integration of MCTF in AVC HW pipeline

    MCTF stage addition into AVC pipeline to provide
    advanced automatic filtering capabilities to AVC
    HW encoder when content is extremely noisy.

    Issue: N\A
    Tests: manual testing
```

Bisect log:
```
$ git bisect start
[...]

PR #2414, opened 2020-10-23
[MCTF] Segmentation fault bug fix related to missing output pointer

spossos:
A segmentation faul was happening because the ouput pointer was not
being assigned when the previous frame was assigned as a P frame to
be denoised and the next frame happened to be and I frame which
also needed to be denoised. The second frame (I frame) was not able
to get the actual output pointer for itself and that caused the
segmentation fault.

The next part addresses the DEVICE_FAILED issue, for this system
checks what is the platform using the filter and based on that
compares if the input width and height are supported, if not
then MCTF disables itself.

Issue: https://github.com/Intel-Media-SDK/MediaSDK/issues/2412
Test:  Manual testing based on MDP-64761

### DIFF
--- a/_studio/mfx_lib/mctf_package/mctf/src/mctf_common.cpp
+++ b/_studio/mfx_lib/mctf_package/mctf/src/mctf_common.cpp
@@ -366,6 +366,13 @@ mfxStatus CMC::MCTF_RELEASE_FRAME(
         MFX_CHECK(mco, MFX_ERR_UNDEFINED_BEHAVIOR);
         mco = nullptr;
         MctfState = AMCTF_NOT_READY;
+        if (mco2)
+        {
+            mco = mco2;
+            idxMco = idxMco2;
+            mco2 = nullptr;
+            idxMco2 = nullptr;
+        }
     }
     return MFX_ERR_NONE;
 }
@@ -604,8 +611,8 @@ mfxStatus CMC::MCTF_SetMemory(
             res = IM_SURF_SET(&mco, &idxMco);
             MCTF_CHECK_CM_ERR(res, MFX_ERR_DEVICE_FAILED);
         }
-        res = IM_SURF_SET(&mco2, &idxMco2);
-        MCTF_CHECK_CM_ERR(res, MFX_ERR_DEVICE_FAILED);
+        mco2 = nullptr;
+        idxMco2 = nullptr;
 
         //Setup for 2 references
         res = GEN_SURF_SET(&mv_1, &mvSys1, &idxMv_1);
@@ -3241,10 +3248,18 @@ mfxStatus CMC::MCTF_PUT_FRAME(
         if (isCmSurface)
         {
             res = IM_SURF_PUT(QfIn[bufferCount].frameData, (CmSurface2D*)frameData);
-            if ((countFrames == 0 || !(countFrames % MCTFCADENCE)) && !mco)
+            if ((countFrames == 0 || !(countFrames % MCTFCADENCE)))
             {
-                mco = (CmSurface2D*)frameData;
-                mco->GetIndex(idxMco);
+                if (mco && isSceneChange)
+                {
+                    mco2 = (CmSurface2D*)frameData;
+                    mco2->GetIndex(idxMco2);
+                }
+                else
+                {
+                    mco = (CmSurface2D*)frameData;
+                    mco->GetIndex(idxMco);
+                }
             }
         }
         else
@@ -3322,7 +3337,13 @@ mfxStatus CMC::MCTF_DO_FILTERING_IN_AVC()
             case 0:
             {
                 MCTF_UpdateANDApplyRTParams(1);
-                res = (this->*(pMCTF_func))(false);
+                if (QfIn[1].scene_idx != QfIn[0].scene_idx)
+                {
+                    res = MCTF_RUN_AMCTF(1);
+                    RotateBuffer();
+                }
+                else
+                    res = (this->*(pMCTF_func))(false);
                 CurrentIdx2Out = DefaultIdx2Out;
                 break;
             }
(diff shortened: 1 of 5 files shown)


spossos:
The second commit addresses the DEVICE_FAILURE issue, the system determines the platform and based on that decides if the filter should be enabled or not.

spossos:
The second commit dacd63a5e728b4cc4bbc54ec661fdc8a7351ffc5 fixes the DEVICE_FAILURE issue and selectively enables or disables MCTF based on the platform limitations.

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -1391,14 +1391,37 @@ bool MfxHwH264Encode::IsCmNeededForSCD(
...
 #if defined(MXF_ENABLE_MCTF_IN_AVC)
+    mfxU16
+        max_height = 0;
+    switch (platform)
+    {
+    case MFX_HW_ICL:
+    case MFX_HW_ICL_LP:
+    case MFX_HW_TGL_LP:
+    case MFX_HW_DG1:
+        max_height = 2160;
```
dvrogozh, line 1408: We again hardcode some limitations... Is there any way to query these limitations from underlying driver or CM? At least: can we point to the CM/driver source code where it works with these limitations? For example, 1696 is absolutely non-trivial number which I simply don't understand where it came from...

dvrogozh (changes requested):
You ended up with 2 commits which has same commit message: please, adjust a message!

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
spossos: I did the test until I found the actual limit, in this case if you use 1698 (even 1697 which is odd but anyway), you get the device_failure. now 2160 is the limit for MCTF and not for the driver.

dvrogozh: Well, this is weird. So, effectively we were guessing. But what if BXT has different Limit? or ICL? I guess you did not try all the platforms. So, we can accept this change as a workaround, but we need to find out the actual limit driven by implementation (driver or cm or whatever). And optimally this limit should be reported thru caps.

dmitryermilov: Hi @spossos ,

> you get the device_failure

Is this device_failure failure on memory allocation and from kernel execution or ?

dvrogozh: Talked with @spossos  offline. The talk is about this call:
https://github.com/Intel-Media-SDK/MediaSDK/blob/e7b1363357e22b778e8d0706fdc842a45c2f8580/_studio/mfx_lib/mctf_package/mctf/src/mctf_common.cpp#L3161
It seems we don't understand its limitations. And @spossos guessed them.

I wonder whether we can use msdk internal copy here which hopefully does not have these limitations:
https://github.com/Intel-Media-SDK/MediaSDK/blob/e7b1363357e22b778e8d0706fdc842a45c2f8580/_studio/shared/src/cm_mem_copy.cpp#L1447

dvrogozh: That's wrong assumption that this gpu copy depends on resolution. See https://github.com/Intel-Media-SDK/MediaSDK/issues/2412#issuecomment-716719434: we clearly fail on the comparison of input and output resolution, output resolution is lower by height. Hence the failure. MCTF simply does not account for possible surface padding alignment.

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
```diff
@@ -1399,6 +1399,7 @@ bool MfxHwH264Encode::IsMctfSupported(
     mfxExtCodingOption2 const & extOpt2 = GetExtBufferRef(video);
     isSupported = (IsOn(extOpt2.ExtBRC) &&
         IsExtBrcSceneChangeSupported(video) &&
+        (video.mfx.FrameInfo.Width <= 3840 && video.vpp.In.Height <= 2160) &&
```
dmitryermilov, line 1402: Sebastian, could you say where it comes from?

dvrogozh, line 1402: where this comes from? Do we have some MCTF limitations which we did not handle before?

dvrogozh (changes requested):

_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
spossos: Yes, this is an actual limitation for MCTF, it can only denoise frames up to 3840x2160.

spossos: This is an actual limitation for MCTF, it can denoise frames up to 3840x2160.

dvrogozh: Ok, that's fine then, but this needs to be yet another commit!

_studio/mfx_lib/mctf_package/mctf/src/mctf_common.cpp
```diff
@@ -487,7 +487,7 @@ mfxStatus CMC::IM_SURF_SET(
     SurfaceIndex ** p_idxSurf
 )
 {
-    res = device->CreateSurface2D(p_ctrl->width, p_ctrl->height, CM_SURFACE_FORMAT_NV12, *p_surface);
+    res = device->CreateSurface2D(ROUNDUP(p_ctrl->width,64), ROUNDUP(p_ctrl->height,64), CM_SURFACE_FORMAT_NV12, *p_surface);
```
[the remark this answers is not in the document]
dvrogozh: It occurs that we rewrite input surface w/ mctf output. I don't think we are supposed to do that. Proper way would be to allocate pool of surfaces for mctf output and eventually use them for encoder input. This is pretty big change however. So, my proposal is to currently disable implicit mctf within extbrc avc, refactor implementation, than enable it back.

dvrogozh (changes requested):

dvrogozh (approved):

dmitryermilov (approved):

PR #2420, opened 2020-10-26
Replace outdated 01org links

onabiull:


dvrogozh (approved):
I think we need to cherry-pick this change to the latest release branch.

PR #2429, opened 2020-10-28
MCTF: DEVICE_FAILURE fix related to copy of frames

spossos:
Removed enqueueCopyGPUToGPU as actual allocated surface
dimensions for input frame are unknown at the time of
adding frames into the denoising buffer. While stream
information about frame size is valid, the input frame
surface can have a variation because of different platform/OS
constrains/requirements for padding which causes mismatch
between the surfaces involved (SrcSurface > DstSurface).

Issue: MDP-64761
Test:  Manual testing


spossos:
Cleaning up the commits to just have one main

PR #2434, opened 2020-10-28
DROP ME: try aenc

dvrogozh:
Don't merge, this PR is to run MFX_ENABLE_AENC=ON thru CI


builder/FindPackages.cmake
```diff
@@ -514,3 +514,16 @@ if( Linux )
   endif()
 endif()
 
+if (MFX_ENABLE_AENC)
+  find_library( AENC_LIBRARY libaenc.a PATHS ${CMAKE_HOME_DIRECTORY}/contrib/aenc/ )
+  if(NOT ${AENC_LIBRARY} MATCHES NOTFOUND)
```
onabiull, line 519: Looks it doesn't work this way. You made this check before defining MFX_ENABLE_AENC. Note absence "found here" message in the log.

onabiull (changes requested):

PR #2436, opened 2020-10-29
[samples] Support machines without adapters

Alexandr-Konovalov:
Issue: MDP-64867

### DIFF
--- a/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
+++ b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************\
-Copyright (c) 2005-2019, Intel Corporation
+Copyright (c) 2005-2020, Intel Corporation
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -717,6 +717,12 @@ mfxStatus Launcher::QueryAdapters()
     mfxStatus sts = MFXQueryAdaptersNumber(&num_adapters_available);
     MFX_CHECK_STS(sts);
 
+    // no adapters on the machine, able to use software implementation
+    if (!num_adapters_available)
+    {
+        return MFX_ERR_NONE;
+    }
+
```
AntonGrishin, line 725: Based on your comment, seems that we should add a check for implementation type (hw/sw).

AntonGrishin (changes requested):

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
Alexandr-Konovalov: @AntonGrishin , ok, let’s fail more gracefully than just `MFX_ERR_UNSUPPORTED` from `InitEx()`.

AntonGrishin (approved):

dmitryermilov (approved):

PR #2439, opened 2020-10-30
[hevce] prevent HRD parameters submitting for CQP cases

vilichev:
HRD parameters cause switching to CBR in media driver for HEVCe FEI

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
@@ -246,12 +246,8 @@ void AddVaMiscHRD(
 {
     auto& hrd = AddVaMisc<VAEncMiscParameterHRD>(VAEncMiscParameterTypeHRD, buf);
 
-    uint32_t bNeedBufParam =
-        par.mfx.RateControlMethod != MFX_RATECONTROL_CQP
-        && par.mfx.RateControlMethod != MFX_RATECONTROL_ICQ;
-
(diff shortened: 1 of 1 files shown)


lakulako (commented):
AddVaMiscHRD already considers CQP case. Please check why it doesn't work. Anyway better to avoid doubled check.

vilichev:
@lakulako AddVaMiscHRD just sets HRD parameters depending on CQP. But we need to avoid submitting of such buffer to driver for CQP case.

lakulako:
@vilichev, may be it would be more correct not to switch to CBR if HRD buffer is attached. Anyway CBR is not the only mode that uses HRD.

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2020 Intel Corporation
+D/ Copyright (c) 2019-2020 Intel Corporation
```
lakulako, line 1: D/  - misprint?

vilichev:
> @vilichev, may be it would be more correct not to switch to CBR if HRD buffer is attached. Anyway CBR is not the only mode that uses HRD.
@lakulako I think we should fix it in MSDK too, since HRD does not intend for CQP/ICQ cases.

_studio/mfx_lib/encode_hw/hevc/linux/base/hevcehw_base_va_packer_lin.cpp
vilichev: Thanks! Done

saosipov (approved):

lakulako (approved):

dmitryermilov (approved):

PR #2460, opened 2020-11-06
[sample_vpp] Enable runtime HSBC

MicroYY:
Enable runtime HSBC.

### DIFF
--- a/samples/sample_vpp/src/sample_vpp.cpp
+++ b/samples/sample_vpp/src/sample_vpp.cpp
@@ -731,58 +731,34 @@ int main(int argc, msdk_char *argv[])
             else
             {
 #ifdef ENABLE_VPP_RUNTIME_HSBC
-                auto procAmp = pOutSurf->AddExtBuffer<mfxExtVPPProcAmp>();
-                // set default values for ProcAmp filters
-                procAmp->Brightness = 0.0F;
-                procAmp->Contrast   = 1.0F;
-                procAmp->Hue        = 0.0F;
-                procAmp->Saturation = 1.0F;
-
-                if (Params.rtHue.isEnabled)
+                if (Params.rtHue.isEnabled || Params.rtSaturation.isEnabled ||
+                    Params.rtBrightness.isEnabled || Params.rtContrast.isEnabled)
                 {
-                    if((nOutFrames / Params.rtHue.interval & 0x1) == 0)
-                    {
-                        procAmp->Hue = Params.rtHue.value1;
-                    }
-                    else
-                    {
-                        procAmp->Hue = Params.rtHue.value2;
-                    }
-                }
+                    auto procAmp = pOutSurf->AddExtBuffer<mfxExtVPPProcAmp>();
+                    // set default values for ProcAmp filters
+                    procAmp->Brightness = 0.0F;
+                    procAmp->Contrast   = 1.0F;
+                    procAmp->Hue        = 0.0F;
+                    procAmp->Saturation = 1.0F;
 
-                if (Params.rtSaturation.isEnabled)
-                {
-                    if((nOutFrames / Params.rtSaturation.interval & 0x1) == 0)
+                    if (Params.rtHue.isEnabled)
                     {
-                        procAmp->Saturation = Params.rtSaturation.value1;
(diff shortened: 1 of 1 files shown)


dmitryermilov (changes requested):
pOutSurf->AddExtBuffer<mfxExtVPPProcAmp> is the right way to handle ExBuffer. You don't need these std::map

MicroYY:
Hi @dmitryermilov 
So what' the meaning of  Params.rtHue.isEnabled? 
https://github.com/Intel-Media-SDK/MediaSDK/blob/8f6bcd030918c4b67d73572c31ee594454030438/samples/sample_vpp/src/sample_vpp.cpp#L741
If not enabled, the correct procamp params will be overwritten by default values in 
https://github.com/Intel-Media-SDK/MediaSDK/blob/8f6bcd030918c4b67d73572c31ee594454030438/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp#L3777
Could we remove it?

MicroYY:
Hi @dmitryermilov
With `pOutSurf->AddExtBuffer`, if no runtime procamp params are usesd (just enable this macro),  the correct params will be overwritten by default values in
https://github.com/Intel-Media-SDK/MediaSDK/blob/8f6bcd030918c4b67d73572c31ee594454030438/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp#L3777
because a valid `mfxExtVPPProcAmp `pointer is returned.
https://github.com/Intel-Media-SDK/MediaSDK/blob/8f6bcd030918c4b67d73572c31ee594454030438/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp#L3771

AntonGrishin:
> Hi @dmitryermilov
> With `pOutSurf->AddExtBuffer`, if no runtime procamp params are usesd (just enable this macro but do nothing), the correct params will be overwritten by default values in
> https://github.com/Intel-Media-SDK/MediaSDK/blob/8f6bcd030918c4b67d73572c31ee594454030438/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp#L3777
> 
> 
> because a valid `mfxExtVPPProcAmp`pointer is returned.
> https://github.com/Intel-Media-SDK/MediaSDK/blob/8f6bcd030918c4b67d73572c31ee594454030438/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp#L3771

So, seems that to avoid this issue, you add 
```
                if (Params.rtHue.isEnabled || Params.rtSaturation.isEnabled ||
                    Params.rtBrightness.isEnabled || Params.rtContrast.isEnabled)
                {
```

Can we just move this check and AddExtBuffer only if one of these options enabled?

dmitryermilov:
Hi @MicroYY .



> With `pOutSurf->AddExtBuffer`, if no runtime procamp params are usesd (just enable this macro but do nothing), the correct params will be overwritten by default values in

Anton already answered: 
> So, seems that to avoid this issue, you add
```
                if (Params.rtHue.isEnabled || Params.rtSaturation.isEnabled ||
                    Params.rtBrightness.isEnabled || Params.rtContrast.isEnabled)
                {
                       auto procAmp = pOutSurf->AddExtBuffer<mfxExtVPPProcAmp>();
                  .....
                }
```
> Can we just move this check and AddExtBuffer only if one of these options enabled?

MicroYY:
Patch update.

samples/sample_vpp/src/sample_vpp.cpp
```diff
@@ -731,58 +731,62 @@ int main(int argc, msdk_char *argv[])
...
+                    procAmp->Contrast   = 1.0F;
+                    procAmp->Hue        = 0.0F;
+                    procAmp->Saturation = 1.0F;
+
+                    if (Params.rtHue.isEnabled)
                     {
-                        procAmp->Hue = Params.rtHue.value1;
+                        if((nOutFrames / Params.rtHue.interval & 0x1) == 0)
+                        {
+                            procAmp->Hue = Params.rtHue.value1;
```
dmitryermilov, line 748: @MicroYY , generally the change is right. But since you're touching this code :) , can you please re-factor it a bit , this way:
`procAmp->Hue = ((nOutFrames / Params.rtHue.interval & 0x1) == 0) ?  Params.rtHue.value1 : Params.rtHue.value2`
?
and the same for other procAmp settings.

AntonGrishin (approved):

dmitryermilov (approved):

Issue #2013, opened 2020-03-14
allow ROI rectangle top/left/bottom/right to be 0 for query


guoyejun:
i submitted a PR https://github.com/Intel-Media-SDK/MediaSDK/pull/1856 for this issue, and after discussion, i hoped to pass the ball back to msdk team.

I just checked the latest code and found that we still need to set a non-empty rectangle for the query, so file this issue to check it. Hope it can be resolved before next release, thanks.

guoyejun:
any update? thanks.

guoyejun:
any update? thanks

daleksan:
@vilichev please follow up

vilichev:
@yskuridi Please investigate

yskuridi:
proposed a solution.
waiting for confirmation from architects

dmitryermilov:
Dear @guoyejun ,

Sorry for delay. We're discussed internally. The plan is the following:
Per the msdk manual, Query shall return max number of supported ROI if app passes 256 to input params:
`NumROI - Number of ROI descriptions in array. The Query function mode 2 returns maximum supported value (set it to 256 and Query will update it to maximum supported value).`

But currently AVC and HEVC encoders treats zero size rect's as error. We'll fix it. Query will fill rect sizes with min supported sizes. 

@yskuridi , could you implement a fix?

dmitryermilov:
Commit above fixes HEVC. We need to fix AVC as well. So re-opening

PR #2480, opened 2020-11-13
[hevce] Expand zero ROI to minimum supported size

yskuridi:
Fixes [#2013](https://github.com/Intel-Media-SDK/MediaSDK/issues/2013).

Change (0 0 0 0) ROIs instead of removing them.

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_roi.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_roi.cpp
@@ -50,6 +50,13 @@ static mfxStatus CheckAndFixRect(
     changed += AlignUp(rect.Right, blockSize);
     changed += AlignUp(rect.Bottom, blockSize);
 
+    if(rect.Left == 0 && rect.Top == 0 && rect.Right == 0 && rect.Bottom == 0)
+    {
+        rect.Right  = blockSize;//expand zero rectangle to min supported size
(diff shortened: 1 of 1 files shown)


dmitryermilov (changes requested):
@yskuridi , if this commit is intended to fix https://github.com/Intel-Media-SDK/MediaSDK/issues/2013, then you need to put "Fixes #2013" in commit message. 
Also, does this commit enable filling NumROI in the Query mode 2 ( from spec: _The Query function mode 2 returns maximum supported value (set it to 256 and Query will update it to maximum supported value)_)? Or it was already there?

yskuridi:
> @yskuridi , if this commit is intended to fix #2013, then you need to put "Fixes #2013" in commit message.
Done.

> Also, does this commit enable filling NumROI in the Query mode 2 ( from spec: _The Query function mode 2 returns maximum supported value (set it to 256 and Query will update it to maximum supported value)_)? Or it was already there?
Yes, it was already there.

dmitryermilov:
> > @yskuridi , if this commit is intended to fix #2013, then you need to put "Fixes #2013" in commit message.
> 
> Done.


I don't see an update in commit message, only in PR description .

yskuridi:
> I don't see an update in commit message, only in PR description .

Fixed

vilichev (approved):

dmitryermilov (approved):

dmitryermilov:
@yskuridi , just merged. Please don't forget to backport the change.

PR #2484, opened 2020-11-16
Update common_utils.h

samuel100u:
LoadRawFrame 's input is YU12 not YV12.

### DIFF
--- a/tutorials/common/common_utils.h
+++ b/tutorials/common/common_utils.h
@@ -90,8 +90,8 @@ void CloseFile(FILE* fHdl);
 
 using fileUniPtr = std::unique_ptr<FILE, decltype(&CloseFile)>;
 
-// LoadRawFrame: Reads raw frame from YUV file (YV12) into NV12 surface
-// - YV12 is a more common format for for YUV files than NV12 (therefore the conversion during read and write)
+// LoadRawFrame: Reads raw frame from YUV file (YU12) into NV12 surface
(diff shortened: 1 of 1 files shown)


tutorials/common/common_utils.h
```diff
@@ -90,8 +90,8 @@ void CloseFile(FILE* fHdl);
 
 using fileUniPtr = std::unique_ptr<FILE, decltype(&CloseFile)>;
 
-// LoadRawFrame: Reads raw frame from YUV file (YV12) into NV12 surface
-// - YV12 is a more common format for for YUV files than NV12 (therefore the conversion during read and write)
+// LoadRawFrame: Reads raw frame from YUV file (YU12) into NV12 surface
```
dmitryermilov, line 93: what is YU12? I've never heard of it.

dmitryermilov (changes requested):

tutorials/common/common_utils.h
samuel100u: https://wiki.videolan.org/YUV#YV12

Like the like above says

YV12

YV12 is exactly like I420, but the order of the U and V planes is reversed. In the name, "YV" refers to the plane order: Y, then V (then U). "12" refers to the pixel depth: 12-bits per pixel as for I420.


You are right. YU12 is not formally recognised by every one, but apparently loadrawframe is not for YV12 because its plane order is Y, then U.

dmitryermilov: I410 is 8 bit format not 12 bit format.

PR #2486, opened 2020-11-16
Fix  3 AV1 decode issues with system memory

weizhu-intel:
Following 3 issues are fixed in this PR:
1) error buffer status, which cause buffer be unmmaped twice.
2) For all key frame stream,  buffer be overwrited when it is still be used by APP side.
3) When non display frame be repeated, frame should wait for it be repeated, then release.


wangyan-intel:
@dmitryermilov @onabiull @dvrogozh Could you please review this PR? Thanks.

dmitryermilov (approved):

dvrogozh (approved):

PR #2487, opened 2020-11-16
[avce] Expand zero ROI to minimum supported size

dmitryermilov:
Fixes: #2013

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp
@@ -4484,6 +4484,15 @@ mfxStatus MfxHwH264Encode::CheckVideoParamQueryLike(
 
     for (mfxU16 i = 0; i < extRoi->NumROI; i++)
     {
+        auto& rect = extRoi->ROI[i];
+        if (rect.Left == 0 && rect.Right == 0 && rect.Top == 0 && rect.Bottom == 0)
+        {
+            // Allow zero rect sizes at input. Output min supported sizes.
+            rect.Right  = 16;
+            rect.Bottom = 16;
+            changed = true;
+        }
+
         sts = CheckAndFixRoiQueryLike(par, (mfxRoiDesc*)(&(extRoi->ROI[i])), extRoi->ROIMode);
         if (sts < MFX_ERR_NONE)
             unsupported = true;


dmitryermilov:
@lakulako , @saosipov , could you review?

lakulako (commented):
May be put this check to CheckAndFixOpenRectQueryLike?

dmitryermilov:
> May be put this check to CheckAndFixOpenRectQueryLike?

@lakulako , I did it at first. But then I noticed that the same CheckAndFixOpenRectQueryLike is called from run-time part (https://github.com/Intel-Media-SDK/MediaSDK/blob/0765c10a80823fee3d3240cffadde3719389fd69/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils_new.cpp#L2345 ). 
So moved new condition outside.

lakulako:
> > May be put this check to CheckAndFixOpenRectQueryLike?
> ... So moved new condition outside.
Shouldn't Init behave the same?

dmitryermilov:
> Shouldn't Init behave the same?

I assume while for Query zero values (like no values) makes sense, for processing part zero values are useless, rather invalid.

dmitryermilov:
BTW, let's align HEVC and AVC here. Actually I thought HEVC behaved as I described above but now I'm not sure:). @yskuridi , could you say how after your ROI change HEVC behaves:

- does it accept zero ROI values at Init and modifies them internally to min supported ROI rects (like Query does) 

- or at Init zero rect values are considered as errors (or simply ignored )?

dmitryermilov:
> > * does it accept zero ROI values at Init and modifies them internally to min supported ROI rects (like Query does)
> 
> Yes.

@eistarov , can you please comment? Is it correct behavior?

eistarov:
> > > * does it accept zero ROI values at Init and modifies them internally to min supported ROI rects (like Query does)
> > 
> > 
> > Yes.
> 
> @eistarov , can you please comment? Is it correct behavior?

I think, correct behavior would be to consider zero-rects as valid at Query and ignore at Init/run-time. Exception for Query required only due to "special" max supported NumROI report mechanic:
> The Query function mode 2 returns maximum supported value (set it to 256 and Query will update it to maximum supported value).

dmitryermilov:
> > > > * does it accept zero ROI values at Init and modifies them internally to min supported ROI rects (like Query does)
> > > 
> > > 
> > > Yes.
> > 
> > 
> > @eistarov , can you please comment? Is it correct behavior?
> 
> I think, correct behavior would be to consider zero-rects as valid at Query and ignore at Init/run-time. Exception for Query required only due to "special" max supported NumROI report mechanic:
 The Query function mode 2 returns maximum supported value (set it to 256 and Query will update it to maximum supported value).

Thanks @eistarov !

PR #2492, opened 2020-11-17
Fixed wrong UV order

samuel100u:
Fixed wrong UV order.

### DIFF
--- a/tutorials/common/common_utils.cpp
+++ b/tutorials/common/common_utils.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Intel Corporation
+// Copyright (c) 2019-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -173,12 +173,12 @@ mfxStatus LoadRawFrame(mfxFrameSurface1* pSurface, FILE* fSource)
     if (w > 2048)
         return MFX_ERR_UNSUPPORTED;
 
-    // load U
-    sts = ReadPlaneData(w, h, buf, ptr, pitch, 0, fSource);
+    // load V
+    sts = ReadPlaneData(w, h, buf, ptr, pitch, 1, fSource);
     if (MFX_ERR_NONE != sts)
         return sts;
-    // load V
-    ReadPlaneData(w, h, buf, ptr, pitch, 1, fSource);
+    // load U
+    ReadPlaneData(w, h, buf, ptr, pitch, 0, fSource);
     if (MFX_ERR_NONE != sts)
         return sts;
 


samuel100u:
image

dmitryermilov:
@samuel100u , please squash commits and give  better name to git commit.  "Update common_utils.cpp" doesn't sound good.

dmitryermilov (changes requested):
@samuel100u , please squash commits and give  better name to git commit.  "Update common_utils.cpp" doesn't sound good.

dmitryermilov:
@samuel100u , please squash these 4 commits into one..

dmitryermilov:
@samuel100u  ?

samuel100u:
Hi I am here. Sorry I am busy this week. I will finish it!

tutorials/common/common_utils.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Intel Corporation
+// Copyright (c) 2020 Intel Corporation
```
romka0017, line 1: 2019-2020

tutorials/common/common_utils.cpp
```diff
@@ -173,12 +173,12 @@ mfxStatus LoadRawFrame(mfxFrameSurface1* pSurface, FILE* fSource)
     if (w > 2048)
         return MFX_ERR_UNSUPPORTED;
 
-    // load U
-    sts = ReadPlaneData(w, h, buf, ptr, pitch, 0, fSource);
+    // load V
+    sts = ReadPlaneData(w, h, buf, ptr, pitch, 1, fSource);
```
romka0017, line 177: Why do we need to have offset 1 here but 0 for U?

romka0017 (commented):
And please add info to commit message: Which color format will you fix?

tutorials/common/common_utils.cpp
samuel100u: According to [common_utils.h#L93](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/tutorials/common/common_utils.h#L93), LoadRawFrame read YV12 format into NV12.
YV12 = YYYYYYYYVVUU  ,  NV12 = YYYYYYYYUVUV
That means that read YV12 with U first to NV12 is wrong in YV12 !

romka0017: Ok. If we read from plane format to NV12.

tutorials/common/common_utils.cpp
samuel100u: change to 2019-2020

ok

romka0017 (commented):
Please, update commit message.
Add info about color formats, which you will fix.

romka0017 (approved):

samuel100u:
@dmitryermilov  Done

dmitryermilov (approved):

PR #2500, opened 2020-11-20
Add Android build support for MSDK 20.3

LukeZheZhu:


_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_recon_info.h
```diff
@@ -45,9 +45,13 @@ namespace Base
         {}
 
     protected:
+#ifndef ANDROID
```
dvrogozh, line 48: Could you, please, explain what are you trying to achieve here? From declaration there should be functional implementation somewhere and I don't see that you disable it from compilation.... So, you should have ended up with 2 conflicting implementations...

dvrogozh (changes requested):

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_recon_info.h
dmitryermilov: @LukeZheZhu , can you please reply?

LukeZheZhu: Sorry for the late. I add '{}' because clang report error without '{}'

PR #2503, opened 2020-11-24
[sample_misc] Fix video stuttering during wayland rendering

SiewHoon:
The issue at here when RenderFrame exits, does not mean
corresponding wl_buffer is not used anymore by weston.
So, the wl_buffer has be wait until buffer_release function
call out under wl_buffer_listener get trigger. Than only
consider wl_buffer is no longer used anymore by weston and
safe to use by previous upper component like vpp csc.

The implementation code fixed at here is to prevent surface
get overwrite and used back by vpp csc inside sample_decode
when wl_buffer still not yet release when RenderFrame exits.

Platform: EHL, TGL-U, TGL-H
OS: Yocto native wayland weston
Tested:
./sample_decode h264 -i Puppies_1920x1080.h264 -rwld -rgb4 -f 30

Signed-off-by: Lim Siew Hoon <siew.hoon.lim@intel.com>

### DIFF
--- a/samples/sample_common/src/vaapi_device.cpp
+++ b/samples/sample_common/src/vaapi_device.cpp
@@ -397,7 +397,7 @@ mfxStatus CVAAPIDeviceWayland::RenderFrame(mfxFrameSurface1 * pSurface, mfxFrame
             return mfx_res;
     }
 
-    m_Wayland->RenderBuffer(m_wl_buffer, pSurface->Info.CropW, pSurface->Info.CropH);
+    m_Wayland->RenderBuffer(m_wl_buffer, pSurface);
 
     return mfx_res;
 }
--- a/samples/sample_misc/wayland/include/class_wayland.h
+++ b/samples/sample_misc/wayland/include/class_wayland.h
@@ -28,7 +28,13 @@ extern "C"
 }
 #include <poll.h>
 #include <wayland-client.h>
+#include <list>
 #include "wayland-drm-client-protocol.h"
+#include "mfxstructures.h"
+#include "mfx_buffering.h"
+#include "sample_defs.h"
+
+typedef struct buffer wld_buffer;
 
 /* ShmPool Struct */
 struct ShmPool {
@@ -38,7 +44,7 @@ struct ShmPool {
     unsigned size;
 };
 
-class Wayland {
+class Wayland: public CBuffering {
     public:
         Wayland();
         virtual ~Wayland();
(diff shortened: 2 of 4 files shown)


samples/sample_common/src/vaapi_device.cpp
```diff
@@ -397,7 +397,7 @@ mfxStatus CVAAPIDeviceWayland::RenderFrame(mfxFrameSurface1 * pSurface, mfxFrame
             return mfx_res;
     }
 
-    m_Wayland->RenderBuffer(m_wl_buffer, pSurface->Info.CropW, pSurface->Info.CropH);
+    m_Wayland->RenderBuffer(m_wl_buffer, pSurface->Info.CropW, pSurface->Info.CropH, pSurface);
```
dmitryermilov, line 400: If we pass the whole pSurface, why do we need to pass CropW and CropH separately?

samples/sample_misc/wayland/include/class_wayland.h
```diff
@@ -38,7 +42,14 @@ struct ShmPool {
...
 };
 
-class Wayland {
+/* buffer Struct */
+struct buffer {
+   struct wl_buffer *buffer;
+   mfxFrameSurface1 *pInSurface;
+   bool done;
+};
+
```
dmitryermilov, line 51: it's not used directly in header. Can we make typename buffer in header and declare struct in cpp file?

dmitryermilov (commented):
Generally LGTM except a few things

samples/sample_common/src/vaapi_device.cpp
SiewHoon: I can clean up it. Just direct passing only pSurface. CropW and CropH settle in RenderBuffer function.

samples/sample_misc/wayland/include/class_wayland.h
SiewHoon: it need to be use in listener_wayland.cpp under buffer_release function.

dmitryermilov: Sorry, I didn't get your reply. Yes, it's used in cpp file, that's why I'd propose to move declaration in cpp file :)

samples/sample_common/src/vaapi_device.cpp
dmitryermilov: Thanks

samples/sample_misc/wayland/include/class_wayland.h
SiewHoon: Ok, I already move it to declare in cpp file and typename in header. And I also make a bit code fixed changed as well.  I already push in new changed as well.

dmitryermilov (approved):

dmitryermilov:
@alariono can you please review?

PR #2507, opened 2020-11-25
Revert "Revert "[Media]Implemented dynamic Gpu session priority""

dmitryermilov:
Reverts Intel-Media-SDK/MediaSDK#2470

### DIFF
--- a/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
@@ -38,7 +38,7 @@
 
 #include "umc_jpeg_frame_constructor.h"
 #include "umc_mjpeg_mfx_decode_hw.h"
-
+#include "mfx_session.h"
 
 // Declare skipping constants
 enum
@@ -777,6 +777,11 @@ mfxStatus VideoDECODEMJPEG::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1 
 
     mfxU32 numPic = 0;
     mfxU32 picToCollect = (MFX_PICSTRUCT_PROGRESSIVE == m_vPar.mfx.FrameInfo.PicStruct) ? 1 : 2;
+
+    //gpu session priority 
+    UMC::VideoDecoderParams* videoDecoderParams = &(decoder->umcVideoParams);
+    if (videoDecoderParams != nullptr && videoDecoderParams->pVideoAccelerator != nullptr)
+        videoDecoderParams->pVideoAccelerator->m_ContextPriority = m_core->GetSession()->m_priority;
 
     do
     {
--- a/_studio/shared/umc/codec/vp9_dec/src/umc_vp9_va_packer.cpp
+++ b/_studio/shared/umc/codec/vp9_dec/src/umc_vp9_va_packer.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2020 Intel Corporation
+// Copyright (c) 2020 Intel Corporation
 //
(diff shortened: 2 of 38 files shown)


_studio/shared/umc/codec/vp9_dec/src/umc_vp9_va_packer.cpp
```diff
@@ -105,6 +105,36 @@ void PackerVA::PackAU(VP9Bitstream* bs, VP9DecoderFrame const* info)
...
+    {
+        GpuPriorityBuf_Vp9Decode->context_priority.bits.priority = 0;
+    }
+    else if (priority == MFX_PRIORITY_HIGH)
+    {
+        GpuPriorityBuf_Vp9Decode->context_priority.bits.priority = m_va->m_MaxContextPriority;
+    }
+    else
+    {
+    GpuPriorityBuf_Vp9Decode->context_priority.bits.priority = m_va->m_MaxContextPriority/2;
```
wangyan-intel, line 134: So MSDK only supports 3 levels. But media driver may support more levels. May we need expand it in the future?
Another question: Should we add it for AV1 decoding too? I am not sure.

dmitryermilov: These are questions to @hongbolv  and @dylanintel . Folks, can you please reply?

hongbolv: @wangyan-intel  In the beginning, we design 3 levels to align with graphics side. About the AV1 decoding, we should add it.

wangyan-intel: Another question: The code section seems similar in every codec. So could we use common function for avoiding code duplication? Thanks.

dmitryermilov: @wangyan-intel , It's a very good question. You know, when I reviewed the PR the first time my desire was to propose to decouple the change into some separate code/class. But later decided not to raise/insist on concern because it may require quite a big refactoring which I'm not sure Dylan is ready to make. Ideal solution for me would be to move this GPU priority logic out of codec level code, e.g. to move it to a wrapper under libva level and switch each component to this wrapper instead of libva.  @dylanintel , @hongbolv, please feel free to comment.

hongbolv: Frankly speaking, I am thinking about this issue. The requirement is that priority setting per frame, then we have to pass the priority value through vaRenderPicture calling on current implementation. It made us to change every component because there is no a common libva wrapper in the media SDK. If we plan to add a libva wrapper for each component, i think it is a big refactoring effort too.

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -778,6 +778,11 @@ mfxStatus VideoDECODEMJPEG::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
     mfxU32 numPic = 0;
     mfxU32 picToCollect = (MFX_PICSTRUCT_PROGRESSIVE == m_vPar.mfx.FrameInfo.PicStruct) ? 1 : 2;
 
+    //gpu session priority 
+    UMC::VideoDecoderParams* videoDecoderParams = &(decoder->umcVideoParams);
+    if (videoDecoderParams != nullptr && videoDecoderParams->pVideoAccelerator != nullptr)
```
mgonchar, line 783: no sense to check this pointer, ```videoDecoderParams``` can't be null, since it is address of local variable

PR #2509, opened 2020-11-30
smt: new video memory default

mgonchar:
Default memory type changed from opaq to video
Added option to set opaq

Perf val was launched, no difference on default mode found

### DIFF
--- a/samples/sample_multi_transcode/src/transcode_utils.cpp
+++ b/samples/sample_multi_transcode/src/transcode_utils.cpp
@@ -201,8 +201,10 @@ void TranscodingSample::PrintHelp()
         MSDK_STRING("                In encoding sessions (-o::source) and transcoding sessions \n") \
         MSDK_STRING("                  this parameter limits number of frames sent to encoder.\n"));
 
-    msdk_printf(MSDK_STRING("  -ext_allocator    Force usage of external allocators\n"));
-    msdk_printf(MSDK_STRING("  -sys          Force usage of external system allocator\n"));
+    msdk_printf(MSDK_STRING("  -MemType::video    Force usage of external video allocator (default)\n"));
+    msdk_printf(MSDK_STRING("  -MemType::system   Force usage of external system allocator\n"));
+    msdk_printf(MSDK_STRING("  -MemType::opaque   Force usage of internal allocator\n"));
+
     msdk_printf(MSDK_STRING("  -dec::sys     Set dec output to system memory\n"));
     msdk_printf(MSDK_STRING("  -vpp::sys     Set vpp output to system memory\n"));
     msdk_printf(MSDK_STRING("  -vpp::vid     Set vpp output to video memory\n"));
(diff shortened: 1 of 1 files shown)


samples/sample_multi_transcode/src/transcode_utils.cpp
```diff
@@ -2102,6 +2103,10 @@ mfxStatus CmdProcessor::ParseParamsForOneSession(mfxU32 argc, msdk_char *argv[])
         {
             InputParams.bUseOpaqueMemory = false;
         }
+        else if (0 == msdk_strcmp(argv[i], MSDK_STRING("-opaq")))
+        {
+            InputParams.bUseOpaqueMemory = false;
```
romka0017, line 2108: May be InputParams.bUseOpaqueMemory = true; ?

mgonchar: thanks, Roma!

romka0017 (dismissed):
May be we can remove multiple lines: InputParams.bUseOpaqueMemory = false;
from code.

samples/sample_multi_transcode/src/transcode_utils.cpp
```diff
@@ -201,7 +201,8 @@ void TranscodingSample::PrintHelp()
         MSDK_STRING("                In encoding sessions (-o::source) and transcoding sessions \n") \
         MSDK_STRING("                  this parameter limits number of frames sent to encoder.\n"));
 
-    msdk_printf(MSDK_STRING("  -ext_allocator    Force usage of external allocators\n"));
+    msdk_printf(MSDK_STRING("  -ext_allocator    Force usage of external allocators (default)\n"));
+    msdk_printf(MSDK_STRING("  -opaq         Force usage of OPAQ memory\n"));
```
dmitryermilov, line 205: "Force usage of the opaque memory type"

dmitryermilov, line 205: Why do we have per component sys or vid and global "opaq". I assume we need to align here. Make either all types per component or make them global.

dmitryermilov (changes requested):

samples/sample_multi_transcode/src/transcode_utils.cpp
mgonchar: this "opaq" key is just ability to return previous behavior, so it is global as well as "sys" and "ext_allocator" keys. I don't know how those per-component types should work with this key (it will work as as it did previously actually) and why did we need them. This option is required because user will not have ability to use opaq after default change

mgonchar: ok, I can change wording

mgonchar: So my point is that I don't want to bring redesign of these keys with this current PR, trying to minimize the change. I touched those lines only to group all memory related keys together

dmitryermilov: > So my point is that I don't want to bring redesign of these keys with this current PR, trying to minimize the change. I touched those lines only to group all memory related keys together

I understand that you want to minimize efforts. However do you agree after the change code looks worse?

mgonchar: Not sure if it become worse, it become just a little bit different. I agree that logic of setting memtypes seems to be missleading and a little strange. We need all those keys and functionality because it is used in tests, so your suggestion is to rework this part completely, did I get it right?

dmitryermilov (dismissed):

dmitryermilov (approved):

Issue #2452, opened 2020-11-05
output yuv green screen when decode h264 file using sample decode


lzfgnu:
here is input h264 file
[h264.zip](https://github.com/Intel-Media-SDK/MediaSDK/files/5491918/h264.zip)

platform: windows10 1909
cpu: Intel Core i7-8700
gpu: intel uhd graphics 630
ram: 16gb

wangyan-intel:
What platform?

lzfgnu:
> What platform?

platform: windows10 1909
cpu: Intel Core i7-8700
gpu: intel uhd graphics 630
ram: 16gb

Here is my computer config. I also tested in my laptop thinkpad t480, the same problem appeared

wangyan-intel:
@chuanli1 could you please check it? Thanks.

chuanli1:
sure

dmitryermilov:
@lzfgnu , please post:
1) Driver version
2) command line you used?
3) info if the issue is seen or not with YUV file dumping 

Thanks.

lzfgnu:
> @lzfgnu , please post:
> 
> 1. Driver version
> 2. command line you used?
> 3. info if the issue is seen or not with YUV file dumping
> 
> Thanks.

Thanks for your reply.  Here is my computer information
1. Driver version: Intel UHD Graphics 630 **(26.20.100.7642)**
2. Command line: **sample_decode h264 -i input.h264 -o output.yuv**
3. The issue is seen with yuv file. some frames are normal while some frames are green.
![image](https://user-images.githubusercontent.com/25069103/98216879-a340a900-1f84-11eb-8548-34aa3e3eb6b2.png)

dmitryermilov:
Thanks. Could you try to add "-hw -d3d" or  "-hw -d3d11" cmd option? Will you see the same green frames?

lzfgnu:
> Thanks. Could you try to add "-hw -d3d" or "-hw -d3d11" cmd option? Will you see the same green frames?

I tried and the issuse is still existed

lzfgnu:
> > Thanks. Could you try to add "-hw -d3d" or "-hw -d3d11" cmd option? Will you see the same green frames?
> 
> I tried and the issue is still existed

chuanli1:
- windows 

sample_decode.exe h264 -i data.h264 -o output_d3d.yuv -hw -d3d
**the result**:
[...]

PR #2516, opened 2020-12-08
[h264d]fix reference frame mark to support discontinuous frame_num

chuanli1:


wangyan-intel:
@chuanli1 Could you please add the detail info into commit message like you mentioned in #2452?  You could make summary in commit message. And add one line like "Fixes #2452." which will close the issue automatically when your PR is merged. Thanks.

dmitryermilov (changes requested):
Does the change follow the spec? Please provide a link or wording from AVC spec which say that in case of available ref frames we need to skip frame gaps handling.

dmitryermilov:
@dgurulev , could you review as well?

chuanli1:
thanks @dmitryermilov and @wangyan-intel . 
Rec. ITU-T H.264 pointed out in the Decoding process for gaps in frame_num section that the bitstream shall not contain data that result in the derivation of a reference picture that is marked as "non-existing" in any invocation of the reference picture selection process specified in clause 8.4.2.1.

In order to cover this kind of cases，I judge whether there have the mark reference frames according to count NumShortTermRefs and NumLongTermRefs, if yes, current frame will not use a non-existing frame as reference frame and continue to use the already  marked reference frame.

chuanli1:
@dmitryermilov  this CI build is the environment issue, and there is no relationship with my patch.

dmitryermilov:
> @dmitryermilov this CI build is the environment issue, and there is no relationship with my patch.

@eponomar , can you please help?

eponomar:
> > @dmitryermilov this CI build is the environment issue, and there is no relationship with my patch.
> 
> @eponomar , can you please help?

Checks restarted. It seems it was local issue or force push to driver master.

_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
```diff
@@ -2729,7 +2729,12 @@ Status TaskSupplier::ProcessFrameNumGap(H264Slice *pSlice, int32_t field, int32_
     Status umcRes = UMC_OK;
     const H264SeqParamSet* sps = pSlice->GetSeqParam();
     H264SliceHeader *sliceHeader = pSlice->GetSliceHeader();
-
+	uint32_t NumShortTermRefs, NumLongTermRefs;
+	view.GetDPBList(0)->countActiveRefs(NumShortTermRefs, NumLongTermRefs);
+	if (NumShortTermRefs > 0 || NumLongTermRefs > 0) 
+	{
+		return UMC_OK;
```
dgurulev, line 2736: Here u even don't try to check (see `DetectFrameNumGap`) whether we have gaps or not (so actually throw away all function's code below). The second thing, the behavior of the decoder when [SPS :: gaps_in_frame_num_value_allowed_flag] is well defined (see 8.2.5.2).

dgurulev (changes requested):

chuanli1:
As 8.2.5.2 mentioned, filling the frame_num gap with non-existing frames and canceling the previously marked reference frame, and finally marking the non-existent frame as a reference frame, the result will appear a green screen , I think this is a right phenomenon, because the purpose of filling the non-existent frame is that the current frame cannot find the reference frame, but in order to cover the clip, before inserting this non-existent frame, first check whether there is a marked reference frame.

_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
dmitryermilov: @chuanli1 , yes, it seems the change is not according to the spec. By “if (NumShortTermRefs > 0 || NumLongTermRefs > 0)” you disable all further actions even if gaps_in_frame_num_value_allowed_flag is allowed. Which is not right. However even if input stream is corrupted we need to at least try to reach better error tolerances in the decoder.  I mean, if possible we need to try to find a WA which on the one hand doesn’t contradict the spec, and on the other hand resolves the issue.

BTW, could you check behavior with other decoders (reference JM, libav, vaapi, something else)?

dmitryermilov (changes requested):

wangyan-intel:
@chuanli1 Have you done something wrong？ Your commit is pushed another PR?

chuanli1:
sorry, submit wrong.

PR #2517, opened 2020-12-08
[HEVCe] Handle NumMbPerSlice with LowPower on

JeremyShang:
HEVC encoder supports row slice with LowPower on


_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
```diff
@@ -3860,6 +3860,8 @@ mfxStatus Legacy::CheckSlices(
 
     if (bCheckNMB)
     {
+        MFX_CHECK(!IsOn(par.mfx.LowPower), MFX_ERR_UNSUPPORTED);
```
lakulako, line 3863: It is the case for incompatible_param. The function should not exit immediately, but change pCO2->NumMbPerSlice to zero, ++changed, and continue to check other parameters.
User will now the feature is not supported by seeing NumMbPerSlice has changed to 0.

lakulako (changes requested):

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
JeremyShang: Thanks @lakulako. I will upload a commit to fix that.

JeremyShang:
Draft PR. Still WIP.

PR #2520, opened 2020-12-09
[h264d] Unmark the duplicated frame_num with the coming frame in DPB

dylanintel:
Fix the issue about some h264 streams which have a wrong way to calculate frame_num.

Issue: MDP-64211, MDP-48207
Signed-off-by: Wang Dylan <dylan.wang@intel.com>

### DIFF
--- a/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
+++ b/_studio/shared/umc/codec/h264_dec/include/umc_h264_task_supplier.h
@@ -598,6 +598,7 @@ protected:
     virtual void OnFullFrame(H264DecoderFrame * pFrame);
     virtual bool ProcessNonPairedField(H264DecoderFrame * pFrame) = 0;
 
+    void DPBSanitize(H264DecoderFrame * pDPBHead, const H264DecoderFrame * pFrame);
     void DBPUpdate(H264DecoderFrame * pFrame, int32_t field);
 
     virtual void AddFakeReferenceFrame(H264Slice * pSlice);
--- a/_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
+++ b/_studio/shared/umc/codec/h264_dec/src/umc_h264_task_supplier.cpp
@@ -3591,6 +3591,20 @@ H264Slice * TaskSupplier::DecodeSliceHeader(NalUnit *nalUnit)
     return pSlice;
 }
 
+void TaskSupplier::DPBSanitize(H264DecoderFrame * pDPBHead, const H264DecoderFrame * pFrame)
+{
+    for (H264DecoderFrame *pFrm = pDPBHead; pFrm; pFrm = pFrm->future())
+    {
+        if ((pFrm != pFrame) &&
+            (pFrm->FrameNum() == pFrame->FrameNum()) &&
+             pFrm->isShortTermRef())
+        {
+            pFrm->SetErrorFlagged(ERROR_FRAME_SHORT_TERM_STUCK);
+            AddItemAndRun(pFrm, pFrm, UNSET_REFERENCE | FULL_FRAME | SHORT_TERM);
+        }
+    }
+}
+
 Status TaskSupplier::AddSlice(H264Slice * pSlice, bool force)
 {
     if (!m_accessUnit.GetLayersCount() && pSlice)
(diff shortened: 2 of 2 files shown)


dmitryermilov (commented):
@dylanintel , thank you for the PR. 
As far I understand this change is almost the same as https://github.com/Intel-Media-SDK/MediaSDK/pull/166 (except "Frame_type != B_PICTURE" additional condition) which later was removed by https://github.com/Intel-Media-SDK/MediaSDK/pull/585. Right ? So now we're reapplying  it back. Okay. Why not.
But I'd like to propose a few things:
1) Please explain in code comment (please feel free to wrote as many line as needed) why now we have this "lastSlice->GetSeqParam()->gaps_in_frame_num_value_allowed_flag != 1 **&& Frame_type != B_PICTURE**" condition . It's not clear to me right now. And I'm afraid even if you describe it in a github reply, later if we come back to this topic again, no one will remember the reasons.
2) Let's run h264d validation with all types of conformance test cases and corrupted streams previously collected. I'll help with to arrange it. Plus, please also **manually** test that this change doesn't cause a regression with the bitstream from MDP-48189/MDP-46557. Asking because PR #585 was for these MDP-48189/MDP-46557.

hongbolv:
> @dylanintel , thank you for the PR.
> As far I understand this change is almost the same as #166 (except "Frame_type != B_PICTURE" additional condition) which later was removed by #585. Right ? So now we're reapplying it back. Okay. Why not.
> But I'd like to propose a few things:
> 
> 1. Please explain in code comment (please feel free to wrote as many line as needed) why now we have this "lastSlice->GetSeqParam()->gaps_in_frame_num_value_allowed_flag != 1 **&& Frame_type != B_PICTURE**" condition . It's not clear to me right now. And I'm afraid even if you describe it in a github reply, later if we come back to this topic again, no one will remember the reasons.
> 2. Let's run h264d validation with all types of conformance test cases and corrupted streams previously collected. I'll help with to arrange it. Plus, please also **manually** test that this change doesn't cause a regression with the bitstream from MDP-48189/MDP-46557. Asking because PR #585 was for these MDP-48189/MDP-46557.

@dmitryermilov Yes, this change is almost same except (except "Frame_type != B_PICTURE" additional condition).
1.  In the #166, the patch is to fix the corrupted stream which has a frame number gap. The frame number gap caused a reference frame to always stay in DPB buffer. The #166 patch unmarks the reference frame in DPB buffer when the future frame with the same frame number is coming. 
    But in the another corrupted stream, the frame number is wrong. In the correct mode, in the encoding sequence, the B frame's frame_num should be previous reference frame's frame_num + 1; but the corrupted stream's B frame's frame_num is same as the previous reference frame's frame_num.  When the patch #166 handle this case, when the B frame arrived, the previous reference frame is always unmarked because of the same frame_num. 

Normal frame number:
I     P     B     P     B     P     B     P
0    1      2     2     3     3     4     4

Corrupted stream 0:
I     P     B     P     B     P     B     P
0    1      2     2     3     3     5     5
                      I
                      I
                      This frame always stays in the DPB buffer. Patch #166 will unmark it when the next frame which frame_num is 2 coming.
Corrupted stream1 :
I     P     B     P     B     P     B     P
0    1     1      2     2     3     3     4
              I
              I
              With the patch #166, this B frame will unmark the previous reference frame in the DPB buffer. In fact, it should not do it. So we add the condition (Frame_type != B_PICTURE) to deal with this case.

dmitryermilov (approved):

dmitryermilov:
Thank

> > @dylanintel , thank you for the PR.
> > As far I understand this change is almost the same as #166 (except "Frame_type != B_PICTURE" additional condition) which later was removed by #585. Right ? So now we're reapplying it back. Okay. Why not.
> > But I'd like to propose a few things:
> > 
> > 1. Please explain in code comment (please feel free to wrote as many line as needed) why now we have this "lastSlice->GetSeqParam()->gaps_in_frame_num_value_allowed_flag != 1 **&& Frame_type != B_PICTURE**" condition . It's not clear to me right now. And I'm afraid even if you describe it in a github reply, later if we come back to this topic again, no one will remember the reasons.
> > 2. Let's run h264d validation with all types of conformance test cases and corrupted streams previously collected. I'll help with to arrange it. Plus, please also **manually** test that this change doesn't cause a regression with the bitstream from MDP-48189/MDP-46557. Asking because PR #585 was for these MDP-48189/MDP-46557.
> 
> @dmitryermilov Yes, this change is almost same except (except "Frame_type != B_PICTURE" additional condition).
> 
> 1. In the #166, the patch is to fix the corrupted stream which has a frame number gap. The frame number gap caused a reference frame to always stay in DPB buffer. The #166 patch unmarks the reference frame in DPB buffer when the future frame with the same frame number is coming.
>    But in the another corrupted stream, the frame number is wrong. In the correct mode, in the encoding sequence, the B frame's frame_num should be previous reference frame's frame_num + 1; but the corrupted stream's B frame's frame_num is same as the previous reference frame's frame_num.  When the patch #166 handle this case, when the B frame arrived, the previous reference frame is always unmarked because of the same frame_num.
> 
> Normal frame number:
> I P1 B2 P2 B3 P3 B4 P4
> 0 1 2 2 3 3 4 4
> 
> Corrupted stream 0:
> I P1 B1 P2 B2 P3 B3 P4
> 0 1 2 2 3 3 5 5
> This frame P2 always stays in the DPB buffer. Patch #166 will unmark it when the next frame with frame_num =2 is coming.
> Corrupted stream1 :
> I P1 B1 P2 B2 P3 B3 P4
> 0 1 1 2 2 3 3 4
> With the patch #166, this B1 frame will unmark the previous reference frame P1 in the DPB buffer. In fact, it should not do it. So we add the condition (Frame_type != B_PICTURE) to deal with this case.

Thank you. Also it seems internal validations results are fine. 
Let's merge the PR but at first let's make sure CI "test" failed not sue to this PR . @dylanintel , can you please rebase the PR to trigger CI again?  If it doesn't help, probably need to look into the issue manually. There is the wiki page: https://github.com/Intel-Media-SDK/MediaSDK/wiki/Run-CI-smoke-tests

dmitryermilov:
@dylanintel , the same CI issue is seen with other PRs, e.g. https://github.com/Intel-Media-SDK/MediaSDK/pull/2531. We're trying to address it. JFYI

dmitryermilov (approved):

PR #2528, opened 2020-12-10
[sample_encode] Init mfxExtMVCSeqDesc for vpp

Alexandr-Konovalov:
Since a2f2115c mfxExtMVCSeqDesc in m_mfxVppParams is empty.

Issue: MDP-65303

### DIFF
--- a/samples/sample_encode/src/pipeline_encode.cpp
+++ b/samples/sample_encode/src/pipeline_encode.cpp
@@ -342,15 +342,13 @@ mfxStatus sTask::Reset()
     return MFX_ERR_NONE;
 }
 
-mfxStatus CEncodingPipeline::AllocateExtMVCBuffers()
+void CEncodingPipeline::InitExtMVCBuffers(mfxExtMVCSeqDesc *mvcBuffer) const
 {
     // a simple example of mfxExtMVCSeqDesc structure filling
     // actually equal to the "Default dependency mode" - when the structure fields are left 0,
     // but we show how to properly allocate and fill the fields
 
     mfxU32 i;
-
-    auto mvcBuffer = m_mfxEncParams.AddExtBuffer<mfxExtMVCSeqDesc>();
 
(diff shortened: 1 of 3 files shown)


samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -343,15 +343,21 @@ mfxStatus sTask::Reset()
 }
 
 mfxStatus CEncodingPipeline::AllocateExtMVCBuffers()
```
dmitryermilov, line 345: can we fully drop this function? It's used in one place.

dmitryermilov, line 345: Can we fully delete this function? It's used in one place only.

dmitryermilov (changes requested):

samples/sample_encode/src/pipeline_encode.cpp
Alexandr-Konovalov: [CRegionEncodingPipeline::Init()](https://github.com/Intel-Media-SDK/MediaSDK/blob/c22544bcf12de3f7c4ab00276d0fb604fea5892b/samples/sample_encode/src/pipeline_region_encode.cpp#L420) is another place. Not that one is very useful.

dmitryermilov: Why another? As far as I see this is the only place where it's called. I propose to remove

Alexandr-Konovalov: @dmitryermilov 1st place is in [CEncodingPipeline::Init()](https://github.com/Intel-Media-SDK/MediaSDK/blob/c22544bcf12de3f7c4ab00276d0fb604fea5892b/samples/sample_encode/src/pipeline_encode.cpp#L1829). Sure, let's remove.

dmitryermilov (approved):

PR #2529, opened 2020-12-11
[HEVCe] Avoid nullptr dereferenced

JeremyShang:


onabiull:
@mgonchar and @eistarov isn't it a known false positive?

dmitryermilov (commented):
It seems to be a false positive.

JeremyShang:
If that's the case, I can close this PR

Issue #2392, opened 2020-10-01
h265 to vp9e transcoding failure on 8K resolution stream when vaSyncBuffer is used instead of vaSyncSurface


aobolensk:
Transcoding using sample_multi_transcode:
Input stream: h265
Output stream: vp9
Resolution: 7680x4320 (for input and output)
Tested on: #2351 (vaSyncSurface is replaced with vaSyncBuffer)
There is no such problem in master branch.
```
./sample_multi_transcode -i::h265 input.265 -o::vp9 output.vp9 -f 25 -b 4600000 -u 1 -hw -n 300 -async 1 -lowpower:on -w 7680 -h 4320
```
Result:
```
[ERROR], sts=MFX_ERR_DEVICE_FAILED(-17), PutBS, Encode: SyncOperation failed at /home/user/MediaSDK/samples/sample_multi_transcode/src/pipeline_transcode.cpp:1934

[ERROR], sts=MFX_ERR_ABORTED(-12), Transcode, PutBS failed at /home/user/MediaSDK/samples/sample_multi_transcode/src/pipeline_transcode.cpp:1896

[ERROR], sts=MFX_ERR_ABORTED(-12), Run, CTranscodingPipeline::Run::Transcode() [0x55e7b0eaef40] failed at /home/user/MediaSDK/samples/sample_multi_transcode/src/pipeline_transcode.cpp:4538


 session 0 [0x55ac4f784f40] failed with status MFX_ERR_ABORTED shutting down the application...

session [0x55ac4f784f40] m_bForceStop is set

Transcoding finished

Common transcoding time is 0.322258 sec
-------------------------------------------------------------------------------
*** session 0 [0x55ac4f784f40] FAILED (MFX_ERR_ABORTED) 0.322074 sec, 1 frames, 3.105 fps
-i::h265 input.265 -o::vp9 output.vp9 -f 25 -b 4600000 -u 1 -hw -n 300 -async 1 -lowpower:on -w 7680 -h 4320

-------------------------------------------------------------------------------

The test FAILED

[ERROR], sts=MFX_ERR_ABORTED(-12), main, transcode.ProcessResult failed at /home/user/MediaSDK/samples/sample_multi_transcode/src/sample_multi_transcode.cpp:1165
```

aobolensk:
This issue seems to be not reproducible on the latest media driver.

dmitryermilov:
We saw it again on latest driver.

[...]

PR #2351, opened 2020-09-11
[VP9e] Use vaSyncBuffer in VAAPIEncoder::QueryStatus

aobolensk:

PR #2530, opened 2020-12-14
[VP9e] Use vaSyncBuffer in VAAPIEncoder::QueryStatus

IlyaSergeyev:
originaly introduced in  #2351

### DIFF
--- a/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
+++ b/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
@@ -1324,6 +1324,19 @@ mfxStatus VAAPIEncoder::QueryStatus(
 
     VASurfaceStatus surfSts = VASurfaceSkipped;
 
+#if VA_CHECK_VERSION(1,9,0)
+    vaSts = vaSyncBuffer(m_vaDisplay, codedBuffer, VA_TIMEOUT_INFINITE);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    vaSts = vaQuerySurfaceStatus(m_vaDisplay, waitSurface, &surfSts);
+    MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
+
+    if (VASurfaceReady == surfSts)
+    {
+        UMC::AutomaticUMCMutex guard(m_guard);
+        m_feedbackCache.erase(m_feedbackCache.begin() + indxSurf);
+    }
+#else
     vaSts = vaSyncSurface(m_vaDisplay, waitSurface);
     MFX_CHECK_WITH_ASSERT(VA_STATUS_SUCCESS == vaSts, MFX_ERR_DEVICE_FAILED);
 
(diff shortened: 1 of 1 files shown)


_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -1324,6 +1324,18 @@ mfxStatus VAAPIEncoder::QueryStatus(
 
     VASurfaceStatus surfSts = VASurfaceSkipped;
 
+#if VA_CHECK_VERSION(1,9,0)
+    vaSts = vaSyncBuffer(m_vaDisplay, codedBuffer, VA_TIMEOUT_INFINITE);
```
dvrogozh, line 1328: you reduce code robustness by overriding `vaSts`

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
```diff
@@ -1324,6 +1324,18 @@ mfxStatus VAAPIEncoder::QueryStatus(
...
+
+    vaSts = vaQuerySurfaceStatus(m_vaDisplay, waitSurface, &surfSts);
```
dvrogozh, line 1330: Is this call attempt to resolve https://github.com/Intel-Media-SDK/MediaSDK/issues/2392? If yes, then this is incorrect fix for the reason that `vaQuerySurfaceStatus` is an optional call. Effectively, if vaSyncBuffer (or surface) returned OK, then user can proceed and everything should work. `vaQuerySurfaceStatus` is kind of alternative way to sync in a polling fashion.

If this code really fix the issue then you make some machine state update in `vaQuerySurfaceStatus` which is missed in `vaSyncBuffer`  - and this is an actual bug.

dvrogozh (changes requested):

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
IlyaSergeyev: Yes, this resolves #2392

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
IlyaSergeyev: done

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
dvrogozh: Ok. As I noted, this looks to be a wrong fix. It merely highlights an issue in the driver than properly resolves it. Which operations are hidden inside vaQuerySurfaceStatus? why they were not executed within vaSyncBuffer?

dvrogozh (changes requested):

_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_vaapi.cpp
IlyaSergeyev: There are the same difference as between syncSurface and syncBuffer - waiting for pCurrentFrameSemaphore

dmitryermilov: It doesn't make sense from MSDK perspective. Here MSDK VP9e code waits for output resource - bitstream.

IlyaSergeyev: syncSurface actually do DdiMedia_StatusCheck first, then the same as syncBuffer

PR #2531, opened 2020-12-15
Add the conversion from UMC::GRAY to MFX_FOURCC_YUV444

wangyan42164:
There is only the conversion from MFX_FOURCC_YUV444 to UMC::GRAY.
It will cause UMC:GRAY is converted to MFX_FOURCC_NV12 and fire
the assertion in debug build.

### DIFF
--- a/_studio/mfx_lib/shared/src/mfx_common_decode_int.cpp
+++ b/_studio/mfx_lib/shared/src/mfx_common_decode_int.cpp
@@ -255,6 +255,7 @@ mfxU32 ConvertUMCColorFormatToFOURCC(UMC::ColorFormat format)
         case UMC::YUV411:  return MFX_FOURCC_YUV411;
         case UMC::YUV444:  return MFX_FOURCC_YUV444;
         case UMC::UYVY:    return MFX_FOURCC_UYVY;
+        case UMC::GRAY:    return MFX_FOURCC_YUV400;
 
         default:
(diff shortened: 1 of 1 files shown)


wangyan42164:
@eponomar Could you please check CI build. buildbot/test and buildbot/test-api-next error seems not be related with my PR. It failed when encoding test. Thanks.

dmitryermilov:
Hi @wangyan42164 , Eugeny is on vacation. @aukhina is a backup. I also worry about broken CI. It blocks other commits as well. We'll have a meeting today to address it. Current suspicious is it's due to driver update, this commit https://github.com/intel/media-driver/commit/533ceb90eab54d89a37222abadbb483cc981737d .But it should be double check and if so , it should be routed to Carl.

wangyan42164:
> Hi @wangyan42164 , Eugeny is on vacation. @aukhina is a backup. I also worry about broken CI. It blocks other commits as well. We'll have a meeting today to address it. Current suspicious is it's due to driver update, this commit [intel/media-driver@533ceb9](https://github.com/intel/media-driver/commit/533ceb90eab54d89a37222abadbb483cc981737d) .But it should be double check and if so , it should be routed to Carl.
@dmitryermilov Sure. I can sync this with Carl tomorrow.

dmitryermilov (approved):

PR #2547, opened 2020-12-22
[HEVCe] Hard code caps report for YUV422ReconSupport

JeremyShang:


dmitryermilov:
@JeremyShang , why is it a draft?

JeremyShang:
> @JeremyShang , why is it a draft?

Because it's not changing the current hardcode caps report value(stay as 0), this change just to make it friendly readable, not sure this PR is needed. It's under discussion.

PR #2555, opened 2020-12-31
[samples]: Enable vacopy for sample_encode and sample_decode.

ZhongXiaoxia:
Add "-vacopy" option to command line for sample_encode and sample_decode
to use vaapi vaCopy interface for high efficiency raw data copy between
VA surface and user buffer.

Sample usage:
./sample_decode h264 -i in.264 -o out.yuv -n 5 -hw -vaapi -vacopy
./sample_encode h264 -i in.yuv -o out.264 -w 1280 -h 720 -n 5 -nv12 -hw -vaapi -vacopy

Signed-off-by: Zhong, Xiaoxia <xiaoxia.zhong@intel.com>

### DIFF
--- a/samples/sample_common/include/vaapi_utils.h
+++ b/samples/sample_common/include/vaapi_utils.h
@@ -114,6 +114,10 @@ namespace MfxLoader
                                                   VAConfigID config_id);
         typedef VAStatus (*vaDestroyContext_type) (VADisplay dpy,
                                                    VAContextID context);
+        typedef VAStatus (*vaCopy_type) (VADisplay dpy,
+                                         VACopyObject * dst,
+                                         VACopyObject * src,
+                                         VACopyOption option);
 
         VA_Proxy();
         ~VA_Proxy();
@@ -139,6 +143,7 @@ namespace MfxLoader
         const vaCreateContext_type          vaCreateContext;
         const vaDestroyConfig_type          vaDestroyConfig;
         const vaDestroyContext_type         vaDestroyContext;
+        const vaCopy_type					vaCopy;
     };
 #endif
 
@@ -430,6 +435,18 @@ namespace MfxLoader
 
 } // namespace MfxLoader
 
+typedef struct _SurfaceInfo
+{
+    uint32_t    width;
+    uint32_t    height;
+    uint32_t    fourCC;
+    uint32_t    format;
+    uint32_t    memtype;
+    uint32_t    alignsize;
+    void        *pBuf;
+    uint8_t     *pBufBase;
+    uintptr_t   ptrb;
+} SurfaceInfo;
 
 class CLibVA
 {
@@ -448,6 +465,22 @@ public:
         VASurfaceID /*srf1*/,
         VADisplay dpy2,
         VASurfaceID srf2);
+
+#if defined(LIBVA_DRM_SUPPORT)
+	VAStatus AcquireUserSurface(
+		VADisplay va_dpy,
+		VASurfaceID *p_surface_id,
+		SurfaceInfo &surf);
+
+	VAStatus ReleaseUserSurface(
+		VADisplay va_dpy,
+		VASurfaceID *p_surface_id);
+
+	VAStatus VACopy(VADisplay va_dpy,
+		VASurfaceID in_surface_id,
+		VASurfaceID out_surface_id,
+		mfxI32 vacopy_mode);
+#endif
 
     inline int getBackendType() { return m_type; }
     VADisplay GetVADisplay() { return m_va_dpy; }
--- a/samples/sample_common/src/vaapi_device.cpp
+++ b/samples/sample_common/src/vaapi_device.cpp
@@ -419,11 +419,30 @@ CHWDevice* CreateVAAPIDevice(void)
 CVAAPIDeviceDRM::CVAAPIDeviceDRM(const std::string& devicePath, int type)
     : m_DRMLibVA(devicePath, type)
     , m_rndr(NULL)
+    , m_sysSurfaceID(VA_INVALID_SURFACE)
+
 {
+    memset(&m_sysSurfInfo, 0, sizeof(m_sysSurfInfo));
 }
 
 CVAAPIDeviceDRM::~CVAAPIDeviceDRM(void)
 {
+    VAStatus va_status = VA_STATUS_ERROR_INVALID_PARAMETER;
+
+    VADisplay dpy = m_DRMLibVA.GetVADisplay();
+
+    if (NULL != m_sysSurfInfo.pBuf) {
+        free(m_sysSurfInfo.pBuf);
+        m_sysSurfInfo.pBuf = NULL;
+    }
+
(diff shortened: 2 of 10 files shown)


samples/sample_common/src/vaapi_device.cpp
```diff
@@ -454,6 +473,88 @@ mfxStatus CVAAPIDeviceDRM::RenderFrame(mfxFrameSurface1 * pSurface, mfxFrameAllo
...
+         case MFX_FOURCC_NV12:
+             m_sysSurfInfo.fourCC = VA_FOURCC_NV12;
+             m_sysSurfInfo.format = VA_FOURCC_NV12;
+             break;
+         default :
+             return MFX_ERR_UNSUPPORTED;
+        }
+
+        m_sysSurfInfo.memtype = VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR;
+        m_sysSurfInfo.alignsize = 128;
```
SofyaBalandina, line 503: 128 looks incomprehensible, please add a named variable created here or or more globally like enum

samples/sample_decode/src/pipeline_decode.cpp
```diff
@@ -1566,12 +1595,21 @@ mfxStatus CDecodingPipeline::DeliverOutput(mfxFrameSurface1* frame)
 
     if (m_bExternalAlloc) {
         if (m_eWorkMode == MODE_FILE_DUMP) {
-            res = m_pGeneralAllocator->Lock(m_pGeneralAllocator->pthis, frame->Data.MemId, &(frame->Data));
```
SofyaBalandina, line 1569: Shouldn't we execute this line in case of VACopy?

aobolensk (commented):
1) This change does not work on gen9 and gen11.
2) Please, fix commit message since -vacopy flag accepts number after it: `-vacopy 0`

dmitryermilov:
@ZhongXiaoxia , please also say in https://github.com/intel/media-driver/issues/1123 if you expected that vaCopy should work on all platforms.

samples/sample_common/include/vaapi_utils.h
```diff
@@ -430,6 +435,18 @@ namespace MfxLoader
 
 } // namespace MfxLoader
 
+typedef struct _SurfaceInfo
```
onabiull, line 438: Please use just struct. Typedef is only reasonable for shared C/C++ headers. But here we have C++-only code.

samples/sample_encode/src/pipeline_encode.cpp
```diff
@@ -2009,18 +2014,29 @@ void CEncodingPipeline::FreeFileWriters()
...
-            mfxStatus sts = m_pMFXAllocator->Lock(m_pMFXAllocator->pthis, surface->Data.MemId, &surface->Data);
-            MSDK_CHECK_STATUS(sts, "m_pMFXAllocator->Lock failed");
-            sts = m_FileReader.LoadNextFrame(surface);
-            MSDK_CHECK_STATUS(sts, "m_FileReader.LoadNextFrame failed");
-            sts = m_pMFXAllocator->Unlock(m_pMFXAllocator->pthis, surface->Data.MemId, &surface->Data);
-            MSDK_CHECK_STATUS(sts, "m_pMFXAllocator->Unlock failed");
+#if defined(LIBVA_DRM_SUPPORT)
+            if (m_nVACopy != -1) // vacopy mode
+            {
+                sts = LoadFrameWithVACopy(surface, m_nVACopy);
```
onabiull, line 2027: What if X11 backend was selected?

onabiull (changes requested):
Please answer comments and update https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/samples/readme-encode_linux.md with description of the new parameter.

dvrogozh:
Why do we introduce new command line option at the first place?

dmitryermilov:
> 1. This change does not work on gen9 and gen11.
> 2. Please, fix PR message since -vacopy flag accepts number after it: `-vacopy 0`

@ZhongXiaoxia , please also say in intel/media-driver#1123 if you expected that vaCopy should work on all platforms.

ZhongXiaoxia:
> > 1. This change does not work on gen9 and gen11.
> > 2. Please, fix PR message since -vacopy flag accepts number after it: `-vacopy 0`
> 
> @ZhongXiaoxia , please also say in [intel/media-driver#1123](https://github.com/intel/media-driver/issues/1123) if you expected that vaCopy should work on all platforms.

Actually, I only need this feature on DG1/SG1 according to meet customer's requirement.

dmitryermilov:
> > > 1. This change does not work on gen9 and gen11.
> > > 2. Please, fix PR message since -vacopy flag accepts number after it: `-vacopy 0`
> > 
> > 
> > @ZhongXiaoxia , please also say in [intel/media-driver#1123](https://github.com/intel/media-driver/issues/1123) if you expected that vaCopy should work on all platforms.
> 
> Actually, I only need this feature on DG1/SG1 according to meet customer's requirement.

Okay. But we have to focus on meeting requirements of wide range or customers.  Also taking into account others concern that vaCopy doesn't need separate cmd setting we can't accept the change.

samples/sample_encode/src/pipeline_encode.cpp
ZhongXiaoxia: This PR only implemented for DRM backend according to customer's requirement. For other backends, this option will be disabled by default. If we need to support X11 backend, we may add vaCopy related functions to CVAAPIDeviceX11 in future.

samples/sample_decode/src/pipeline_decode.cpp
ZhongXiaoxia: No, we needn't. This called function will execute vaMapBuffer and vaDeriveImage to map data between VA surface and user buffer. But VACopy doesn't need vaMapBuffer or vaDeriveImage. It copies data between two surfaces directly.

ZhongXiaoxia:
> > > > 1. This change does not work on gen9 and gen11.
> > > > 2. Please, fix PR message since -vacopy flag accepts number after it: `-vacopy 0`
> > > 
> > > 
> > > @ZhongXiaoxia , please also say in [intel/media-driver#1123](https://github.com/intel/media-driver/issues/1123) if you expected that vaCopy should work on all platforms.
> > 
> > 
> > Actually, I only need this feature on DG1/SG1 according to meet customer's requirement.
> 
> Okay. But we have to focus on meeting requirements of wide range or customers. Also taking into account others concern that vaCopy doesn't need separate cmd setting we can't accept the change.

As we see, current vaCopy can't support all platforms. It's implemented for Gen12 specially.  And I only verified it with DRM backend on DG1/SG1. I'm not sure whether it can work for other configuration or environment. It's not safe to reuse and replace the previous processing in encoding and decoding samples. Thus, just like what we did for gpucopy in samples, I added a new command option to avoid of breaking previous functions.

dvrogozh:
> It's implemented for Gen12 specially.

PR #2563, opened 2021-01-12
Disable vp8 hw decoding from TGL platform

wangyan42164:


_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_common.cpp
```diff
@@ -165,6 +165,9 @@ mfxStatus MFX_VP8_Utility::Query(VideoCORE *p_core, mfxVideoParam *p_in, mfxVide
         return Query(p_core, &in1, p_out, type);
     }
 
+    if (p_core->GetPlatformType() == MFX_PLATFORM_HARDWARE && p_core->GetHWType() >= MFX_HW_TGL_LP)
```
dvrogozh, line 168: emmm. Here: https://github.com/intel/media-driver#decodingencoding-features, we see "VP8 decoding is only supported on TGL platform". So, why do you disable something which is supposed to work and which landed into the releases? I don't think we can do that now.

dvrogozh (changes requested):

PR #2567, opened 2021-01-13
[smt] Fix ext allocator parameters for multi GPUs case

ndomnina:
Since the allocator parameters (including device handle) for all
sessions were taken from the 0th session, the smt options -iGfx
and -dGfx for the 1st and subsequent sessions were ignoring

### DIFF
--- a/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
+++ b/samples/sample_multi_transcode/src/sample_multi_transcode.cpp
@@ -110,7 +110,9 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
     SafetySurfaceBuffer* pBuffer = NULL;
     mfxU32 BufCounter = 0;
     mfxHDL hdl = NULL;
+    std::vector<mfxHDL> hdls;
     sInputParams    InputParams;
+    bool bNeedToCreateDevice = true;
 
     //parent transcode pipeline
     CTranscodingPipeline *pParentPipeline = NULL;
@@ -125,7 +127,6 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
         // There's no error in parameters parsing, but we should not continue further. For instance, in case of -? option
         return sts;
     }
-
 
     // get parameters for each session from parser
     while(m_parser.GetNextSessionParams(InputParams))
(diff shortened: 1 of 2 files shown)


dmitryermilov (approved):

dmitryermilov (dismissed):

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -149,146 +150,194 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
...
+                pD3DParams->pManager = (IDirect3DDeviceManager9*)hdl;
+
+                m_pAllocParams.push_back(std::shared_ptr<mfxAllocatorParams>(pAllocParam));
+                m_hwdevs.push_back(std::move(hwdev));
+                hdls.push_back(hdl);
+            }
+            else
+            {
+                m_pAllocParams.push_back(m_pAllocParams.back());
+                hdls.push_back(hdls.back());
```
onabiull, line 198: Not sure that I've get the idea here. We creating a copy of shared pointer, thus pinning object. Btw, are we sure, that we have at least one element in both m_pAllocParams and hdls?

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -149,146 +150,194 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
...
 
+                m_pAllocParams.push_back(std::shared_ptr<mfxAllocatorParams>(pAllocParam));
+                m_hwdevs.push_back(std::move(hwdev));
+                hdls.push_back(hdl);
             }
-            else if (params.libvaBackend == MFX_LIBVA_X11)
+            else
             {
-                pVAAPIParams->m_export_mode = vaapiAllocatorParams::PRIME;
+                m_pAllocParams.push_back(m_pAllocParams.back());
```
onabiull, line 238: Same as above.

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
```diff
@@ -149,146 +150,194 @@ mfxStatus Launcher::Init(int argc, msdk_char *argv[])
...
-        }
-        else /* NO RENDERING*/
-        {
-            m_hwdev.reset(CreateVAAPIDevice(InputParams.strDevicePath));
-            if(!m_hwdev.get()) {
-                msdk_printf(MSDK_STRING("error: failed to initialize VAAPI device\n"));
-                return MFX_ERR_DEVICE_FAILED;
+            else
+            {
+                m_pAllocParams.push_back(m_pAllocParams.back());
```
onabiull, line 328: Same as above.

onabiull (changes requested):

samples/sample_multi_transcode/src/sample_multi_transcode.cpp
ndomnina: > Not sure that I've get the idea here. We creating a copy of shared pointer, thus pinning object.

Yes, that's right. For example, in the case of 1-N, the object for alloc parameters will be created once for the zero session. For the remaining sessions, this object will be reused. Instead of creating new objects, the number of references to the first one will increase.

> Btw, are we sure, that we have at least one element in both m_pAllocParams and hdls?

I added checks

onabiull (dismissed):

onabiull:
As 6f32f07a6fe31efe0feb2b0674610c3bf429d0eb merged, rebase may fix clang-11 check.

ndomnina:
> As [6f32f07](https://github.com/Intel-Media-SDK/MediaSDK/commit/6f32f07a6fe31efe0feb2b0674610c3bf429d0eb) merged, rebase may fix clang-11 check.

Thanks, rebase helped

dmitryermilov (approved):

PR #2570, opened 2021-01-15
[smt] Fix the unsupported issue of transcoding from I420/nv12 to HEVC

dylanintel:
1. MFX_ERR_UNSUPPORTED will be reported when transcoding from I420/nv12 to HEVC.
2. It will print MFX_ERR_MORE_DATA, but this error doesn't effect the result, So fix it.

Issue: VCD-2398, VCD-2486

Signed-off-by: Wang Dylan <dylan.wang@intel.com>

### DIFF
--- a/samples/sample_multi_transcode/src/pipeline_transcode.cpp
+++ b/samples/sample_multi_transcode/src/pipeline_transcode.cpp
@@ -550,7 +550,8 @@ mfxStatus CTranscodingPipeline::DecodeOneFrame(ExtendedSurface *pExtSurface)
         {
             pExtSurface->pSurface = GetFreeSurface(false, MSDK_SURFACE_WAIT_INTERVAL);
             sts = m_pBSProcessor->GetInputFrame(pExtSurface->pSurface);
-            MFX_CHECK_STS(sts);
+            if(sts != MFX_ERR_NONE)
+                return sts;
         }
         else if (MFX_WRN_DEVICE_BUSY == sts)
         {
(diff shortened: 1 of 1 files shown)


dylanintel:
hi @onabiull , There is a failing check about cmake. Could you help me to fix the cmake issue please? Thanks.

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -550,7 +550,8 @@ mfxStatus CTranscodingPipeline::DecodeOneFrame(ExtendedSurface *pExtSurface)
         {
             pExtSurface->pSurface = GetFreeSurface(false, MSDK_SURFACE_WAIT_INTERVAL);
             sts = m_pBSProcessor->GetInputFrame(pExtSurface->pSurface);
-            MFX_CHECK_STS(sts);
+            if(sts != MFX_ERR_NONE)
```
dmitryermilov, line 553: @dylanintel , could you do it via "MSDK_BREAK_ON_ERROR(sts);" ?

dmitryermilov (commented):
LGTM but could you fix one comment?

onabiull:
Strangely enough clang-11 disappeared from ubuntu-latest. @dvrogozh did you met such issue previously? PR is actual, in a sense that it is based on 77f673853c23a9124174b1e4981b0785b03922db - current HEAD@master

samples/sample_multi_transcode/src/pipeline_transcode.cpp
dylanintel: hi @dmitryermilov , The issue is that sample_multi_transcode don't  support from i420/nv12 to hevc, m_mfxDecParams.mfx.FrameInfo.PicStruct = MFX_PICSTRUCT_PROGRESSIVE; The code aims to solve this issue. In addition, there is another issue, it will print the MFX_ERR_MORE_DATA error, But the task could be finished successfully. The root cause is that it will return  MFX_ERR_MORE_DATA when reading the end of yuv file. I think it is the normal phenomenon. Because it will deal with the error and set the bEndOfFile as true in CTranscodingPipeline::Transcode function. The function MFX_CHECK_STS will print the error if the status !=  MFX_ERR_NONE. So I  return an error instead of printing and returning an error.

dmitryermilov (approved):

PR #2577, opened 2021-01-22
[sample_encode] SW HEVCe encodes NV12 for YUY2 input

Alexandr-Konovalov:
VPP is used under the hood.

Issue: MDP-65557


dmitryermilov (changes requested):
Shouldn't it be doe this way:
https://github.com/Intel-Media-SDK/MediaSDK/pull/2575 
?

Alexandr-Konovalov:
> Shouldn't it be doe this way:
> #2575
> ?

@dmitryermilov , good point. If someone need 4:2:0, one must ask explicitly via -ec::nv12. So, let's abandon.

Alexandr-Konovalov:
Superseded by -ec::nv12.

PR #2600, opened 2021-02-19
Default memtype is system for file dumping

dmitryermilov:

### DIFF
--- a/samples/sample_decode/include/pipeline_decode.h
+++ b/samples/sample_decode/include/pipeline_decode.h
@@ -57,9 +57,9 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #endif
 
 enum MemType {
-    SYSTEM_MEMORY = 0x00,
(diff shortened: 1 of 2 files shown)


samples/sample_decode/include/pipeline_decode.h
```diff
@@ -57,9 +57,9 @@ or https://software.intel.com/en-us/media-client-solutions-support.
 #endif
 
 enum MemType {
-    SYSTEM_MEMORY = 0x00,
-    D3D9_MEMORY   = 0x01,
-    D3D11_MEMORY  = 0x02,
+    SYSTEM_MEMORY = 0x01,
```
dvrogozh, line 60: what was the reason to +1 values?

dmitryermilov: please see change below. To detect if user set some mem type value ("pParams->memType").

onabiull (approved):

PR #2612, opened 2021-03-05
Add MFX_DEPRECATED and add deprecation attributes in API headers

aobolensk:


dmitryermilov (changes requested):
@aobolensk , @dvrogozh ,@dgurulev I think we need to mark this API as deprecated since 1.35.  By this PR, it seems we mark them deprecated in 1.34.

onabiull (dismissed):
Technically this change do not change API. No new features, no new structures, no even new return codes from existing functions. So arguably this change may be treated as part of 1.34.

api/include/mfxdefs.h
```diff
@@ -103,6 +103,52 @@ extern "C"
 
 #define MFX_INFINITE 0xFFFFFFFF
 
+#if !defined(MFX_DEPRECATED_OFF) && (MFX_VERSION < MFX_VERSION_NEXT)
+#define MFX_DEPRECATED_OFF
+#endif
+
+#ifndef MFX_DEPRECATED_OFF
```
dvrogozh, line 110: I don't understand how samples got built successfully in our API=latest CI configuration since they should have stepped into warnings, but it seems they did not. Did `-DMFX_DEPRECATED_OFF` in dispatcher influenced samples, tools and totorials build?

dvrogozh (changes requested):

api/include/mfxdefs.h
aobolensk: Samples and tutorials have "-Wno-deprecated-declarations" flag. That's why build doesn't fail.

PR #2601, opened 2021-02-20
[HEVCe] Fix temporal layers reset issue

JeremyShang:

PR #2615, opened 2021-03-08
[HEVCe] Fix temporal layers reset issue

JeremyShang:
Previously, #2601 got reverted due to incorrect option inherting condition for Scale

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2020 Intel Corporation
+// Copyright (c) 2019-2021 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


JeremyShang:
The condition to inherit option for Layer[i].Scale from init par to reset par:  Layer structure is empty (all zero)

_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy.cpp
```diff
@@ -539,6 +542,30 @@ void Legacy::SetInherited(ParamInheritance& par)
...
+        bool bEmptyLayer = true;
+        for (mfxU32 i = 0; i < 7; i++)
+        {
+            if (ebReset.Layer[i].Scale > 0) bEmptyLayer = false;
+        }
+        if (bEmptyLayer)
+        {
+            for (mfxU32 i = 0; i < 7; i++)
+            {
+                INHERIT_OPT(Layer[i].Scale);
```
vilichev, line 564: Do we really need to inherit init values on reset?

JeremyShang: In case when we reset other parameter, if Layer.Scale in Init Par is 1,2,0,0,0,0 for example, and not inherited to reset par. Then, we can't see temporal layers during reset, is that expected?

dmitryermilov: @JeremyShang , do you plan to address the comments?

dmitryermilov: @JeremyShang  ?

JeremyShang: Hi @dmitryermilov, sorry for late response. I have offline talk with them, it is concluded that it is not a bug to set AvcTemporalLayers in every reset, so I think I can close this PR. Please leave comments if you still have any. Thanks!

PR #2617, opened 2021-03-09
[GH Actions] Added linux build with clang 12

eponomar:


.github/workflows/ubuntu.yml
```diff
@@ -6,6 +6,70 @@ env:
   CFLAGS: -O2 -Wformat -Wformat-security -Wall -Werror -D_FORTIFY_SOURCE=2 -fstack-protector-strong
 
 jobs:
+  clang12-api-latest:
```
dvrogozh, line 9: @onabiull : do we need to be over paranoid and have both clang-11 and clang-12 at the same time? I initially thought to just switch clang-11 configuration to clang-12...

dvrogozh (changes requested):

.github/workflows/ubuntu.yml
eponomar: In this case we need remove clang 10 as well...
I thought that we only ADD builds when found it.

dvrogozh: After offline talk: we'll go with modifying clang-11 configuration

eponomar: DONE

dvrogozh (approved):

dmitryermilov (approved):

PR #2619, opened 2021-03-12
vaapi: use VA FourCC codes

xhaihao:
It is to test the FourCC code of a VA Image in the switch statement


dmitryermilov (commented):
Thanks for the patch, @xhaihao !

dmitryermilov (approved):

PR #2621, opened 2021-03-15
[MPEGe][Linux] pass target bitrate to driver for VBR

lakulako:
Now bitrate in VBR mode is close to target, but not to maximum.


dmitryermilov (approved):

uartie:
This patch regresses MPEG CQP encode quality (low PSNR and visual artifacts) when using ffmpeg-qsv and gst-msdk.  I will file issue soon...

PR #2635, opened 2021-03-24
[hevce] Disable bit depth conversion

yskuridi:
VME driver doesn't support 10b->8b and 8b->10b conversions.

Issue: MDP-65185
Test: manually

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
@@ -2569,6 +2569,14 @@ public:
         CheckBD(pCO3->TargetBitDepthLuma == 10);
 
         MFX_CHECK(!invalid, MFX_ERR_UNSUPPORTED);
+
+        auto&  fi = par.mfx.FrameInfo;
+        mfxU32 changed = 0;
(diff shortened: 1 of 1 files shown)


dmitryermilov:
@thomasli21801 , @vilichev , please review

dmitryermilov (dismissed):

thomasli21801 (changes requested):
Hi Yuliya and Dmitry,
Bit depth conversion can be supported by HEVC VDEnc.
Can you help to make the change for VME specific?

Thanks,
Thomas

dmitryermilov:
@thomasli21801 , you said HW can support this. but you also mentioned it's not covered in validation. So do suggest to allow VDEnc case here (even it's not validated)? Just want to be on the same page.

dmitryermilov:
> > @thomasli21801 , you said HW can support this. but you also mentioned it's not covered in validation. So do suggest to allow VDEnc case here (even it's not validated)? Just want to be on the same page.
> 
> Hi Dmitry,
> Yes. HW can support this and it's not covered in validation since we thought there was no usage. But I would suggest to keep VDEnc to provide the ability to community and support bug fixing if any. (might be low risk as well).

Thanks, Thomas. It's fine to me.

dmitryermilov:
@yskuridi , LGTM but can you please update the comment that this change for VME only.

JeremyShang (approved):

vilichev (approved):

thomasli21801 (approved):

PR #2650, opened 2021-04-01
[jpegd] Check that number of scans is not exceeding maximum

filippenkos:


_studio/shared/umc/codec/jpeg_dec/src/mfx_mjpeg_task.cpp
```diff
@@ -249,6 +249,10 @@ mfxStatus CJpegTask::AddPicture(UMC::MediaDataEx *pSrcData,
             m_pics[m_numPic]->scanOffset[numScans] = pAuxData->offsets[i];
             m_pics[m_numPic]->scanSize[numScans] = chunkSize;
             numScans += 1;
+            if (numScans >= maxNumScans)
+            {
+                return MFX_ERR_MORE_DATA;
```
wangyan-intel, line 254: MFX_ERR_MORE_DATA is OK for this situation? BTW, is there JIRA for tracking this issue? Thanks.

filippenkos: This status will be returned from DecodeFrameAsync eventually. As per documentation we are limited with it's returned status. I asked Dmitry Ermilov and he said that we should return MFX_ERR_MORE_DATA in this case. There is no JIRA for this issue currently.

dmitryermilov: > I asked Dmitry Ermilov and he said that we should return MFX_ERR_MORE_DATA in this case.

@filippenkos , I said another thing. I meant that when decoder detects fully wrong syntax, we skip such unit (SPS/PPS, slice, etc) and look for another unit  in the provided bitstream. If we parsed the whole bitstream object when decoder returns MFX_ERR_MORE_DATA.

filippenkos: @dmitryermilov , sorry, may be I misunderstood your words. This function call is wrapped in MFX_SAFE_CALL and then this result goes straight to MFX_CHECK in DecodeFrameAsync, so any error code we return from this function will be immediately returned by DecodeFrameAsync. What do you suggest to return in this case?

dmitryermilov: > so any error code we return from this function will be immediately returned by DecodeFrameAsync

"immediately" .. How it will work if mfxBitstream contains more than 1 jpeg picture? Do you skip 2nd one?

_studio/mfx_lib/decode/mjpeg/src/mfx_mjpeg_dec_decode.cpp
```diff
@@ -857,7 +857,17 @@ mfxStatus VideoDECODEMJPEG::DecodeFrameCheck(mfxBitstream *bs, mfxFrameSurface1
         m_isHeaderFound = false;
         m_isHeaderParsed = false;
 
-        MFX_SAFE_CALL(decoder->AddPicture(pSrcData, numPic));
+        try
+        {
+            MFX_SAFE_CALL(decoder->AddPicture(pSrcData, numPic));
+        }
+        catch(UMC::eUMC_Status sts)
```
dmitryermilov, line 864: Let's do it this way:
```
catch(const UMC::eUMC_Status & sts)
{
if (UMC::UMC_ERR_INVALID_STREAM == sts)
    continue
else
    return sts;
}

dmitryermilov (approved):

_studio/shared/umc/codec/jpeg_dec/src/mfx_mjpeg_task.cpp
dmitryermilov: Now LGTM . @wangyan42164 , I appreciate your review.

wangyan-intel: @dmitryermilov No problem. :)

wangyan-intel (approved):

PR #2653, opened 2021-04-02
Add error message for trying to run av1d with d3d9

SofyaBalandina:
Issue: None
Test: manually

### DIFF
--- a/samples/sample_decode/src/pipeline_decode.cpp
+++ b/samples/sample_decode/src/pipeline_decode.cpp
@@ -441,6 +441,14 @@ mfxStatus CDecodingPipeline::Init(sInputParams *pParams)
     sts = m_mfxSession.QueryIMPL(&m_impl); // get actual library implementation
     MSDK_CHECK_STATUS(sts, "m_mfxSession.QueryIMPL failed");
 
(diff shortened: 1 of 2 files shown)


samples/sample_decode/src/sample_decode.cpp
```diff
@@ -749,6 +749,14 @@ mfxStatus ParseInputString(msdk_char* strInput[], mfxU8 nArgNum, sInputParams* p
     }
 #endif
 
+#if defined(_WIN32) || defined(_WIN64)
+    if (pParams->videoType == MFX_CODEC_AV1 && pParams->memType == D3D9_MEMORY)
```
dmitryermilov, line 753: It will fail even with system memory but if MSDK session was initialized dx9 path. Can you please move this check to the place right after `mfxInit(); mfxQueryIMPL(&impl_type);`  :
e.g.
```
mfxInit()
mfxQueryIMPL(&impl_type);
if (AV1 && (impl_type | DX9)) {
  printf("blah blah blah\n");
  return error;
}
```

dmitryermilov (approved):

dmitryermilov (approved):

PR #2669, opened 2021-04-21
Fixed HEVCe interlace in sample_encode

dmitryermilov:
sample_encode h265 -i <yuv> -w 720 -h 480 -b 1500 -tff -extbrc:on


Fantast34:
Hi @dmitryermilov , why HEVCe interlace works with extbrc only? Our customers uses AVC/HEVC encoding with hardware ratecontrol including interlace cases. At this moment we are using patch for just removing this restriction. We are expecting the same behavior as for AVC. Also extbrc disabled at all in oneVPL sample, so interlace encoding is impossible there.

dmitryermilov:
Because HW is not aware of interlace in HEVC. Even in the spec interlace is defined as a sort of extension on top of regular HEVC progressive.  That's why there's a field splitting VPP filter, BTW.
As for the VPL samples, please open at issue (feature request) at https://github.com/oneapi-src/oneVPL against VPL samples. VPL run-time library supports extBRC.

PR #2676, opened 2021-05-04
Set frame rate from IVF to decoder's params

alexelizarov:


samples/sample_common/src/sample_utils.cpp
```diff
@@ -779,6 +779,7 @@ mfxStatus CJPEGFrameReader::ReadNextFrame(mfxBitstream *pBS)
 CIVFFrameReader::CIVFFrameReader()
 {
     MSDK_ZERO_MEMORY(m_hdr);
+    MSDK_ZERO_MEMORY(m_hInfo);
```
mgonchar, line 782: nit: I'd rather added default in-place initialization for ```BitrstreamHeaderInfo``` class

mgonchar (approved):

samples/sample_common/src/sample_utils.cpp
alexelizarov: Done

samples/sample_multi_transcode/src/pipeline_transcode.cpp
```diff
@@ -2265,15 +2265,14 @@ mfxStatus CTranscodingPipeline::InitDecMfxParams(sInputParams *pInParams)
     {
         ConvertFrameRate(pInParams->dDecoderFrameRateOverride, &m_mfxDecParams.mfx.FrameInfo.FrameRateExtN, &m_mfxDecParams.mfx.FrameInfo.FrameRateExtD);
     }
-    // if frame rate not specified and input stream header doesn't contain valid values use default (30.0)
-    else if (0 == (m_mfxDecParams.mfx.FrameInfo.FrameRateExtN * m_mfxDecParams.mfx.FrameInfo.FrameRateExtD))
+    else if(0 == (m_mfxDecParams.mfx.FrameInfo.FrameRateExtN * m_mfxDecParams.mfx.FrameInfo.FrameRateExtD))
```
dmitryermilov, line 2268: Isn't it a bug in decoder implementation that reported (from DecodeHeader) frame rate is zero while container has some frame rate?

alexelizarov: DecodeHeader can't return frame rate from IVF because IVF splitted in CIVFFrameReader::ReadHeader.
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/samples/sample_common/src/sample_utils.cpp#L804

dmitryermilov: No, I mean that decode DecodeHeader should report actual frame rate value, if it's in syntax bitstream, and some default (like 30/0) value if it's absent . And I see AV1 decode even doesn't try: 
https://github.com/Intel-Media-SDK/MediaSDK/blob/a9069d858fb044f5f3c911ca6210f6e9f76b80eb/_studio/shared/umc/codec/av1_dec/src/umc_av1_bitstream.cpp#L84-L85

This info is not to fill FrameRateN/D.
The proposed PR is rather a workaround. I propose to have a proper fix in component at first

dmitryermilov (changes requested):
see comments

samples/sample_multi_transcode/src/pipeline_transcode.cpp
alexelizarov: AV1's DecodeHeader even return 1 for frame rate. https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/shared/src/mfx_common_decode_int.cpp#L326

If it returned 0 than sample could convert it to default. But AV1 decoder occasionally set frame rate to 1/1. I agree that it is bug in decoder.

But sample also have an issue.
The priority for frame rate should be:
1) -override_decoder_framerate cmd parameter
2) frame rate in bitstream if 1 doesn't exist
3) frame rate in IVF if 1 and 2 doesn't exist
4) default value 30/1 if 1, 2 and 3 doesn't exist
Now sample completely ignore frame rate from IVF.
And I think it can be independent fixes.

dmitryermilov: @wangyan-intel , can someone from decoder team look at AV1 framerate reporting issue?

dmitryermilov (approved):

samples/sample_multi_transcode/src/pipeline_transcode.cpp
wangyan-intel: > @wangyan-intel , can someone from decoder team look at AV1 framerate reporting issue?

I will take a look and assign to someone. Thanks.

mgonchar (approved):

PR #2679, opened 2021-05-08
Reworked recent fix for "VDBOX SFC output frame order issue"

shawnli2:
A proper fix is to keep assignment(I mean the place where umc decoder starts
using the surfaces) of internal surface (for original) resolution and of
external surface (for resized resolution) in the same function
(which actually means at the same timeframe).


wangyan-intel (approved):

wangyan-intel:
@dmitryermilov Could you please review this PR? Thanks.

dmitryermilov (approved):

PR #2705, opened 2021-06-02
[HEVCE] enable GPU weighted prediction

Tianhaol:
Enable GPU weighted prediction by passing size and offset of weighted table
and the flag.


dmitryermilov:
@thomasli21801 , please take a look

thomasli21801 (approved):

Issue #2597, opened 2021-02-11
Fill HDR Side Data for HEVC Decoding


softworkz:
Structures for HDR side data have recently been added for HEVC encoding: https://github.com/Intel-Media-SDK/MediaSDK/commit/a9685bf658d1ec97ff44b614dcf75818ecc62819 : 

- mfxExtContentLightLevelInfo 
- mfxExtMasteringDisplayColourVolume

What missing is to fill those structures when decoding HEVC video.

Could this be added?

dmitryermilov:
@softworkz , just to be on the same page: would you like to have this reporting from DecodeHeader and together with each output frame (if there're per frame SEI), right?

softworkz:
> @softworkz , just to be on the same page: would you like to have this reporting from DecodeHeader and together with each output frame (if there're per frame SEI), right?

Correct. From all videos that I've seen so far, even with static HDR (like HDR10), the data is in every frame (SEI type 137 and 144 IIRC).

Purposes:
- Re-Encoding while keeping HDR data
  e.g. just scaling down while retaining color format and HDR (you've already done the encoding side)
- The data is required for tone mapping
  - It's a prerequisite for this: https://github.com/Intel-Media-SDK/MediaSDK/issues/1949
  - But even right now it's a problem when doing tone mapping for example with OpenCL, because QuickSync decoders (unlike VAAPI) do bitstream and codec parsing, so it's not possible to get at this data (e.g. using ffmpeg)

dmitryermilov:
Thanks for details. Yes, we can do it but please expect some delay, ~2 weeks.

softworkz:
2 weeks? That's not a delay - that would be awesome 😆

wangyan-intel:
@pengxin99 Could you please take a look? Thanks.

softworkz:
> Thanks for details. Yes, we can do it but please expect some delay, ~2 weeks.

@dmitryermilov  @pengxin99 - Is there any progress in sight on this issue?

wangyan-intel:
@softworkz Sorry for slow response. As you know, we are transferring from MSDK legacy to oneVPL (https://github.com/oneapi-src/oneVPL-intel-gpu). For legacy branch, it has been implemented by @pengxin99 with extension buffer. But this flow cannot be used in VPL architecture. We are evaluating the path and will update it ASAP.

softworkz:
[...]

PR #2726, opened 2021-07-06
[HEVCd] Added HDR SEI info parsing for HEVC decoding

pengxin99:
For #2597 

Attach the **MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME**  and **MFX_EXTBUFF_CONTENT_LIGHT_LEVEL_INFO** to outputSurface when you need this info for HEVC HDR transfer.

``` 
     mfxFrameSurface1& surf;
    // Mastering Display Colour Volume SEI info
    mfxExtMasteringDisplayColourVolume* pExtPar_MasteringDisplay = new mfxExtMasteringDisplayColourVolume{};
    pExtPar_MasteringDisplay->Header.BufferId = MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME;
    pExtPar_MasteringDisplay->Header.BufferSz = sizeof(mfxExtMasteringDisplayColourVolume);
    
    // Mastering Content Light Level info
    mfxExtContentLightLevelInfo* pExtPar_ContentLight = new mfxExtContentLightLevelInfo{};
    pExtPar_ContentLight->Header.BufferId = MFX_EXTBUFF_CONTENT_LIGHT_LEVEL_INFO;
    pExtPar_ContentLight->Header.BufferSz = sizeof(mfxExtContentLightLevelInfo);

    surf.Data.ExtParam[0] = reinterpret_cast<mfxExtBuffer*>(pExtPar_MasteringDisplay);
    surf.Data.ExtParam[1] = reinterpret_cast<mfxExtBuffer*>(pExtPar_ContentLight);
```

### DIFF
--- a/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
@@ -1384,6 +1384,26 @@ void VideoDECODEH265::FillOutputSurface(mfxFrameSurface1 **surf_out, mfxFrameSur
             info->FrameType |= MFX_FRAMETYPE_REF;
    }
 
+    mfxExtMasteringDisplayColourVolume* dispaly_colour = (mfxExtMasteringDisplayColourVolume*)GetExtendedBuffer(surface_out->Data.ExtParam, surface_out->Data.NumExtParam, MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME);
+    if (dispaly_colour && pFrame->m_mastering_display.payLoadSize > 0)
+    {
+        for (size_t i = 0; i < 3; i++)
+        {
+            dispaly_colour->DisplayPrimariesX[i] = (mfxU16)pFrame->m_mastering_display.SEI_messages.mastering_display.display_primaries[i][0];
+            dispaly_colour->DisplayPrimariesY[i] = (mfxU16)pFrame->m_mastering_display.SEI_messages.mastering_display.display_primaries[i][1];
+        }
+        dispaly_colour->WhitePointX = (mfxU16)pFrame->m_mastering_display.SEI_messages.mastering_display.white_point[0];
+        dispaly_colour->WhitePointY = (mfxU16)pFrame->m_mastering_display.SEI_messages.mastering_display.white_point[1];
+        dispaly_colour->MaxDisplayMasteringLuminance = (mfxU32)pFrame->m_mastering_display.SEI_messages.mastering_display.max_luminance;
+        dispaly_colour->MinDisplayMasteringLuminance = (mfxU32)pFrame->m_mastering_display.SEI_messages.mastering_display.min_luminance;
+    }
+    mfxExtContentLightLevelInfo* content_light = (mfxExtContentLightLevelInfo*)GetExtendedBuffer(surface_out->Data.ExtParam, surface_out->Data.NumExtParam, MFX_EXTBUFF_CONTENT_LIGHT_LEVEL_INFO);
+    if (content_light && pFrame->m_content_light_level_info.payLoadSize > 0)
+    {
+        content_light->MaxContentLightLevel = (mfxU16)pFrame->m_content_light_level_info.SEI_messages.content_light_level_info.max_content_light_level;
+        content_light->MaxPicAverageLightLevel = (mfxU16)pFrame->m_content_light_level_info.SEI_messages.content_light_level_info.max_pic_average_light_level;
+    }
+
 }
 
 // Wait until a frame is ready to be output and set necessary surface flags
--- a/_studio/shared/umc/codec/h265_dec/include/umc_h265_bitstream_headers.h
+++ b/_studio/shared/umc/codec/h265_dec/include/umc_h265_bitstream_headers.h
@@ -351,6 +351,10 @@ protected:
     int32_t pic_timing(const HeaderSet<H265SeqParamSet> & sps, int32_t current_sps, H265SEIPayLoad *spl);
     // Parse recovery point SEI data
     int32_t recovery_point(const HeaderSet<H265SeqParamSet> & sps, int32_t current_sps, H265SEIPayLoad *spl);
+    // Parse mastering display colour volume SEI data
+    int32_t mastering_display_colour_volume(const HeaderSet<H265SeqParamSet>& sps, int32_t current_sps, H265SEIPayLoad* spl);
+    // Parse content light level info SEI data
+    int32_t content_light_level_info(const HeaderSet<H265SeqParamSet>& sps, int32_t current_sps, H265SEIPayLoad* spl);
 
     // Skip unrecognized SEI message payload
     int32_t reserved_sei_message(const HeaderSet<H265SeqParamSet> & sps, int32_t current_sps, H265SEIPayLoad *spl);
--- a/_studio/shared/umc/codec/h265_dec/include/umc_h265_dec_defs.h
+++ b/_studio/shared/umc/codec/h265_dec/include/umc_h265_dec_defs.h
@@ -328,6 +328,7 @@ typedef enum
     SEI_KNEE_FUNCTION_INFO                          = 141,
     SEI_COLOUR_REMAPPING_INFO                       = 142,
     SEI_DEINTERLACED_FIELD_IDENTIFICATION           = 143,
+    SEI_CONTENT_LIGHT_LEVEL_INFO                    = 144,
 
     SEI_LAYERS_NOT_PRESENT                          = 160,
     SEI_INTER_LAYER_CONSTRAINED_TILE_SETS           = 161,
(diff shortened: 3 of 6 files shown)


_studio/shared/include/mfx_config.h
```diff
@@ -85,6 +85,10 @@
     //#define MFX_ENABLE_HEVCE_DIRTY_RECT
 #endif
 
+#if defined(MFX_ENABLE_H265_VIDEO_DECODE)
+    #define MFX_ENABLE_HEVCE_HDR_SEI
```
dmitryermilov, line 89: MFX_ENABLE_HEVC**E**_HDR_SEI ??

dmitryermilov (changes requested):

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -1384,6 +1384,38 @@ void VideoDECODEH265::FillOutputSurface(mfxFrameSurface1 **surf_out, mfxFrameSur
...
    }
 
+#ifdef MFX_ENABLE_HEVCE_HDR_SEI
+    mfxExtMasteringDisplayColourVolume* dispaly_colour = (mfxExtMasteringDisplayColourVolume*)GetExtendedBuffer(surface_out->Data.ExtParam, surface_out->Data.NumExtParam, MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME);
+    if (pFrame->m_mastering_display.payLoadSize != 0)
+    {
+        if (dispaly_colour && pFrame->m_mastering_display.payLoadSize > 0)
+        {
+            dispaly_colour->Header.BufferId = MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME;
+            dispaly_colour->Header.BufferSz = (mfxU32)pFrame->m_mastering_display.payLoadSize;
```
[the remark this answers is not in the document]
pengxin99: Have removed this in new commit.

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -1384,6 +1384,38 @@ void VideoDECODEH265::FillOutputSurface(mfxFrameSurface1 **surf_out, mfxFrameSur
             info->FrameType |= MFX_FRAMETYPE_REF;
    }
 
+#ifdef MFX_ENABLE_HEVCE_HDR_SEI
```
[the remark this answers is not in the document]
pengxin99: Have removed all **MFX_ENABLE_HEVCE_HDR_SEI** Marco for this SEI is a pure SW feature.

_studio/shared/include/mfx_config.h
pengxin99: Have been removed because there is no need for Marco protection in the pure SW feature.

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -1384,6 +1384,38 @@ void VideoDECODEH265::FillOutputSurface(mfxFrameSurface1 **surf_out, mfxFrameSur
...
+    mfxExtMasteringDisplayColourVolume* dispaly_colour = (mfxExtMasteringDisplayColourVolume*)GetExtendedBuffer(surface_out->Data.ExtParam, surface_out->Data.NumExtParam, MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME);
+    if (pFrame->m_mastering_display.payLoadSize != 0)
+    {
+        if (dispaly_colour && pFrame->m_mastering_display.payLoadSize > 0)
+        {
```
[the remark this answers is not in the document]
pengxin99: Now, the related ExtBuffer will use the default value `0` for the frame not include HDR SEI. And can do modify according to the ARB discussion

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -1384,6 +1384,33 @@ void VideoDECODEH265::FillOutputSurface(mfxFrameSurface1 **surf_out, mfxFrameSur
             info->FrameType |= MFX_FRAMETYPE_REF;
    }
 
+    mfxExtMasteringDisplayColourVolume* dispaly_colour = (mfxExtMasteringDisplayColourVolume*)GetExtendedBuffer(surface_out->Data.ExtParam, surface_out->Data.NumExtParam, MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME);
+    if (pFrame->m_mastering_display.payLoadSize != 0)
+    {
+        if (dispaly_colour && pFrame->m_mastering_display.payLoadSize > 0)
+        {
+            dispaly_colour->InsertPayloadToggle = MFX_PAYLOAD_IDR;
```
dmitryermilov, line 1392: Could you explain why it's always MFX_PAYLOAD_IDR ?

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
```diff
@@ -1384,6 +1384,33 @@ void VideoDECODEH265::FillOutputSurface(mfxFrameSurface1 **surf_out, mfxFrameSur
             info->FrameType |= MFX_FRAMETYPE_REF;
    }
 
+    mfxExtMasteringDisplayColourVolume* dispaly_colour = (mfxExtMasteringDisplayColourVolume*)GetExtendedBuffer(surface_out->Data.ExtParam, surface_out->Data.NumExtParam, MFX_EXTBUFF_MASTERING_DISPLAY_COLOUR_VOLUME);
+    if (pFrame->m_mastering_display.payLoadSize != 0)
+    {
+        if (dispaly_colour && pFrame->m_mastering_display.payLoadSize > 0)
```
dmitryermilov, line 1390: There is always != 0 payLoadSize check above.

_studio/mfx_lib/decode/h265/src/mfx_h265_dec_decode.cpp
pengxin99: If the frames do not have `MasteringDisplayColourVolume` and `ContentLightLevelInfo` SEI, **InsertPayloadToggle** is default set as `0` in output surface during decoding steps, this SEI will not be used in encoding steps.

And if parse these two SEIs during decoding steps, we need to set the  **InsertPayloadToggle** to `MFX_PAYLOAD_IDR`, then this SEI can be used in encoding steps.

Refers to [mfxstructures.h](https://github.com/intel-innersource/drivers.gpu.media.sdk.lib/blob/86b711813aef29045e36c374792d84aaf37996cc/api/vpl/mfxstructures.h#L2804)
>  If the application attaches this structure to the mfxEncodeCtrl structure at runtime, the encoder inserts the HDR SEI message for the current frame and ignores InsertPayloadToggle.
If the application attaches this structure to the mfxVideoParam structure during initialization or reset, the encoder inserts the HDR SEI message based on InsertPayloadToggle.

dmitryermilov: But it's an encoder input flag. It doesn't seems that decoder should control encoder behavior.

pengxin99: Do you mean the encoder will set this flag to decide which frame needs to use HDR SEI?

dmitryermilov: It's discussible. I think a user should do it. But there might be other options like the one you proposed. Anyway, we need to get aligned on how users will operate with the feature.

softworkz: I wonder whether this is even worth any discussion:

- For decoding, the two new structures are provided as part of the frame data
- For encoding, the data needs to be either attached once at init to mfxVideoParam (constant values which are repeated) or to mfxEncodeCtrl for each frame that is submitted for encoding
- As the SDK consumer doesn't control the lifetime of the memory for those structures coming from the decoder, it cannot simply use these structures as-is and attach them to the encoding parameters.
  That means: it will be required to allocate new structures and copy the individual values manually - so it's all up to the developer's coding and neither the flag nor even the structures will automatically be propagated to the encoder

This looks pretty good to me already..

pengxin99: Thanks for the advice, we will soon close the discussion and then update this PR. 
@softworkz, can you provide some HEVC clips contain both mastering_display and content_Light_Level SEI? So i can do some tests for this RP, thank you in advance~

softworkz: You can try one of those:
- https://4kmedia.org/ultra-hd-hdr-samsung-4k-demo-wonderland/
- https://4kmedia.org/sony-swordsmith-hdr-uhd-4k-demo/

They have Mastering Display data for sure, CLL most likely but don't remember exactly.

softworkz:
Hi, what happened to this PR?

It doesn't look like there would be any showstopping reason?

dmitryermilov (approved):
LGTM but we need to decide if we need to update manual, especially for VPL. I'll ping Igor.

PR #2692, opened 2021-05-17
IMPORTANT: load VPL runtime by MSDK dispatcher

aobolensk:
PR description provided by @dvrogozh...

This PR highlights important change in overall Media SDK program and introduces related changes into Media SDK code base.

Media SDK lifetime comes to an end in a form it exists right now. It will be maintained to support existing applications and "legacy" platforms (see below notes on runtime). Future development is planned to happen within [oneVPL](https://github.com/oneapi-src/oneVPL) library and its runtime implementations (here is [gen runtime](https://github.com/oneapi-src/oneVPL-intel-gpu)). While the name changes, API exposed by oneVPL is actually next 2.x version of the API exposed by Media SDK. Pay attention that 2.x API is not (eventually) backward compatible with 1.x API. However, backward incompatibility comes from [1.x features which got dropped](https://spec.oneapi.com/versions/latest/elements/oneVPL/source/appendix/VPL_intel_media_sdk.html), but those structures and functions which remained has preserved their layouts and signatures – this is important to highlight for runtime discussion below.

There are few things which complicate the whole Media SDK to oneVPL migration story. Long story short, generally speaking it’s not possible to fully drop Media SDK and completely switch to oneVPL. I would like to pay attention on the following 2 aspects:

* Media SDK runtime library (libmfxhw64.so.1) is still required to support “legacy” GPU platforms since new VPL runtime (libmfx-gen.so.1) **does not** support these platforms. That’s the case, for example, for SKL/KBL/CFL.
* Some frameworks which uses Media SDK did not yet switch to oneVPL. For example, ffmpeg-qsv plugins still require Media SDK (on the contrary gstreamer msdk plugins from gst-plugins-bad set which I believe has enabled support for oneVPL already).

Overall, pay attention that oneVPL library will load different runtimes depending of the underlying GPU platform:

* It will load new VPL runtime (libmfx-gen.so.1) on a new GPU platforms (like TGL)
* It will load old VPL runtime (libmfxhw64.so.1) on an older GPU platforms (like SKL)

This PR is adding similar behavior to Media SDK dispatcher library which also will load different runtimes depending on the underlying GPU platform – see PR for details and examples.

Summarizing, the following ingredients are required to support Intel media stack thru oneVPL/MSDK:
| Repo | Produced artifacts | Comments |
| ---- | ------------------ | -------- |
[...]

PR #2730, opened 2021-07-09
Disable ADL-S platform support

egorovdanil:


dvrogozh (changes requested):
This is a NO GO CHANGE!

ADL-S support went to the mediasdk release:
* https://github.com/Intel-Media-SDK/MediaSDK/blob/24b964e32692f68ed01aad73850bdaa440c2fcd7/api/include/mfxcommon.h#L183

You 100% can NOT revert changes (at least) in:
* API headers
* Documentation

And I strongly suggest to avoid modifying source code either. What we can do is to redirect execution to oneVPL, i.e. see #2692. As for this change, just drop it.

PR #2755, opened 2021-08-13
Fixed tile modifier issue for NV12 format

VUNNYSODHI:
If we are using I915_FORMAT_MOD_Y_TILED which is needed for
NV12 format then modifiers need to be enabled which can be passed
as flags argument to ioctl.

Issue: None
Test: Manually

Signed-off-by: Sodhi, Vunny <vunny.sodhi@intel.com>


dmitryermilov (approved):

VUNNYSODHI:
Thanks @dmitryermilov for merge.

dmitryermilov:
> Thanks @dmitryermilov for merge.

I shall thank you:)

Issue #2721, opened 2021-06-30
MFX_ERR_INCOMPATIBLE_VIDEO_PARAM error when implement adding source session dynamically  by reset mfxvpp in sink session to let it work


nickifu:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/fuqiaoni/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<link rel=themeData
href="file:///C:/Users/fuqiaoni/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/fuqiaoni/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>

</head>

<body lang=EN-US style='tab-interval:.5in;word-wrap:break-word'>

<p class=MsoNormal>We reference the sample MediaSDK Sample_multi_transcode to
implement the NVR use case, we meet MFX_ERR_INCOMPATIBLE_VIDEO_PARAM error
when<span style='mso-spacerun:yes'>  </span>implement the adding source session
dynamically <span style='mso-spacerun:yes'> </span>and try to reset mfxvpp in
sink session to let it work(removing source session dynamically works well), below
are the details:<o:p></o:p></p>

<p class=MsoNormal><b>The detail of the use case:<o:p></o:p></b></p>

<ol style='margin-top:0in' start=1 type=1>
 <li class=MsoListParagraphCxSpFirst style='margin-left:0in;mso-add-space:auto;
     mso-list:l0 level1 lfo1'>The pipeline including multiple input streaming from
     IP cameras and corresponding multiple decode + vpp sessions and then
     composition to one channel and encode it.<o:p></o:p></li>
</ol>

<p class=MsoListParagraphCxSpLast><o:p>&nbsp;</o:p></p>

<p class=MsoNormal style='margin-left:.5in'>N source sessions to 1 sink
session:<o:p></o:p></p>

[...]

PR #2758, opened 2021-08-19
[VPP] Fix increasing num of input channels doesn't work

MicroYY:
Fix: https://github.com/Intel-Media-SDK/MediaSDK/issues/2721

Number of input composite layers can be dynamically changed at runtime, but should not exceed the initial counts.
Signed off by: jay.yang@intel.com

### DIFF
--- a/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
+++ b/_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
@@ -5869,17 +5869,16 @@ mfxStatus ConfigureExecuteParams(
                         mfxExtVPPComposite* extComp = (mfxExtVPPComposite*) videoParam.ExtParam[i];
                         StreamCount = extComp->NumInputStream;
 
-                        if (!executeParams.dstRects.empty())
+                        if (executeParams.dstRects.empty())
                         {
-                            if (executeParams.dstRects.size() < StreamCount)
-                                return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
+                            executeParams.initialStreamNum = StreamCount;
                         }
+                        else
+                        {
+                            MFX_CHECK(StreamCount <= executeParams.initialStreamNum, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
+                        }
+                        executeParams.dstRects.resize(StreamCount);
 
-                        if (executeParams.dstRects.size() != StreamCount)
-                        {
-                            executeParams.dstRects.clear();
-                            executeParams.dstRects.resize(StreamCount);
-                        }
 #if MFX_VERSION > 1023
                         executeParams.iTilesNum4Comp = extComp->NumTiles;
 #endif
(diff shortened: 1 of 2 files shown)


dmitryermilov (changes requested):
Hi Jay,

It doesn't look as a valid fix. Please see https://github.com/Intel-Media-SDK/MediaSDK/issues/2721#issuecomment-883578697

MicroYY:
Hi Dima.
From my understanding, we just need to allocate the `executeParams.dstRects` in the first time. If reset num of input stream exceeds it's size, then we return error.

dmitryermilov:
> Hi Dima.
> From my understanding, we just need to allocate the `executeParams.dstRects` in the first time. If reset num of input stream exceeds it's size, then we return error.

It will work for a num of channels decrease?

MicroYY:
> > Hi Dima.
> > From my understanding, we just need to allocate the `executeParams.dstRects` in the first time. If reset num of input stream exceeds it's size, then we return error.
> 
> Will it work for a num of channels decrease? Please consider e.g. the following cases:
> 
> * 10 -> 5
> * 10 -> 11> * 10 -> 5 > 6 -> 10
> * 10 -> 5 > 6 -> 11

Updated PR.

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5871,15 +5871,15 @@ mfxStatus ConfigureExecuteParams(
 
                         if (!executeParams.dstRects.empty())
                         {
-                            if (executeParams.dstRects.size() < StreamCount)
+                            if (StreamCount > executeParams.initialStreamNum)
```
dmitryermilov, line 5874: Thank you. Now it looks correct. But this will look better:
```
                        if (executeParams.dstRects.empty())
                        {
                            executeParams.initialStreamNum = StreamCount;
                        }
                        else
                        {
                            if (StreamCount > executeParams.initialStreamNum)
                                return MFX_ERR_INCOMPATIBLE_VIDEO_PARAM;
                        }
                        executeParams.dstRects.resize(StreamCount);
```

dmitryermilov (approved):

_studio/mfx_lib/vpp/src/mfx_vpp_hw.cpp
```diff
@@ -5869,17 +5869,16 @@ mfxStatus ConfigureExecuteParams(
...
                         }
-
-                        if (executeParams.dstRects.size() != StreamCount)
+                        else
                         {
-                            executeParams.dstRects.clear();
-                            executeParams.dstRects.resize(StreamCount);
+                            MFX_CHECK(StreamCount <= executeParams.initialStreamNum, MFX_ERR_INCOMPATIBLE_VIDEO_PARAM);
                         }
+                            executeParams.dstRects.resize(StreamCount);
```
dmitryermilov, line 5880: please remove extra space

dmitryermilov (approved):

PR #2759, opened 2021-08-20
Fix the name for oneVPL RT on Linux

xhaihao:
There isn't libmfx-gen.so.1 when building the library from
https://github.com/oneapi-src/oneVPL-intel-gpu

Install the project...
/usr/bin/cmake -P cmake_install.cmake
-- Install configuration: "Debug"
-- Up-to-date: /usr/lib/x86_64-linux-gnu/libmfx-gen.so.1.2.3
-- Up-to-date: /usr/lib/x86_64-linux-gnu/libmfx-gen.so.1.2
-- Up-to-date: /usr/lib/x86_64-linux-gnu/libmfx-gen.so
-- Up-to-date: /usr/lib/x86_64-linux-gnu/pkgconfig/libmfx-gen.pc

### DIFF
--- a/api/mfx_dispatch/linux/mfxloader.cpp
+++ b/api/mfx_dispatch/linux/mfxloader.cpp
@@ -46,7 +46,7 @@ namespace MFX {
     #else
         #define LIBMFXSW "libmfxsw32.so.1"
         #define LIBMFXHW "libmfxhw32.so.1"
-        #define ONEVPLRT "libmfx-gen.so.1"
+        #define ONEVPLRT "libmfx-gen.so"
     #endif
 #elif defined(__x86_64__)
     #ifdef ANDROID
@@ -55,7 +55,7 @@ namespace MFX {
     #else
         #define LIBMFXSW "libmfxsw64.so.1"
         #define LIBMFXHW "libmfxhw64.so.1"
-        #define ONEVPLRT "libmfx-gen.so.1"
+        #define ONEVPLRT "libmfx-gen.so"
     #endif
 #else
     #error Unsupported architecture
--- a/tests/unit/suites/mfx_dispatch/linux/mfx_dispatch_test_cases_libs.cpp
+++ b/tests/unit/suites/mfx_dispatch/linux/mfx_dispatch_test_cases_libs.cpp
@@ -85,7 +85,7 @@ TEST_P(DispatcherLibsTestParametrized, ShouldEnumerateCorrectLibNames)
 #elif defined(__x86_64__)
     const std::string LIBMFXSW("libmfxsw64.so.1");
     const std::string LIBMFXHW("libmfxhw64.so.1");
-    const std::string ONEVPLRT("libmfx-gen.so.1");
+    const std::string ONEVPLRT("libmfx-gen.so");
 #endif
 
     std::string modules_dir(MFX_MODULES_DIR);


api/mfx_dispatch/linux/mfxloader.cpp
```diff
@@ -46,7 +46,7 @@ namespace MFX {
     #else
         #define LIBMFXSW "libmfxsw32.so.1"
         #define LIBMFXHW "libmfxhw32.so.1"
-        #define ONEVPLRT "libmfx-gen.so.1"
+        #define ONEVPLRT "libmfx-gen.so"
```
dvrogozh, line 49: Well, yes, this is a problem. However, I would argue that: 1) .so is a linker name and should not be used in dlopen() calls, 2) libmfx-gen should be fixed to have proper .1 soname instead of 1.2.

Also, you have CI failing on this pr...

dvrogozh (changes requested):

api/mfx_dispatch/linux/mfxloader.cpp
xhaihao: Is libmfxhw64.so not a linker name? libmfxhw64.so is used for MSDK RT

uartie: OneVPL dispatcher uses "libmfx-gen.so.1.2".  The second number is the "mfx major version".

uartie: ... that is, it is the API major version.

xhaihao: We will have to change libmfx code in the future if using libmfx-gen.so.1.2 after bumping a new major version for libvpl

uartie: It's unclear to me why "1" is the prefix.

uartie: The first number should be "major"

uartie: In current version scheme, if mfx major changes but not library major, then we expect it's still backwards compatible?

xhaihao: @dvrogozh I don't understand why this change cause CI failing, maybe this is a case issue. Is there a way to run CI locally?

uartie: > We will have to change libmfx code in the future if using libmfx-gen.so.1.2 after bumping a new major version for libvpl

OneVPL dispatcher has the same problem.

dvrogozh: Yes, of course, just run `make test`.

uartie: > In current version scheme, if mfx major changes but not library major, then we expect it's still backwards compatible?

If so, then @dvrogozh original #2 should be the fix.  But also would require vpl dispatcher fix.
If not, then the "1." should be dropped in the libmfx-gen version.  And still requires vpl dispatcher fix, too.

The only way to avoid vpl dispatcher change is to load libmfx-gen.so.1.2 in mfx dispatcher.

uartie: However, I think @dvrogozh suggested fix is the "right" fix.  And, we will need to do some damage control with vpl dispatcher.

xhaihao: @jonrecker any comment for oneVPL library name?

uartie (changes requested):
Also adding my "request-changes" flag... because I have to agree with @dvrogozh.  Loading just the `.so` is dangerous because it does not guarantee ABI compatibility.  SONAMES (i.e. library name with version after `.so.`... https://en.wikipedia.org/wiki/Soname) indicate ABI compatibility/incompatibility and should be used instead.

Most libraries that I've looked at on my Linux machine only use a `libname.so.<X>` as SONAME.  I found a few that use `libname.so.<X>.<Y>` as SONAME.  Nonetheless, the standard convention is that anytime the version part changes (anything after the `.so.`) means ABI **incompatibility**.

Currently, `<library major>.<mfx major>` is the ABI compatibility version for libmfx-gen.  So, the question is whether this is actually correct?  If there is a scenario where `<mfx major>` changes but not `<library major>` and is still ABI compatible, then the current ABI versioning scheme is wrong.

xhaihao:
@uartie actually libvpl.so.1.2 is not backward compatible with libmfx.so.1 because some features are removed.

uartie:
> @uartie actually libvpl.so.1.2 is not backward compatible with libmfx.so.1 because some features are removed.

But now you're comparing 2 different libraries.  Not the same thing as my previous argument.

xhaihao:
@uartie libmfx.so should not load libmfx-gen.so.1.2 if we consider the ABI compatibility because a feature user is using might not be implemented in libmfx-gen.so.1.2

uartie:
> @uartie libmfx.so should not load libmfx-gen.so.1.2 if we consider the ABI compatibility because a feature user is using might not be implemented in libmfx-gen.so.1.2

So loading `libmfx-gen.so` is any better? No.

uartie:
> > @uartie libmfx.so should not load libmfx-gen.so.1.2 if we consider the ABI compatibility because a feature user is using might not be implemented in libmfx-gen.so.1.2
> 
> So loading `libmfx-gen.so` is any better? No.

That is, what about a feature a user is using _is implemented_, but the ABI for that feature changes.  IDK, maybe I'm crazy 😝

xhaihao:
> > @uartie libmfx.so should not load libmfx-gen.so.1.2 if we consider the ABI compatibility because a feature user is using might not be implemented in libmfx-gen.so.1.2
> 
> So loading `libmfx-gen.so` is any better? No.

Considering MediaSDK is going to be supported in maintainence mode, I think libmfx-gen.so is better. If so, we needn't change libmfx in the future. But I'm fine to use libmfx-gen.so.1, libmfx-gen.so.1.2, libmfx-gen.so.2 or others. I will close this PR if oneVPL may provide libmfx-gen.so.1

xhaihao:
@uartie @dvrogozh @dmitryermilov are there more comments for this PR ?

uartie:
> @uartie @dvrogozh @dmitryermilov are there more comments for this PR ?

I rest my case... and now defer to @dvrogozh @dmitryermilov

dmitryermilov:
> @uartie @dvrogozh @dmitryermilov are there more comments for this PR ?



> y case... and now defer to @dvrogozh @dmitryermilov

We're going to discuss it tomorrow with @varistar . @dvrogozh  is on vacation.

dmitryermilov:
> > @uartie @dvrogozh @dmitryermilov are there more comments for this PR ?
> 
> > y case... and now defer to @dvrogozh @dmitryermilov
> 
> We're going to discuss it tomorrow with @varistar . @dvrogozh is on vacation.

Here is the PR reflecting the decision made: https://github.com/Intel-Media-SDK/MediaSDK/pull/2771

PR #2770, opened 2021-08-31
Fix idr frame can't be set except the first GOP when idr_interval use…

TangZhiZhen:
… default 0

Signed-off-by: Zhizhen Tang <zhizhen.tang@intel.com>

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
@@ -1069,7 +1069,7 @@ public:
         mfxU32 gopOptFlag = par.mvp.mfx.GopOptFlag;
         mfxU32 gopPicSize = par.mvp.mfx.GopPicSize;
         mfxU32 gopRefDist = par.mvp.mfx.GopRefDist;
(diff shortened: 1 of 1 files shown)


TangZhiZhen:
according to doc/samples/readme-encode_linux.md.
[-idr_interval size]| idr interval, default 0 means every I is an IDR, 1 means every other I frame is an IDR etc

dmitryermilov (changes requested):
According to MediaSDK manual:
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#mfxinfomfx
`For HEVC, if IdrInterval = 0, then only first I-frame is an IDR-frame. If IdrInterval = 1, then every I-frame is an IDR-frame. If IdrInterval = 2, then every other I-frame is an IDR-frame, etc.`

TangZhiZhen:
> According to MediaSDK manual:
> https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#mfxinfomfx
> `For HEVC, if IdrInterval = 0, then only first I-frame is an IDR-frame. If IdrInterval = 1, then every I-frame is an IDR-frame. If IdrInterval = 2, then every other I-frame is an IDR-frame, etc.`

Got it, Thank Dmitry.

PR #2780, opened 2021-09-08
[Decode]Enable AVC CSC via sfc pipeline

shawnli2:
AVC SFC CSC is supported

### DIFF
--- a/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
+++ b/_studio/mfx_lib/decode/h264/src/mfx_h264_dec_decode.cpp
@@ -927,8 +927,10 @@ mfxStatus VideoDECODEH264::QueryIOSurf(VideoCORE *core, mfxVideoParam *par, mfxF
         // need to substitute output format
         // number of surfaces is same
         request->Info.FourCC = videoProcessing->Out.FourCC;
-
         request->Info.ChromaFormat = videoProcessing->Out.ChromaFormat;
+        sts = UpdateCscOutputFormat(par, request);
+        MFX_CHECK_STS(sts);
+
         request->Info.Width = videoProcessing->Out.Width;
         request->Info.Height = videoProcessing->Out.Height;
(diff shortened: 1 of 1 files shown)


dmitryermilov (commented):
@shawnli2 , are there any HW limitations for CSC?

shawnli2:
> @shawnli2 , are there any HW limitations for CSC?
I 
@dmitryermilov  Yes, platform before TGL only support 8 bit output format, UpdateCscOutputFormat() will query and check. If do not support, will return MFX_ERR_INVALID_VIDEO_PARAM.

dmitryermilov:
> > @shawnli2 , are there any HW limitations for CSC?
> > I
> > @dmitryermilov  Yes, platform before TGL only support 8 bit output format, UpdateCscOutputFormat() will query and check. If do not support, will return MFX_ERR_INVALID_VIDEO_PARAM.
> > https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/shared/src/mfx_common_int.cpp#L305

Thanks. But MFX_ERR_INVALID_VIDEO_PARAM seems to be incorrect status. Can you please change it to MFX_ERR_UNSUPPORTED in another PR?

dmitryermilov (approved):

shawnli2:
> > > @shawnli2 , are there any HW limitations for CSC?
> > > I
> > > @dmitryermilov  Yes, platform before TGL only support 8 bit output format, UpdateCscOutputFormat() will query and check. If do not support, will return MFX_ERR_INVALID_VIDEO_PARAM.
> > > https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/shared/src/mfx_common_int.cpp#L305
> 
> Thanks. But MFX_ERR_INVALID_VIDEO_PARAM seems to be incorrect status. Can you please change it to MFX_ERR_UNSUPPORTED in another PR?

Thanks for review. Sure, I will!

PR #2782, opened 2021-09-08
Fixed wstring into string conversion warning

dmitryermilov:


dvrogozh (commented):
Imho, correct solution is to:
1. REMOVE all wchar, tchar and other Windows chars from samples
2. REMOVE all wrappers over trivial functions

PR #2785, opened 2021-09-13
[Decode CSC] return unsupport status when query unsupport output format

shawnli2:
MFX_EMFX_ERR_INVALID_VIDEO_PARAM seems to be incorrect status;
return MFX_ERR_UNSUPPORTED when output format is unsupported.

request from: #2780


shawnli2:
@dmitryermilov Please help review, request from #2780 , thanks!

dmitryermilov (approved):

dmitryermilov:
Thank you!

PR #2789, opened 2021-09-16
[Hevcd] Check the integrity of the SEI

chuanli1:
NALU contain NAL_UT_SEI_SUFFIX header, but part of
Supplemental enhancement information is missing

Issue: MDP-67689
Test: manually


dmitryermilov (changes requested):
As we discussed in internal thread, let's make bits check inside GetBits()/PeakNextBits() functions, not around each GetBits()/PeakNextBits() call

chuanli1:
> As we discussed in internal thread, let's make bits check inside GetBits()/PeakNextBits() functions, not around each GetBits()/PeakNextBits() call

Okay, I have closed this change, then I will use better solution to fix these issues.

PR #2791, opened 2021-09-22
VPL RT is a PoR on ADL-S

dmitryermilov:


daleksan (approved):
confirm that ADL is POR for VPL platform

dvrogozh (approved):
LGTM

PR #2795, opened 2021-09-27
[Decode] Enable HEVC CSC via sfc pipeline

cyew3:
HEVC SFC CSC is supported

Issue: MDP-65483
Test: -


wangyan-intel:
@shawnli2 Could you please take a look?

PR #2843, opened 2021-12-09
[smt] Add support of p010 input stream

sbalandi:


dmitryermilov:
@sfominx , can you please help with CI?

PR #2874, opened 2022-01-24
[Decode][hevcd] Add hevc FrameRate parameters parsing from VideoParamSet

pengxin99:
In HEVC MFXVideoDECODE_DecodeHeader(), the framerate is calculated from SPS, but the framerate should combine the info from VPS and SPS according to the spec. This PR fixes this.

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_supplier.cpp
+++ b/_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_supplier.cpp
@@ -258,13 +258,7 @@ UMC::Status MFXTaskSupplier_H265::FillVideoParam(mfxVideoParam *par, bool full)
     if (!seq)
         return UMC::UMC_ERR_FAILED;
 
-    if (vps != nullptr && (vps->getTimingInfo()->vps_timing_info_present_flag || full))
-    {
-        par->mfx.FrameInfo.FrameRateExtD = vps->getTimingInfo()->vps_num_units_in_tick;
-        par->mfx.FrameInfo.FrameRateExtN = vps->getTimingInfo()->vps_time_scale;
-    }
-
-    if (MFX_Utility::FillVideoParam(seq, par, full) != UMC::UMC_OK)
+    if (MFX_Utility::FillVideoParam(vps, seq, par, full) != UMC::UMC_OK)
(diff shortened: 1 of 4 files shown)


pengxin99:
@wangyan-intel pls help to review

wangyan-intel (approved):

_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_supplier.cpp
```diff
@@ -252,10 +252,18 @@ void MFXTaskSupplier_H265::SetVideoParams(mfxVideoParam * par)
...
+
     if (!seq)
         return UMC::UMC_ERR_FAILED;
 
+    if (vps != nullptr && (vps->getTimingInfo()->vps_timing_info_present_flag || full))
+    {
+        par->mfx.FrameInfo.FrameRateExtD = vps->getTimingInfo()->vps_num_units_in_tick;
+        par->mfx.FrameInfo.FrameRateExtN = vps->getTimingInfo()->vps_time_scale;
+    }
+
```
dmitryermilov, line 266: This contradicts the rest code. I mean we fill all params inside FillVideoParam() and only FrameRate here. Any special reason for it?

dmitryermilov (changes requested):

_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_supplier.cpp
pengxin99: The FrameRate can get from VPS or SPS, but FillVideoParam() only gets SPS as a parameter. So add FrameRate set based on VPS here.

dmitryermilov: > The FrameRate can get from VPS or SPS, but FillVideoParam() only gets SPS as a parameter. 

Why not to add VPS as well?

pengxin99: Add VPS into FillVideoParam() done.

dmitryermilov (approved):

wangyan-intel (approved):

Issue #2627, opened 2021-03-17
mfxExtPictureTimingSEI and interlaced


mliu2020:
This is a [question from IDZ forum](https://community.intel.com/t5/Media-Intel-oneAPI-Video/mfxExtPictureTimingSEI-and-interlaced/m-p/1252921#M13544) about mfxExtPictureTimingSEI() and I believe this is a good place to ask:

Customer has following scenario:
"I'm using mfxExtPictureTimingSEI on the mfxEncodeCtrl struct that I send to MFXVideoENCODE_EncodeFrameAsync, in order to insert pic_timing info on the frames.

For progressive video, everything works OK, but for interlaced, both fields have identical pic_timing values, while I need the time_offset of the first field to be 0, and the second field to be 1.

I've noticed that mfxExtPictureTimingSEI has 3 timestamp structs, but whatever I set on the two latter doesn't seem to have any effect. Only the first timestamp is used for both fields.

Can you help me find out if there's a way to set different data on each field? I wouldn't like to modify the bytes after encoding to find the pic_timing and change the 0 for a 1."

"- What's the point of having the timestamp as an array of three timestamps? Only TimeStamp[0] seems to be used. What should I set in TimeStamp[1] and TimeStamp[2]. The manual doesn't say anything about it, and I couldn't find it used on the samples.

- How can I make top and bottom field to have different values? They both have the values set in TimeStamp[0]."

Mark

dmitryermilov:
@mliu2020 , it seems it's about old https://community.intel.com/t5/Media-Intel-oneAPI-Video/H264-Timecode-absent-in-QSV/m-p/950913

mliu2020:
Thanks Dmitry,

I read through all the post and scenario is similar. If my understanding is correct, mfxExtPictureTimingSEI is not very relevant to this case. The customer in new post seems more confused by mfxExtPictureTimingSEI.

- What's the point of having the timestamp as an array of three timestamps? Only TimeStamp[0] seems to be used. What should I set in TimeStamp[1] and TimeStamp[2]. The manual doesn't say anything about it, and I couldn't find it used on the samples.

I have forwarded this post to customer.

Mark

Fernickk:
Hello, I'm the "customer" ^_^. Thank you for taking the time to address my issue.

In my case, the timecode is applied to the hardware-encoded video. It's just a case of a single parameter (time_offset), that I can't generate as I like.

[...]

PR #2875, opened 2022-01-25
Fix mfxExtPictureTimingSEI for interlaced

kovakimy:
Fixes: #2627

### DIFF
--- a/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils.cpp
+++ b/_studio/mfx_lib/encode_hw/h264/src/mfx_h264_encode_hw_utils.cpp
@@ -5635,9 +5635,10 @@ namespace
 {
     mfxExtPictureTimingSEI const * SelectPicTimingSei(
         MfxVideoParam const & video,
-        DdiTask const &       task)
+        DdiTask const &       task,
+        mfxU32                fieldId)
     {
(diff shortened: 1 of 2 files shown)


dmitryermilov (commented):
Hi @kovakimy ,

Did you test this change? Asking because I guess it won't work since we have a special function which checks if extBuffers can be paired. See https://github.com/Intel-Media-SDK/MediaSDK/blob/2ddcdc9485a705b37b7cf734b884ecb11714d6ef/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L1808

Please try to modify some simple tutorial in order to verify the change works fine. There's can be other placed I'm not aware of like the one above.

kovakimy:
> Hi @kovakimy ,
> 
> Did you test this change? Asking because I guess it won't work since we have a special function which checks if extBuffers can be paired. See
> 
> https://github.com/Intel-Media-SDK/MediaSDK/blob/2ddcdc9485a705b37b7cf734b884ecb11714d6ef/_studio/mfx_lib/shared/src/mfx_h264_enc_common_hw.cpp#L1808
> 
> Please try to modify some simple tutorial in order to verify the change works fine. There's can be other placed I'm not aware of like the one above.

Thank you, @dmitryermilov !
Added buffer to 2 verification funcs, tested on simple_3_encode with some changes related to added pair of buffers in ctrl.

dmitryermilov (approved):

PR #2877, opened 2022-02-03
Scan all render nodes in dispatcher

onabiull:
In certain configurations, notably in containers it is possible to have /sys/class/drm/renderD128 and 
/sys/class/drm/renderD129 but no /dev/dri/renderD128. This fix enumerate all renderD128+(0..63) instead of stopping
at first failure.

### DIFF
--- a/api/mfx_dispatch/linux/device_ids.h
+++ b/api/mfx_dispatch/linux/device_ids.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Intel Corporation
+// Copyright (c) 2022 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -19,6 +19,7 @@
 // SOFTWARE.
 
 #include <stdio.h>
+#include <fcntl.h>
 #include <sys/stat.h>
 #include <algorithm>
 #include <string>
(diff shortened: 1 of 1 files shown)


api/mfx_dispatch/linux/device_ids.h
```diff
@@ -415,19 +415,19 @@ std::vector <Device> get_devices() {
...
             continue;
         }
         path = std::string(dir) + "/renderD" + std::to_string(128 + i) + device_id_file;
         file = fopen(path.c_str(), "r");
-        if (!file) break;
+        if (!file) continue;
         err = fscanf(file, "%x", &device.device_id);
         fclose(file);
-        if (err == EOF) break;
+        if (err == EOF) continue;
```
dvrogozh, line 430: This still won't work for the docker use case I have in mind. If system has 2 gpus and user explicit adds only one with  `docker --device /dev/dri/renderD129` as an example, we will still have an issue because docker will have full access to sysfs entries, i.e. /sys/class/drm/* which we actually parse, but it will map only /dev/dri/renderD129 which is what application opens. So, we will populate the table with something which is not actually accessible, i.e. with /dev/dri/renderD128.

I still think that this change makes a sense, but we need to slightly modify it with something like:
```
    for (; i < 64; ++i) {
        Device device;
        std::string path = std::string(dir) + "/renderD" + std::to_string(128 + i) + vendor_id_file;
        FILE *file = fopen(path.c_str(), "r");
        if (!file) continue;
        err = fscanf(file, "%x", &device.vendor_id);
        fclose(file);
        if (err == EOF) continue;
        if (device.vendor_id != 0x8086) {  // Filter out non-Intel devices
            continue;
        }
        path = std::string(dir) + "/renderD" + std::to_string(128 + i) + device_id_file;
        file = fopen(path.c_str(), "r");
        if (!file) continue;
        err = fscanf(file, "%x", &device.device_id);
        fclose(file);
        if (err == EOF) continue;

        path = "/dev/dri/" + "/renderD" + std::to_string(128 + i);       <<<<<<<<<<<<<<<<<<<< NEW
        int fd=open(path.c_str());
        if (fd < 0) continue; //device not accessible

        device.platform = get_platform(device.device_id);
        result.emplace_back(device);
```

dvrogozh (changes requested):

api/mfx_dispatch/linux/device_ids.h
onabiull: Thank you! Change adjusted as  suggested.

api/mfx_dispatch/linux/device_ids.h
```diff
@@ -413,21 +413,29 @@ std::vector <Device> get_devices() {
...
+
+        // if user only mapped /dev/dri/renderD129 in container, need to skip /dev/dri/renderD128
+        path = "/dev/dri/renderD" + node_num;
+        int fd=open(path.c_str());
```
dvrogozh, line 435: mode is missing:
```
int fd=open(path.c_str(), O_RDWR));
```

api/mfx_dispatch/linux/device_ids.h
```diff
@@ -413,21 +413,29 @@ std::vector <Device> get_devices() {
...
-        if (!file) break;
+        if (!file) continue;
         err = fscanf(file, "%x", &device.device_id);
         fclose(file);
-        if (err == EOF) break;
+        if (err == EOF) continue;
+
+        // if user only mapped /dev/dri/renderD129 in container, need to skip /dev/dri/renderD128
+        path = "/dev/dri/renderD" + node_num;
+        int fd = open(path.c_str(), O_RDWR);
```
dvrogozh, line 435: open on Linux needs these headers, some are missing:
```
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
```

dvrogozh (approved):
Looks good now.

PR #2881, opened 2022-02-10
[hevce] Only VDEnc mode is available on JSL & EHL

xhaihao:
The SDK always use VDEnc mode for AVC & VP9 encoding on JSL & EHL by
default. To keep the hehavior consistent across codecs, use VDEnc mode
by default for HEVC encoding as well


xhaihao:
@thomasli21801  This repo is for MediaSDK, not oneVPL. According to https://github.com/Intel-Media-SDK/MediaSDK#media-sdk-support-matrix, oneVPL runtime is not available on JSL/EHL.

dmitryermilov:
> BTW, this change is for VPL and you might need to create another PR for 1_x branch, right?

It's handled semi automatically.

PR #2887, opened 2022-02-16
[hevce] Use VDEnc by default for A2RGB10 input

xhaihao:
VME doesn't support A2RGB10 input

### DIFF
--- a/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
+++ b/_studio/mfx_lib/encode_hw/hevc/agnostic/base/hevcehw_base_legacy_defaults.cpp
@@ -1459,6 +1459,7 @@ public:
 #if (MFX_VERSION >= 1027)
                || fcc == MFX_FOURCC_Y410
 #endif
+               || fcc == MFX_FOURCC_A2RGB10
                   )));
 
         return mfxU16(


dmitryermilov:
@thomasli21801 , can you please review?

thomasli21801:
Quick question, how about ARGB8?

xhaihao:
Our testing shows VME supports RGB4

thomasli21801:
> Our testing shows VME supports RGB4

Thanks Haihao, I mean from perf/quality wise or solution consolidation wise, will you prefer to direct rgb 8/10bit to VDEnc as default?

xhaihao:
> > Our testing shows VME supports RGB4
> 
> Thanks Haihao, I mean from perf/quality wise or solution consolidation wise, will you prefer to direct rgb 8/10bit to VDEnc as default? Also, may I know which platforms this change targets?

It seems MSDK prefers VME, I don't think we need to change the default behavior.  But for rgb10bit, it is impossible to use VME.

dmitryermilov:
> > > Our testing shows VME supports RGB4
> > 
> > 
> > Thanks Haihao, I mean from perf/quality wise or solution consolidation wise, will you prefer to direct rgb 8/10bit to VDEnc as default? Also, may I know which platforms this change targets?
> 
> It seems MSDK prefers VME, I don't think we need to change the default behavior. But for rgb10bit, it is impossible to use VME.

It's just a legacy behavior we have to keep on already released platfroms.

thomasli21801:
got it. thanks for explanation. The change looks good to me.

thomasli21801:
> > > Our testing shows VME supports RGB4
> > 
> > 
> > Thanks Haihao, I mean from perf/quality wise or solution consolidation wise, will you prefer to direct rgb 8/10bit to VDEnc as default? Also, may I know which platforms this change targets?
> 
> It seems MSDK prefers VME, I don't think we need to change the default behavior. But for rgb10bit, it is impossible to use VME.

This preference is for ICL and TGL, right?

dmitryermilov:
> > > > Our testing shows VME supports RGB4
> > > 
> > > 
> > > Thanks Haihao, I mean from perf/quality wise or solution consolidation wise, will you prefer to direct rgb 8/10bit to VDEnc as default? Also, may I know which platforms this change targets?
> > 
> > 
> > It seems MSDK prefers VME, I don't think we need to change the default behavior. But for rgb10bit, it is impossible to use VME.
> 
> This preference is for ICL and TGL, right?

If a user doesn't force LowPower to some value, then AFAIR on all platforms where VME is available encoders uses VME by default.  

> Any customer is using RGB 10bit?
I don't know exactly but I assume we can build logic to prefer VME/VDEnc if user requested specific FOURCC supported only by VME/VDEnc.

thomasli21801:
> > > > > Our testing shows VME supports RGB4
> > > > 
> > > > 
> > > > Thanks Haihao, I mean from perf/quality wise or solution consolidation wise, will you prefer to direct rgb 8/10bit to VDEnc as default? Also, may I know which platforms this change targets?
> > > 
> > > 
> > > It seems MSDK prefers VME, I don't think we need to change the default behavior. But for rgb10bit, it is impossible to use VME.
> > 
> > 
> > This preference is for ICL and TGL, right?
> 
> If a user doesn't force LowPower to some value, then AFAIR on all platforms where VME is available encoders uses VME by default.
> 
> > Any customer is using RGB 10bit?
> > I don't know exactly but I assume we can build logic to prefer VME/VDEnc if user requested specific FOURCC supported only by VME/VDEnc.

 Thanks Dmitry. I'm wondering if any exposure. The change looks good to me.

PR #2888, opened 2022-02-16
[Decode] Handling the abnormal state

chuanli1:
Issue: MDP-68765
Test: manual

### DIFF
--- a/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
+++ b/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
@@ -946,7 +946,9 @@ namespace UMC_MPEG2_DECODER
         if (sliceCount)
         {
             EliminateSliceErrors(frame, fieldIndex);
-            Submit(frame, fieldIndex);
+            UMC::Status sts = Submit(frame, fieldIndex);
+            if (UMC::UMC_OK != sts)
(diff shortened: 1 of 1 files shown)


_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -946,7 +946,9 @@ namespace UMC_MPEG2_DECODER
         if (sliceCount)
         {
             EliminateSliceErrors(frame, fieldIndex);
-            Submit(frame, fieldIndex);
+            UMC::Status sts = Submit(frame, fieldIndex);
+            if (UMC::UMC_OK != sts)
+                throw mpeg2_exception(UMC::UMC_ERR_INVALID_STREAM);
```
dmitryermilov, line 951: It's an error from driver not parsing error. I assume it's better to return something like DEVICE_FAILED or similar.

dmitryermilov (approved):

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
chuanli1: Thanks for your comment, DEVICE_FAILED is not defined in the state of the UMC, and I use UMC_ERR_FAILED instead of the previous state.

wangyan-intel:
LGTM

PR #2892, opened 2022-02-21
Revert "[Decode] Handling the abnormal state"

chuanli1:
Reverts Intel-Media-SDK/MediaSDK#2888


chuanli1:
because of some platform regression, I need to revert this PR.

PR #2893, opened 2022-02-21
[Decode] Reasonably assign the first frame slice

chuanli1:
Issue: MDP-68765
Test: manual

### DIFF
--- a/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
+++ b/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
@@ -498,10 +498,8 @@ namespace UMC_MPEG2_DECODER
                 return UMC::UMC_ERR_NOT_ENOUGH_BUFFER;
             }
         }
-
-        const auto picExt = *m_currHeaders.picExtHdr;
-        uint32_t fieldIndex = m_currFrame->GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
-        MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(fieldIndex);
+    
+        MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(m_currFrame->firstFieldIndex);
 
         // Add the slice to the picture
         info.AddSlice(slice);
@@ -792,8 +790,10 @@ namespace UMC_MPEG2_DECODER
                 }
             }
         }
-
+        //The flag of each slice of current frame, 0 means first field (in decoded order) or frame, 1 means 2nd field (in decoded order)
         frame.currFieldIndex = frame.GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
+        //The flag of the first slice of current frame, 0 means first field (in decoded order) or frame, 1 means 2nd field (in decoded order)
+        frame.firstFieldIndex = frame.currFieldIndex;
     }
 
     MPEG2DecoderFrame* MPEG2Decoder::StartFrame(MPEG2Slice * slice)
@@ -843,7 +843,7 @@ namespace UMC_MPEG2_DECODER
 
     bool MPEG2Decoder::IsFieldOfCurrentFrame() const
     {
-        const auto firstFrameSlice = m_currFrame->GetAU(0)->GetSlice(0);
+        const auto firstFrameSlice = m_currFrame->GetAU(m_currFrame->firstFieldIndex)->GetSlice(0);
         const auto picHdr = firstFrameSlice->GetPicHeader();
         const auto picExtHdr = firstFrameSlice->GetPicExtHeader();
         const auto newPicHdr = *m_currHeaders.picHdr.get();
--- a/_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
+++ b/_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_frame.h
@@ -231,7 +231,7 @@ namespace UMC_MPEG2_DECODER
         FrameType                         frameType;
 
         int8_t                            currFieldIndex; // 0 means first field (in decoded order) or frame, 1 means 2nd field (in decoded order)
-
+        uint32_t                          firstFieldIndex;
         virtual void Free()
         {
             auto fd = GetFrameData();


_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
```diff
@@ -212,6 +212,7 @@ namespace UMC_MPEG2_DECODER
         std::mutex                      m_guard;
         UMC::FrameAllocator*            m_allocator;
         uint32_t                        m_counter;
+        uint32_t                        fieldIndex;
```
wangyan-intel, line 215: m_fieldIndex？

wangyan-intel (changes requested):

_studio/shared/umc/codec/mpeg2_dec/include/umc_mpeg2_decoder.h
chuanli1: Hi Yan, code updated.

wangyan-intel (approved):

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -500,8 +500,8 @@ namespace UMC_MPEG2_DECODER
         }
 
         const auto picExt = *m_currHeaders.picExtHdr;
-        uint32_t fieldIndex = m_currFrame->GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
```
dmitryermilov, line 503: Can we do it only once on the first slice, not when adding each new slice?

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -500,8 +500,8 @@ namespace UMC_MPEG2_DECODER
...
-        MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(fieldIndex);
+        m_fieldIndex = m_currFrame->GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
+        MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(m_fieldIndex);
```
dmitryermilov, line 504: Not fully clear what the values if m_fieldIndex is. Can you please explain?

chuanli1: when fieldIndex=0, this is a top field or frame,  `slicesInfo` will be assigned by `MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(fieldIndex)`
when m_fieldIndex = 1, this is a bottom field, `slicesInfoBottom` will be assigned by `MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(fieldIndex)`
when run firstFrameSlice = m_currFrame->`GetAU(0)`->GetSlice(0), the value of `fieldIndex` defaults to 0 in `GetAU(0)`, but we get this `slicesInfoBottom` within a frame, this pointer will return null, so I use xx to get a valid `slicesInfo` or `slicesInfoBottom`.

dmitryermilov: Sorry, probably we're talking about different things. That is the value of class member variable instead of local variable? What is the reason to keep this "local" frame specific member in "global" MPEG2Decoder class in may contain lots of frames with a different field parity?

dmitryermilov (changes requested):

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
chuanli1: > Sorry, probably we're talking about different things. That is the value of class member variable instead of local variable? What is the reason to keep this "local" frame specific member in "global" MPEG2Decoder class which may contain lots of frames with a different field parity?

because this error clip only has the top field and no bottom field, the top field information should be used in [MPEG2Decoder::IsFieldOfCurrentFrame](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp#L846), but for the program to run, I use the bottom field information, we get the top field, bottom field and frame slice info through this local variable fieldIndex, in order to get the slice information in [MPEG2Decoder::IsFieldOfCurrentFrame](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp#L846), I need to define it in the class.
Or according to this error clip with no bottom field, msdk can directly return error, could you give the better suggestions? Thanks.

dmitryermilov: My point is:
1) You shouldn't define **frame** level variable into **super-class** MPEG2Decoder. If you need a new member, you can define it inside MPEG2DecoderFrame, not globally
2) You can field index locally:
```
    bool MPEG2Decoder::IsFieldOfCurrentFrame() const
    {
        auto fieldIndex = m_currFrame->GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
        const auto firstFrameSlice = m_currFrame->GetAU(fieldIndex )->GetSlice(0);
```
1st point is a strong request. 2nd point is optional.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -500,8 +500,8 @@ namespace UMC_MPEG2_DECODER
         }
 
         const auto picExt = *m_currHeaders.picExtHdr;
-        uint32_t fieldIndex = m_currFrame->GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
-        MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(fieldIndex);
+        m_currFrame->fieldIndex = m_currFrame->GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
```
dmitryermilov, line 503: Can we do it only **once** on the first slice, not when adding **each new** slice?

chuanli1: It doesn't seem feasible.
1. after `MPEG2DecoderFrameInfo & info = *m_currFrame->GetAU(fieldIndex)`, we can get the slice size according to `info.GetSliceCount()`
2. according to `slices.size() = 0`, we can get the first slice.
And there is a judgment of the first slice in [MPEG2DecoderFrameInfo::AddSlice](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_frame.cpp#L64)

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
chuanli1: thanks, code updated.

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
dmitryermilov: Sorry, I don't follow what is not feasible here. 
See https://github.com/Intel-Media-SDK/MediaSDK/blob/b4cfab66cfe92958fed6bff30691ed11d8748eb9/_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp#L723

dmitryermilov (approved):

wangyan-intel (dismissed):

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
```diff
@@ -794,6 +792,7 @@ namespace UMC_MPEG2_DECODER
         }
 
         frame.currFieldIndex = frame.GetNumberByParity(picExt.picture_structure == BOTTOM_FLD_PICTURE);
+        frame.fieldIndex = frame.currFieldIndex;
```
dmitryermilov, line 795: Need explanation what currFieldIndex and fieldIndex. What the difference is.

dmitryermilov (changes requested):

_studio/shared/umc/codec/mpeg2_dec/src/umc_mpeg2_decoder.cpp
chuanli1: Thanks for your good suggestion, code updated, pls help review it again.

dmitryermilov (approved):

PR #2896, opened 2022-02-22
[Decode] scan component check code refactor

chuanli1:
Issue: MDP-68552
Test: manual

### DIFF
--- a/_studio/shared/umc/codec/jpeg_dec/src/jpegdec.cpp
+++ b/_studio/shared/umc/codec/jpeg_dec/src/jpegdec.cpp
@@ -2371,10 +2371,6 @@ JERRCODE CJPEGDecoder::DecodeHuffmanMCURowBL(int16_t* pMCUBuf, uint32_t colMCU, 
 
     for(n = m_curr_scan->first_comp; n < m_curr_scan->first_comp + m_curr_scan->ncomps; n++)
     {
-      if(m_curr_scan->ncomps < 1 || m_curr_scan->ncomps > MAX_COMPS_PER_SCAN)
-      {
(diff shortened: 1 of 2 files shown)


chuanli1:
code review trace to https://github.com/intel-innersource/drivers.gpu.media.sdk.lib/pull/4439

wangyan-intel (approved):

_studio/shared/umc/codec/jpeg_dec/src/jpegdec.cpp
```diff
@@ -2371,10 +2371,6 @@ JERRCODE CJPEGDecoder::DecodeHuffmanMCURowBL(int16_t* pMCUBuf, uint32_t colMCU,
 
     for(n = m_curr_scan->first_comp; n < m_curr_scan->first_comp + m_curr_scan->ncomps; n++)
     {
-      if(m_curr_scan->ncomps < 1 || m_curr_scan->ncomps > MAX_COMPS_PER_SCAN)
-      {
-        return JPEG_ERR_SOS_DATA;
-      }
```
dmitryermilov, line 2377: Can you please say why we don't need this check anymore?

chuanli1: Hi, from @dgurulev comment:
`The check is not depend on loop variable, I'd rather move it out of the for loop. Moreover, I believe this check is needed at the place where m_curr_scan->ncomps is evaluated` in https://github.com/intel-innersource/drivers.gpu.media.sdk.lib/pull/4439, so I updated my PR.

PR #2908, opened 2022-03-14
[Decode] Check if the bitstream length is valid

chuanli1:
Issue: MDP-68850
Test: manual

### DIFF
--- a/_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_common.cpp
+++ b/_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_common.cpp
@@ -73,7 +73,8 @@ namespace VP8DecodeCommon
         {
             // set offset, but leave last six bytes
             // since they can be start bytes of start code and frame tag
-    
+            if(n_bytes_offset < 6)
+                return MFX_ERR_MORE_BITSTREAM;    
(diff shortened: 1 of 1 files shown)


wangyan-intel (changes requested):
Could we add check at parsing not use this value?

_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_common.cpp
```diff
@@ -73,7 +73,8 @@ namespace VP8DecodeCommon
         {
             // set offset, but leave last six bytes
             // since they can be start bytes of start code and frame tag
-    
+            if(n_bytes_offset < 6)
```
wangyan-intel, line 76: Could we add check at parsing not use this value?

wangyan-intel (changes requested):

_studio/mfx_lib/decode/vp8/src/mfx_vp8_dec_decode_common.cpp
chuanli1: Hi yan, thanks for your review.
`n_bytes_offset` is obtained here, msdk don't find the start code when parser this error clip, so we need to set offset, but leave last six bytes, because they can be start bytes of start code and frame tag.

wangyan-intel (approved):

PR #2909, opened 2022-03-14
Revert "[Decode] Reasonably assign the first frame slice"

chuanli1:
Reverts Intel-Media-SDK/MediaSDK#2893


chuanli1:
Because of some platform regression, need to revert it.

dmitryermilov (approved):

Issue #2935, opened 2022-05-26
Full CMake project for Window


TKNgu:
Can add CMakeLists.txt for mfx_dispatch/windows ? Now ms visual studio support CMake project.

Biswa96:
I have compiled only the dispatcher with cmake in mingw toolchain. Do you have any project to test the library?

TKNgu:
I create new CMakeLists.txt and add to root CMake project: add_subdirectory(lib/extention/intel/)

```
#Cmake version.
cmake_minimum_required(VERSION 3.10.2)

# C++ std 14.
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_EXE_LINKER_FLAGS)

# VideoCore
project(WindowIntelMFX VERSION 1.0.0 LANGUAGES CXX C DESCRIPTION "Lib intel decode window")

# Source code
file(GLOB SRC
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/main.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_critical_section.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_dispatcher.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_dispatcher_log.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_driver_store_loader.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_dxva2_device.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_function_table.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_library_iterator.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_load_dll.cpp
	${CMAKE_CURRENT_SOURCE_DIR}/MediaSDK/api/mfx_dispatch/windows/src/mfx_load_plugin.cpp
[...]

PR #2943, opened 2022-06-07
[Dispatcher] Add cmake files for Windows platform

Biswa96:


wangyan-intel:
Thanks for your contribution. In general, this repo is for Linux only. It doesn't include MSDK windows runtime.

Biswa96:
The dispatcher library only has Visual Studio project files. With CMake, one can compile or cross compile it without Visual Studio. It will not require MS c/c++ runtime and will link with any other c/c++ libraries that comes with compiler toolchain e.g. gcc, clang etc.

wangyan-intel:
So you just need compile dispatcher for Windows? Windows Intel media driver folder should include MSDK built runtime and dispatcher which are close source.

Biswa96:
> So you just need compile dispatcher for Windows?

Yes.

wangyan-intel:
This repo is not used to accept the patch for Windows code.

Biswa96:
CMake project was also requested here https://github.com/Intel-Media-SDK/MediaSDK/issues/2935

Issue #2983, opened 2022-08-25
invalid fd close cause issues when application already open the fd 0, in VA-surface transfer.


monstercy:
## System information
- CPU information(`cat /proc/cpuinfo | grep "model name" | uniq`): all cpu
- GPU  information(`lspci -nn | grep -E 'VGA|isplay`): all gpu
- Display server if rendering to display(X or wayland):X11

### Describe the current behavior
in file https://github.com/Intel-Media-SDK/MediaSDK/blob/51ffdf59f281a8604f6f10cf3775cf276ff1d30d/samples/sample_common/src/vaapi_utils.cpp#L454
it will close fd "close(ctx->fd);" which has not been assign value.  most of times the ctx->fd is 0, it will cause trouble, if appellation already open fd  0 for other purpose.
### Describe the expected behavior
files with fd as 0  will be closed. 

## Debug information
- What's libva/libva-utils/gmmlib/media-driver/Media SDK version? latest version.
- Could you confirm whether GPU hardware exist or not by `ls /dev/dri`? no
- Could you attach dmesg log if it's GPU hang by `dmesg >dmesg.log 2>&1`? no
- Could you provide vainfo log if possible by `vainfo -a >vainfo.log 2>&1`? no
- Could you provide strace log if possible by `strace YOUR_CMD >strace.log 2>&1`? no
- Could you provide libva trace log if possible? Run cmd `export LIBVA_TRACE=/tmp/libva_trace.log` first then execute the case.
- Media SDK tracer output (https://github.com/Intel-Media-SDK/MediaSDK/blob/master/tools/tracer/README.md)?
- Do you want to contribute a PR? (yes/no): yes

PR #2984, opened 2022-08-25
fix: remove the invalid close fd

monstercy:
fix: remove the invalid close fd, AcquireVASurface is used for surface to surface. no fd is needed.

### DIFF
--- a/samples/sample_common/src/vaapi_utils.cpp
+++ b/samples/sample_common/src/vaapi_utils.cpp
@@ -355,7 +355,6 @@ CLibVA* CreateLibVA(const std::string& devicePath, int type)
 
 struct AcquireCtx
 {
-    int fd;
     VAImage image;
 };
 
@@ -451,7 +450,6 @@ void CLibVA::ReleaseVASurface(
         AcquireCtx* ctx = (AcquireCtx*)actx;
         if (ctx) {
             m_libva.vaDestroySurfaces(dpy2, &srf2, 1);
-            close(ctx->fd);
             m_libva.vaReleaseBufferHandle(dpy1, ctx->image.buf);
             m_libva.vaDestroyImage(dpy1, ctx->image.image_id);
             free(ctx);


samples/sample_common/src/vaapi_utils.cpp
```diff
@@ -451,7 +451,6 @@ void CLibVA::ReleaseVASurface(
         AcquireCtx* ctx = (AcquireCtx*)actx;
         if (ctx) {
             m_libva.vaDestroySurfaces(dpy2, &srf2, 1);
-            close(ctx->fd);
```
lwan108, line 454: When is this closed?

monstercy: > When is this closed?
 AcquireVASurface is just handle surface thing, No fd is exposed to outside, so we don't need to close it.

monstercy: https://github.com/Intel-Media-SDK/MediaSDK/issues/2983

lwan108: Prefer below change:
if (ctx->fd) {
close(ctx->fd);}

monstercy: sure.  I updated it based on your feedback.

lwan108: Thanks for the change.

lwan108 (dismissed):
LGTM

monstercy:
> LGTM

Thank you for your approval, it seems it still needs more reviewer for merge, 
@lwan108 would you help with it, please?

lwan108:
@Sherry-Lin, need your help to merge the change.

samples/sample_common/src/vaapi_utils.cpp
```diff
@@ -451,7 +451,9 @@ void CLibVA::ReleaseVASurface(
...
+            if (ctx->fd) {
```
dvrogozh, line 454: This is wrong check. You must check for `>=0` since that's a range of valid fds:
```
if (ctx->fd >= 0) {
    close(ctx->fd);
}
```

dvrogozh (changes requested):

samples/sample_common/src/vaapi_utils.cpp
monstercy: the current issue is that default ctx->fd value is 0, but it is a invalid fd, if we closed it here, it has chance to close the fd 0 which is the file opened by other component and cause trouble to overall appliction.

dvrogozh: > the current issue is that default ctx->fd value is 0

Where it was set to be 0? Can we change the default to -1 which is more appropriate?

monstercy: @dvrogozh  I updated the code based on your feedback.

samples/sample_common/src/vaapi_utils.cpp
```diff
@@ -400,6 +400,8 @@ VAStatus CLibVA::AcquireVASurface(
     ctx = (AcquireCtx*)calloc(1, sizeof(AcquireCtx));
     if (!ctx) return VA_STATUS_ERROR_OPERATION_FAILED;
 
+    ctx->fd = -1;
```
dvrogozh, line 403: I am trying to search where this `fd` is actually used and I actually don't see usages. Is it actually used somewhere? If not - we can just delete the code entirely.

monstercy: > I am trying to search where this `fd` is actually used and I actually don't see usages. Is it actually used somewhere? If not - we can just delete the code entirely.
both the modifications can resolve the issues, the fd exists may because it have  other reasons. Shall we have the patch in to so that we can fix current issue.

monstercy: > I am trying to search where this `fd` is actually used and I actually don't see usages. Is it actually used somewhere? If not - we can just delete the code entirely.
both the modifications can resolve the issues, the fd exists may because it have  other reasons. Shall we have the patch in to so that we can fix current issue.

monstercy: OK, how about keeping AcquireCtx as wrapper and remove the member of fd, in case new member added in future?

dvrogozh: I am fine with that. Still can't believe `fd` is not actually used... Let's see if build will pass.

dvrogozh: so, build is ok.

monstercy:
@dvrogozh I merged the 5 commit into one, but i closed the pull by accident, can you help to open the pr?

monstercy:
> 5 commits won't do. Please, rebase and squash all commits into a single one. Prepare it for merge.

done

dvrogozh (approved):
looks good

monstercy:
@lwan108  could you kindly review the new code updated?

lwan108 (approved):
LGTM

PR #2768, opened 2021-08-27
vaapi: recognize VA_FOURCC_XYUV VAImage

xhaihao:
In FFmpeg, VA_FOURCC_XYUV is used for 4:4:4 8bit content when vaapi is
    used for hardware acceleration, hence we will have to use VA_FOURCC_XYUV too
    for 4:4:4 8bit content when qsv is used on Linux because qsv is based on
    vaapi, and map VA_FOURCC_XYUV to MFX_FOURCC_AYUV in FFmpeg. So mapping
    VA_FOURCC_XYUV is needed in the SDK.
    
    $ ffmpeg -y -hwaccel qsv -i in444.mp4 -vf "hwdownload,format=vuyx" -f
    null -

PR #2985, opened 2022-09-08
vaapi: recognize P012, Y212, and Y412 VAImages (#2768)

xhaihao:
In FFmpeg VAAPI path, the following fourcc codes are used for 12bit contents

VA_FOURCC_P012 4:2:0 12bit
VA_FOURCC_Y212 4:2:2 12bit
VA_FOURCC_Y412 4:4:4 12bit

We should use these codes too in FFmpeg QSV path, hence add mapping P012/Y212/Y412 here

### DIFF
--- a/_studio/shared/src/libmfx_allocator_vaapi.cpp
+++ b/_studio/shared/src/libmfx_allocator_vaapi.cpp
@@ -568,14 +568,17 @@ mfxStatus mfxDefaultAllocatorVAAPI::SetFrameData(const VAImage &va_image, mfxU32
     case VA_FOURCC_P010:
 #if (MFX_VERSION >= 1031)
     case VA_FOURCC_P016:
+    case VA_FOURCC_P012:
 #endif
-        if (mfx_fourcc != va_image.format.fourcc) return MFX_ERR_LOCK_MEMORY;
-
+        if ((mfx_fourcc == MFX_FOURCC_P010 && va_image.format.fourcc == VA_FOURCC_P010) ||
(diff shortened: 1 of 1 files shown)


XinfengZhang (approved):

_studio/shared/src/libmfx_allocator_vaapi.cpp
```diff
@@ -619,14 +625,15 @@ mfxStatus mfxDefaultAllocatorVAAPI::SetFrameData(const VAImage &va_image, mfxU32
 
 #if (MFX_VERSION >= 1031)
     case VA_FOURCC_Y416:
-        if (mfx_fourcc != va_image.format.fourcc) return MFX_ERR_LOCK_MEMORY;
-
+    case VA_FOURCC_Y412:
+        if (mfx_fourcc == MFX_FOURCC_Y416)
```
wangyan-intel, line 629: Need also add MFX_FOURCC_Y412?

wangyan-intel (changes requested):

_studio/shared/src/libmfx_allocator_vaapi.cpp
xhaihao: No. There isn't `MFX_FOURCC_Y412`, Here add mapping between  VA_FOURCC_Y412 VA Image  and `MFX_FOURCC_Y416` mfx surface

wangyan-intel: Thanks for clarification.

wangyan-intel (approved):

PR #2998, opened 2022-11-28
add missing <stdint.h> to mfxparser.cpp

trofi:
Without the change build fails on `gcc-13` as:

    MediaSDK/api/mfx_dispatch/linux/mfxparser.cpp: In function 'std::string MFX::printCodecId(mfxU32)':
    MediaSDK/api/mfx_dispatch/linux/mfxparser.cpp:60:3: error: 'uint8_t' was not declared in this scope
       60 |   uint8_t* data = reinterpret_cast<uint8_t*>(&id);
          |   ^~~~~~~


lwan108 (approved):
LGTM

PR #3003, opened 2023-01-13
hevce: use Low Power mode for RGB encoding by default

wenbinc-Bin:
According to: https://github.com/intel/media-driver/blob/master/docs/media_features.md, Only VDENC supports RGB encoding, so use lowpower when input RGB data.


wenbinc-Bin:
This pr align with [https://github.com/oneapi-src/oneVPL-intel-gpu/pull/256](https://github.com/oneapi-src/oneVPL-intel-gpu/pull/256).

thomasli21801 (approved):

lwan108 (approved):

PR #3013, opened 2023-02-28
[Decode]Fix hevc decode issue

Davis-HXX:
The “m_maxBsSize” is data of type “int”, and BitsDecoded() returns data of type “unsigned long long”, when the latter is large enough, the former minus the latter may get a positive number, although the former is smaller than the latter, which will causes no exception to be thrown(even if there is not enough data to read).

### DIFF
--- a/_studio/shared/umc/codec/h265_dec/include/umc_h265_bitstream_headers.h
+++ b/_studio/shared/umc/codec/h265_dec/include/umc_h265_bitstream_headers.h
@@ -549,7 +549,12 @@ inline size_t H265BaseBitstream::BytesLeft() const
 // Throw exception if left bits not enough to decode
 inline void H265BaseBitstream::CheckBitsLeft(const uint32_t nbits)
 {
-    if ((int32_t)m_maxBsSize * 8 - BitsDecoded() < nbits)
+    if((int32_t)m_maxBsSize > 0 && (size_t)m_maxBsSize * 8 > BitsDecoded())
+    {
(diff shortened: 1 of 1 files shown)


_studio/shared/umc/codec/h265_dec/include/umc_h265_bitstream_headers.h
```diff
@@ -549,7 +549,12 @@ inline size_t H265BaseBitstream::BytesLeft() const
 // Throw exception if left bits not enough to decode
 inline void H265BaseBitstream::CheckBitsLeft(const uint32_t nbits)
 {
-    if ((int32_t)m_maxBsSize * 8 - BitsDecoded() < nbits)
+    if((int32_t)m_maxBsSize > 0 && (size_t)m_maxBsSize * 8 > BitsDecoded())
```
lwan108, line 552: when will m_maxBsSize be less then 0?

Davis-HXX: m_maxBsSize will be recalculated when AllocFrames, It is uncertain whether the value will be changed to negative due to the user's hacker operation, therefore, protection is carried out in advance.
Just as BitsDecoded () here returns an abnormal value, when the user decodes with a manually customized stream, it may result in an abnormal assignment, resulting in an out-of-bounds read operation(which we want to avoid).

lwan108: I still can't get it. 
Anyway, approved.

lwan108 (approved):

